Issue #27, opened 2014-12-30
Update data loader to use postgres copy csv to table function


kaitlin:
Kelly confirms that we can wipe the database with every load, so I think we should convert the incremental loader to generating a postgres formatted CSV that we can copy directly into the table. This should make sure we don't have any vestigial records and make the loader much faster.

PR #30, opened 2014-12-31
Fix Issue #27 and fix requirements for python 3

arowla:


kaitlin:
Also, I think the switch to the batch update means that the text search index doesn't get updated. I noticed that when I merged these changes and ran the loader, the text search no longer worked. So for instance this query:
http://localhost:8000/api/rates/?q=engineer

Returns lots of results under the old loader, but 0 under the new one. My guess is the indexing happens on save,  but save doesn't get called under bulk_create.

arowla:
Interesting. I knew `save()` didn't get called with `bulk_create`, but I thought the reindexing was triggered within Postgres, not from Django. Will look into it.

arowla:
The latest commits have the improved logging, and text reindexing fixed.

PR #35, opened 2015-01-06
Added greater than and less than price filtering to API

arowla:
- Added the filters
- Wrote tests for each filter
- Replaced fixture with object factory to speed up tests


kaitlin:
Looks good. Could you update so that both a __lt and __gt param can be passed in? The way the trello card is written I think we'll need to allow them to specify both at once.

arowla:
Done.

PR #37, opened 2015-01-06
Improve API test coverage

arowla:
This will be a running PR with the test changes. So far:

- Added tests for multi-word queries
- Added tests for education level
- Added tests for min and max experience


arowla:
We've now got 88% coverage on api/views.py, with 100% of the filtering covered.

PR #55, opened 2015-01-16
Added docstring for queryset and its filter params AND NEW TESTS! TEST REFACTORING! AND MORE!

arowla:
Taking a stab at a Google-formatted docstring. Comments very welcome. Should we be documenting the query params there, as I did?


arowla:
Oops, a bunch of selenium test commits got thrown in here. Guess I should have started a new branch. Derp.

arowla:
There, I changed the title, so that makes it ok.

Issue #44, opened 2015-01-09
Collapsible columns are not sortable on click


shawnbot:
The Schedule column has `class="sortable collapsible"`, but the collapsible click handler overrides the sortable one. @brethauer, do you have thoughts on how to make this work? I was thinking that a collapsed sortable column would just show the ◀ icon, but when it's shown it would look like ▽▶.

brethauer:
What if we use text instead of an icon to show/hide?  
Something along the lines of this:

![show](https://cloud.githubusercontent.com/assets/7635133/5760076/35300e78-9ca0-11e4-896d-31335e56d03a.png)

![hide](https://cloud.githubusercontent.com/assets/7635133/5760081/3a75e0d8-9ca0-11e4-8b90-14307dea9deb.png)

shawnbot:
This is what I have now in [my fix/sortable_and_collapsible branch](https://github.com/shawnbot/hourglass/tree/fix/sortable_and_collapsible):

![hourglass-collapse-fix](https://cloud.githubusercontent.com/assets/113896/5781151/23083296-9d65-11e4-93d4-9ed23c0d57e5.gif)

I guess the left/right arrow is superfluous, huh?

brethauer:
Wanted to run this by you @shawnbot :

![image](https://cloud.githubusercontent.com/assets/7635133/5782986/3da21694-9d8d-11e4-9e59-e59f3e8b338d.png)

PR #56, opened 2015-01-16
Allow columns to be sortable and collapsible

shawnbot:
The Schedule column is now sortable and collapsible, resolving #44:

hourglass-collapse-fix

@brethauer, should I ditch the little left/right triangle in the "show/shide" text and nudge it left by a couple of px, per your sketch?

hide

Also seems like it needs some `text-decoration: underline` and a lighter color, huh?


brethauer:
@shawnbot yeah let's take out the arrow, no need to nudge, but we can lighten it up a few shades, maybe #979797?

PR #81, opened 2015-03-10
Added exclude parameter to API

kaitlin:
- Added `exclude` param to API
- Added a unique `id` field to the api response/serializer 
- Updated README to document usage of `exclude` param
- Added a test
- Updated the `assertResultsEqual` method to test only for keys in the expected result. This will avoid needing to update every test when an attribute is added to the api response.


api/views.py
```diff
@@ -65,9 +66,16 @@ def get_contracts_queryset(request_params, wage_field):
     sort = request_params.get('sort', wage_field)
     # query_type can be: [ match_all (default) | match_phrase | match_exact ]
     query_type = request_params.get('query_type', 'match_all')
+    exclude = request_params.getlist('exclude')
+    if len(exclude) == 1 and ',' in exclude[0]:
```
arowla: Do you even need to do this check? `.split(',')` will return a single-valued list if there is no match for the separator.

api/tests.py
```diff
@@ -598,6 +598,22 @@ def test_filter_by_site(self):
            'contractor_site': 'Q3',
            'business_size': None}])
 
+    def test_exclude_by_id(self):
+        get_contract_recipe().make(_quantity=1, id=100)
```
arowla: I _think_ you should be able to do `get_contract_recipe().make(_quantity=3, id=cycle([100,101,102]))`

api/views.py
kaitlin: Good point. I will remove.

api/tests.py
kaitlin: This is what was throwing the error. It tries to cast a generator object to an int and fails. The error is `TypeError: int() argument must be a string or a number, not 'itertools.cycle'`. I feel like this is a limitation of mommy model so I enumerated the calls as a work around.

arowla: Hm. Maybe primary keys are special. Perhaps you should file a ticket over at model_mommy.

Do you still need the `_quantity=1` param, in that case?

kaitlin: Derp, no I don't. Removed.

PR #91, opened 2015-03-30
logo, header, about page

brethauer:


brethauer:
Spacing fixed.  Here is a shot of some possible logo directions.  I don't love the the vertical overly much, I think the general issue is that spelling it out is so darn wordy.  I'll run it by one of the visual designers to get some more input too.

logo_sketches

Issue #86, opened 2015-03-16
Visual design


shawnbot:
@jehlers gave me some great feedback on the Hourglass design today. Here are my notes:

- needs intro text: what is this, and how do I use it?
- information hierarchy: what's more important, the filters or the chart? how can we make it clear that the filters affect both the chart and the table?
- make it clear that prices are hourly
- label the chart to help people understand it

Jen also shared some inspiration:
- [CFPB's Owning a Home](http://www.consumerfinance.gov/owning-a-home/check-rates/)
- [TRUECar pricing page](https://www.truecar.com/prices-new/volkswagen/golf-pricing/)

shawnbot:
@brethauer and I whipped this up in Murally today:

![image](https://cloud.githubusercontent.com/assets/113896/6718375/932604be-cd70-11e4-949b-6ec4012c51fd.png)

shawnbot:
And here's an idea that I had for a logo:

![image](https://cloud.githubusercontent.com/assets/113896/6718476/2982ffac-cd71-11e4-99d6-9f0b3303fa12.png)

shawnbot:
Assigning @brethauer to think about colors and logo.

brethauer:
I think we can use a similar color palette to Discovery & FOIA.  @shawnbot go ahead and throw in your logo and see how it looks.  
Rough layout concept:

![calc_v1](https://cloud.githubusercontent.com/assets/7635133/6760113/38cbbc0e-cf1c-11e4-8c65-0f0feaa10bdf.png)

brethauer:
Carrying forward the issues Shawn mentioned:

1. *The spacing between the inputs is... yech. I think we'll need to break out of Skeleton's grid to fix this, though.*  **Agreed - not the highest priority but if it seems easy then take a crack at it**
2. *I think the histogram is a bit too big when it's this wide. I've made it 200px tall, and in that screenshot I was playing with increasing the number of bars, but it still feels like too much. Not sure what to do about that.*  **Yeah I think it is too wide - it gets hard to take in the whole thing at once.  Let's bring the width down to maybe 600 or so px and just let it stay left-aligned.**
3. *Some of these elements work really well as the screen width shrinks, and some don't. I think we're going to need a pretty serious style overhaul to make this layout responsive, especially with the tables.*  **Based on our understanding of our users, this is not a huge priority, so let's punt for a future sprint/iteration**

shawnbot:
OK cool, so:

1. Re: spacing, punting.
[...]

PR #92, opened 2015-03-31
Design tweaks

shawnbot:
These address the tasks in #86.

The one thing that I snuck in here is some new positioning for the column labels (the Schedule column's "show"/"hide" link on the right and the "Restore x rows" link on the left):

image

@brethauer let me know if you think the unicode characters are too much, or you want to swap them out for something else. :)


brethauer:
Nice! The repositioning makes a lot of sense.  And :+1: to the unicode chars.

PR #104, opened 2015-04-06
Fix frontend tests, add proper remote testing

shawnbot:
This PR fixes frontend tests so they pass on IE9, Chrome, Firefox, and PhantomJS.

This also introduces a new `make` target and general workflow for testing other browsers on Sauce. I've added some documentation for how this works to [CONTRIBUTING.md](https://github.com/18F/calc/blob/fix-frontend-tests/CONTRIBUTING.md), but the TL;DR is that you need to add your Sauce login info to `hourglass/local_settings.py` and then run:

```sh
make test-sauce
```

There's a bunch of logic in `selenium_tests/tests.py` for [creating the right Sauce configuration for remote testing](https://github.com/18F/calc/blob/fix-frontend-tests/selenium_tests/tests.py#L44-L66) and [getting all of the relevant config options](https://github.com/18F/calc/blob/fix-frontend-tests/selenium_tests/tests.py#L19-L34) from either `settings.REMOTE_TESTING` and/or environment variables. (I chose not to call out Sauce specifically because this configuration could *theoretically* be used to run tests on [BrowserStack](https://browserstack.com) as well.) Because the browser configuration can also be modified by environment variables, we can run multi-browser tests this way:

```sh
# Chrome first, then IE9, then Firefox
REMOTE_TESTING_BROWSER=chrome make test-sauce && \
make test-sauce && \
REMOTE_TESTING_BROWSER=firefox make test-sauce
```

This isn't quite ready for Travis (or another CI service) yet because, although you can provide your Sauce username and access key in the env, the tests are [hard-wired to run locally](https://github.com/18F/calc/blob/fix-frontend-tests/selenium_tests/tests.py#L44-L45) if there are no `settings.REMOTE_TESTING`. The remote testing configuration should be built from the environment if it's not provided in the settings.


kaitlin:
What is `lt-run`? it looks like it is referenced by the makefile and requires a node dependency.

Issue #117, opened 2015-04-14
Removed rows aren't added on clear search


shawnbot:
Steps to reproduce:

1. Remove some rows
2. Press "clear search"
3. Observe that the removed rows aren't added and the "restore rows" link remains

PR #125, opened 2015-04-28
Clear excluded rows on form reset

shawnbot:
This PR includes a fix for #117 and some miscellaneous fixes for the Selenium tests. @kaitlin I could use your eye on the `api/permissions.py` change, specifically. The deal there was that when running the Selenium tests on Sauce through https://localtunnel.me, the `X-Forwarded-For` header is set to a raw IP address, so the line that parses out the comma-separated bits was failing.


kaitlin:
Looks fine to me. I'll make sure to test on dev before deploying to prod though.

PR #128, opened 2015-04-28
Fix 508 sort headers

shawnbot:
Adds `title` attributes to all sortable headers that reflect their current state, for instance when a column is sorted it gets the title, "{title}: sorted {order}, click to sort {opposite order}".

@bristow, we decided to use `title` instead of sr-only spans and replaced one instance of `aria-describedby` that looked wrong. Thanks for taking a look!

Also, we killed the sortable class toggles on un-sortable columns.

### DIFF
--- a/hourglass_site/static/hourglass_site/js/index.js
+++ b/hourglass_site/static/hourglass_site/js/index.js
@@ -131,7 +131,9 @@
     });
 
     var sort = parseSortOrder(data.sort);
+    var sortable = function(d) { return d.sortable; };
     sortHeaders
+      .filter(sortable)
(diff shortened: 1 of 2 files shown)


nickbristow:
@shawnbot yeah, so the title fix is a good fix for this. Go for it.

maya:
Here's a visual of the "years" idea @shawnbot brought up:

**idea** (years text revealed on hover inside cell):
years idea

**now**:
years current

brethauer:
@maya This is a great tweak.  I really like the idea of showing it on hover, which avoids the problem of cluttering up every years row.

PR #140, opened 2015-05-08
Add vendor sortability back

maya:
I added back in sortability for the Vendor column. This is what it looks like:
screen shot 2015-05-08 at 11 38 06 am


theresaanna:
The test failure here is clearly a result of tests hanging and not a change in this PR. Merging.

PR #141, opened 2015-05-08
Remove price filter

maya:
This removes the price filter from the optional filters. This is what it looks like:

screen shot 2015-05-08 at 12 59 16 pm


maya:
I've updated the tests with the following:
- removed the price filter tests
- updated experience test to experience range from min experience
- added schedule tests

PR #142, opened 2015-05-11
Update dynamic sentence

maya:
This updates the dynamic sentence (mad-libs style) created above the histogram when selecting filters. I've removed the min/max price and changed min experience to experience range.

This is what it looks like now:
screen shot 2015-05-11 at 6 41 14 am


theresaanna:
Thanks for catching the experience range update.

PR #145, opened 2015-05-11
Add filter layout

maya:
This implements the new filter layout.

This is what it looks like:
screen shot 2015-05-11 at 3 20 30 pm


maya:
I’ve added in the the average price callout and rate explainer text below it.

Here's how it looks now:
screen shot 2015-05-11 at 5 00 32 pm

Issue #152, opened 2015-05-12
Color contrast issues


brethauer:
1) Need more contrast between average price indicator (blue text) and grey background
2) More contrast on grey "What's this?" text
3) Check out contrast on Xs
4) Check out contrast on blue buttons

![untitled](https://cloud.githubusercontent.com/assets/7635133/7597665/fab0ee32-f8c2-11e4-9d04-60f3d5e55e57.png)

maya:
The grays in the table pass, however let me know if you want to switch it to our new link colors:

![screen shot 2015-05-12 at 2 29 35 pm](https://cloud.githubusercontent.com/assets/5249443/7598986/6224e222-f8b3-11e4-8441-f0a43a666fde.png)

PR #156, opened 2015-05-12
Fix color contrast for 508 accessibility

maya:
While I do love our bright and fresh blue, @brethauer noted they didn't pass color contrast checks for 508 accessibility here: https://github.com/18F/calc/issues/152

Here's how it looks with accessible colors:

screen shot 2015-05-12 at 2 22 55 pm


maya:
Hover color looks like this: 
screen shot 2015-05-12 at 2 00 48 pm

PR #164, opened 2015-05-19
Add button for users to download histogram as an image

theresaanna:
- In modern browsers, the button will automatically save the graph as 'histogram.png'.
- In IE9, a modal will open asking the user to right click and save.

Introduces libraries to convert from svg -> canvas, along with polyfills for the Blob API (https://developer.mozilla.org/en-US/docs/Web/API/Blob) and `saveAs()`. Also introduces the Vex modal library, which only loads in IE9.


theresaanna:
@shawnbot Would you mind reviewing when you get the chance? Thanks.

hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -59,6 +60,9 @@
     submit(true);
   });
 
+  histogramDownloadLink.className = "button button-primary";
```
shawnbot: I think this should just be done in the HTML.

theresaanna: I should comment this. The purpose of this line is to remove a `hidden` class that is in the template.

theresaanna:
FileSaver.js is a dependency of canvg.

theresaanna:
Ah, so this is my non-classList way of doing what you described. We have been looking at supporting IE8, so I didn't want to add more IE8-unfriendly things. The template has `class="hidden button button-primary"` already.

I'll remove the hidden class and add a noscript note. Thanks!

theresaanna:
@maya @brethauer and I talked about this with Kelly and decided that the transparent bg is fine, since the charts are mostly going to be inserted into Word docs.

I tried to do it quickly but had trouble getting the background color to persist through to the image. Do you think its ok to merge and put an item in the backlog to add the bg color?

PR #165, opened 2015-05-20
Add tooltips and use aria-label

maya:
WIP :warning: :construction: 

This uses [tooltipster.js](http://iamceege.github.io/tooltipster/) to display tooltips on table headers and exclude rows. This displays the dynamic content on `aria-label` (was previously `title`) for greater accessibility.

Thanks to @theresaanna and @shawnbot!

Table headers:
screen shot 2015-05-20 at 3 15 25 pm

Exclude results:
screen shot 2015-05-20 at 3 15 52 pm

## Bug :bug: 
There is a bug where bc the “Hide Schedule” button is structured inside a table header, the tooltip is firing on hover, even though it is visually outside of it. One idea is to move the hide button outside of the table header.

screen shot 2015-05-20 at 3 45 50 pm

### DIFF
--- a/hourglass_site/static/hourglass_site/js/index.js
+++ b/hourglass_site/static/hourglass_site/js/index.js
@@ -870,3 +870,8 @@
   }
 
 })(this);
+
+// initialize tooltipster.js
+$(document).ready(function() {
+    $('.tooltip').tooltipster();
+});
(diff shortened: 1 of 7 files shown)


theresaanna:
The bug is fixed with b655e47.

PR #168, opened 2015-05-29
Reverse sort arrows

dhcole:
Looks like the sort arrows are reserved. This sets them so that pointing down is descending and pointing up is ascending.

**Old**
screen shot 2015-05-29 at 12 10 37 pm

**New**
screen shot 2015-05-29 at 12 11 08 pm


brethauer:
Thanks @dhcole !

PR #170, opened 2015-06-04
Update contract data

theresaanna:
This updates the database with a fresh set of data. It also adds some documentation on how to update the data and adds a new db field for `contract_year`, the year of the contract that we are in (values 1-5).

At first I filtered within the API, but it felt out of place. Maybe in the future if we need to do more filtering on the data from a source that isn't user-inputted filters, we should build in a clear place where its done, but for now it felt best to do it on the import side. Especially because the current price is dictated at time of import, not within the API, so that value doesn't change until new data is imported.


theresaanna:
@arowla Would you mind reviewing, please and thanks?

contracts/models.py
```diff
@@ -60,6 +60,7 @@ class Contract(models.Model):
     piid = models.CharField(max_length=128) #index this field
     contract_start = models.DateField(null=True, blank=True)
     contract_end = models.DateField(null=True, blank=True)
+    contract_year = models.DecimalField(max_digits=1, decimal_places=0, null=True, blank=True)
```
arowla: I'm curious about the `max_digits` and `decimal_places` values here and in the migration. Why use a Decimal field (vs Integer) if we won't have any decimal places? And can we expect the `contract_year` to always be less than 10?

contracts/management/commands/load_data.py
```diff
@@ -71,14 +72,14 @@ def handle(self, *args, **options):
...
-                    if contract.contract_end > today and contract.contract_start < today:
-                        #it's a current contract, need to find which year we're in
-                        start_day = contract.contract_start
-                        for plus_year in range(0,5):
-                            if date(year=start_day.year + plus_year, month=start_day.month, day=start_day.day) < today:
-                                contract.current_price = getattr(contract, 'hourly_rate_year' + str(plus_year + 1))
+                    # don't create current price for records where the rate
+                    # is under the federal minimum contract rate
+                    current_price = getattr(contract, 'hourly_rate_year' + str(line[14]))
+                    if current_price and current_price >= 10.10:
```
arowla: It would be nice to see `10.10` pulled out into a constant.

contracts/models.py
theresaanna: Ha. Dumb mistake on my part. We can definitely depend on the contract year to be less than ten. It can only go up to five, as it corresponds with one of the five fields of yearly rate data.

contracts/management/commands/load_data.py
theresaanna: I knew something felt strange here but I didn't know how I should handle it. Thank you!

arowla:
This looks fine to me with the change to `IntegerField` and creation of a constant for `FEDERAL_MIN_CONTRACT_RATE` (or whatever-you-want-to-call-it).

theresaanna:
@arowla I discovered and used `PositiveSmallIntegerField` instead of `IntegerField`. I've also put the rate into a constant. Mind taking a second quick look? Thanks!

theresaanna:
Implemented Alison's changes, so am going to merge myself instead of having to pester her again.

PR #173, opened 2015-06-09
Add multi-select filter

maya:
**WIP** :warning: :construction:

This adds the frontend code for a multi-select filter.

This is what it looks like:
multiselect filter

Live gif: http://cl.ly/bVkn


maya:
Problems are:
1. data is not loading for the histogram
2. selecting multiple things is not getting entered into the main field like below. 
Live example here: http://codepen.io/mbenari/pen/KpmGRJ
screen shot 2015-06-08 at 8 17 48 pm

maya:
The last commit fixes the issue (#2) above, so now you can select a few results and see them visually: http://cl.ly/bXBv

I'm guessing data is not loading bc it needs to be synched up.

maya:
There's an empty HTML anchor that's used to open the dropdown `<a href="#">` that makes the page jump to the top when you click it. 

Could something like this fix it? http://stackoverflow.com/questions/493175/how-can-i-create-an-empty-html-anchor-so-the-page-doesnt-jump-up-when-i-click

maya:
Thanks @xtine! One thing I noticed now is our focus state does not appear. Is there any way to bring that back?
screen shot 2015-06-09 at 12 47 38 pm

maya:
What's better, leaving the anchor link empty or `<a href="#" onclick="return false;">`?

xtine:
Typically I try to avoid any inline JS as much as possible, so the empty link would be preferred.

maya:
The last commits:
- Adds in some accessibility content thanks to @nickbristow. 
- Adds a `sr-only` class to hide content from being visible but still accessible for screen readers. This uses the "Positioned, Clipped, and (almost) Collapsed" method found here which combines a few techniques:
http://snook.ca/archives/html_and_css/hiding-content-for-accessibility

hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -930,4 +930,52 @@
...
+    if (!$clicked.parents().hasClass("dropdown")) $(".dropdown dd ul").hide();
+  });
+
+
+  $('.mutliSelect input[type="checkbox"]').on('click', function () {
+
+      var title = $(this).next().html();
+
+      if ($(this).is(':checked')) {
+        var html = '<span title="' + title + '">' + title + '</span>';
```
theresaanna: Can we define `html` and `ret` at the top of the function?

theresaanna:
This is a WIP until I write unit tests and get the API documentation up to date. It'll also need review from Alison.

hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -930,4 +954,74 @@
...
+    if (!$clicked.parents().hasClass("dropdown")) $(".dropdown dd ul").hide();
+  });
+
+
+  $('.multiSelect input[type="checkbox"]').on('click', function () {
+
+      var title = $(this).next().html();
+
+      if ($(this).is(':checked')) {
+        var html = '<span title="' + title + '">' + title + '</span>';
```
theresaanna: Its my understanding that its good practice to define all vars at the top of a function. @xtine, what do you think about moving `html` and `ret` to be defined at the top?

hourglass_site/templates/index.html
```diff
@@ -194,15 +236,15 @@ <h5 class="filter-title">Optional Filters</h5>
           <div class="highlights-container row">
             <div class="four columns">
               <p>&nbsp</p>
```
arowla: Not part of this changeset, but shouldn't this `&nbsp` be followed by a semicolon?

hourglass_site/static/hourglass_site/js/index.js
arowla: :+1:

arowla:
Aside from the couple line notes, LGTM.

hourglass_site/templates/index.html
theresaanna: I don't see this in the most recent version of the template. I believe this has been removed.

PR #175, opened 2015-06-15
Filter updates

xtine:
Implemented a range slider for years of experience filter. Hooked into API. Only issue so far is that the "experience_range" input doesn't seem to take non-range values (eg: [1,1] will return the entire range rather than just 1 years experience).


maya:
Seems like there are two ways to do a slider: inputs + range (HTML5). I think either could work. This is using noUiSlider right? If so, can we add in reading and writing from an input: http://www.jqueryrain.com/?zUltWnMq

This will make it accessible.

xtine:
Thought of something that might make more sense and still would be accessible: select dropdown inputs for the years of experience range. Would be linked to slider so user could use slider or dropdown to select the year range and either would update accordingly.

maya:
That is an interesting idea @xtine! Never would have thought of that, also I've haven't seen that pattern before, but yes, it's accessible. 

I'm posting screenshots to show what it looks like:
screen shot 2015-06-16 at 8 25 24 am
screen shot 2015-06-16 at 8 25 07 am

maya:
When I enter the full range 0-25 years, it returns 42,346 results, but on the live site we have 48,666 results when we have an "All" value for the years. Are some results not being shown?

xtine:
I noticed this issue and then I figured out because there are listings with over 25 years experience, up to 45 (sort ascending to see). So limiting the search to 0 - 25 yields fewer results than all.

Now is the issue if we just add more years to the years experience range.

maya:
Ok, so should we set the upper limit to 45 years then?

theresaanna:
@xtine it looks like the test failures are real. To run tests locally, install PhantomJS and then run `make test` from the root of your repo. If you need a hand with any of this, lmk.

maya:
@xtine Just got the :+1: from @nickbristow on adding the select input! 

Two things..
- Can we add in a focus state for the slider button. We can just match what we use on the inputs.
- Can we start the initial range 0-45 years so we displaying everything to start?

maya:
Also @nickbristow suggests we add this in for accessibility.

```
<fieldset>
<legend>Experience:</legend>
...div stuff
<select id="minexperience"></select>
<label for="minexperience" class="sr-only">Minimum Years</label>
-
<select id="maxexperience"></select>
<label for="maxexperience" class="sr-only">Maximum Years</label>
years
</fieldset>
```

xtine:
@maya on drag the slider handles have a blue outline now.

maya:
Awesome, thanks @xtine!

theresaanna:
@xtine Looks like there are a few conflicts between this branch and master. Would you mind taking a look at those?

Issue #176, opened 2015-06-15
Design: chart callouts (avg, standard deviation, proposed $)


brethauer:
![image](https://cloud.githubusercontent.com/assets/7635133/8166884/9a9a5c7a-1368-11e5-91ad-da76ecdd1d98.png)

brethauer:
This is a test of how we can use GH issues for design discussions.

brethauer:
How can we differentiate the average from the standard deviation?  They run together a little / have equal weight.  Sort of same w/ proposed price.  Almost makes we wonder if in-chart labeling is the way to go if our callouts are getting too complicated?

maya:
We can try different font sizes like this: 
![standard deviation mockup](https://cloud.githubusercontent.com/assets/5249443/8167150/2cce3cb4-1351-11e5-857d-fc73111caa7a.jpg)

maya:
I think this is a good thing to A/B test @brethauer, but if you need any other versions before, let me know!

PR #178, opened 2015-06-18
[WIP] Add standard deviation callout

maya:
This adds a standard deviation callout below the graph. It also moves the two download buttons together below the callout. This does work around: https://github.com/18F/calc/issues/176.

Here is what it looks like:
screen shot 2015-06-17 at 10 52 32 pm

Great work on the functionality by @theresaanna!


arowla:
LGTM! Just to be sure, since this is marked [WIP], this is ready to merge?

PR #181, opened 2015-06-19
proposed price input

xtine:
Adding Proposed Price input and price callout.

### DIFF
--- a/hourglass_site/static/hourglass_site/js/index.js
+++ b/hourglass_site/static/hourglass_site/js/index.js
@@ -542,7 +542,7 @@
     })
     .html(function(d) {
       var id = d.string.split('-').join('');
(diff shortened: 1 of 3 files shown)


theresaanna:
Looks good from my side. @maya do you want to take a look?

maya:
Looking good @xtine! Is it possible to display the callout when a user hits "enter" too?

One minor thing, can we drop the proposed price callout to its own line when the viewport gets small enough that it gets covered by the others?

maya:
Here is a screenshot so others can see:

screen shot 2015-06-19 at 5 00 30 pm

xtine:
@maya got the enter submission and media query working

maya:
:+1: this looks good to me @xtine!

PR #192, opened 2015-06-24
Fixing Selenium tests

theresaanna:
It seems as though this changeset prevents the test timeouts we've been seeing, but the broken pipe errors may need to wait a bit longer.


theresaanna:
@xtine Would you mind reviewing?

xtine:
I still get some exceptions, but it doesn't take forever to have the tests run and pass. LGTM.

theresaanna:
Yeah, its progress! But we're not entirely there yet. I asked Shawn about this because I can see in the code where he was having the same problem with the broken pipe errors, but its a mystery to both of us. 

Please have the merging honors, if you'd be so kind.

PR #195, opened 2015-06-25
Updating Cloud Foundry buildpack

theresaanna:
The buildpack has changed and the Python version we use in CF needs to change along with it.


theresaanna:
@khandelwal reviewed.

khandelwal:
Yup.

Issue #198, opened 2015-06-25
When there is no data, standard deviation callouts display "$NaN"


theresaanna:
We should probably make it say "unavailable" or something? @maya what do you think?

cc: @xtine 
![screen shot 2015-06-24 at 12 49 24 pm](https://cloud.githubusercontent.com/assets/126935/8364206/478e021a-1b39-11e5-8804-9dbe195156b0.png)

PR #199, opened 2015-06-25
display $0 instead of NaN when no results for standard deviation

xtine:
regarding #198 ... for now it just returns $0 instead of NaN

### DIFF
--- a/hourglass_site/static/hourglass_site/js/index.js
+++ b/hourglass_site/static/hourglass_site/js/index.js
@@ -324,7 +324,8 @@
           .attr("preserveAspectRatio", "xMinYMid meet"),
         formatDollars = function(n) {
           return "$" + formatPrice(n);
-        };
+        },
+        stdMinus, stdPlus;
 
     var extent = [data.minimum, data.maximum],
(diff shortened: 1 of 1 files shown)


hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -337,14 +338,32 @@
...
     console.log('count extent:', countExtent);
 
+
     d3.select("#avg-price-highlight")
       .text(formatDollars(data.average));
 
+    stdMinus = data.average - data.first_standard_deviation;
+    stdPlus = data.average + data.first_standard_deviation;
+
+    if(isNaN(stdMinus)) {
```
theresaanna: How would you feel about using a ternary statement to make the assignment of `stdMinus` and `stdPlus` more concise?
`stdMinus = isNaN(stdMinus) : "$0" ? formatDollars(stdMinus);`

theresaanna: @xtine because this is a stylistic comment and not a dealbreaker, I'm going to merge.

PR #205, opened 2015-06-29
Strip non-alpha numeric or whitespace chars out of search query

theresaanna:
This morning, in production, we had an issue where requests were timing out. After consulting New Relic, it looks like the app errored when a user submitted a couple search terms that included parentheses. 

This pull requests makes it so that, when users search using the default setting, "Contains words", non-query-friendly characters are stripped out of the query, leaving just alphanumeric characters and whitespaces.

### DIFF
--- a/api/tests.py
+++ b/api/tests.py
@@ -63,6 +63,13 @@ class ContractsTest(TestCase):
         self.assertEqual(resp.status_code, 200)
         self.assertEqual(resp.data['results'], [])
 
+    def test_search_results_with_nonalphanumeric(self):
+        # the search should be able to handle queries with non-alphanumeric chars without erroring
+        self.make_test_set()
+        resp = self.c.get(self.path, {'q': 'category (ABC)"^$#@!&*'})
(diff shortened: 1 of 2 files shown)


cmc333333:
This tests that things don't explode, which is good. It'd be a little better if there was an explicit test on the transformation in `convert_to_tsquery` (i.e. that those chars were stripped). Important PR, though, so I'll merge after travis

theresaanna:
@cmc333333 Agreed. I wanted to think it through more before deviating from the patterns established in the tests. 

I'm looking at the need to use `convert_to_tsquery` and `raw=True` when performing the search, and, from my testing and the documentation you so nicely found, it looks as though we may not need to use either. That paired with the knowledge that our search functionality is changing means I did indeed just want to get this PR through so that prod doesn't choke again. 

Good suggestion.

cmc333333:
Not sure what's up with travis, but the push branch should be the same as the PR. So 1/2 works for me.

PR #208, opened 2015-06-30
Rework the method by which raw queries are formatted

theresaanna:
Looking at the errors users encountered that have been logged in New Relic, I noticed that they all had double spaces in the search term, which caused `convert_to_tsquery` to format queries like `senior & & engineer:*`, which causes errors as it is invalid syntax.

I rewrote `convert_to_tsquery` to address this. I also added prefix matching to each chunk of the query. So, for example, if someone searches "staffing consultant" with default search settings, they now get results for categories like "staff consultant". Previously, only the last term would have prefix matching applied.

### DIFF
--- a/api/tests.py
+++ b/api/tests.py
@@ -75,6 +75,7 @@ class ContractsTest(TestCase):
         self.make_test_set()
         resp = self.c.get(self.path, {'q': 'category (ABC)"^$#@!&*'})
         self.assertEqual(resp.status_code, 200)
+        self.assertEqual(resp.data['results'], [])
 
     def test_search_results_with_extra_spaces(self):
         # the search should insert the correct number of ampersands in the right locations
--- a/api/tests.py
+++ b/api/tests.py
@@ -3,6 +3,8 @@ from model_mommy import mommy
 from model_mommy.recipe import seq
 from contracts.models import Contract
 from contracts.mommy_recipes import get_contract_recipe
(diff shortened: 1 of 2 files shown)


api/tests.py
```diff
@@ -68,7 +75,12 @@ def test_search_results_with_nonalphanumeric(self):
         self.make_test_set()
         resp = self.c.get(self.path, {'q': 'category (ABC)"^$#@!&*'})
         self.assertEqual(resp.status_code, 200)
-        self.assertEqual(resp.data['results'], [])
```
arowla: Should this line be removed?

arowla: To elaborate, I guess it may not be strictly necessary, but it seems like there's some value in testing that no results are returned when you're not expecting any results from the query.

arowla:
Aside from the line note(s), LGTM. :+1:

api/tests.py
theresaanna: Yeah, I left it in for that reason, but thinking about it now, there is a search miss test. I'll remove it. Thanks!

arowla: By remove it do you mean put it back in? Sorry, talking about negating a negative is confusing.

theresaanna: I'm sorry, I'm foggy brained this morning. I thought the entire situation and your line notes were the opposite. Adding back in instead.

Issue #215, opened 2015-07-07
Limit proposed price decimal places on histogram update


xtine:
The histogram update function currently doesn't limit decimal places (to 2) on proposed price input.

Issue #214, opened 2015-07-07
Support proposed price decimal places on histogram


xtine:
Currently histogram rounds any decimal places for proposed price on the histogram.

PR #221, opened 2015-07-09
proposed price with decimal support in histogram

xtine:
Addresses both #214 and #215

### DIFF
--- a/hourglass_site/static/hourglass_site/js/index.js
+++ b/hourglass_site/static/hourglass_site/js/index.js
@@ -57,6 +57,20 @@
       return true;
   }
 
+  function getSigFigs(num) {
+    if (!isFinite(Number(num))) {
+      return -1;
+    }
+    var n = String(num).trim(),
+    FIND_FRONT_ZEROS_SIGN_DOT_EXP = /^[\D0]+|\.|([e][^e]+)$/g,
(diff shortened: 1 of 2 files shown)


hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -57,6 +57,20 @@
       return true;
   }
 
+  function getSigFigs(num) {
```
theresaanna: @xtine Would you mind commenting this function a bit? Its not immediately clear how it does what it does, particularly given the reg exes.

hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -1298,6 +1321,12 @@ if(getUrlParameterByName('proposed-price').length) {
     }
   });
 
+ $('.two-decimal-places').keyup(function(){
+    if(!(/^\d+(\.{0,1}\d{0,2})?$/.test(this.value))){
```
theresaanna: @xtine Can we also comment this regex?

xtine:
Thanks for the catch @theresaanna. Cleaned up the code and removed an overly complex function for what could have been done in a cleaner one line. Also added comment about the two decimal place regex functionality.

PR #223, opened 2015-07-13
Update README.md

wslack:
updated README that CALC is in beta

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![Build Status](https://travis-ci.org/18F/calc.svg)](https://travis-ci.org/18F/calc)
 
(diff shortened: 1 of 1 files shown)


theresaanna:
Good call! This should definitely change. I'm not sure that we're in beta, though, so I talked to @brethauer and it seems like it'd make the most sense if we changed that line to "The tool is live at https://calc.gsa.gov."

If you'd like to make the change, go for it and I'll merge this. Otherwise, we can close this and I'll make the change. You have until Monday. :)

wslack:
Done! this is ready to merge @theresaanna

wslack:
Also the dashboard says Beta :)

theresaanna:
Oh, huh, I've never seen that. @brethauer, what do you think?

Thanks, Will!

brethauer:
Yep I think we should also update the dashboard w/ "live" status  @wslack @theresaanna

PR #224, opened 2015-07-13
Add ethnio screener

brethauer:


hourglass_site/templates/base.html
```diff
@@ -138,6 +138,7 @@ <h2 class="description-secondary">CALC lets you conduct market research on profe
       }
     </script>
     <![endif]>
-
+     <!-- Ethnio Activation Code -->
+    <script type="text/javascript" language="javascript" src="//ethn.io/88263.js" async="true" charset="utf-8"></script>
```
theresaanna: Protocol relative URLs (the lack of an `http:` or `https:` at the beginning of the URL) have become an anti-pattern. Paul Irish nicely summarizes at the top of the article here: http://www.paulirish.com/2010/the-protocol-relative-url/, which mentions the work of our very own @konklone, I should point out.

I'll check out this branch, change the URL and test.

PR #228, opened 2015-07-14
DB indexes and New Relic on dev

theresaanna:
This adds db indexes for the fields that the API filters on. Looking in New Relic, requests don't normally spend an inordinate amount of time in Postgres, but it could stand to be cut down.

This also adds New Relic monitoring to dev so that we can, in theory, compare these changes to what is in production.

This PR should be reviewed by @arowla.


newrelic.ini
```diff
@@ -140,6 +140,7 @@ thread_profiler.enabled = true
 #
 
 [newrelic:development]
+app_name = CALC (development)
```
arowla: Unless there's some other reason to include this here, I think I'd just let this get defined by `NEW_RELIC_APP_NAME` in the manifest, to avoid duplication and possible confusion.

arowla:
Besides the one line note, LGTM!

newrelic.ini
theresaanna: I can see that. My thinking was that, in the event that it wasn't defined in the manifest for whatever reason, it'd still be differentiated from production. But you're right, I'll fix. Thanks!

PR #229, opened 2015-07-14
Extend search

xtine:
Autocomplete on labor category search field, separated by commas. I'd like input on this functionality before moving onto making tokenized tags or not.


hourglass_site/static/hourglass_site/js/vendor/jquery.auto-complete.js
```diff
@@ -1,8 +1,8 @@
 /*
```
theresaanna: Should we extend this documentation to describe the modifications we've made?

hourglass_site/static/hourglass_site/js/vendor/jquery.auto-complete.js
```diff
@@ -61,33 +61,40 @@
...
+                    o.onSelect(e, v, item);
+                    that.sc.hide();
+                }
             });
 
-            that.blur(function(e){
-                that.last_val = that.val();
-                that.sc.hide();
+            that.on('blur.autocomplete', function(){
+                try { over_sb = $('.autocomplete-suggestions:hover').length; } catch(e){ over_sb = 0; } // IE7 fix :hover
```
theresaanna: We support IE9+. When users in IE8- come to the site, I believe they get a message instead of content. 

Do you think we should leave this in just the same? We've talked about supporting IE8 before but I can't see us stretching back to 7.

theresaanna:
Needs review from @arowla, if she's available.

hourglass_site/static/hourglass_site/js/vendor/jquery.auto-complete.js
xtine: Ah, I forgot to mention all I did was just upgrade the auto-complete library.

Issue #222, opened 2015-07-10
Design: A way to see out-year pricing data


brethauer:
Here's one idea I had.  Please comment/critique/upload your own sketch

![img_2673](https://cloud.githubusercontent.com/assets/7635133/8622891/c933cc96-26fb-11e5-8df6-4f3397463eb8.JPG)

maya:
@brethauer is -2 and +2 the max range of year data we'll have to show?

KellyBailey:
Yes it is.

brethauer:
Another idea to handle this would be a slider with "notches" living just above the histogram - see sketch
![img_2676](https://cloud.githubusercontent.com/assets/7635133/8628464/492f3ad2-2721-11e5-8e10-fad063894ff3.JPG)

KellyBailey:
I am not sure that we would need previous years data. Maybe we should exclude the -1 and -2 because people could accidentally be looking at old data. I can't think of reasons they would be looking backwards.

theresaanna:
Here is a little bit of a clunky, but maybe worth considering, solution.

The idea is that you'd click the button next to the price to slide down an expanded view with year-related data.

![img_20150710_143022](https://cloud.githubusercontent.com/assets/126935/8629500/b3557e6e-2710-11e5-9919-b868a6acb7c4.jpg)

maya:
This is meant to be a filter in the right filter sidebar.

This shows the -2 and +2 years:

![fullsizerender](https://cloud.githubusercontent.com/assets/5249443/8663143/6e6102e2-297c-11e5-98a2-ad494e42cbf1.jpg)

This shows the current and year-out for 2 years:

![fullsizerender 1](https://cloud.githubusercontent.com/assets/5249443/8663140/6abc5916-297c-11e5-83b3-bf7eae309733.jpg)

theresaanna:
How would the table change as this filter changed?

brethauer:
@theresaanna My thinking is that when you select a different year, the table would need to update w/ the new prices (and potentially other updates?).  Perhaps certain rows would even drop out if you were already looking at year 5 for that particular contract.

theresaanna:
:+1: Would the table header need to change or would we need to do something else to indicate that the prices are not current year prices? Also, should we add a column to the table to show the current contract year?

brethauer:
[...]

PR #231, opened 2015-07-16
Add year out filter

maya:
This adds year out pricing for current, 1 year out, and 2 years out.

As described in: #222

This is what it looks like:

:camera: 
screen shot 2015-07-17 at 11 22 13 am

:movie_camera: 

Back end notes (Theresa):
I chose to add two fields to the model - `next_year_price` and `second_year_price`. Their values are set in the `load_data` script. This prevents us from having to introduce logic to the API view in order to determine the correct rate based on the `contract_year` field, and then filter on those that had a valid rate for that year. It also sounds rather inefficient to me.
Additionally, the way `wage_field` is passed around, it looks like the original author had an idea for how this type of functionality might work, and I wanted to try to stay within their flow.


theresaanna:
Needs review by @arowla.

PR #232, opened 2015-07-17
Dynamic sentence (results description) refactor

xtine:


hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -1012,102 +1011,58 @@
...
+
+    // first count of results
+    d3.select('#description-count')
+      .text(formatCommas(res.results.length));
+
+    if(laborCategory) {
+      var laborEl = $(document.createElement('span')).addClass('filter');
+      filters.append(laborEl);
+      laborEl.append(
+        $(document.createElement('a')).addClass('focus-input').attr('href', '#').html(laborCategory)
```
theresaanna: This is a small thing, but is there any reason we shouldn't do `$('a')` instead of creating the element and then turning it into a jQuery object?

hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -1012,102 +1011,58 @@
...
-    // update all of the generic descriptive elements
-    desc.selectAll('[data-key]')
-      .datum(function() {
-        return this.getAttribute('data-key');
-      })
-      .text(function(key) {
-        return keys[key] || '';
-      });
+    // loop through optional filters for description setting
+    $.each(data, function(key, value) {
```
theresaanna: I wonder if there's a way that we can condense this loop. What would you think about creating an object where you could lookup the `.html()` value, class name, etc by key name? Something like:
```
lookupObject = {
  'education': {
    'label': "Education level: ",
    'class': 'education-levels',
    'value': $('.multiSel').html()
    }
  }
...
}
```
Then we could potentially remove the inner conditional structure and create and append all of the elements with the same code. Then, the outer conditional can say something like `if lookupObject.hasOwnProperty(key)` to skip over all those but the ones we need to deal with.

Does that make sense?

hourglass_site/static/hourglass_site/js/index.js
xtine: if you are thinking of
`$('a').addClass('focus-input').attr('href', '#').html(laborCategory)`, it doesn't work.

`$('a')` selects all <a> elements on the dom.

The options for append() is to write out the actual HTML like `<a class="focus-input" href="..."` or to create an element inside like I did. Not that we are overly concerned about JS speed, but also creating an element is the fastest way to render it as well.

theresaanna: :+1: well-explained. Clearly I am rusty! Thank you.

xtine:
Refactored updateDescription function per @theresaanna good suggestions.

Issue #235, opened 2015-07-22
Scrolling down on multi-search removes 1st search term


maya:
After searching and selecting a first term and then search for a second term, when i scroll down and click enter to select it, the search changes to only the last term and the first is removed.

See here:
![](http://g.recordit.co/CmGHwiyOnO.gif)

Issue #233, opened 2015-07-21
Commas causing errors


maya:
After I enter a search term and click enter commas are inserted and cause internal server errors.

theresaanna:
The search technically works now, but there is the issue of user experience around the commas. @maya do you want to document that here?

maya:
Sometimes after I hit "enter", new commas appear: 

<img width="662" alt="screen shot 2015-07-22 at 8 13 57 am" src="https://cloud.githubusercontent.com/assets/5249443/8828932/6d6f941a-3049-11e5-9136-9e9813809a12.png">

Issue #236, opened 2015-07-22
Removing search terms does not remove them from results


maya:
After i select a few and then remove them and click enter or search, they are still in the search.

Issue #234, opened 2015-07-21
Autocomplete not always working


maya:
Sometimes autocomplete stops working after I enter the second search term. Sometimes it only shows after the 3rd letter typed but not after.

PR #238, opened 2015-07-28
multi-search updates

xtine:
- removed input delay for faster response
- edited autocomplete library to have better handling of keyboard selections (can now remove search results)

@maya : can you check out if you still encounter the same bugs you found before?


xtine:
Resolves: #235 , #236 and hopefully #234

maya:
LGTM! I think this also resolves #233.

hourglass_site/static/hourglass_site/js/index.js
```diff
@@ -160,7 +161,8 @@
           '</div>'
         ].join("");
       },
-      onSelect : function (e, term, item) {
+      onSelect : function (e, term, item, autocompleteSuggestion) {
```
theresaanna: Do you think it might read more clearly if we used the conditionals to set a var to the desired value and then update at the end like `$("$labor_category").val(theValue)`? That way, the else clause could be the default, like `var theValue = $("#labor_category").val() + ", "` and we'd only need two conditional checks.

Also, it may make sense to `var $laborCategory = $("#labor_category")`.

What do you think? Not a huge deal, just a suggestion.

hourglass_site/templates/index.html
```diff
@@ -452,7 +452,7 @@ <h5 class="proposed-price-highlight"></h5>
 </form>
 
 <![if gt IE 8]>
-<script src="{% static 'hourglass_site/js/vendor/jquery.auto-complete.min.js' %}"></script>
```
theresaanna: What do you think about minifying our version?

xtine: When I put the auto-complete.min file back to avoid static file server issues, I minified the version I modified, so once issues are clear I'll minify again and using that instead.

hourglass_site/static/hourglass_site/js/index.js
xtine: Good suggestions, let me clean it up a bit.

PR #243, opened 2015-08-07
Update .about.yml

mbland:
Now that https://team-api.18f.gov/ is about ready for production, I've run the `scripts/about/validate.rb` script from 18F/team-api#84 to ensure this file is now up-to-spec (given the state of the spec today).

cc: @brethauer

### DIFF
--- a/.about.yml
+++ b/.about.yml
@@ -1,7 +1,6 @@
 ---
-name: CALC
+name: calc
 full_name: CALC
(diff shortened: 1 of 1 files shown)


mbland:
Pretty sure the Travis failure has nothing to do with this change...

brethauer:
@mbland yep don't worry about it - thanks for this!

PR #245, opened 2015-08-17
Fix focus state

maya:
This adds a fix for the focus state so the text doesn't shift when you trigger the focus state. It also makes the focus state consistent across all inputs.

This is what it looks like:
screen shot 2015-08-17 at 9 44 59 am


theresaanna:
Test failures have nothing to do with this change set, so I'm merging.

Issue #255, opened 2015-10-24
Load Schedule 70 data into CALC. >>> Current bid: 1 <<<


vzvenyach:
As [previously explained](https://18f.gsa.gov/2015/10/13/open-source-micropurchasing/), this issue has been tagged as a "micro-purchase" issue. The opening "bid" for this issue is $3,499, and vendors may offer lower bids. The vendor with the lowest bid at closing time will have 10 working days to provide code that satisfies the requirements and acceptance criteria below. Upon successful completion, the vendor will be paid. If the vendor fails to complete the requirements and satisfy the acceptance criteria within the 10 working day period, then the vendor with the next lowest bid shall have 10 working days to provide the code. If the acceptance criteria are not satisfied, vendors shall not receive payment pursuant to this competition.

## Description of issue

Currently, CALC uses a [management command to load data](https://github.com/18F/calc/blob/master/contracts/management/commands/load_data.py) from 7 different schedules. We want to load the Schedule 70 data provided into CALC (note, the data will likely need to be reformatted). The data is available here: https://github.com/18F/calc/tree/s70/contracts/docs/s70

## Acceptance criteria

1. Either a new Django management command or modification to https://github.com/18F/calc/blob/master/contracts/management/commands/load_data.py to load the Schedule 70 as Contract objects in the database.
2. Whether a new command or a modification to the existing command, the rating from https://codeclimate.com/github/18F/calc for the file must be at a B or better rating, and test coverage for the command at 90% or better upon completion. 
3. Front-end changes to ensure the ability to filter results by schedule type (e.g. Schedule 70 data or others) and easily see the schedule source for each row in the results.

## Bid submission deadline

Bids for this issue close at 12 p.m. EST on Thursday, Oct. 29, 2015. To bid on the issue, click here: https://docs.google.com/a/gsa.gov/forms/d/1eRFX0hSTTXMc2FulK6kPP2P02ZApQqlYZL7oVDghJJo/viewform

vzvenyach:
QUESTION: How many times is a company allowed to submit a bid on this form?
ANSWER: Multiple bids are permissible.

pavellishin:
> QUESTION: How many times is a company allowed to submit a bid on this form?
> ANSWER: Multiple bids are permissible.

So wouldn't it make sense for a company to enter three low bids, varying by a penny, and thus get 30 working days to get a fix in?

Connery-LCSolutions:
[...]

PR #256, opened 2015-11-05
Load Schedule 70 data into CALC

brendansudol:
This PR addresses [issue #255](https://github.com/18F/calc/issues/255), the "micro-purchase" experiment. This code adds Schedule 70 data support to CALC.

A summary of the additions & changes:

#### New management command: `load_s70`

We chose to add a new management command to load the Schedule 70 contracts, leaving the existing `load_data` command as is to reduce the number of changes required.

Standard usage is simply:

`./manage.py load_s70`

This replaces all existing Schedule 70 contracts with new ones from the default dataset, located at `contracts/docs/s70/s70_data.csv`.

There are also a few options to the new command:

* `-f --filename`: specify an alternate filename to load.
* `-a, --append`: this controls whether to append or replace data; the default is to replace.
* `-s, --strict`: this controls whether to abort the load if any input data fails validation; the default is to simply skip (and count) the bad rows.

For more information, run `./manage.py load_s70 --help`

#### Loader behavior

The new loader attempts to behave in a similar way to the existing `load_data` command, specifically:

* Input rows without price data are skipped and not loaded.
* Labor categories are cleaned up, e.g. by replacing `'\n'` with `' '`.
* Rows with no specified minimum years experience specified are treated as requiring zero.
* Rows with no specified contract year are assumed to be in their first year.
* Display prices are not set when row prices are below the federal minimum contract rate.

Any contracts in the input file that are not of schedule type `"IT Schedule 70"` are also skipped. In addition, the single price provided in the file is assumed to be the price for all years of the contract.

#### Tests

New unit tests for this command and all related methods can be found in `contracts/tests/test_load_s70.py`.

They can be invoked on their own with `./manage.py test contracts` or with the rest of the test suite with `make test-backend` or `make test`.

All new code is [PEP 8](https://www.python.org/dev/peps/pep-0008/) compliant.

#### Other changes

Besides the load command and related loader class, changes also include:

* Adding `IT Schedule 70` as an option to the Schedule dropdown on the CALC interface  (`hourglass_site/templates/index.html`).
* Adding a paragraph about IT Schedule 70 to the about page (`hourglass_site/templates/about.html`).
* Adding documentation about the new dataset to `updating_data.md` and `README.md`


contracts/management/commands/load_s70.py
```diff
@@ -0,0 +1,184 @@
...
+            idv_piid=row[6],
+            contract_start=cls.parse_date(row[12]),
+            contract_end=cls.parse_date(row[13]),
+            contract_year=cls.int_or_fallback(row[11], 1),
+            vendor_name=row[7],
+            labor_category=row[1].strip().replace('\n', ' '),
+            education_level=cls.model.get_education_code(row[2]),
+            min_years_experience=cls.int_or_fallback(row[3]),
+            hourly_rate_year1=price,
+            hourly_rate_year2=price,
```
theresaanna: @brendansudol Using the site, when I filter by Schedule 70 and then select next or second year pricing, I see the same result set as year one with the same prices. 

What do you think about keeping `hourly_rate_year` 2-5 empty, along with `next_year_price` and `second_year_price`? This is how we handled records in the other schedules for which we didn't have full pricing data.

brendansudol: Hey @theresaanna, thanks for the note. I went back and forth on how to handle this. I wasn't sure because for some entries, the `contract_year` was > 1 and so I didn't know if I should leave the `hourly_rate_year` 2-5 empty in those cases (even though we only have one price field for this dataset) and thus defaulted to filling them all in. But I think your solution is better. I will change!

brendansudol:
Hmm, tests passed locally but it looks like something hung on Travis CI...

brendansudol:
Hmm, tests passed locally after my latest commit, but it looks they hung on Travis CI (It says "No output has been received in the last 10 minutes"). Not sure if you all have seen this before.

vzvenyach:
@brendansudol it's happened to me before. travis lets you restart the script from the dashboard. tried it and looks like it passed.

brendansudol:
Ah, cool -- thanks @vzvenyach!

theresaanna:
@vzvenyach This gets my :+1: We will need to talk about deployment next week.

theresaanna:
@brendansudol Apologies for the test issues. Our Selenium tests hang sometimes for a reason that no one has yet figured out. :/

brendansudol:
No worries :smile: thanks for the code review @theresaanna!

theresaanna:
@brendansudol Very welcome!. This is going to make it possible for a bunch of folks to do their jobs more effectively and save taxpayers $$! Your effort will be very much appreciated by the users.

Scotchester:
@18F Sending much love from your @cfpb colleagues for the great experiment!

You rock, too, @brendansudol!

theresaanna:
Thanks, @Scotchester! :wave:

PR #259, opened 2016-01-14
Update .about.yml

melodykramer:
>updated description
> updated impact
> rewrote milestones to be complete sentences
> add blog tags


mbland:
Looks like those builds hiccuped due to some Travis issues; I had a similar timeout happen recently on another project. I restarted the builds, and they've passed now.

cc: @brethauer.

PR #261, opened 2016-01-25
Extending about.yml services

DavidEBest:
Making the services entry match the services listed in the README.


jseppi:
Closing since `.about.yml` has been updated in #352

PR #265, opened 2016-06-07
Add initial docker support.

toolness:
This is based largely off of https://github.com/18F/projects/pull/3.

To do:

- [x] Add instructions to readme.
- [x] ~~Figure out how to make the selenium/phantomjs tests succeed. Not sure if this is actually a docker issue or not, as there seem to have been [issues w/ selenium in the past](https://github.com/18F/calc/pull/256#issuecomment-154494713).~~ Confirmed from @jseppi that there are problems w/ selenium tests independent of docker, so we'll punt on this for now.


manage.py
```diff
@@ -2,9 +2,93 @@
...
+            print("Attempting to connect to database.")
+
+    print("Connection to database established.")
+
+
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "hourglass.settings")
 
-    from django.core.management import execute_from_command_line
+    if os.getcwd() == '/calc':
```
jseppi: It is probably very unlikely to ever be an issue, but it seems like checking that this is running in the `/calc` path leaves some potential for problems depending on how the app is deployed.

jseppi: Perhaps another way to do it would be to set an environment variable in the Dockerfile like `ENV IS_DOCKER 1` and then just check for the existence of that env var.

toolness: Good call, will do!

toolness:
Ok, I've added a `IS_RUNNING_IN_DOCKER` env var and used it to make a few things more robust!

hourglass/settings.py
```diff
@@ -179,9 +179,10 @@
...
-    pass
-
-
+if 'IS_RUNNING_IN_DOCKER' in os.environ:
+    from hourglass.docker_settings import *
+else:
+    try:
+        from hourglass.local_settings import *
+    except ImportError:
+        pass
```
toolness: Sooo I ended up just ditching `local_settings` if we're running in docker. My thinking is that it's probably better to migrate this app's configuration to be as [12 factor](http://12factor.net/) as possible anyways, so any additional configuration settings we want to control can be documented as environment variables in the `README` and set via docker-compose's `.env` file?

jseppi:
👍 Looks good to me.
I like this little trick in manage.py: https://github.com/18F/calc/pull/265/files#diff-632b651e5579af0e928ed63482e9bc77R81

toolness:
Thanks! I will merge now.

Issue #266, opened 2016-06-08
Fix selenium/WebDriver issues


toolness:
This occurs in Travis CI on https://github.com/18F/calc/pull/256#issuecomment-154494713 and #265.

I *thought* it was consistently reproducible locally, but apparently it's not--sometimes running `manage.py test` takes a little while to run all the integration tests (about 9 seconds) but they all succeed.  Other times it appears to time out, while other times it stalls but continues with only 1 failure, while other times it runs with lots of failures. I guess there might be a race condition or something somewhere?

A current workaround for this is to do `make test-backend` (or `manage.py test api contracts`) so that only the non-selenium tests run.

toolness:
Ok, even though the solution in #267 isn't perfect, I think it's good enough to close this issue. Will re-open if we start getting random failures again though.

PR #267, opened 2016-06-08
Attempt to make selenium tests more reliable

toolness:
This is an attempt to help with #266, though I don't feel comfortable saying it actually *fixes* things. It just deals with the symptoms in a more fault-tolerant way.

Basically, it's hard to diagnose what's going wrong in #266. There's a lot of different technologies involved here: a python webdriver client is talking to a ghostdriver-based remote webdriver server embedded in a headless webkit browser, which is talking to a django-based web server. There's a high degree of concurrency, and very little visibility into each component's internal state.

Google searches haven't helped much.

The only thing I *did* manage to determine was that [`selenium.webdriver.remote.webdriver.WebDriver.get()`](http://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webdriver.WebDriver.get) was hanging.

I added a workaround so that we basically provide a fairly brief (3-second) socket timeout to ensure that `get()` throws a `socket.timeout` exception instead of hanging. If a timeout occurs, instead of immediately giving up, we retry the `get()` up to 10 times.

Tests always pass after re-running the test suite many times; `get()` only times out at most 3 or 4 times every few runs.

I still don't know what's causing the root problem, though, which bugs me. My guess is that it's due to https://github.com/detro/ghostdriver/issues/334, because I've used PhantomJS from the command-line a lot and it's never given me problems, and Django's `LiveServerTestCase` seems pretty reliable too. In fact, all the technologies in the pipeline are pretty well-maintained, except for GhostDriver, which has been flagging [according to its README](https://github.com/detro/ghostdriver#help-needed). I could easily be wrong though.

### DIFF
--- a/selenium_tests/tests.py
+++ b/selenium_tests/tests.py
@@ -29,11 +29,15 @@ from itertools import cycle
 import re
 import time
 import os
+import socket
 from datetime import datetime
 
 TESTING_KEY = 'REMOTE_TESTING'
 REMOTE_TESTING = hasattr(settings, TESTING_KEY) and getattr(settings, TESTING_KEY) or {}
 TESTING_URL = os.environ.get('LOCAL_TUNNEL_URL', REMOTE_TESTING.get('url'))
+
+PHANTOMJS_TIMEOUT = 3
+WEBDRIVER_TIMEOUT_LOAD_ATTEMPTS = 10
 
 def _get_testing_config(key, default=None):
     return REMOTE_TESTING.get(key, os.environ.get('%s_%s' % (TESTING_KEY, key.upper()), default))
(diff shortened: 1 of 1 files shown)


toolness:
Yay, travis didn't hang!

:tada: :dancer:

jseppi:
👍 Can confirm that this worked locally (I think, still lots of exceptions  thrown during `make test-frontend`, but maybe they're expected?)

toolness:
Correct--I wish there was an easier way to make that evident to the person running the tests... I believe it's Django's live test server thread reporting the aborted connections that are occurring when our webdriver code is telling GhostDriver/Phantom to forget about the last request (which is timing out) and  to start a new one.

Issue #12, opened 2014-12-17
Consider django-pipeline for assets


shawnbot:
http://django-pipeline.readthedocs.org/

Minifying and concatenating JavaScript can really speed up front-end loads. I'm doing Sass compilation as I work on things right now (`cd hourglass_site/static/hourglass_site; make watch-style`), but it might also be good for the compilation to happen transparently.

Developing with compiled assets can be a major PITA, but [source maps](http://www.html5rocks.com/en/tutorials/developertools/sourcemaps/) might help there. Sass creates them alongside the compiled CSS now.

toolness:
I'm getting the impression that we will probably just want to replace WhiteNoise with this. While some folks have had experience with [combining the two](https://github.com/mozilla/playdoh/issues/173#issuecomment-440440560), I'm worried about incompatibilities down the line and there don't seem to be many people combining the two anyways.

toolness:
One concern I have about using this package is how it will affect our deploys.

For example, it looks as though its SASS support relies on the [`sass` binary being present on the host system](https://django-pipeline.readthedocs.io/en/latest/compilers.html#sass-compiler), rather than using [python's `libsass` binding](https://pypi.python.org/pypi/libsass).  This implies that we'd probably have to make the deployer's machine (e.g. travis, 18F employee's laptop) responsible for running `manage.py collectstatic`, rather than cloud.gov.

That's not to say we shouldn't use this package, of course--in fact we will probably eventually have to build the assets on the deployer's machine if we want to use any of the major JS-bundling tools out there, too.

toolness:
Hey, we decided to go with gulp for asset compilation in #367, so I'm closing this!

PR #268, opened 2016-06-09
Add sass integration to docker

toolness:
I noticed #12 and figured that while integrating django-pipeline would take a little while, simply adding sass integration to our docker setup wouldn't be hard.

However, after adding the docker integration, I discovered that CALC doesn't actually seem to *use* the `screen.css` file being built by sass.

I verified this with `grep -r screen\.css *` from the root of the repository.

@shawnbot I know it's been a really long time since you filed #12, but do you have any idea whether any of the SASS in this project is actually being used?  If it's not, I guess we should replace this PR with one that removes `screen.css` and `hourglass_site/static/hourglass_site/style/sass` from the repo to avoid confusion...


toolness:
Ack. This PR is against the `master` branch because it was made before we moved development to `develop`, so I'm going to close it and re-open a new one.

PR #271, opened 2016-06-09
Show warning if user runs 'manage.py runserver' manually.

toolness:
Because of the eccentricities of docker/docker-compose, running `python manage.py runserver` usually won't do what the developer expects (the server will run, but it won't be accessible on the docker host).  So here we'll detect to see if the user is doing that and display a warning if so.

The warning will *not* show up if the user runs `docker-compose up`.


jseppi:
👍

Issue #213, opened 2015-07-07
Missing image in static when switching between branches


theresaanna:
Sometimes, when switching branches on your dev environment, you get a message that an asset in the static dir isn't available. Running collectstatic doesn't seem to work, you have to `touch` each file. 

I suspect that its something to do with the white noise package?

toolness:
Hmm, I'm running into this now too.

jseppi:
I think I also ran into this and did a `make clean`, which seemed to solve it

toolness:
Sorry about that, missed the "Comment" button, hit "Close and comment".

Anyways, I'm running into this now too. It's particularly weird because the file it's looking for is one that was removed in a recent commit.

It looks like we're using whitenoise 1.0.6, but the latest version is 3.2. They've also added some nice features that make using it with Django a lot easier. Upgrading *might* get rid of this weird problem.

toolness:
Ok, I did a bit of digging and it looks like--at least on my end--this is actually an issue with Django's [`ManifestStaticFilesStorage`](https://docs.djangoproject.com/en/1.9/ref/contrib/staticfiles/#manifeststaticfilesstorage) backend, which white noise inherits from.

As its docs mention, you *must* run `manage.py collectstatic` whenever your static files have changed, so that the md5 hashes of all files can be recalculated and `staticfiles.json` can be rebuilt.

I was running into a problem where static assets couldn't be found while running tests, and `make clean` wasn't helping, but I ran `manage.py collectstatic` and it fixed things for me.

I think an alternative workaround might be to use `django.contrib.staticfiles.storage.StaticFilesStorage` whenever `DEBUG=True` (i.e., during development) but I'm not sure how much I like this from a dev/prod parity standpoint.

toolness:
Oh, I just noticed that `make test` and its related commands actually *do* run `collectstatic` before running the tests, so that explains why this kind of bug was never run into by the other devs; I was running `manage.py test` directly.

PR #272, opened 2016-06-10
Upgrade to whitenoise 3.2.

toolness:
This upgrades us to whitenoise 3.2.

I was hoping this would help with #213, but it doesn't really. Nonetheless, it seems like a good idea to upgrade to the latest version anyways, for bugfixes and such.


jseppi:
👍 that looks like much better django integration

toolness:
Woot thanks!!

PR #273, opened 2016-06-10
Factor out docker_django_management.py.

toolness:
At this point I've added Docker support to [calc](https://github.com/18F/calc/pull/265), [tock](https://github.com/18F/tock/pull/359), and [projects](https://github.com/18F/projects/pull/3). They all involve very similar code, since all three projects are Django-based.

Rather than copy-pasting a bunch of code and find-replacing some strings, I thought it might be useful to factor out all the common code into a single well-documented, self-contained module. This will make adding Docker support to any Django project easier.  It will also allow us to easily upgrade the module between different projects as further enhancements/bugfixes are made.

**Note:** I'd kinda like to follow Python best practices and make this an actual pip package, but the problem with that approach is that it'd require developers to set up and enter a virtualenv just so they could use this one module, which is kind of a pain, especially since a big benefit of using docker is to obviate the need for things like virtualenv.  So it seems easier to just have a self-contained, well-documented, versioned file that can be plopped next to a project's `manage.py`.


jseppi:
This looks pretty good to me. It again has the problem, however, that if you already have `local_settings.py` defined, then `docker-compose up` will hang because it will be trying to connect to the database defined there instead of in the container.

I think you just need to include your previous solution of importing `hourglass.docker_settings` when `IS_RUNNING_IN_DOCKER` is true.

Ref to previous solution https://github.com/18F/calc/blob/87014ce860bf3d2c55e53a2996df8418fd6e4c0f/hourglass/settings.py#L177

toolness:
Oh right, good catch!  I think I just fixed it.

jseppi:
👍 lgtm

PR #277, opened 2016-06-13
Specify package versions

jseppi:
I'm not totally sure about best practices in Python-land, but it seemed odd to me package versions weren't specified for several of CALC's dependencies.

I added minor version matching (ex: `>=1.1,<1.2`) for all the listed packages, and I also changed the `djorm-ext-pgfulltext` dependency to come from pypi instead of from a git repo.

@toolness: please review


toolness:
Hmm, I definitely agree that we should be specifying version numbers for the unversioned stuff, but I'm actually a bit conflicted about minor version matching. I've never actually seen it widespread in python-land, and I once saw a presentation at EmpireJS stressing the benefits of pinning on exact version numbers (I also once ran into a problem on a nodeJS project where a minor version update broke semver and therefore our whole build, but that's only happened to me once).

It looks like Heroku's [pip best practices](https://devcenter.heroku.com/articles/python-pip#the-basics) suggest that version numbers be explicitly specified... I'm curious what the folks on #python in Slack think.

jseppi:
I'd be fine with pinning to exact versions.

toolness:
Ok, cool, I just [asked on Slack](https://18f.slack.com/archives/python/p1465851842000030) and it looks like the general consensus is that for apps (vs. libraries) we should pin to exact version numbers.

jseppi:
Cool, I'll make that change.

toolness:
Awesome! Feel free to go ahead and merge once you make that change.

PR #278, opened 2016-06-13
Switch to py.test

jseppi:
Switching from `nose` to `py.test` since nose is currently unsupported and py.test seems like the current best option.
 
Still to-do:
- [x] modify the Makefile to use `py.test` instead of `manage.py test`
- ~~figure out the Selenium tests (?). They don't work locally for me ever.~~ They work.
- [x] figure out if anything needs to change for Travis test running

cc @toolness


jseppi:
Looks like Travis is running the tests a-ok. Ready for review, @toolness.

jseppi:
As a note, I tried to change as little as possible to get this running. To make that the case, I had to slightly modify how py.test discovers tests to bring it inline with how (I assume) `nose` did it, which is why I had to make some customizations to `python_files` and `python_classes` in `pytest.ini` (https://github.com/18F/calc/pull/278/files#diff-513c9b6861fc981fb2c708db477f542cR3)

toolness:
Thanks!

Hmm, so I'm noticing when running `make test` on the `develop` branch that the final output is:

```
Ran 69 tests in 11.996s
```

but when I run this PR, I get:

```
68 passed in 11.64 seconds
```

Any idea where that one test went?

(I wish there were an easy way to get a "diff" of the tests being run between the two different tools, but unfortunately I don't know of any way to do that.)

Also, when running `make test-frontend` I get the following:

```
python manage.py collectstatic --noinput --link > /dev/null
py.test selenium_tests \
        -x --noinput \
        --liveserver=localhost:8081 \

usage: py.test [options] [file_or_dir] [file_or_dir] [...]
py.test: error: unrecognized arguments: --noinput
  inifile: /calc/pytest.ini
  rootdir: /calc
Makefile:19: recipe for target 'test-frontend' failed
make: *** [test-frontend] Error 2
```

Removing the `--noinput` from the Makefile seems to make things work ok, but I'm not sure what `--noinput` actually did for nose, and whether we need some kind of analogous flag for py.test...

One thing I like about py.test is that it appears to squelch stdout/stderr *unless* the test actually fails, which (I think) means no more spam from the Django live server as mentioned in https://github.com/18F/calc/pull/267#issuecomment-224929157.

However, I noticed in your PR description you wrote:

> - ~~figure out the Selenium tests (?). They don't work locally for me ever.~~ They work intermittently, doesn't seem to have anything to do with the test runner.

I was hoping this was fixed in #267? Do those tests still sometimes fail on your end? Since #267 was merged, they haven't failed for me, though sometimes they have taken a while to run due to the weird timeout they're working around.

jseppi:
I think I hadn't merged #267 when I wrote that bullet point about the Selenium tests, then I merged in the `develop` branch to this one is when it started working. I've run it several times this morning and it works as expected each time. Sorry for the confusion.

I tracked down the "missing test" to `api/tests.py`. `manage.py test api/tests.py` (which uses nose) reports that there are 38 in that file, while `py.test api/tests.py` only counts 37. A simple count of `def test_` methods only gives 37 tests. It appears that nose is counting the helper method called `def make_test_set` as a test (probably because it has `test` in the name). I verified by changing its name and re-running the tests with nose, which then counted 37 tests, all passing. 
Good catch on noticing that difference in the number! I thought I had checked that the test counts were the same, but I missed it.

I removed `--noinput` from the `make test-frontend` task, and it works without issue. It appears to not have had an impact on those tests. I verified by removing `--noinput` from the task in the `develop` branch, and it still worked properly there as well.

toolness:
Yay awesome!!!  Thanks for fixing those, I think this is 👍 to merge.

PR #281, opened 2016-06-14
Remove Sauce Labs testing support

jseppi:
Remove support for Sauce Labs testing. We will need to revisit cross-browser testing, but right now this doesn't work and it relies on an unapproved tunnel module.

cc @toolness


toolness:
Thanks!

PR #283, opened 2016-06-15
Use SASS & refactor some CSS

hbillings:
Some highlights:
- `main.css` and `index.css` both should get written by SASS now. (I still need to make this happen as part of the `runserver` command, I think. **Update:** nah we can just deal with it via docker for now, until we fix it for good with #12.)
- There seem to be two JS libs that had styles in `index.css` originally (`tooltipster` and `noUi` -- latter is for the range slider). I factored these into their own CSS files in `vendor` and included them in `index.scss`. Maybe instead we should concat/minify all three of those as part of the build? **Update:** We can deal with this in #12, probably.
- Colors have been abstracted into `variables.scss`. Font sizes and other standards will go here eventually.

New 6/21: Deleted alllllll the unused sass


toolness:
Awesome thanks!

Do you have any suggestions on how best to QA this? I'm terrible at reviewing large CSS refactorings.

Also, once we merge this, #268 will actually be useful! Yay!

hbillings:
@toolness Ideally, you should be able to run it locally and have nothing look broken (I can't figure out a way to write that so it doesn't sound sarcastic, heh). Mostly I haven't changed the actual styles; I've just refactored them. I updated this issue's description with the major things I did and will do that again when I finish ripping stuff apart, so maybe it will be easier for you to look at!

hbillings:
Okay, this should be good to go!

jseppi:
Just to make sure (since I am not familiar with django's static file stuff), does it automatically do the sass-to-css conversion?

toolness:
Nice, this looks great! I'm also able to make trivial modifications to `main.scss` and `index.scss` and they show up fine in the generated CSS.

@jseppi Django doesn't do the sass-to-css conversion automatically by default; there are third-party libraries that integrate with [libsass](https://pypi.python.org/pypi/libsass) to make that possible, but for now it's probably easiest to just use ruby sass and commit the generated CSS into the repository as @hbillings has done in this PR.  We'll just want to let non-docker users know what to do in the README. (For Docker users, this will be taken care of transparently by #268.)

toolness:
Oh, I just realized that Shawn actually has already created an issue for making our asset pipeline more robust, in #12. This will take care of @jseppi's question about making SASS compilation completely transparent, and I *think* it will also take care of @hbillings' second bullet point about concatenating/minifying all our CSS (including noUi and tooltipster).

toolness:
Ok, this looks good to go--thanks for the PR @hbillings !

Issue #253, opened 2015-10-14
Delete old feature branches at https://github.com/18F/calc


mshenfield:
Hey! I was interested in contributing to this project. I appreciate the awesome service this provides.

There are several feature branches that don't seem to correspond to specific issues, or cards on the [Trello board](https://trello.com/b/LjXJaVbZ/calc-project-board), and haven't been updated in 3+ months. This makes it feel a little perilous for new users to make changes, and goes against one of the recommendations outlined in the [internal contribution guide](https://github.com/18F/calc/blob/master/CONTRIBUTING-internal.md#git-workflow).

You can run the command

```
for k in `git branch -r | perl -pe 's/^..(.*?)( ->.*)?$/\1/'`; do echo -e `git show --pretty=format:"%Cgreen%ci %Cblue%cr%Creset" $k -- | head -n 1`\\t$k; done | sort -r
```

to see the current list of remote branches and their last commit date ([thanks VonC](http://stackoverflow.com/questions/2514172/listing-each-branch-and-its-last-revisions-date-in-git)).

Here's the output when I run it:

```
2015-10-13 15:09:15 -0700 2 hours ago	origin/master
2015-10-13 15:09:15 -0700 2 hours ago	origin/HEAD
2015-09-25 11:59:29 -0700 3 weeks ago	origin/aug-data
2015-08-25 20:48:43 +0000 7 weeks ago	origin/selenium-fix
2015-08-17 17:16:01 +0000 8 weeks ago	origin/new-search-features
2015-08-04 11:32:09 -0700 2 months ago	origin/update-description
2015-07-31 19:10:33 +0000 2 months ago	origin/fix-radio-buttons
2015-07-30 20:23:22 +0000 3 months ago	origin/add-year-out
2015-07-16 19:47:29 +0000 3 months ago	origin/public-docs
2015-07-14 14:55:41 +0000 3 months ago	origin/indexes
2015-07-13 19:54:41 +0000 3 months ago	origin/graph-img
2015-07-09 13:55:32 -0700 3 months ago	origin/add-data-readme
2015-07-09 13:15:50 -0700 3 months ago	origin/update-deploy-docs
2015-07-08 14:56:49 +0000 3 months ago	origin/selenium
2015-07-07 09:36:17 -0700 3 months ago	origin/decimal
2015-06-24 19:39:31 +0000 4 months ago	origin/fix-tests
2015-06-23 12:24:46 -0400 4 months ago	origin/docs
[...]

PR #284, opened 2016-06-16
[WIP] Public docs (swagger)

toolness:


toolness:
I've memorialized this in #586 and will re-close this PR now.

PR #285, opened 2016-06-16
[WIP] New search features

toolness:


toolness:
I'm filing this as #643 and closing this.

PR #286, opened 2016-06-16
[WIP] adding august data

toolness:


jseppi:
Closing sinceit is superseded by https://github.com/18F/calc/pull/444

PR #292, opened 2016-06-17
Upgrade to Django REST Framework 3.1

### LINKED ISSUE #290: Django REST framework logs RemovedInDjango19Warning
As mentioned in #279 and #288:

> Our version 3.0.5 of `rest_framework` is using `django.utils.importlib`, which is removed in 1.9. Unfortunately, upgrading it requires dealing with the fact that [`PaginationSerializer` was deprecated in 3.1](http://stackoverflow.com/questions/31500192/django-rest-framework-error-cannot-import-name-paginationserializer/31500287#31500287), which will require refactoring our code.

toolness:
This fixes #290 by upgrading us to Django REST Framework 3.1.

Note that the latest version of DRF is 3.3, and I doubt we should run into any major barriers upgrading it all the way there, but I'd rather save it for a separate PR.

As mentioned in #290, upgrading to DRF 3.1 was non-trivial because of its architectural overhaul of pagination.

I believe this PR has feature parity with the existing API, with the following changes:

* On the existing API, submitting an API request for a non-integer page, e.g. `page=blarg`, results in a 500 internal server error.  This PR fixes the bug by delegating all pagination to DRF's `PageNumberPagination`, which returns the following response:

  ```http
HTTP 404 Not Found
Vary: Accept
Content-Type: application/json
Allow: GET, HEAD, OPTIONS

{
    "detail": "Invalid page \"u\": That page number is not an integer."
}
```

* On the existing API, the `next` and `previous` keys in the response are relative URLs, e.g. `?page=2`. In this PR, they are now absolute URLs.

### DIFF
--- a/api/pagination.py
+++ b/api/pagination.py
@@ -5,11 +5,10 @@ from django.conf import settings
 
 
 class ContractPagination(pagination.PageNumberPagination):
-    page_size = settings.PAGINATION
-
     def __init__(self, context):
         super().__init__()
         self.context = context
+        self.page_size = settings.PAGINATION
 
     def get_paginated_response(self, data):
         return Response(OrderedDict([
--- a/api/pagination.py
+++ b/api/pagination.py
@@ -1,0 +1,40 @@
+from collections import OrderedDict
+from rest_framework import pagination
+from rest_framework.response import Response
+from django.conf import settings
+
+
+class ContractPagination(pagination.PageNumberPagination):
+    page_size = settings.PAGINATION
+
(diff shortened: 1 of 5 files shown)


jseppi:
I haven't dug into too much, but it doesn't look like there are any tests for pagination in `api/tests.py`. What do you think about adding some along with this PR?

toolness:
Ok, just added some tests for pagination!

jseppi:
👍

Issue #289, opened 2016-06-16
Table and chart break with filter set to Schedule 70 and Contract Year = +1 or +2


jseppi:
Wanted to log this before I forgot about it. Following these steps, I get a JS error that causes the table and chart to break on the main page of CALC:

1. Set Schedule to "IT Schedule 70"
1. Select "+1" or "+2" for Contract Year
1. Open your browser's developer console, where you will see:

<img width="383" alt="screen shot 2016-06-16 at 5 28 34 pm" src="https://cloud.githubusercontent.com/assets/697848/16135253/d756df62-33e7-11e6-9c36-14aa058bd60e.png">

toolness:
Nice catch!

I think this might actually be a specific case of #244, as the same error (from the same line number) appears to be logged if searching with a query that produces no results, e.g. https://calc.gsa.gov/?q=waegaweg.

PR #294, opened 2016-06-17
Fix empty price histogram

### LINKED ISSUE #244: Page doesn't clear when no results are returned
Replication:
1) Do a search or filter and get a result set back
2) Do a search that will produce no results

You should see that the histogram and table will be populated with the previous query's response. I imagine we should reset/empty these so as not to confuse users.

toolness:
This fixes #244 and #289.

I'm not really sure what I'm doing here, though, as I'm totally unfamiliar with the front-end code. I basically just added the minimum amount of code to ensure that the browser doesn't throw an exception or log any SVG attribute errors.

From an end-user perspective, though, this solution seems nice because one can now enter queries that return no results, and the graph is updated (it previously would remain unchanged due to the JS exception being thrown).  That said, the graph isn't updated to anything *particularly* user-friendly:

empty-price-histogram

But hey, it's still better than the graph not updating *at all*, right? :grin:

### DIFF
--- a/hourglass_site/static/hourglass_site/js/index.js
+++ b/hourglass_site/static/hourglass_site/js/index.js
@@ -350,6 +350,7 @@
         maximum: .001,
         average: 0,
         count: 0,
+        proposedPrice: 0,
         results: [],
         wage_histogram: [
           {count: 0, min: 0, max: 0}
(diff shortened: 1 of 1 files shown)


jseppi:
Yeah, I agree, it is a decent immediate fix over having no feedback.

toolness:
Cool,  should we go ahead and merge it?

jseppi:
👍

PR #296, opened 2016-06-17
[WIP] Simple explainers for header icons

hbillings:
Elaborate on what these terms mean. Needs help with what the text actually says, because I'm not sure these (especially Worldwide Rates) are totally accurate.
screen shot 2016-06-17 at 9 49 19 am

### DIFF
--- a/hourglass_site/static/hourglass_site/style/main.css
+++ b/hourglass_site/static/hourglass_site/style/main.css
@@ -170,9 +170,14 @@ header {
 }
 
 .price-callout p {
-  position: relative;
-  top: 50%;
+  margin-bottom: 0.8rem;
   transform: translateY(-50%);
+}
+
+.price-callout p:last-child {
+  color: rgba(255,255,255,.6);
+  margin-bottom: 2.5rem;
+  font-weight: 300;
 }
 
 h1 {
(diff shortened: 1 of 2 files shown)


toolness:
Nice! Do you think we should bring in the [18F Writing Lab](https://18f.gsa.gov/2016/01/22/18f-writing-lab/) for help on the copy?

hbillings:
This is good to go! I think the language will probably change slightly as we understand better what the crap we're actually doing, heh. (And yes, the writing lab would probably be very helpful!)

jseppi:
👍 Looks good, feel free to merge!
If we wait until #346 is merged then we will see the auto-deploy work (hopefully!)

toolness:
@jseppi, I just merged #346, so feel free to merge this one and see if auto-deploy works!

jseppi:
I'm guessing something is going on with travis, because I have not seen this error during a build before: https://travis-ci.org/18F/calc/builds/142477895#L396

jseppi:
New deploy error appears to be something amiss with travis's `dpl` gem: https://github.com/travis-ci/dpl/pull/474 Hopefully a fix will be live soon.

jseppi:
🎆 Looks like the fix to that `dpl` bug in travis is fixed now. The automatic deploy worked and this PR is now live on staging.

Issue #293, opened 2016-06-17
Create a POST API for submitting contract data, with two integrations


toolness:
So we've currently got #274 and #275, which are prototypes/spikes/proofs-of-concept for submitting data into CALC--one for Region 10's workflow, another for Schedule 70's, respectively. Both of these prototypes are built as Django views in the CALC project, and they're tightly coupled to CALC internals (which is fine for our purposes).

That said, at the same time, we're also thinking about [investigating establishing a data standard for other schedules/vehicles to get their data into CALC](https://trello.com/c/2yKaLkTT/11-investigate-establishing-a-data-standard-for-other-schedules-vehicles-to-get-their-data-into-calc). As folks from the CALC 1 team mentioned at yesterday's meeting, this is potentially where CALC's biggest impact may lie, as it'd allow CALC's dataset to scale dramatically.

Without a data standard and API, every contracting schedule/vehicle will need to strongly couple itself into CALC's Django app in order to get their data in. This requires a knowledge of Python and Django, which not every client may have in-house. One the other hand, an API combined with a data standard, as [18F has done with the DATA Act via its broker and schema](https://18f.gsa.gov/2016/06/14/prototype-early-prototype-often-lesson-from-the-data-act/), could be a more scalable strategy, as any client or vendor could use the technologies they're most comfortable with to get their data into CALC.

The [18F API Standards](https://github.com/18F/api-standards#using-ones-own-api) state:

> The #1 best way to understand and address the weaknesses in an API's design and implementation is to use it in a production system.
>
> Whenever feasible, design an API in parallel with an accompanying integration of that API.

Given this, I'd like to propose that we create an API + data standard for submitting data into CALC, and refactor/rebuild our two prototypes to work as *integrations* of the API + standard. This would allow us to investigate the feasibility of such an approach using two real-world scenarios.

Thoughts?

toolness:
Oh, I should note that creating this API will probably require some user research; as Kelly mentioned to us, he currently reviews all new data submitted to CALC before committing it, so it's likely that the API will need to be structured in such a way that it plays nicely with that workflow.

jseppi:
Sweet idea! Totally agree that tightly coupled django integrations are not the best path forward.

I suppose the review flow could even just be made part of an integration, with the standardized data format not getting POSTed to the CALC endpoint (with admin credentials/token) until the end of that flow.

toolness:
[...]

PR #297, opened 2016-06-20
Upgrade to Django REST framework 3.3.3.

toolness:
This gets us up to date with the latest version of Django REST framework, which paves the way for #293!


toolness:
The only change required was that [`request.QUERY_PARAMS` was deprecated in favor of `request.query_params`](http://www.django-rest-framework.org/topics/3.2-announcement/#deprecations).

PR #298, opened 2016-06-20
[DO NOT MERGE] CoD Export PriceList model discussion

jseppi:
First stab at PriceList and PriceRow models, meant to model the fields of the CoD Export proposed price list uploads. This is based off the example Excel document of contractor prices (not Schedule 70) that was provided to us.

These are very rough and just meant to serve as a discussion starter really. If we follow the proposal in #293 (an API for contracts data), then these models (or their replacements) would be part of the "upload price list xls integration," not the core CALC codebase.

Some questions that this exercise brought up for me:
* What do we call the class of price list that this is based on? It is not Sched 70, it is the combined list that Region 10 deals with.
* Do any of the rates other than the GSA rate matter? I think they do not since they are not currently used in CALC.
* Should the escalation rate be included somewhere? It is not in the example XLS.
* What exactly is a SIN? Will SINs ever need to be linked across price lists/contracts?

### DIFF
--- a/data_capture/models.py
+++ b/data_capture/models.py
@@ -28,8 +28,8 @@ class PriceList(models.Model):
     approval_date = models.DateField(null=True, blank=True)
 
 
-# This model is based on the Sched 70 price proposal list excel sheet that
-# was provided to us as an example.
+# This model is based on the contractor (not sched 70) price proposal
+# excel sheet that was provided to us as an example.
 # It should be trimmed down to only the necessary fields, possibly those
 # described by FLIFCALC.
 # https://github.com/18F/calc/wiki/Federal-Labor-Interchange-Format-for-CALC
--- a/data_capture/models.py
+++ b/data_capture/models.py
@@ -1,0 +1,88 @@
+from django.db import models
+
+from contracts.models import EDUCATION_CHOICES
+
+
+# PriceList is a model for storing the data when a price list XLS is uploaded
+# A PriceList has many PriceRows
+# None of the fields described in this model are in the example XLS,
+# so they would probably need to be gathered through a form.
+class PriceList(models.Model):
+    idv_piid = models.CharField(
+        help_text="Contract Number"
+    )
+
+    schedule = models.CharField(
+        help_text="Schedule"
+    )
+
+    contract_start = models.DateField(null=True, blank=True)
+    contract_end = models.DateField(null=True, blank=True)
+    contract_year = models.IntegerField(null=True, blank=True)
+
+    is_approved = models.BooleanField(default=False)
+
+    # TODO: Uncomment once User models are added
+    # owner = models.ForeignKey(User)
+    # approver = models.ForeignKey(User, null=True)
+    approval_date = models.DateField(null=True, blank=True)
+
+
+# This model is based on the Sched 70 price proposal list excel sheet that
+# was provided to us as an example.
+# It should be trimmed down to only the necessary fields, possibly those
+# described by FLIFCALC.
+# https://github.com/18F/calc/wiki/Federal-Labor-Interchange-Format-for-CALC
+class PriceRow(models.Model):
+    sin = models.CharField(
+        help_text="Proposed SIN(s)"  # TODO: What does SIN stand for?
+    )
+    labor_category = models.CharField(
+        help_text="Labor Category"
+    )
+    education_level = models.CharField(
+        db_index=True, choices=EDUCATION_CHOICES, max_length=5,
+        null=True, blank=True, help_text="Minimum Education Level")
+    minimum_education_years = models.IntegerField()
+    contractor_site = models.CharField()
+    domesticity = models.CharField()  # domestic only, overseas only, worldwide
+
+    # Not in the example XLS, but perhaps should be as it would capture
+    # the yearly percentage increase of the negotiated rate
+    escalation_rate = models.DecimalField(
+        help_text="Escalation Rate"
+    )
+
+    commercial_price_per_hour = models.DecimalField(
+        help_text="Commercial Price per Hour"
+    )
+    most_favored_customer = models.CharField(
+        help_text="Most Favored Customer Name"
+    )
+
+    mfc_discount = models.DecimalField(
+        help_text="Most Favored Customer Discount (Percent)"
+    )
+
+    mfc_price_per_hour = models.DecimalField(
+        help_text="Most Favored Customer Price per Hour"
+    )
+
+    gsa_commercial_discount = models.DecimalField(
+        help_text="Discount to GSA off of Commercial Rate (Percent)"
+    )
+    gsa_mfc_discount = models.DecimalField(
+        help_text="Discount to GSA off of Most Favored Customer Rate (Percent)"
+    )
+
+    # This might be the only price field necessary to keep for CALC's purpose
+    gsa_price_per_hour = models.DecimalField(
+        help_text="Price offered to GSA"
+    )
+
+    supporting_document_number = models.CharField(
+        help_text="Support Document Number"
+    )
+
+    # PriceList has many PriceRows
+    price_list = models.ForeignKey(PriceList, on_delete=models.CASCADE)
--- a/data_capture/models.py
+++ b/data_capture/models.py
@@ -2,6 +2,18 @@ from django.db import models
 
 from contracts.models import EDUCATION_CHOICES
 
+DOMESTICITY_CHOICES = (
+    ('D', 'Domestic Only'),
+    ('O', 'Overseas Only'),
+    ('W', 'Worldwide')
+)
+
+SITE_CHOICES = (
+    ('CON', 'Contractor Site'),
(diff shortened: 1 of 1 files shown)


toolness:
> What exactly is a SIN? Will SINs ever need to be linked across price lists/contracts?

SIN stands for **Special Item Number**. From the [Welcome to GSA Schedules](http://www.gsa.gov/portal/category/100615) page:

> Each GSA Schedule is composed of Special Item Numbers (SINs). It’s a categorization method that groups similar products, services, and solutions together to aid in the acquisition process.

So it's roughly like a category ID of sorts. For example, [here's all the SINs in Schedule 70](http://www.gsaelibrary.gsa.gov/ElibMain/scheduleSummary.do?scheduleNumber=70) (the "category" column is the SIN column, with a dash instead of a space between the two numbers, I believe).

A related categorization system that the Census Bureau uses is the **North American Industry Classification System** but there's no "bridge" or "rosetta stone" that maps NAICS to SIN or vice versa. I forget why this particular piece of information is important, but one of the people we interviewed thought it was...

I believe Kelly said on one of our calls that the "Schedule" pull-down on the CALC filter list will eventually change to be a SIN pull-down or text field:



So in that sense I *think* that SINs will need to be linked across price lists/contracts? We might want to make sure though.

toolness:
> Should the escalation rate be included somewhere? It is not in the example XLS.

So an interesting thing about the way CALC is currently storing this information is not as a percentage rate, but as the actual rate for the next five years, represented in the model as [`hourly_rate_year1` thru `hourly_rate_year5`](https://github.com/18F/calc/blob/0e9e3fc0d67359e3d3841ae36dce9d44411260f4/contracts/models.py#L68-L72).  Technically the current model is storing more information, since it allows the escalation rate to change year after year.  In fact, looking at the powerpoint Kelly sent us, Region 10's internal MS Access-based tool has the following form:



So it looks like they are capturing a different escalation rate for each year, and probably converting that to actual hourly rate prices when importing the data into CALC?

toolness:
> Do any of the rates other than the GSA rate matter? I think they do not since they are not currently used in CALC.

I think these are the ones that contractors are sensitive about?  I suppose that leaves the question of whether we should store the information internally but simply not expose it to end-users, in the anticipation that the information may *eventually* be made public, or whether we should leave it out of the database entirely...

Also, for the Schedule 70 data, this rate is in a column called **PRICE OFFERED TO GSA (including IFF)**.  I'm not sure what "IFF" stands for, and whether it's a Schedule 70-specific concept or a GSA-wide concept.

By the way, thanks for including the `help_text` in the model fields!  :tada:  Since it's actually [unescaped HTML](https://docs.djangoproject.com/en/1.9/ref/models/fields/#help-text), we can even include links for the readers to learn more about some of the terms, which might be helpful.

toolness:
> What do we call the class of price list that this is based on? It is not Sched 70, it is the combined list that Region 10 deals with.

At the risk of over-generalizing, we might want to actually call it a `UniversalPriceList` or something, since Nick Brethauer mentioned in the meeting last week that he thinks the same model could probably be used unmodified across all contracting shops?

Regardless of what we choose, we should probably mention it in the docstring for the class!

data_capture/models.py
```diff
@@ -0,0 +1,88 @@
...
+from contracts.models import EDUCATION_CHOICES
+
+
+# PriceList is a model for storing the data when a price list XLS is uploaded
+# A PriceList has many PriceRows
+# None of the fields described in this model are in the example XLS,
+# so they would probably need to be gathered through a form.
+class PriceList(models.Model):
+    idv_piid = models.CharField(
+        help_text="Contract Number"
```
toolness: I would really, really like to figure out what `idv_piid` actually means.  If we can't figure it out, I kinda want to propose just changing the name of the field to `contract_id`... Maybe I'm nitpicking though.

data_capture/models.py
```diff
@@ -0,0 +1,88 @@
...
+
+    schedule = models.CharField(
+        help_text="Schedule"
+    )
+
+    contract_start = models.DateField(null=True, blank=True)
+    contract_end = models.DateField(null=True, blank=True)
+    contract_year = models.IntegerField(null=True, blank=True)
+
+    is_approved = models.BooleanField(default=False)
```
toolness: Does this effectively mean "whether Kelly has approved it for inclusion in the CALC database"?  Or is it some other level of approval?

data_capture/models.py
```diff
@@ -0,0 +1,88 @@
...
+    # owner = models.ForeignKey(User)
+    # approver = models.ForeignKey(User, null=True)
+    approval_date = models.DateField(null=True, blank=True)
+
+
+# This model is based on the contractor (not sched 70) price proposal
+# excel sheet that was provided to us as an example.
+# It should be trimmed down to only the necessary fields, possibly those
+# described by FLIFCALC.
+# https://github.com/18F/calc/wiki/Federal-Labor-Interchange-Format-for-CALC
```
toolness: Wooo FLIFCALC!  We're totally inventing our own government acronym! :bowtie:

data_capture/models.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    labor_category = models.CharField(
+        help_text="Labor Category"
+    )
+    education_level = models.CharField(
+        db_index=True, choices=EDUCATION_CHOICES, max_length=5,
+        null=True, blank=True, help_text="Minimum Education Level")
+    minimum_education_years = models.IntegerField()
+    contractor_site = models.CharField()
+    domesticity = models.CharField()  # domestic only, overseas only, worldwide
```
toolness: Should we use a [`choices` keyword arg](https://docs.djangoproject.com/en/1.9/ref/models/fields/#choices) for this?

data_capture/models.py
```diff
@@ -0,0 +1,88 @@
...
+    # approver = models.ForeignKey(User, null=True)
+    approval_date = models.DateField(null=True, blank=True)
+
+
+# This model is based on the contractor (not sched 70) price proposal
+# excel sheet that was provided to us as an example.
+# It should be trimmed down to only the necessary fields, possibly those
+# described by FLIFCALC.
+# https://github.com/18F/calc/wiki/Federal-Labor-Interchange-Format-for-CALC
+class PriceRow(models.Model):
```
toolness: Do you think `PriceListRow` or `PriceListItem` might be a clearer name, since the prefix implies it's a part of a `PriceList`?  Dunno, just thinking out loud.

data_capture/models.py
```diff
@@ -0,0 +1,88 @@
...
+        db_index=True, choices=EDUCATION_CHOICES, max_length=5,
+        null=True, blank=True, help_text="Minimum Education Level")
+    minimum_education_years = models.IntegerField()
+    contractor_site = models.CharField()
+    domesticity = models.CharField()  # domestic only, overseas only, worldwide
+
+    # Not in the example XLS, but perhaps should be as it would capture
+    # the yearly percentage increase of the negotiated rate
+    escalation_rate = models.DecimalField(
+        help_text="Escalation Rate"
```
toolness: As I mentioned [here](https://github.com/18F/calc/pull/298#issuecomment-227299065), we probably need to store multiple fields for different years of the escalation rate...  Or just do what CALC does and store it as the actual computed hourly rate for years 1-5.

data_capture/models.py
toolness: Hmm, I'm realizing now that there seem to be at least two levels of approval... one is the approval of the CALC data administrator, i.e. Kelly, and then there's the approval of the "Acquisition Center Data Administrator", e.g. the admin at Schedule 70, who makes sure that the internal COs entered the right data...

Since you mentioned that in the post-API world of #293, these models wouldn't actually live in CALC, I'm guessing this means that `is_approved` is referring to the acquisition center data administrator?  And that we'd probably need to add an `is_approved` field to our own data model in CALC to indicate whether Kelly has approved the data?

data_capture/models.py
jseppi: Individual Price Item Id? Just a guess...

data_capture/models.py
jseppi: definitely

data_capture/models.py
jseppi: I think what we end up going with is going to need to be what is easiest or what most closely mirrors the contracting officers' current way of working. This is probably a good thing for our user research to explore.

data_capture/models.py
jseppi: I am actually not totally clear on if there needs to be two levels of approval actually. Definitely something that needs to be figured out in user research.
You're right that if there are the two levels, then that flag or something similar would need to be in both apps.

toolness:
Um... what does CoD mean?

jseppi:
During yesterday's meeting, I asked Kelly what would be a good name for the price list he generates from the Region 10 internal tool. "CoD Export" was his answer - I think it stands for "Contracts Database Export".

jseppi:
Good to close!

PR #300, opened 2016-06-21
Take SECRET_KEY from environment.

### LINKED ISSUE #280: SECRET_KEY should be defined by environment
@jseppi noticed that `SECRET_KEY` is currently hard-coded into the base settings file. This probably isn't a big deal right now since we don't think there's any user accounts on the system, so no way to abuse knowledge of the secret key; but we should probably make the value of this setting come from the environment, and set it to a known default if `DEBUG` is true.

toolness:
This fixes #280.

Note that in production, this will require configuring the cloud.gov environment to set a `SECRET_KEY` environment variable. According to the [documentation](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#env-block) I think this can probably be done with something like:

```
cf set-env SECRET_KEY boop
```

However, depending on how we deploy the app, we might need to somehow ensure that this environment variable is still defined whenever we re-deploy... This could potentially get complicated with a zero-downtime tool like [autopilot](https://github.com/concourse/autopilot), I'm not sure though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,12 @@ can just run `python manage.py` directly from outside the container--the
 `manage.py` script has been modified to run itself in a Docker container
 if it detects that Django isn't installed.
 
+## Environment Variables
+
+* `SECRET_KEY` is a large random value corresponding to Django's
+  [`SECRET_KEY`][] setting. It is automatically set to a known, insecure
+  value when `DEBUG` is true.
+
 ## API
 
 If you're interested in the underlying data, please see https://github.com/18F/calc/blob/master/updating_data.md
@@ -239,3 +245,4 @@ for other than small business.
 [Docker]: https://www.docker.com/
 [Docker Compose]: https://docs.docker.com/compose/
 [Docker goes native]: https://blog.docker.com/2016/03/docker-for-mac-windows-beta/
+[`SECRET_KEY`]: https://docs.djangoproject.com/en/1.9/ref/settings/#secret-key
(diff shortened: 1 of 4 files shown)


jseppi:
I think using a Custom User Provided Service (CUPS) for environment variables is the preferred way to persist environment variables across deploys in Cloud Foundry: https://docs.cloudfoundry.org/devguide/services/user-provided.html
I did a short write-up in the team-api-server README: https://github.com/18f/team-api-server#on-cloudgov

jseppi:
It might be a good idea to throw an exception if SECRET_KEY (or any other required environment variables) are not defined. That way any deploys to cloud.gov will fail and can show an informative error. Otherwise a bad version could get deployed and start running without some vital piece of config set.

cmc333333:
For reference, autopilot will indeed drop any `set` env vars (as it's using the project's manifest file). Some folks have gotten around this by running `create-app-manifest`within the deployment script, but +1 to using a user service instead.

toolness:
> It might be a good idea to throw an exception if SECRET_KEY (or any other required environment variables) are not defined.

Oh, I implemented this with that in mind--if `SECRET_KEY` isn't defined, then `os.environ['SECRET_KEY']` will be consulted, and that will throw a `KeyError` if the env var is missing.

jseppi:
Oh, cool, I forgot python does that.

jseppi:
👍 to merge

PR #304, opened 2016-06-21
[WIP] Cloud.gov UAA-brokered Authentication

jseppi:
Work-in-progress PR for adding UAA-based authentication


hourglass/settings.py
```diff
@@ -158,26 +159,36 @@
...
+# List of required configuration variables
+required_config_vars = [
+    'SECRET_KEY',
+    'ADMIN_EMAIL',
+    'UAA_OAUTH_APP_NAME',
+    'UAA_OAUTH_SECRET'
+]
+
+# set all required config variables to ''
+for required_var in required_config_vars:
```
jseppi: @toolness: could you review what I'm doing here? I'm not sure if this is 'pythonic'.
We're also going to need to do come up with a way to deal with private config values in the docker environment. Any ideas for that?

toolness: Sure, that seems reasonable. Another option is for us to do away with the `local_settings`/`docker_settings` thing and just require developers to define *everything* via the environment (or `VCAP_SERVICES`).

I guess alternatively, since you're going to be using `globals()` anyways, you could just skip the setting of every required option to the empty string, and test for the existence of each required config variable down below too--that way all the "required config variable logic" stays in one place, at least?

Regarding docker, typically the private config value and user-configurable stuff is done via an [environment file](https://docs.docker.com/compose/env-file/).  We do this with the `EXTERNAL_PORT` env var in Tock (see its  [`.env.sample`](https://github.com/18F/tock/blob/master/tock/.env.sample), [`docker-compose.yml`](https://github.com/18F/tock/blob/master/docker-compose.yml)). Just need to remember to document it in the readme.

hourglass/settings.py
```diff
@@ -187,5 +186,18 @@
...
+    'UAA_OAUTH_SECRET'
+]
+
+# Check that each required config variable has been set,
+# if not, attempt to get the value from environment variable of the same name.
+# This will throw a KeyError if a value is not found.
+# TODO: this will eventually need to be modified to work with VCAP_SERVICES
+for required_var in required_config_vars:
+    if required_var not in globals():
+        globals()[required_var] = os.environ[required_var]
```
toolness: In #310 I've added a new module called `settings_utils`, so another option here is to create a function like `load_required_settings_from_environment(setting_names, globs)` there and call it from here. One advantage of this is that you can add unit tests for the function, but it's up to you.

jseppi:
I am stumped :( Hoping @toolness can get this working.
@toolness: Let's pair tomorrow so I can explain where things are.

jseppi:
Made it slightly further than I was earlier.
This is the OpenId Connect provider I'm running locally to develop against: https://www.npmjs.com/package/oidc-provider

jseppi:
Closing this mess in favor of rolling 🎢 our own now that we worked out all the details with cloud.gov.

PR #305, opened 2016-06-21
[DO NOT MERGE] Non-functional selenium server integration.

toolness:
Basically I was trying to see if I could work-around our problems with PhantomJS selenium integration (see #266, #267, etc) by using [Docker images for standalone Selenium](https://github.com/SeleniumHQ/docker-selenium). What I've got right now works *in theory*--the app container is able to connect to the selenium server and pass it its IP address, and the selenium server is able to connect back--but there are lots of problems, e.g.:

* The user has to run `docker-compose up selenium` in a separate terminal window, which is inconvenient.
* Sometimes `docker-compose up selenium` [randomly fails](https://github.com/18F/calc/pull/305#issuecomment-227479583).
* The docker-selenium container image is large and takes a very long time to download even on a fast internet connection, for some reason (that could've just been something weird going on with my internet connection, I suppose).

Aside from that, there's lots of [issues in docker-selenium's GitHub](https://github.com/SeleniumHQ/docker-selenium/issues) that seem to indicate lots of possibilities for the same kinds of problems we're running into with PhantomJS, so I'm closing this PR unmerged.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,21 +1,34 @@
-app:
-  build: .
-  volumes:
-    - .:/calc
-  links:
-    - db
-  working_dir: /calc
-  entrypoint: python /calc/docker_django_management.py
-  environment:
-    - DDM_HOST_USER=calc_user
-    - DDM_IS_RUNNING_IN_DOCKER=yup
-    - PYTHONUNBUFFERED=yup
-    - DATABASE_URL=postgres://calc_user@db/calc
-  command: python manage.py runserver 0.0.0.0:8000
-  ports:
-    - "8000:8000"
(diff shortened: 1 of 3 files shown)


toolness:
Also, every so often `docker-compose up selenium` gives me this:

```
Starting calc_selenium_1
Attaching to calc_selenium_1
selenium_1  | xvfb-run: error: Xvfb failed to start
calc_selenium_1 exited with code 1
```

Strangely, this seems to get fixed by making trivial, minor tweaks to the `docker-compose.yml`. For example, simply adding the following:

```yml
environment:
  - BOOP=1
```

will fix it.

PR #307, opened 2016-06-21
Don't use globals() when looking up SECRET_KEY.

toolness:
#300 was a bit silly. This is way less hacky!


jseppi:
👍 This is more understandable to me

PR #308, opened 2016-06-21
Add sass integration to docker

toolness:
(Note that this PR was originally filed as #268, but I had to close that and create a new one because it was targeted against `master` instead of `develop`.)

This PR adds docker support for SASS and also adds documentation for SASS setup to the README.

@hbillings or @jseppi, can you review this real quick?


docker-compose.yml
```diff
@@ -20,3 +20,10 @@ db:
   environment:
     - POSTGRES_DB=calc
     - POSTGRES_USER=calc_user
+sass:
+  build: .
+  volumes:
+    - .:/calc
+  working_dir: /calc/hourglass_site/static/hourglass_site
+  entrypoint: python /calc/entrypoint.py
```
jseppi: This line is preventing the container from starting since that path doesn't exist in it. Looks like removing it makes it work  as expected.

jseppi:
It seems a little heavy-handed or overkill or maybe just weird to me have a container running just to rebuild the sass files. I'm not totally against it, especially given that I am not really familiar with best practices for builds/asset pipelines in pythondjangoland, though.

docker-compose.yml
toolness: Oh shoot, good catch--that went away with #273. Just fixed it (note we want to use `docker_django_management.py` as the entrypoint solely because it sets things up so that the SASS files created aren't owned by root).

toolness:
Yeah, it's definitely suboptimal--I think the real fix will come with #12, this is just a quick stopgap to make development not-too-painful in the interim.

jseppi:
Cool, 👍 from me then.

toolness:
Woot thanks! I'll merge once travis gives the :+1:

PR #309, opened 2016-06-21
Increase phantomJS timeout to 15s on travis

toolness:
A number of travis builds were failing earlier today due to hitting the 3-second timeout so many times (see #267) during the selenium tests.  I re-ran one of them just now and it worked fine, though, so my guess is that Travis was just under a lot of load and taking a really long time to run our tests. So this PR changes the timeout to 15s on Travis while keeping it at a brisk 3s for local development.


jseppi:
👍

PR #312, opened 2016-06-22
Use Web Design Standards

hbillings:
Mostly this only affects the colors used (including a new header background image). I think it makes sense not to change the fonts. Might change the table styles in the future. Looks like this: 
screencapture-0-0-0-0-8000-1467238900296


toolness:
Thanks, this looks great!

Just as a workflow thing, I suggest committing the "compiled" changes as a separate commit, if possible, as this makes it easier to review your code by just focusing on the commits that change the "real" files rather than the ones auto-generated by SASS.

That said, I think the problem will go away when we move to django-pipeline (#12) and/or libsass, since it will mean that our SASS gets compiled on-the-fly instead of being stored in the repo. 😁

PR #314, opened 2016-06-23
Use branch=develop for travis CI badge

toolness:
It seems as though https://travis-ci.org/18F/calc.svg just shows the build status of the most recent build, even if it's e.g. a WIP PR that's currently broken.  If we explicitly add `?branch=develop` to the URL then we'll get the status of the current `develop` branch.


jseppi:
👍

PR #316, opened 2016-06-24
Updated docker docs in README.

toolness:
This removes references to docker-machine, since Docker has gone native for OS X / Windows (although it is still in beta).  It also links to the 18F docker guide and documents updating the container.


jseppi:
👍

PR #317, opened 2016-06-24
Add support for cloud.gov UAA authentication.

toolness:
## Auth Flow

**Note: The following UI isn't exactly ideal, but I'm leaning towards staying with what we've got for now and filing issues to improve things later, to limit the scope of this PR.**

Because very few people will actually need to log in to the site, a nondescript login link is in the *footer*:

> login-link

Clicking on the "Log in" link redirects users to cloud.gov for 2FA login. Once they've authenticated there, CALC will examine the email address they've logged in with.  If no user accounts with that email address exist in the CALC database, the user will *not* be logged-in; instead, they will be shown an error page with the following content:

  > **Login Error**
  >
  > Either you don't have permission to log in to this system, or an error occurred when trying to log you in.

Otherwise, they will be logged in, and they'll see their login status in the header, as well as a link to the admin panel:

> logout-link

Clicking on "Site administration" will take them to Django's standard admin UI, for the time being. We'll probably change it to a more custom admin view later (though we'll also want to make Django's admin UI available then, too, since superusers will probably want to use it to add new users).

Clicking the "Log out" link currently takes the user to a page with the following content:

> You are now logged out.

## Fake UAA Provider

This PR adds a `fake_uaa_provider` app, available in debug mode only, that simulates the cloud.gov/cloudfoundry UAA server. Users can enter any email address, and the client will be told that the user with that email address has been authenticated:

> fake-uaa-provider

This provider can successfully be authenticated with via @jseppi's [cloud.gov UAA client proof-of-concept](https://gist.github.com/jseppi/8a92014396b746cd74fd330eb4ceb110) (just change `AUTH_URL` and `TOKEN_URL` to point at your development server; the relative paths remain unchanged).

This provider is automatically enabled if debug mode is active and no client secret has been configured.

## Security Considerations

* We enforce use of the [`state` parameter](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html) to prevent exploits.

* At present, we set the user's session expiry to the expiry of the access token we receive from cloud.gov.  Given that the access tokens cloud.gov provides currently seem to have a lifetime of about 12 hours, this means that user sessions won't last long.  If this ends up being a significant hassle, we can add code to automatically refresh the token later (even though we don't actually use the token to access any protected resources, it's important to refresh it as a way of ensuring that the user's account is still valid, e.g. that they haven't left GSA or something).

* Our `/auth/login` endpoint supports a `next` parameter representing the URL to redirect to after the user has logged in, but we use `django.utils.http.is_safe_url()` to ensure that it cannot lead to arbitrary third-party sites.

## To Do

- [x] Add a `uaa_client` app that implements a Django authentication backend. See [mozilla/teach-api's `new_webmaker.py`](https://github.com/mozilla/teach-api/blob/master/teach/new_webmaker.py) for an example of an auth backend that uses OAuth2.
- [x] We probably should use a `state` parameter with our OAuth2 handshake. I forget what the security implications are--[here's a blog post of them](http://www.twobotechnologies.com/blog/2014/02/importance-of-state-in-oauth2.html), though.
- [x] Add a `next` argument to the `/auth/login` endpoint/redirector, so that pages which have a login link can ensure that they're navigated back-to after login.
- [x] Add a human-friendly error page when the OAuth2 handshake fails.
- [x] Add proper login/logout links to the base template.
- [x] Use the `expires_in` field of the authentication token to set our app's session lifetime~~; or at the very least, align our Django app's session lifetime with the current session lifetime of the tokens cloud.gov serves (it appears to be about 12 hours)~~. We might also want to ping the cloud.gov or compliance folks about best practices around this.
- [x] Document how users are created in the readme (put simply, one needs to create users via `manage.py createsuperuser` or the admin UI and associate their email address w/ their gov-related email, so that they'll be able to log in via cloud.gov UAA).
- [x] Add unit tests for `uaa_client.authenticate`; see [mozilla/teach-api's `test_new_webmaker.py`](https://github.com/mozilla/teach-api/blob/master/teach/tests/test_new_webmaker.py) for an example.
- [x] Add unit tests for `uaa_client.views`; see [mozilla/teach-api's `test_views.py`](https://github.com/mozilla/teach-api/blob/master/teach/tests/test_views.py) for an example.
- [x] Add some kind of setting(s) to configure the auth backend to use cloud.gov or `fake_uaa_provider` and document it/them. We should probably do #319 before this.
- [x] Make it so users never see the django-admin login screen, but are rather redirected to cloud.gov/UAA login ([here is a potential solution](http://stackoverflow.com/a/13186337)).


toolness:
In the future, it might be worth spinning this kind of code out into a separate python package so that other Django-based gov projects can use them.  I don't think we need to build for that right this minute--other gov projects may likely need a lot more UAA functionality than we do--but it'd be nice to keep in the back of our minds as we develop.

toolness:
I have some questions regarding a login link. It occurs to me that very few people will actually be logging into CALC, which makes me wonder if we should actually feature it *at all* in CALC's base template, as it will be irrelevant and potentially even confusing for the vast majority of users.  

Another option is to simply tell administrators/staff, out-of-band (e.g. during some kind of training), to visit https://calc.gsa.gov/admin/ to log in.  Once someone's logged in, it probably makes sense to let them know they're logged in by indicating it in the base template--which is fine, since we'd only be giving the information to folks who will find it useful at that point.

Thoughts @hbillings ?

toolness:
grumble grumble selenium grumble travis 😭

hbillings:
@toolness I agree that it probably will be confusing for users who can't log in to have a login link.  At the same time, relying on folks' memories (and/or desire to read documentation) has some drawbacks, too.  I wonder if perhaps a good solution would be to include a login link in the footer, where most casual users will never see it?  We could also take the language in a direction like, "Registered user login" or something else that implies that one has to have been granted access already?

I don't have super-strong opinions either way, though.  I think maybe it's worth asking some folks to see if they need or want a login link.

toolness:
Ah good idea about the footer link! I'll try that, thanks!

jseppi:
Yay, CodeClimate integration works 🎉

jseppi:
👍 good to merge!

PR #320, opened 2016-06-27
Use python-dotenv and .env file.

### LINKED ISSUE #319: Use python-dotenv for all customizable settings
Currently the system whereby `local_settings.py` is used to change setting requires some really weird code in our `settings.py` and is also orthogonal to the environment-based 12-factor setup for cloud.gov deploys, which also means we have to write lot more documentation (or leave lots of things undocumented).

Instead of this, we should just use [python-dotenv](https://github.com/theskumar/python-dotenv) to make it easy for devs to use a `.env` file, at which point they can just follow the same environment variable documentation used for cloud.gov deploys.

toolness:
This fixes #319.


jseppi:
👍 Looks good.
I made some small pep8/flake8 fixes.

Issue #321, opened 2016-06-27
Add pep8 to continuous integration


jseppi:
Either CodeClimate or Travis.

jseppi:
The final step for this that I just finished was adding the CodeClimate webhook to the repo settings, as documented here https://docs.codeclimate.com/docs/github#pull-requests

PR #322, opened 2016-06-27
Enable pep8 in CodeClimate

jseppi:
I looked into adding `flake8` to Travis, but right now the codebase has a lot of errors so I think it would be burdensome (and probably ignored). Instead, I think we should focus on adhering to pep8 on any new code, and cleaning up existing code as we do any modifications.

This PR adds pep8 checking to CodeClimate.

Ref #321


toolness:
Nice! I'm not familiar with CodeClimate or its configuration settings, so if this looks good to you, feel free to merge it...

jseppi:
Cool. I'm not that familiar with it either. It has GitHub commit/PR integration I think, so I need to figure out how to enable that still.

jseppi:
Unfortunately you can't change the default branch in CodeClimate once a repo has been added. Instead, the repo has to be removed an re-added. Since we've changed our default branch to `develop`, we will have to do that, but we need a CodeClimate admin to help us out.

jseppi:
The default branch has been updated in CodeClimate, so this is good to go. I never got a response from the email I sent to CodeClimate, so I'm not sure how the change happened. It may have been automatic, or someone there may have manually assisted.

PR #323, opened 2016-06-27
use preferred url import syntax, fixes #318; pep8

### LINKED ISSUE #318: Consistently import urls
To be consistent with Django documentation, import application urls using the string parameter form, ex:

```python
url(r'^author-polls/', include('polls.urls', namespace='author-polls'))
```

instead of 

```python
from .polls import urls as polls_urls
...
url(r'^author-polls/', include(polls_urls, namespace='author-polls'))
```

jseppi:


toolness:
Nice! As long as travis gives this 👍 , feel free to merge!

PR #327, opened 2016-06-27
add .env to cfignore

jseppi:
Don't push local `.env` file when doing a cf deploy.


toolness:
Ah good idea! Feel free to merge (don't even worry about travis' arbitrary failing here, it's clearly unrelated to your changes).

jseppi:
Heh, this was actually the source of a problem that took me a long time to solve today unfortunately. The `DATABASE_URL` from my local `.env` was clobbering the one automatically set by the buildpack, so the app would never start because it couldn't connect to the database. However, the fact that it couldn't connect to the database was not very apparent to me as the source of the issue, so it took me a long time to track it down.

toolness:
Oh man, that sucks!! That would've taken me forever to figure out 😣

PR #328, opened 2016-06-27
match cfignore to gitignore

jseppi:
Since I was looking at `.cfignore` I actually realized that all kinds of stuff could/should not be included in the deploy pushes. This reduces the amount of files, both in size and count, that need to be uploaded, in turn making deploys a little bit faster.

I also made more lenient rules for ignoring `*manifest*.yml` and `*credentials*.json` files in both `.gitignore` and `.cfignore`.

before:
`Uploading 13.6M, 1273 files`
after:
`Uploading 8.6M, 179 files`


.cfignore
```diff
@@ -1,3 +1,26 @@
-.env
-hourglass/local_settings.py
+/static
```
toolness: Great catch!  This `/static` is my only concern though--does cf automatically run `manage.py collectstatic` itself like heroku does, or are we expected to run it locally and then have `cf push` push the static files to cloud.gov along with the rest of the app?

jseppi: It does run `collectstatic` (ref https://github.com/cloudfoundry/python-buildpack/blob/master/bin/steps/collectstatic)

Issue #330, opened 2016-06-28
PhantomJS-based Selenium tests randomly fail on CircleCI


toolness:
I'm not sure when this started. For a while, after #267, we had no problems. Then we had to do #309. About a week ago we seem to be getting [random failures like this](https://travis-ci.org/18F/calc/builds/140786815#L2392), where, as far as I can tell, the following happens:
1. Some number of selenium tests actually succeed.
2. At some point in a selenium test, `http.client` raises a `http.client.BadStatusLine` error with an empty string as the content. The module's code indicates that "Presumably, the server closed the connection before sending a valid response."
3. All future tests fail with a `urllib.error.URLError` of `Connection refused`.
4. The final test errors while trying to take a screenshot via webdriver, with the same exception as above.

The particularly frustrating thing about this is that, unlike previous selenium problems, I'm unable to reproduce this on my local system.

Probably the best thing to do for now is to just not run selenium tests on Travis. 😞

jseppi:
Good to close, @toolness?

toolness:
Hmm, I feel a bit weird closing it because #331 was more of a workaround than an actual fix... Maybe we should rename it to "Selenium tests fail on travis when enabled" or something?

toolness:
Ok, so in #841 we added support for Travis to run selenium tests via Sauce Labs for *trusted* PRs--however, for *untrusted* ones we can't have our Sauce Labs key in the Travis environment, and therefore there's still a need for falling back on PhantomJS. So we're leaving this issue open!

toolness:
Ok, we're closing this because (A) we're not on Travis anymore and (B) the selenium tests seem to work on Circle.  But if (B) changes we'll re-open this ticket and rename it to mention Circle.

toolness:
well we moved from phantomjs to headless chrome in #1767 so i'm closing this.

PR #331, opened 2016-06-28
Remove makefile, ignore selenium tests on Travis.

toolness:
This is a workaround (not really a fix) for #330.

I also decided to get rid of the makefile, because it wasn't very flexible:

* Unless I'm mistaken, it didn't allow us to run only specific tests like `py.test` does.
* The `test-backend` target explicitly listed all the Python packages that constituted the back-end, which meant that we'd have to update it every time we added new apps.

Instead, I just changed the selenium tests fixture to run `manage.py collectstatic`, which as far as I can tell is the only reason the makefile existed.


jseppi:
Looks good for now. I agree with removing the Makefile.

PR #337, opened 2016-06-29
Replace numpy

### LINKED ISSUE #335: Replace numpy
`numpy`, while great, is also a huge requirement and really slows down builds. It is only be used to generate a histogram and to compute a standard deviation. I suspect both of those methods could be either implemented directly, or other modules could be found that don't have the compilation overhead that `numpy` has.

jseppi:
Fixes #335 

`numpy` was a requirement for this project but it was only used in two places: 1) calculating a standard deviation, and 2) generating simple, fixed-width histogram bins. This PR removes the numpy dependency, replacing the used functionality with native and local code. The point of doing so is primarily to reduce build times, which are generally over 4 minutes long on Travis, with most of the time (200+ seconds) being spent during the `pip install` step.

TODO:
- [x] replacement for `numpy.std`
- [x] replacement for `numpy.histogram`
-  ~~pep8 fixes on all files that are in this PR~~ won't pep8 existing files so that this will be easier to revert if needed
- [x] tests
- [x] verify that for our cases the replacement code returns the same results. This PR includes a comment with my method comparison test harness.

Other findings:
- I'm concerned that the standard deviation calculation, both as it was using `numpy.std` and with the replacement code (`math.sqrt(statistics.pvariance)`) is the _population_ standard deviation instead of the _sample_ standard deviation. That is probably out of scope for this PR and can be validated/discussed independently.


jseppi:
Wow, this brought the `pip install` time on Travis down from 200+ seconds to 21 seconds!

jseppi:
Here is the test harness I wrote to verify that the new methods in this PR are equivalent to the way that the numpy methods were previously used:

```py
import unittest
import random
import statistics
import math
import numpy
import random

# this is this histogram method in this PR
from hist import get_histogram

# this is the old histogram method that uses numpy.histogram
# as shown at:
#  https://github.com/18F/calc/blob/b6adcbdc0a50536fd68a89f9b469c4bc927c6eb5/api/views.py#L146
from hist import get_numpy_histogram

# new standard deviation in this PR to replace use of numpy.std
def stdev(values):
    if (len(values) == 0):
        # mirror numpy.std functionality of returning nan when input is empty
        return float('nan')
    # else standard dev is the square root of the variance
    return math.sqrt(statistics.pvariance(values))


class StdDevComparisonTests(unittest.TestCase):
    def assertStdDevEqual(self, values):
        np_std = numpy.std(values)
        new_std = stdev(values)
        self.assertAlmostEqual(np_std, new_std)

    def test_random_float_values(self):
        values = [random.random() for _ in range (10000)]
        self.assertStdDevEqual(values)

    def test_random_int_values(self):
        values = [random.randrange(1,101,1) for _ in range (10000)]
        self.assertStdDevEqual(values)

    def test_returns_nan_for_empty_values(self):
        values = []
        np_std = numpy.std(values)
        new_std = stdev(values)
        self.assertTrue(math.isnan(np_std))
        self.assertTrue(math.isnan(new_std))


class HistogramComparisonTests(unittest.TestCase):

    def assertBinEquivalence(self, values, num_bins=10):
        new = get_histogram(values, num_bins)
        old = get_numpy_histogram(values, num_bins)
        self.assertEqual(len(new), len(old))
        self.assertEqual(len(new), num_bins)
        for i in range(0, len(old)):
            old_bin = old[i]
            new_bin = new[i]
            self.assertEqual(old_bin['count'], new_bin['count'])
            self.assertAlmostEqual(old_bin['min'], new_bin['min'])
            self.assertAlmostEqual(old_bin['max'], new_bin['max'])

    def test_empty_values(self):
        self.assertBinEquivalence([])

    def test_simple_values(self):
        self.assertBinEquivalence([1, 2, 3])

    def test_random_int_values(self):
        values = [random.randrange(1,101,1) for _ in range (10000)]
        self.assertBinEquivalence(values)

    def test_random_monetary_values(self):
        # values is a list of floats rounded to to a precision of 2
        # to mirror the types of floats we have in CALC
        values = [round(random.random(), 2) for _ in range (10000)]
        self.assertBinEquivalence(values)
```

jseppi:
We're going to ignore the CodeClimate pep8 problems with the existing `api/views.py file` so it will be easier if a revert commit needs to happen from this squashed PR commit.

api/utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+    >>> hist = get_histogram([1, 2, 3], 3)
+    >>> len(hist)
+    3
+    >>> hist[0]
+    {'count': 1, 'max': 1.6666666666666665, 'min': 1.0}
+    >>> hist[1]
+    {'count': 1, 'max': 2.333333333333333, 'min': 1.6666666666666665}
+    >>> hist[2]
+    {'count': 1, 'max': 3.0, 'min': 2.333333333333333}
```
toolness: Wooo doctests! I know this docstring was just copied over from the old version, but do you know if `py.test` is finding and running these?

api/utils.py
```diff
@@ -0,0 +1,72 @@
...
+    """
+
+    # When input array is empty, return array with single empty bin
+    # TODO: or just return an empty array or None?
```
toolness: Sorry, what was the original behavior here?  The `EMPTY_DATA` constant in `index.js` contains this key:

```js
        wage_histogram: [
          {count: 0, min: 0, max: 0}
        ]
```

So I guess what you're doing is the correct behavior... DId you verify by doing a search that gives 0 results?

api/utils.py
jseppi: I didn't even know those were a thing. Apparently they will not run unless specified on the command line
`py.test --doctest-modules`

ref: http://pytest.org/latest/doctest.html

toolness:
Cool, aside from my two questions I think this looks great!

api/utils.py
jseppi: Just tried it. That doctest fails because the keys in the dictionary are in a different order. Doesn't seem like a good test.

api/utils.py
jseppi: Good catch. I had it fixed in my test harness code but didn't modify it here. Will update.

jseppi:
After my most recent commit, this is what the front page looks like with a search that has 0 results. No javascript errors are logged in console, so I think this is good to go. Current production CALC actually throws a console error when a search has 0 results (which was fixed in a previous PR by our team, but is not in prod yet).

screen shot 2016-06-29 at 11 18 25 am

api/utils.py
jseppi: Removed the doctest.

toolness:
Woot!! Looks good to me, feel free to merge!

api/utils.py
toolness: Oh, I think there's ways to work around the ordering issue, actually, but if the same functionality is already being tested against in the unit tests then it's not a big deal!

jseppi: Good to know. It is already being tested in the unit tests, so we're covered.

PR #340, opened 2016-06-29
Coverage cleanup

jseppi:
Small cleanup from stuff I noticed in our coverage report.

- Remove unused view.py files
- Add docker_settings.py to coverage omit list


.coveragerc
```diff
@@ -3,13 +3,14 @@ branch = True
...
-  */migrations/*
-  */mommy_recipes.py
   docker_django_management.py
   manage.py
+  */migrations/*
+  */mommy_recipes.py
   */tests.py
   */tests/*
   hourglass/wsgi.py
+  hourglass/docker_settings.py
```
toolness: Oh shoot! `hourglass/docker_settings.py` actually should've been removed as part of #320... Do you think you could remove that file and this line too?

toolness:
Splendid! Sorry I didn't get rid of `hourglass/docker_settings.py` in #320, if you could get rid of that and the corresponding line in `.coveragerc` then this is good to merge!

toolness:
Nice, this looks great to me!

PR #341, opened 2016-06-29
fix get_histogram to work with Decimals

jseppi:
It turns out #337 would actually not pass the selenium_tests, which I neglected to run locally, and Travis didn't catch since we are ignoring them there 😁. The problem was that `get_histogram` was being called with `Decimal`s, but in the new `get_histogram` there is some float math (like `+ 0.5`) that would fail.

This PR also adds a test to show that an array of `Decimal`s will now work with the new `get_histogram`.


toolness:
Ah, nice catch!! Looks good to me.

toolness:
That's neat, I didn't actually realize Python had a built-in `Decimal` type.

PR #343, opened 2016-06-30
Add support for docker-machine's cloud providers

toolness:
This allows us to more easily deploy CALC to remote cloud providers--specifically, AWS EC2, which we can do at 18f if we ask Jez Humble for an AWS sandbox account. (Note, though, that these AWS sandbox instances are intended to be short-lived, and are wiped at the end of every week!)

## Advantages of doing this

* **Easily showing an experimental build to a client or another team member.** Using Docker on a remote cloud provider is a *lot* faster and more flexible than setting up a custom cloud.gov provider. Since we're not allowed to use tools like ngrok/localtunnel, this could be a nice alternative.
* **Running Selenium tests on Sauce Labs**.  We're not allowed to use [Sauce Connect](https://wiki.saucelabs.com/display/DOCS/Setting+Up+Sauce+Connect) to tunnel in to our local machines, but we can have them tunnel in to our EC2 instances.
* **Bringing our Docker configuration closer to production.** The `docker-compose.cloud.yml` provided in this PR still isn't a production setup, but it brings us closer to it, through its use of `waitress-serve` over `manage.py runserver` and its use of `collectstatic`, which will make our Docker setup more ready for the day when cloud.gov supports Docker.

## Breaking changes

This PR requires a few changes in developer workflow:

* Previously, Docker-based developers could get by without creating an `.env` file. Now they *have* to make one, and it needs to define `DOCKER_EXPOSED_PORT=8000` (or some other number), and it has to define `DEBUG=yup` (unless the developer really wants to run with production settings).
* Local Docker-based developers now need to symlink `docker-compose.local.yml` to [`docker-compose.override.yml`](https://docs.docker.com/compose/extends/#/understanding-multiple-compose-files).

Overall I'm not a fan of these breaking changes, because I think that local developers deserve the smoothest onboarding experience requiring minimal configuration. However, due to the way Docker Compose works, this isn't possible. 😞


.dockerignore
```diff
@@ -0,0 +1,27 @@
...
+/include
+/lib
+pip-selfcheck.json
+/src
+.ruby-version
+.sass-cache/
+*credentials*.json
+*manifest*.yml
+docker-compose.override.yml
+
```
toolness: Note that `.env` is *not* in this list. I feel kind of queasy about this, to be honest... Another alternative would be to exclude `.env` but tell developers to *copy* `docker-compose.cloud.yml` to `docker-compose.override.yml` (instead of symlinking), and then add environment variables to that instead of tinkering with `.env`.

However, this means that they'd have to keep an eye on any major changes made to `docker-compose.cloud.yml`... Ugh. I guess an alternative there might just be to do something with the [`extends` keyword](https://docs.docker.com/compose/extends/#/extending-services).

docker_django_management.py
```diff
@@ -169,7 +169,8 @@ def execute_from_command_line(argv):
     if IS_RUNNING_IN_DOCKER:
         if is_runserver:
             setup_docker_sigterm_handler()
-        wait_for_db()
+        if len(argv) > 1 and argv[1] not in NON_DATABASE_COMMANDS:
+            wait_for_db()
```
toolness: Had to make it possible to skip waiting for the database, since we're now running `manage.py collectstatic` in the `Dockerfile`, when the db container hasn't even been set up yet.

README.md
```diff
@@ -91,6 +91,8 @@ To use it, install [Docker][] and [Docker Compose][] and read the
 Then run:
 
 ```sh
+cp .env.sample .env
+ln -sf docker-compose.local.yml docker-compose.override.yml
 docker-compose build
 docker-compose run app python manage.py syncdb
 docker-compose run app python manage.py load_data
```
jseppi: Might want to also include `docker-compose run app python manage.py load_s70`

jseppi:
Minor comment about loading s70 data, but otherwise this is 👍

toolness:
Realizing now that the Dockerfile will need to bust up `gulp` for the app container to be deployable to the cloud. Hmmmmmm.

toolness:
@jseppi can you review 1de1a44e3772a529e9378d7124882efdbb67d6a3 and 6a766464f3c461cd3ca8590b3c60c470ab6b706c real quick?

Aside from those, I just want to make sure @hbillings has a heads-up on the changes this will introduce (specifically, that she'll need to do `ln -sf docker-compose.local.yml docker-compose.override.yml`) before merging.

jseppi:
👍 LGTM

PR #346, opened 2016-06-30
create manifests/ folder; target specific manifest for auto-deploy;

jseppi:
Attempt at getting our deployment under a little better control. This will likely change in the future, especially once we start pushing to a prod instance, but for now I think this should work well.

`manifest.yml` files are no longer gitignore'd. Rather, there is a `manifests/` folder where we will store manifest files WITHOUT sensitive information. Environment configs should be done instead via User Provided Services.

The `.travis.yml` deploy step has been updated to use the `manifest-staging.yml` manifest to deploy to the `calc-dev` space on pushes to the `develop` branch.

### DIFF
--- a/.cfignore
+++ b/.cfignore
@@ -23,4 +23,4 @@ pip-selfcheck.json
 .ruby-version
 .sass-cache/
 *credentials*.json
-*manifest*.yml
+/manifests
--- a/.gitignore
+++ b/.gitignore
@@ -23,4 +23,3 @@ pip-selfcheck.json
 .ruby-version
 .sass-cache/
 *credentials*.json
-*manifest*.yml
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ deploy:
   api: https://api.cloud.gov
   organization: oasis
   space: calc-dev
+  manifest: manifests/manifest-staging.yml
   on:
     branch: develop
 sudo: required
--- a/manifests/manifest-sandbox.yml
+++ b/manifests/manifest-sandbox.yml
@@ -1,0 +1,15 @@
+---
+services:
+- db-calc-sandbox
+- calc-env
+applications:
+- name: calc-sandbox
+  instances: 1
+  memory: 256M
+  disk_quota: 1024M
+  host: calc-sandbox
+  domain: apps.cloud.gov
+  buildpack: python_buildpack
+  command: bash cf.sh
+  stack: cflinuxfs2
+  timeout: 180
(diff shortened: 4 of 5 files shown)


toolness:
Nice!  Should any of these changes be reflected in [`deploy.md`](https://github.com/18F/calc/blob/develop/deploy.md)?  Or should we add info to that file to prevent the kind of confusion we had yesterday about how deployment works?

jseppi:
Yeah, we definitely need to sort out `deploy.md`. I suppose most of it can be cleared out and rewritten at this point since we are not yet doing blue-green deploys.

jseppi:
Updated `deploy.md` with what the current situation is.

deploy.md
```diff
@@ -1,38 +1,37 @@
...
+
+If you don’t already have one, request a Cloud Foundry account in #devops
+in Slack and download the Cloud Foundry CLI according to the instructions here:
 https://docs.18f.gov/getting-started/setup/
 
+To start, target the org and space you want to work with. For example, if you wanted to work with the staging space:
+`cf target -o oasis -s calc-dev`
+
+Manifest files, which contain import deploy configuration settings, are located
+in the [manifests](manifests/) directory of this project.
```
toolness: Will the `[manifests](manifests/)` link work when it's viewed on GitHub? I'm used to making these kinds of links absolute, but whatever works.

jseppi: Yup! https://github.com/18F/calc/blob/57a99e501951264c6242a185aa336c3717677ab8/deploy.md

https://help.github.com/articles/relative-links-in-readmes/

toolness: Oh awesome! That's really useful.

deploy.md
```diff
@@ -1,38 +1,37 @@
...
+  - calc-dev.apps.cloud.gov -> calc-dev space, calc-dev app
 
-You can edit the app attributes in the `manifest.yml` file, including what domains the site deploys to, how many instances are running, and how much memory they have. 
+TODO: Structure for production deployment once determined
 
 #### Environment Variables
 
 For cloud.gov deployments, this project makes use of a
-[Custom User Provided Service (CUPS)][CUPS] to get its configuration
+[User Provided Service (UPS)][UPS] to get its configuration
```
toolness: Heh, interesting, I didn't realize `cups` stood for "**Create** User Provided Service" and not "**Custom** User Provided Service". Thanks for the correction!

toolness:
This looks great, merging now!

deploy.md
jseppi: Yeah, I had the same misunderstanding originally.

PR #360, opened 2016-07-06
More compliant security

toolness:
This does a few things, most of which deal with [issues raised by Compliance Viewer](https://compliance-viewer.18f.gov/results/calc/current):

* Allows non-`DEBUG` builds to explicitly opt out of HTTPS for development purposes (ok, I cherry-picked this from #343 b/c I didn't want to deal with potential merge conflicts).
* Adds support for the [`X-XSS-Protection`](https://www.virtuesecurity.com/blog/understanding-xss-auditor/) header. If the browser detects the possibility of a reflected XSS attack, it won't render the page at all. (Another option is to tell the browser to sanitize the page itself and display that, but I thought not displaying the page at all would be more likely to eventually alert us of the XSS vulnerability.)
* Adds support for `X-Content-Type-Options: nosniff`, which prevents older versions of Internet Explorer and Chrome to perform MIME-sniffing on the response body.
* Adds support for secure session and CSRF cookies via Django's [`SESSION_COOKIE_SECURE`](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SESSION_COOKIE_SECURE) and [`CSRF_COOKIE_SECURE`](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-CSRF_COOKIE_SECURE) settings.


jseppi:
It looks like instead of using a custom middleware, we could rely on Django's SecurityMiddleware, but just need to enable some of its additional settings:

- nosniff: https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-SECURE_CONTENT_TYPE_NOSNIFF
- xss: https://docs.djangoproject.com/en/1.9/ref/settings/#std:setting-SECURE_BROWSER_XSS_FILTER

Any thoughts on the errors raised by QuantifiedCode, especially the SECURE_PROXY_SSL_HEADER one (https://www.quantifiedcode.com/app/issue_class/70VXf68D)? I'm pretty sure cloud.gov takes care of all the conditions specified in that link.

I think we can ignore (and probably turn off) the "Missing required field in settings" error.

jseppi:
Interestingly, those headers are actually being set already (before this PR) on working pages. It seems that the 404 error page however, is what doesn't include the headers. I found this by checking the raw scan results as json (https://compliance-viewer.18f.gov/results/calc/current?format=json) -- you'll see there that the header problems are being detected when requests `/docs/` and `/sitemap.xml`, which are both 404s for CALC. So, if we do want to use the custom middleware, the tests should probably also include checking the new headers on non-existent routes also.

toolness:
Ah good catch!  I tried searching for those other settings but didn't find them for some reason.

Fixed, and added tests for 404s!

jseppi:
👍 
Also, I disabled the two critical QuantifiedCode issues because they are mitigated/don't apply to us.

PR #361, opened 2016-07-06
reenable default travis email notifications

jseppi:
Remove the explicit disabling of email notifications, which should revert travis back to its default email notification behavior of emailing the committer who broke the build.


toolness:
Nice!

PR #362, opened 2016-07-06
Document API_HOST and WHITELISTED_IPS.

### LINKED ISSUE #301: Document API_HOST and WHITELISTED_IPS environment variables
`settings.py` seems to pull the `API_HOST` setting from the environment, defaulting it to `/api/`.  It looks like it's used to set the URL path to the root of the API but I'm not sure.

[Here's a search for `API_HOST` on the repo's codebase.](https://github.com/18F/calc/search?utf8=%E2%9C%93&q=API_HOST)

toolness:
This fixes #301. Hopefully everything is correct--I'm very new to this api.data.gov stuff, so a review would be helpful!


deploy.md
```diff
@@ -79,5 +79,44 @@ There is an example sandbox manifest at [manifests/manifest-sandbox.yml](manifes
 
 TODO Production application development once determined
 
+### Deploying to api.data.gov
+
+Most federal APIs are actually fronted by an [API Umbrella][] instance which
```
jseppi: I'm not sure if we should make a claim about most federal APIs here.

Consider rewriting:

> CALC's API is fronted by an [API Umbrella][] instance that proxies all requests to its API enpoints. This allows CALC ...

deploy.md
```diff
@@ -79,5 +79,44 @@ There is an example sandbox manifest at [manifests/manifest-sandbox.yml](manifes
 
 TODO Production application development once determined
 
+### Deploying to api.data.gov
```
jseppi: Maybe "Setting up" or "Proxying" instead of "Deploying"?

README.md
```diff
@@ -169,6 +169,16 @@ string), the boolean is true; otherwise, it's false.
   If this is undefined and `DEBUG` is true, then a built-in Fake UAA Provider
   will be used to "simulate" cloud.gov login.
 
+* `WHITELISTED_IPS` is a python-formatted array that specifies IPs that
+  the REST API will accept requests from. Any IPs not in the list attempting
+  to access the API will receive a 403 Forbidden response. Example:
+  `['127.0.0.1', '192.168.1.1']`.
```
jseppi: Given that this setting currently requires an `eval` statement to parse it, which is kind of yucky/dangerous, do you think just a comma-separated string of IP addresses that we could parse into an array would be a better option?

toolness: I agree, maybe we should file a separate issue about that though? Since it will require changing our deployments' environment variables and such?

jseppi: 👍 Issue opened https://github.com/18F/calc/issues/365

jseppi:
Looks good, merging!

PR #366, opened 2016-07-06
expect WHITELISTED_IPS to be a comma-separated string

### LINKED ISSUE #365: WHITELISTED_IPS should be a comma-separated string
Instead of a python-like array that gets parsed via `eval` (which is dangerous), we should make WHITELISTED_IPS a comma-separated string of addresses that gets parsed into an array.

jseppi:
instead of a python-like list that gets eval'd. fixes #365.

Seems slightly overkill to have this in its own method, but that makes it testable.


jseppi:
On merge, I'll need to update the user-provided service value for `WHITELISTED_IPS` to match the new string expectation.

jseppi:
Sorry for silly pep8 fix commits. My editor pep8 integration apparently broke somehow.

toolness:
Nice! I think the code itself is ready for merge--but should anything be changed in the `WHITELISTED_IPS` configuration of prod/staging before we merge?  Just want to make sure that any auto-deployment doesn't break as a result of this merge.

jseppi:
Prod won't be receiving the auto-deploy. I will make the change in staging now. Merging

jseppi:
UPS update and auto-deploy to staging worked 👍

PR #367, opened 2016-07-07
js and sass/css pipeline with gulp

jseppi:
Ready for review again. No longer using `django-compressor` at all, only `gulp`.

- legacy javascript files are now being concatenated (with source maps) via gulp. Moving forward on our data-capture pages, I would recommend we use browserify or something else that supports javascript modules. We will need to build gulp tasks for doing that, but it should be pretty straightforward now.
- SASS source files are now in `hourglass_site/static_source` and are built with `libsass` (via `gulp` and `gulp-sass`). @hbillings - please review this part in particular since you'll be the one dealing with it the most.
- ~~`django-whitenoise` has been replaced with `dj-static`. This really only affects the staging/and prod builds (or anything that runs via `wsgi.py` instead of `manage.py runserver`).~~

TODO:
- [x] change travis `deploy:on` back to `develop` after verifying the changes to travis work
- [x] update CodeClimate config with new `static_source` paths to ignore
- [x] fix broken travis build
- [x] get it working for our SASS files - ended up using `gulp`
- [x] investigate delegating to a separate build system (like `gulp`) as discussed at http://www.revsys.com/blog/2014/oct/21/ultimate-front-end-development-setup/
- [x] document how these changes affect development and production practices.
- [x] make sure it works after `collectstatic` -- had to reinstall `whitenoise`
- [x] get it working with docker
- [x] remove ruby sass stuff from `Dockerfile`
- [x] investigate replacing whitenoise with [dj-static](https://github.com/kennethreitz/dj-static)
- [x] handle conditionally included scripts and stylesheets (such as those at https://github.com/18F/calc/blob/develop/hourglass_site/templates/index.html#L12)
- [x] sourcemaps for minified css
- [x] concat existing javascript with gulp
- [x] remove django-compressor since it will no longer be used
- [x] get hourglass_site/static/js/tests.js working again
- [x] add manifest hashes to built assets for caching


toolness:
Aw, just when I was doing a happy dance for getting rid of whitenoise!

Dockerfile
```diff
@@ -11,8 +11,10 @@ RUN \
   mv /tmp/phantomjs-$PHANTOMJS_VERSION-linux-x86_64/ /srv/var/phantomjs && \
   ln -s /srv/var/phantomjs/bin/phantomjs /usr/bin/phantomjs
 
+RUN curl -sL https://deb.nodesource.com/setup_5.x | bash -
+
 RUN apt-get update && \
-  apt-get install -y ruby-full rubygems && \
+  apt-get install -y ruby-full rubygems nodejs && \
```
toolness: If node is using `libsass`, should we remove `ruby-full` and `rubygems` here, and `gem install sass` below?

docker-compose.yml
```diff
@@ -20,10 +20,10 @@ db:
...
-sass:
+gulp:
   build: .
   volumes:
     - .:/calc
-  working_dir: /calc/hourglass_site/static/hourglass_site
+  working_dir: /calc/
   entrypoint: python /calc/docker_django_management.py
-  command: make watch-style
+  command: sh -c "npm install && npm run gulp"
```
toolness: As an aside, this isn't ideal for development because the initial process ID of this command (`sh`) won't forward the SIGINT it receives to `npm run gulp`, which means that devs will have to wait a while for everything to gracefully shut down when they press CTRL-C.  Not a big deal though, I can try to fix it in a later PR.

gulpfile.js
```diff
@@ -0,0 +1,51 @@
...
+        .pipe(gzip({
+            threshold: '1kb',
+            gzipOptions: {
+                level: 9
+            }
+        }))
+        .pipe(gulp.dest(dirs.dest.style));
+});
+
+// watch files for chnages
```
toolness: Minor comment typo, `chnages` -> `changes`

hourglass/settings.py
```diff
@@ -122,13 +123,26 @@
 STATIC_URL = '/static/'
 STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 
+# if not DEBUG:
+#     STATICFILES_STORAGE = ('whitenoise.storage.'
+#                            'CompressedManifestStaticFilesStorage')
```
toolness: I guess this should probably just be removed entirely before merging this PR?

.travis.yml
```diff
@@ -7,12 +7,13 @@ matrix:
...
 before_install:
 - pip install codeclimate-test-reporter
+- npm install -g gulp-cli
 after_success:
 - codeclimate-test-reporter
 install:
 - pip install -r requirements.txt
 - npm install
-- npm run gulp -- build
+- gulp build
```
jseppi: The line that was replaced here did seem to be the problem. I think the `npm run gulp -- build` way of calling gulp as an npm script doesn't work until node. The `--` wasn't introduced until npm 2.0, and I have no idea what version Travis is running.

jseppi: It is guaranteed to have some version of node though: https://docs.travis-ci.com/user/ci-environment/#Runtimes

docker-compose.yml
jseppi: Cool, yeah, I couldn't figure out a better way to do it so that can be an exercise for the docker expert 😄

toolness:
Oh! Regarding not knowing what version of `npm` is running on travis, perhaps it might be useful to add a `npm --version` somewhere in our `.travis.yml`?  That way at least the version would get logged in the build log...

jseppi:
Ugh so many bad commits 😭

jseppi:
The node and npm versions in Travis's python environment are pretty out-of-date. That could be a potential source of build trouble in the future :\

screen shot 2016-07-08 at 4 19 25 pm

toolness:
Wow, that really is ancient!

Maybe we should be installing the same version in the `Dockerfile` to maintain, uh, dev-travis-CI parity? :grin:

jseppi:
See https://github.com/18F/calc/pull/371 for how we can install more recent versions of `node` and `npm`.

gulpfile.js
```diff
@@ -0,0 +1,43 @@
...
+gulp.task('build', ['sass']);
+
+// compile SASS sources
+gulp.task('sass', function () {
+    return gulp.src(dirs.src.style + paths.sass)
+        .pipe(sass())
+        .pipe(gulp.dest(dirs.dest.style))
+        .pipe(rename({suffix: '.min'}))
+        .pipe(minifycss())
+        .pipe(gulp.dest(dirs.dest.style));
```
toolness: Will this be building sourcemaps too?  It would be awesome but not required...  If we're not doing it, might be nice to file a separate issue for it.

manifests/manifest-sandbox.yml
```diff
@@ -13,3 +13,5 @@ applications:
   command: bash cf.sh
   stack: cflinuxfs2
   timeout: 180
+  env:
+    DISABLE_COLLECTSTATIC: 1
```
toolness: I wonder if it'd be useful to add a comment documenting why we're doing this, for future readers? e.g. something like:

```yaml
# Normally the python buildpack runs `manage.py collectstatic` itself, but
# because of our custom static asset pipeline, we'll be building the static
# assets on the deployer's machine and shipping them to cloud foundry instead.
```

Um... assuming that comment is accurate.

requirements.txt
```diff
@@ -16,7 +18,6 @@ pytest-django==2.9.1
 selenium==2.53.5
 six==1.10.0
 waitress==0.8.10
-whitenoise==3.2
```
toolness: 💃 🎉 💥 💃 🎉 💥 💃 🎉 💥

manifests/manifest-sandbox.yml
jseppi: Good idea to include the comment. The assets are not built on the deployer's machine, but rather on the target instance as part of the `cf.sh` script. See here: https://github.com/18F/calc/pull/367/files#diff-eb18ef1432f5da9e066368521d387d15R10

toolness: Ah yeah, nice! But to clarify, the CSS files *are* built on the deployer's machine, right?  They're just built to the `static` directories of the relevant django apps and then collected by `collectstatic`?

jseppi: Sorry, you're totally right. Already forgot what I did...

toolness:
I'm getting the following warnings when running `docker-compose up`, presumably because they're coming from `npm install`:

```
gulp_1  | npm WARN deprecated gulp-minify-css@1.2.4: Please use gulp-clean-css
gulp_1  | npm WARN deprecated graceful-fs@3.0.8: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
gulp_1  | npm WARN deprecated minimatch@2.0.10: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
gulp_1  | npm WARN deprecated minimatch@0.2.14: Please update to minimatch 3.0.2 or higher to avoid a RegExp DoS issue
gulp_1  | npm WARN deprecated lodash@1.0.2: lodash@<3.0.0 is no longer maintained. Upgrade to lodash@^4.0.0.
gulp_1  | npm WARN deprecated graceful-fs@1.2.3: graceful-fs v3.0.0 and before will fail on node releases >= v7.0. Please update to graceful-fs@^4.0.0 as soon as possible. Use 'npm ls graceful-fs' to find it in the tree.
gulp_1  | npm WARN prefer global node-gyp@3.4.0 should be installed with -g
gulp_1  | npm WARN optional Skipping failed optional dependency /chokidar/fsevents:
gulp_1  | npm WARN notsup Not compatible with your operating system or architecture: fsevents@1.0.12
```

Regarding `prefer global node-gyp`, we could potentially squelch that by installing it globally in our `Dockerfile`.  Not sure if it's worth it though, especially since the warning only appears during the first `npm install`/`docker-compose up` and not subsequently.

In fact, the only warning that appears during repeated `docker-compose up`s is the one about skipping the optional failed dependency.  I wish there was a way to squelch that, but if there isn't, no big deal.

What do you think of the deprecation warnings?

gulpfile.js
jseppi: Yeah, can do that no problem in this PR.
Do you think once we have sourcemaps that we should instead use the minified css files in our templates?

toolness:
So one funky thing that now happens is that, because gulp is now responsible for actually building the CSS, running `py.test` on a fresh checkout actually makes the `test_histogram_download_canvas_hidden` selenium test fail (which is understandable, since the CSS that should make the canvas hidden can't be found).  I think we just need to call out to gulp in the `setUpClass` method of the selenium tests.  I suggest either doing that in this PR, or just filing a separate issue about it.

toolness:
This all looks great but I am super bummed about no sourcemaps. 😭 Maybe we could try out the custom sourcemaps thing you mentioned in the original description?

toolness:
Or alternatively, I guess we could just avoid minifying the JS for now... blegh.

gulpfile.js
toolness: Sure! Hooray for dev/prod parity, if there aren't any major downsides!

jseppi:
Yeah, we can try that.
Any new JS will have sourcemaps since we'll be using gulp + browserify (or something else).

hbillings:
Yaaaasss, `libsass`!! The SASS part looks great to me -- love that you rejiggered the folder structure.  Will there be sourcemaps for the CSS, also, or just the JS? I couldn't quite tell (not a dealbreaker if there aren't, fo sho).

jseppi:
@hbillings - yup, source maps for CSS will be in (almost finished implementing).

docker-compose.yml
toolness: **Update:** I fixed this in #373 if you want to just merge that into this PR!

jseppi:
@toolness: regarding the deprecations. I got rid of one of them by switching to `gulp-clean-css`. The other warnings are somewhere down in the dependency tree. I am comfortable not worrying about them.

jseppi:
@toolness: Another option regarding sourcemaps for existing javascript is to ditch django-compressor altogether, and instead use something like [gulp-concat](https://www.npmjs.com/package/gulp-concat), which supports sourcemaps.

The downside is that instead of specifying the list of javascript sources in the html templates, is that they would have to be specified in the gulpfile. Not really a deal-breaker in my opinion.

toolness:
Yes!! Would love to just use gulp-concat. My mind can reason about how gulp works and how it makes static files, but django + static files has always been a mysterious weird frustrating thing that I do not grok.

jseppi:
ok, on it.

jseppi:
Hrm, we are going to want to add manifest hashes to the built js and css files for production caching. @toolness: ideas for doing this?

jseppi:
Ah, looks like we would use Django's [ManifestStaticFileStorage](https://docs.djangoproject.com/en/1.9/ref/contrib/staticfiles/#manifeststaticfilesstorage)

toolness:
Yeah, since we're using `ManifestStaticFileStorage` again, we might just want to switch back to whitenoise, bleh.  It was doing all the fun optimizations required to make those awesome. We should also only enable its `CompressedManifestStaticFileStorage` (or whatever it was called) when `DEBUG` is falsy to avoid [development hell](https://github.com/18F/calc/pull/334).

toolness:
Yeah! I think we should just revert back to the original `whitenoise` setup and we should be good to go!!!!

jseppi:
Failed CodeClimate check is mostly on legacy JS and CSS code, with a few minor issues on the gulpfile (mostly trailing commas). The issues found on the gulpfile will be obviated once we setup eslint with the airbnb config (#374).

Other tests are passing.

PR #373, opened 2016-07-11
Slightly improved node/gulp setup for django compressor PR (take two)

toolness:
This does ~~two~~ three main things:

* Makes the node dependencies of the project part of the container image. This seems to make sense since the Python packages are also part of the container image. It also means we don't have to run `npm install` every time we want to use `gulp` (or one of our other node dependencies) in a Docker container.
* Ensures that [`gulp` gracefully exits when CTRL-C is pressed during `docker-compose up`](https://github.com/18F/calc/pull/367#discussion_r70111105).
* [Fixes selenium tests to run gulp.](https://github.com/18F/calc/pull/367#issuecomment-231758520)

The one non-standard thing is that our node dependencies are no longer stored in a `node_modules` subdirectory of the project. But as long as we don't write any code that assumes it exists, we should be good. I think.


jseppi:
This means we have to rebuild when `package.json` is modified, right?

toolness:
Yup! Essentially we treat `requirements.txt` and `package.json` the same way, in that sense.

Issue #363, opened 2016-07-06
Parse error on XSS Protection header


jseppi:
<img width="290" alt="screen shot 2016-07-06 at 1 26 16 pm" src="https://cloud.githubusercontent.com/assets/697848/16629862/44f0b850-437d-11e6-8ea6-af1a89a2403a.png">

Looks like there is a problem with the XSS protection header, as shown in Chrome console screencap above. This is on staging instance.

Probably related to changes from #360

jseppi:
Looks like we have two XSS headers coming through on `/`, as shown in `curl` output below. I bet this is the problem.

<img width="382" alt="screen shot 2016-07-06 at 1 42 28 pm" src="https://cloud.githubusercontent.com/assets/697848/16630335/9311e05c-437f-11e6-9e31-3873f8ac6fa1.png">

jseppi:
Only one XSS header is coming through on the 404 response though:

<img width="434" alt="screen shot 2016-07-06 at 1 50 37 pm" src="https://cloud.githubusercontent.com/assets/697848/16630609/b0f5d06e-4380-11e6-93f2-e24aed5d9f17.png">

jseppi:
After some local testing, I think it is actually cloud.gov adding the `X-XSS-Protection` (all caps) header. It seems strange that it does add it to the `200` responses but not to the `404` responses.

toolness:
Yeah, that is definitely odd--maybe we should file a bug for cloud.gov?  Or multiple bugs:

* If a server is already delivering a `X-Xss-Protection` header, don't append stuff to it.
* If you do add an `X-Xss-Protection` header, add it to all requests, not just 404's.

jseppi:
Asked in #cloud-gov-support

jseppi:
Here is where cloud.gov is adding it https://github.com/18F/cg-secureproxy-boshrelease/blob/master/jobs/secureproxy/templates/config/nginx.conf.erb#L107

It looks like the only way around the double header are for us to only add X-XSS-Protection to 4xx responses (which the cloud.gov proxy isn't doing).

jseppi:
`X-Content-Type-Options` is also duplicated on `200` responses. It is also not duplicated on `40X`.

<img width="443" alt="screen shot 2016-07-12 at 8 08 42 am" src="https://cloud.githubusercontent.com/assets/697848/16768021/f4999428-4807-11e6-8e6e-6280a1a12b66.png">

PR #375, opened 2016-07-12
only add security-related headers to error responses

jseppi:
This seems weird to me, but if/until cloud.gov changes the way its nginx proxy works, I think we have to do it this way.

@toolness: Do you think we should put the behavior of only adding the headers to error responses behind a config value, and default to adding them to all responses?

Verified via my cloud.gov sandbox instance that the duplicated headers are no longer there, and that the header parsing error is no longer being thrown in Chrome.

screen shot 2016-07-12 at 10 41 44 am
 
Fixes https://github.com/18F/calc/issues/363


toolness:
Oy this is so weird! I really wish cloud.gov's thingys just worked in a more sensible way. It will make it especially cumbersome for folks outside of the federal government (e.g. state agencies) to reuse CALC.

hourglass/settings.py
```diff
@@ -67,7 +67,7 @@
 )
 
 MIDDLEWARE_CLASSES = (
-    'djangosecure.middleware.SecurityMiddleware',
+    'hourglass.middleware.ComplianceMiddleware',
```
toolness: Oh, um, could you also remove `django-secure==1.0.1` from the `requirements.txt`?  It turns out that I should've actually done that when upgrading to django 1.8 because django-secure was actually [merged into Django 1.8](https://docs.djangoproject.com/en/1.8/releases/1.8/#security-enhancements)... but we might as well do it now since you're removing the last vestiges of it from our codebase (I think?).

toolness:
> @toolness: Do you think we should put the behavior of only adding the headers to error responses behind a config value, and default to adding them to all responses?

Actually, I think that is a very good idea.  That way the project will work as expected for anyone who tries using this outside of the cloud.gov world. Also, presumably cloud.gov will eventually stop being weird, and when that happens it will be a lot easier to undo this hack, since we just disable the config value (and optionally remove the code associated with it).

hourglass/middleware.py
```diff
@@ -0,0 +1,22 @@
...
+
+    If the SECURITY_HEADERS_ON_ERROR_ONLY setting is True, then the headers
+    will only be added to error responses (400 and above). This behavior is
+    needed for cloud.gov deployments because its proxy adds the headers
+    to 200 responses, but not to error responses.
+
+    Otherwise, the headers will be added to all responses.
+    '''
+    def process_response(self, request, response):
+        if (not settings.SECURITY_HEADERS_ON_ERROR_ONLY or
```
toolness: oh, technically the "best practices"-y way of doing this actually to [raise a `MiddlewareNotUsed` exception](https://docs.djangoproject.com/en/1.8/topics/http/middleware/#marking-middleware-as-unused) during construction. But whatevs, it's no big deal.

toolness:
This looks great! You're welcome to bust up the `MiddlewareNotUsed` exception on construction if you want, or not. I'm cool with merging either way.

hourglass/middleware.py
jseppi: It will always be used, right?
If `SECURITY_HEADERS_ON_ERROR_ONLY` is `True`, it adds the headers to `4XX` and `5XX` responses.
If `SECURITY_HEADERS_ON_ERROR_ONLY` is False, it adds the headers to all responses.

toolness: Ohhhhhhh my bad!  Cool, merge away then!

PR #376, opened 2016-07-12
Run QUnit tests in phantomjs

toolness:
This should have been really easy but it took me a while because `whitenoise` is confusing.

I copied `runner.js` from [qunit-phantomjs-runner](https://github.com/jonkemp/qunit-phantomjs-runner) because:

1. Finding the file when it's installed via `npm` is actually non-trivial due to #373.
2. It's just one dinky file that doesn't change often.

If we really want, though, I can install it via `npm` and add code to locate the `node_modules` directory it's located in.

### DIFF
--- a/qunit_tests/tests.py
+++ b/qunit_tests/tests.py
@@ -1,8 +1,10 @@
 import os
 import subprocess
 
-from django.core import management
 from django.test import LiveServerTestCase
+
+from selenium_tests.utils import build_static_assets
+
 
 # This file was taken from https://github.com/jonkemp/qunit-phantomjs-runner.
 RUNNER_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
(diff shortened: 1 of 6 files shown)


toolness:
Uh, QuantifiedCode says "QuantifiedCode — We cannot find the commit for this pull requests."  I guess because GitHub is being weird right now?

jseppi:
I don't mind having `runner.js` committed, but I think we should include the text of its [LICENSE](https://github.com/jonkemp/qunit-phantomjs-runner/blob/master/LICENSE) and where it came from as a comment at the top of the file.

toolness:
@jseppi good idea--I just added it to the directory so that upgrading will be super easy (just copy two files instead of manually reconstructing the `runner.js` one). Also added a readme in that directory to explain our decision...

jseppi:
👍

PR #377, opened 2016-07-12
Eslint config

### LINKED ISSUE #374: Setup eslint
- Needs to handle (or ignore) existing legacy JavaScript sources and any new JavaScript files.
- Should use airbnb config as recommended by the front end guild https://github.com/18F/frontend/issues/98

jseppi:
Once we get settled on how we want to build our new script files, we can add a new `.eslintrc` to the corresponding subdirectory.

As mentioned in the line comment, I've `.estlintignore`'d the CALC 1.0 script files because they will require a lot of modifications to bring into line.

Fixes #374


.eslintignore
```diff
@@ -0,0 +1,9 @@
+static/
+htmlcov/
+hourglass_site/static/hourglass_site/js/built
+hourglass_site/static/hourglass_site/js/vendor
+hourglass_site/static_source/js/vendor
+
+# Ignoring legacy (CALC 1.0) JS files for now
```
jseppi: I've gone ahead and ignored the original CALC 1.0 scripts (for now). We can open another issue if we think they should be linted, but it is going to involve a wholte lot of refactoring to get it to meet even `airbnb/legacy`.

toolness:
Nice!

How will folks eslint their code? Would be great if there were a built-in way to do this (e.g. via test suite, or at least part of the docker container) so folks w/o editor support could easily know whether their PR will pass codeclimate/quantifiedcode checks...

jseppi:
I added a note about eslint and how to run it to CONTRIBUTING.md. Do you think that is sufficient? I always use an editor plugin so I'm not sure what other folks would prefer when not using one.

Once we have some new JavaScript to build and lint, we could also add it to the gulp workflow with [gulp-eslint](https://www.npmjs.com/package/gulp-eslint).

jseppi:
I went ahead and added `gulp-eslint`. It won't do anything now really since all the existing JS files are being ignored, but it will work for new files.

toolness:
Awesome, thanks for adding instructions to contributing!

In the near future, it'd be nice if we could offer a command that ran everything by all the linters, as well as the test suite, so you could use it as a general pre-check before submitting a PR that travis, codeclimate, and quantifiedcode will all be happy.

toolness:
I think this is good to go! 👍

jseppi:
I suppose we could add a django management command to do that relatively easily.

PR #378, opened 2016-07-12
Add super simple styleguide

toolness:
This adds a super simple styleguide at `/styleguide/`:

> screen shot 2016-07-12 at 6 17 59 pm

This PR isn't intended to add a full styleguide--more just a place where we can start documenting the various styles we're adding, which is useful because:

1. Documentation is awesome (as long as it's not out-of-date!)
2. It gives styles a concrete place to "live" if they're added to the codebase before the views that use them are ready.

The style guide uses [Prism](http://prismjs.com/index.html) for syntax highlighting. The general design is based off [Bootstrap's style guide](http://getbootstrap.com/css/#buttons) (I like how everything starts out un-collapsed, as opposed to the USWDS's, which hides the code snippets behind an accordion).

~~The actual CSS/JS of the styleguide is all inline and kind of janky--since it's a development aid rather than production-ready, I decided to opt for ease-of-implementation over best practices, but I can fix that up too.~~

Currently the `/stylesheet/` endpoint is always public, even on production, just like `/tests/` is.  There's some discussion on making `/tests/` inaccessible on production in https://github.com/18F/calc/pull/14#issuecomment-226478277; regardless of the outcome, I believe the styleguide's fate should be bound to that of the test suite.


styleguide/templates/styleguide.html
```diff
@@ -0,0 +1,75 @@
...
+$(function() {
+  $('.styleguide-example').each(function() {
+    var header = $('<h3>Example</h3>');
+    var html = $('pre code', this).text();
+    var renderedDiv = $('<div class="rendering"></div>').html(html);
+
+    renderedDiv.prepend(header);
+    $(this).prepend(renderedDiv);
+  });
+});
```
toolness: If we really, want, I think we could actually convert this JS into a [custom Django template tag](https://docs.djangoproject.com/en/1.9/howto/custom-template-tags/#writing-custom-template-tags), since it's just a hack to stay DRY. I was just lazy, basically.

styleguide/templates/styleguide.html
```diff
@@ -0,0 +1,75 @@
...
+.styleguide-example pre[class*="language-"] {
+  margin: 0;
+}
+
+.styleguide-example pre > code {
+  padding: 0;
+  margin: 0;
+  border: none;
+}
+</style>
```
toolness: I guess we could move this CSS into SASS, if we really want. I was lazy.

Lots of laziness going on in this PR.

styleguide/templates/styleguide.html
```diff
@@ -0,0 +1,75 @@
...
+  <h2>Primary Buttons</h2>
+
+  <p>
+    The color of primary buttons matches the
+    <a href="https://standards.usa.gov/visual-style/#palette">primary colors</a>
+    of the USWDS.
+  </p>
+
+  <div class="styleguide-example">
+    <pre><code class="language-html">{% filter force_escape %}<a class="button button-primary" href="#">Link</a>
```
toolness: I dislike that Django doesn't have a `{%- -%}` variant like Jinja2, that makes it so the newline a template tag is on gets removed from the HTML output. Because this is a `<pre>` it means that we need to smoosh everything into one line. I guess it's avoidable if we either make our own template tag or do some string trimming in the page's JS.

toolness:
Hmm I guess I will have to figure out how to tell codeclimate to ignore the styleguide.

jseppi:
👍 I hadn't seen those custom tags before. Pretty neat!

toolness:
Hooray! 🎉 

Yeah, I hadn't had experience with them either before this. At first I thought Django's template system was super minimal, but it's actually really extensible, and nicely documented too!

PR #385, opened 2016-07-13
set csrf cookie to have HttpOnly flag

### LINKED ISSUE #384: Compliance - Cookie no HttpOnly Flag
Caught during ZAP scan after manual seeding through login, admin screens, and logout.

> A cookie has been set without the HttpOnly flag, which means that the cookie can be accessed by JavaScript. If a malicious script can be run on this page then the cookie will be accessible and can be transmitted to another site. If this is a session cookie then session hijacking may be possible.

This is happening on the csrf cookie and the session cookie, but only on the session cookie when it is reset by logout. According to this stackoverflow [answer](http://stackoverflow.com/questions/33786736/session-cookie-httponly-flag-not-set-on-response-from-logout-django), that should be safe (meaning we should be able to mark it as a False Positive). We can set the csrf cookie to have the flag via https://docs.djangoproject.com/en/1.8/ref/settings/#csrf-cookie-httponly

Here is what they look like in the ZAP scan:
* session cookie on logout: `Set-Cookie: sessionid=; expires=Thu, 01-Jan-1970 00:00:00 GMT; Max-Age=0; Path=/`
* CSRF cookie: `Set-Cookie: csrftoken=<token>; expires=Wed, 12-Jul-2017 17:13:07 GMT; Max-Age=31449600; Path=/; Secure`

jseppi:
Simple change to set `CSRF_COOKIE_HTTPONLY` to `True`, described at https://docs.djangoproject.com/en/1.8/ref/settings/#csrf-cookie-httponly

The caveat to this is that any AJAX scripting that needs the csrf token will have to grab it from a hidden form field or something similar.


toolness:
Looks good, thanks!

PR #387, opened 2016-07-14
Add upload widget

### LINKED ISSUE #389: New JavaScript pipeline
Meta-issue for discussing new JS pipeline/build/style/etc.

My preference would be to use ES6 with npm-installed modules, built nicely into browser-ready ES5 via Babel and Browserify (as gulp tasks). However, that might be overkill, and we could potentially just go with what we're already doing, which is ES5, vendor'd libraries, and simple concatenation (via gulp).

Thoughts, @toolness @hbillings?

toolness:
**[The current work-in-progress can temporarily be viewed here.](http://54.87.247.188/styleguide/#upload)**

## To Do

- [x] Add unit tests. Maybe take inspiration from [bootstrap's test suite](https://github.com/twbs/bootstrap/tree/master/js/tests#readme)?
- [x] ~~Consider integrating [dragula](https://github.com/bevacqua/dragula).~~ Looks like it'll take some time to understand dragula and integrate it, so I'd rather do that in a separate PR.
- [x] ~~Consider filing a separate PR that fixes #389, and filing this on top of that.~~ Nah, we decided it's fine to rebase #391 atop this.
- [x] Maybe add `aria-hidden="true"` to `<span class="js-only">or drag and drop here.</span>` since it's not really relevant to non-sighted users (I think?).

## Notes

* One frustrating thing about using HTML5 Drag-and-Drop is that once you opt-in to it, you also have to do any file uploads via Ajax, because `<input type="file">` is read-only.  So we'll have to support that out-of-the-box too.
* This widget provides visual feedback when the user is dragging a file over it, it's just currently very subtle (switches background from off-white to white).  We can make it more obvious easily, though.
* The widget works without degradation in IE11 and the latest versions of Chrome, Firefox, Edge, and Safari.
* The widget reads fine on NVDA and VoiceOver. We dynamically add `aria-live="polite"` to the upload widget so that its changes are narrated to the user.

### DIFF
--- a/hourglass_site/static_source/js/common/upload.js
+++ b/hourglass_site/static_source/js/common/upload.js
@@ -51,7 +51,7 @@ $(document).ready(function () {
       $el.append(current);
     }
 
-    $el.data('upload', self);
+    $input.data('upload', self);
 
     if (!browserSupportsAdvancedUpload() ||
         this.hasAttribute('data-force-degradation')) {
--- a/hourglass_site/static_source/js/common/upload.js
+++ b/hourglass_site/static_source/js/common/upload.js
@@ -32,6 +32,7 @@ $(document).ready(function () {
     var $el = $(this);
     var $input = $('input', $el);
     var self = {
+      input: $input[0],
       isDegraded: false,
       file: null
     };
@@ -70,7 +71,7 @@ $(document).ready(function () {
       var files = dt.files;
 
       if (files.length > 0) {
-        $el.trigger('changefile', files[0]);
+        $input.trigger('changefile', files[0]);
       }
     });
 
@@ -78,11 +79,11 @@ $(document).ready(function () {
       var selectedFile = this.files[0];
 
       if (selectedFile) {
-        $el.trigger('changefile', selectedFile);
+        $input.trigger('changefile', selectedFile);
       }
     });
 
-    $el.on('changefile', function (e, file) {
+    $input.on('changefile', function (e, file) {
       self.file = file;
       $input.val('');
       setCurrentFilename(file.name);
--- a/hourglass_site/static_source/js/common/upload.js
+++ b/hourglass_site/static_source/js/common/upload.js
@@ -53,6 +53,12 @@
       $el.append(current);
     }
 
+    function setFile(file) {
+      if (file) {
+        $input.trigger('changefile', file);
+      }
+    }
+
     if ($input.data('upload')) {
       // We've already been uploadified!
       return;
@@ -88,20 +94,11 @@
       stopAndPrevent(e);
       $el.removeClass('dragged-over');
 
-      var dt = e.originalEvent.dataTransfer;
-      var files = dt.files;
-
-      if (files.length > 0) {
-        $input.trigger('changefile', files[0]);
-      }
+      setFile(e.originalEvent.dataTransfer.files[0]);
     });
 
     $input.on('change', function () {
-      var selectedFile = this.files[0];
-
-      if (selectedFile) {
-        $input.trigger('changefile', selectedFile);
-      }
+      setFile(this.files[0]);
     });
 
     $input.on('changefile', function (e, file) {
--- a/hourglass_site/static_source/js/common/upload.js
+++ b/hourglass_site/static_source/js/common/upload.js
@@ -30,6 +30,22 @@ $(document).ready(function () {
   $('.upload').each(function () {
     var self = $(this);
 
+    function setCurrentFilename(filename) {
+      $('input', self).nextAll().remove();
+
+      var id = $('input', self).attr('id');
+      var current = $(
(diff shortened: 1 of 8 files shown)


hourglass_site/static_source/js/common/upload.js
```diff
@@ -0,0 +1,119 @@
+/* global $ */
+/* eslint-disable prefer-arrow-callback, func-names */
+
+(function ($) {
+  // The following feature detectors are ultimately pulled from Modernizr.
+
+  function browserSupportsDragAndDrop() {
+    var div = document.createElement('div');
+    return ('draggable' in div) || ('ondragstart' in div && 'ondrop' in div);
```
toolness: Um, so this actually *isn't* in Modernizr, at least not anymore. It looks complicated: https://github.com/Modernizr/Modernizr/issues/57

hourglass_site/static_source/style/widgets/_upload.scss
```diff
@@ -0,0 +1,70 @@
...
+  text-align: center;
+
+  .upload-chooser {
+    display: inline;
+  }
+
+  input {
+    // TODO: This is just clearing out the margin-bottom from our standard
+    // CSS, which feels like a hack.
+    margin-bottom: 0;
```
toolness: Not sure what to do about this, as we're writing CSS that overrides and "cancels out" our own global CSS here.

There were a few similar situations in #381 too actually, but I forgot to mention them...

hbillings: I'd like to modularize our SASS files a bit more (and hopefully have time to start working on that this afternoon). Maybe we can address some of this sort of thing by doing that?

toolness: Sure sounds good to me!

hourglass_site/static_source/js/common/upload.js
```diff
@@ -0,0 +1,123 @@
...
+    var self = {
+      input: $input[0],
+      isDegraded: false,
+      file: null
+    };
+
+    function setCurrentFilename(filename) {
+      $('input', $el).nextAll().remove();
+
+      var id = $('input', $el).attr('id');
```
jseppi: I saw this and thought "there must be a jquery shortcut to get the 'id' of an element" ...but there's not apparently. I'm also not the only one to expect it: http://stackoverflow.com/questions/3239598/how-can-i-get-the-id-of-an-element-using-jquery#comment34799406_3239600

styleguide/templates/styleguide.html
```diff
@@ -19,6 +19,109 @@
...
+            XLS, XLSX, or CSV format, please.
+          </div>
+        </div>
+    </div>
+  </div>
+
+  <h3>Form Submission</h3>
+
+  <p>
+    Due to the technical limitations of HTML5, when the upload widget isn't
```
jseppi: I think it's more due to 'security concerns', right?

styleguide/templates/styleguide.html
```diff
@@ -19,6 +19,109 @@
...
+  <p>
+    The upload widget's <code>&lt;input&gt;</code> element does not currently
+    support the <code>multiple</code> attribute, however&mdash;it is intended
+    for single-file uploads only.
+  </p>
+
+  {% example %}
+  <form enctype="multipart/form-data">
+    <div class="upload">
+      <input type="file" name="file" id="file">
```
jseppi: Maybe consider adding the `accept` attribute to lock it to specific file types? I think that would be nice to have in the example.
I think you should be able to grab that attribute and check the file object against it in the `drop` handler.

jseppi:
This looks pretty darn good!
My biggest suggestion is to do some validation against the `input`'s `accepts` attribute (if it is present). You could add some classes and/or a message when an invalid file is dropped.

styleguide/templates/styleguide.html
toolness: Oh, I didn't realize that... I'm wondering what the security concern is though? Basically the thing that makes things hard is that `<input type="file">` is read-only... If it were possible to assign an HTML5 `File` object to it then that would make it really easy for us to just use standard HTML form behavior instead of Ajax. I understand why one wouldn't want to allow `fileInput.file = 'c:\users\secret_file.txt'` but in the case of `File` objects, we're guaranteed that it's something that either the app has auto-generated or the user has dragged into the app... and we're still sending it to the server regardless, only via Ajax instead of a POST request... So I'm not sure what the difference is?

styleguide/templates/styleguide.html
toolness: Good idea! Mind if I file that as a separate issue that we can add to this later? I'm concerned that this PR is already kinda hefty as it is.

jseppi: Sure thing!

PR #388, opened 2016-07-14
New Relic documentation and config

jseppi:
- Documented New Relic stuff in `deploy.md`.
- Updated `manifest-staging.yml` to include non-secret New Relic env vars
- Manually deployed to cloud.gov and verified that "CALC (staging)" is showing up in our New Relic console.

### DIFF
--- a/deploy.md
+++ b/deploy.md
@@ -57,6 +57,23 @@ cf uups calc-env -p credentials-staging.json
 cf restage calc-dev
 ```
 
+##### New Relic Environment Variables
+
+To enable integrated New Relic monitoring, cloud.gov-deployed versions of the
+the application are launched with `newrelic-admin` (see [`cf.sh`](/cf.sh)).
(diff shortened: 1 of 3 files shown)


deploy.md
```diff
@@ -57,6 +57,23 @@ cf uups calc-env -p credentials-staging.json
 cf restage calc-dev
 ```
 
+##### New Relic Environment Variables
+
+To enable integrated New Relic monitoring, cloud.gov-deployed versions of the
+the application are launched with `newrelic-admin` (see [`cf.sh`](/cf.sh)).
+Basic New Relic configuration is done in [`newrelic.ini`](/newrelic.ini), but
+additional settings need to be provided as "real" environment variables, i.e.,
+not through the user provided service. This is because UPS parsing is done in
```
toolness: At first I was confused about what UPS meant and thought it could use an `<abbr>` or something, but in context it looks like it's right after the section that explains what a `UPS` is, so all good.

toolness:
Nice, looks good!

PR #395, opened 2016-07-15
Add table of contents to style guide

toolness:
This gives us a handy table of contents to the style guide, and interactive anchors next to sections that we can use to generate hyperlinks:

> toc

In the above screenshot, the `#` in the margin (across from the "Upload Widget" heading) is being displayed because my cursor is over it; otherwise it doesn't appear. This is based on the interaction model of GitHub READMEs.

It's pretty barebones right now; we can't have subsections (so the subsections of the Upload Widget section are just static headings) and the TOC doesn't stay with you as you scroll along the page or anything, but it's a start! We can upgrade it later as our style guide becomes more complex.

### DIFF
--- a/styleguide/static/styleguide/styleguide.css
+++ b/styleguide/static/styleguide/styleguide.css
@@ -1,3 +1,34 @@
+.styleguide-toc {
+  background-color: #f0f0f0;
+  padding: 1em;
+}
+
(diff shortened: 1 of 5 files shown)


jseppi:
👍 Nice!

toolness:
Thanks!

PR #399, opened 2016-07-16
[DO NOT MERGE] Ultratap

jseppi:
Here's another idea I just had and coded up real quick. I use `tap` formatters for the tests, which is easily parseable in a consistent way. Then depending on the `--verbose` flag, only print errors or print all output.

Run via `python manage.py ultratap [--verbose]`

I couldn't find a tap output for `flake8`, but I didn't look very hard yet.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,9 +1,9 @@
 static/
 htmlcov/
+vendor/
+
 hourglass_site/static/hourglass_site/js/built
-hourglass_site/static/hourglass_site/js/vendor
 hourglass_site/static/hourglass_site/tests
-hourglass_site/static_source/js/vendor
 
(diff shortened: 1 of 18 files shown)


toolness:
Cool! Is there a reason the vast majority of these commits are by me, though, and already present in our `develop` branch? :)

jseppi:
Hrm, that's weird. I opened it against my `ultratest` branch instead of against `develop`.

jseppi:
Looks like `py.test` has pep8 and pyflakes plugins, and it also has a tap output formatter (already used in this PR), so we might be able to use `py.test` for all of the python checks.

jseppi:
Closing this now. I did more exploration and couldn't find TAP formatters for flake8 or pep8/pyflakes.

PR #402, opened 2016-07-18
Move upload.js to data-capture and ES2015-ify it

### LINKED ISSUE #400: Convert upload widget JS into an ES2015 module
Now that #391 is merged, I should refactor the widget JS from #387 into an ES2015 module and remove all linting issues (I think).

toolness:
This fixes #400.


hourglass_site/static_source/js/data-capture/upload.js
```diff
@@ -0,0 +1,122 @@
...
+
+  $input.on('changefile', (e, file) => {
+    self.file = file;
+    $input.val('');
+    setCurrentFilename(file.name);
+  });
+}
+
+$.fn.uploadify = function uploadify() {
+  this.each(function activate() {
```
toolness: Because this function requires `this` to be passed in by jQuery, it can't be an arrow function, and because of our linter settings, it needs to be given a name. Feels a bit weird.

toolness: This seems to be a recurring thing in general with jQuery-based code... because jQuery constantly passes `this` in to functions, a lot of code we write will often need to be old-style function expressions rather than arrow functions. Not sure if we want to relax the "all function expressions must be given names" requirement here or not?

styleguide/templates/styleguide.html
```diff
@@ -179,4 +179,5 @@ <h3 class="example-heading">Example (no JavaScript)</h3>
   {% endguide %}
 </div>
 <script src="{% static 'styleguide/vendor/prism.js' %}"></script>
+<script src="{% static 'hourglass_site/js/built/data-capture/index.min.js' %}"></script>
```
toolness: It might be nice to document somewhere (the style guide or something? I dunno) that the upload widget is part of the data-capture JS bundle...

toolness:
@jseppi any idea what's up with codeclimate here? When I click "Details" I get a codeclimate page that just says "No analysis results present."  Was hoping this PR would finally put an end to our codeclimate warnings for now...

jseppi:
I don't know. I'm finding it very annoying myself - it seems to have random failures quite frequently.
I wonder if we can get rid of CodeClimate since QuantifiedCode seems to meet our ATO needs on its own (https://pages.18f.gov/before-you-ship/security/static-analysis/).
I like having test coverage reported though :\

styleguide/templates/styleguide.html
jseppi: I think the style guide would be a nice place for info like that.

hourglass_site/static_source/js/data-capture/upload.js
jseppi: While I agree that this rule feels a bit weird in this case, one of the points of standardizing on a linter config is to avoid worrying about these kinds of relatively inconsequential style issues. I think I'd rather try to stick with the config defaults as much as we can.

jseppi:
Hrm, I think I may have found the problem with Code Climate. Looks like we have to use a specific "channel" in our code climate config. Will try it in another PR.

hourglass_site/static_source/js/data-capture/upload.js
toolness: Ah yeah, that makes sense! Sounds good.

jseppi:
Confirmed locally that the passes our `eslint`, so merge away if you are ready.

PR #403, opened 2016-07-18
enable eslint-2 channel in code climate

jseppi:
ref https://github.com/18F/calc/pull/402#issuecomment-233353335


jseppi:
The CodeClimate build log is saying that this is in fact working, but the integration is still reporting an error.

Ref: https://codeclimate.com/github/18F/calc/builds/199

jseppi:
I think it is (strangely) reporting an error because an old build (on the develop branch) is erroring, but once merged this _should_ fix it

screen shot 2016-07-18 at 11 06 16 am

PR #409, opened 2016-07-19
Ignore 'node_modules' in docker volume mount.

### LINKED ISSUE #401: Gulp explodes if node_modules is installed in calc folder on docker host
More details in https://github.com/18F/calc/pull/398#issuecomment-233340149.

It might be nice if we checked to see if we were in docker *and* `/calc/node_modules` exists; if that's the case, then we'll likely be loading modules installed by the host OS rather than by the docker container, in which case weird things can happen.

I wish there were a way to tell node to ignore the `node_modules` directory of a project under certain circumstances, as this would make the problem go away in docker, but there doesn't seem to be a way to do that.

Another slightly unfriendly solution would be to `rm -rf /calc/node_modules` in the entrypoint but that might be weird.

toolness:
This fixes #401.

Now only one annoyance remains when using nodeJS in Docker, which is that [ESLint doesn't respect `NODE_PATH`](https://github.com/benmosher/eslint-plugin-import/issues/142), so it will think that all our `require()` statements are bogus, but whatever.

### DIFF
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -1,9 +1,12 @@
 app:
   volumes:
     - .:/calc
+    # http://stackoverflow.com/a/37898591
+    - /calc/node_modules/
   command: python manage.py runserver 0.0.0.0:${DOCKER_EXPOSED_PORT}
 gulp:
   build: .
   volumes:
     - .:/calc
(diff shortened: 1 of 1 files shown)


jseppi:
Tested locally, looks good

PR #410, opened 2016-07-19
Ignore unresolved imports from eslint for Docker support.

toolness:
This is a workaround for https://github.com/benmosher/eslint-plugin-import/issues/142. I figure it's OK to ignore this warning since there's a very high likelihood that our code simply won't run / compile if there are any unresolved imports.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,8 +1,3 @@
 {
-  "extends": "airbnb-base",
-  "rules": {
-    // Until https://github.com/benmosher/eslint-plugin-import/issues/142 is
-    // fixed, we need to disable the following rule for Docker support.
-    "import/no-unresolved": "off"
-  }
+  "extends": "airbnb-base"
 }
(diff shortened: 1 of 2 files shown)


toolness:
Thoughts on this @jseppi ?  It's an annoying trade-off, for sure... another possibility might be to have ultratest check to see if its running in docker and supply the ignore information [via command-line](http://eslint.org/docs/user-guide/command-line-interface#rule) if so, so that non-docker tests don't suffer because of this.

jseppi:
I think I'd prefer it via command-line in `ULTRATEST` because I've actually had this rule be helpful when run eslint is running via in my editor's eslint plugin.

toolness:
Cool sounds good, I will bust out a change that does it via command-line!

toolness:
@jseppi ok I think this is ready now! Originally I was hoping eslint could take in configuration rules from the OS environment, in which case I was just going to add the rule to `Dockerfile` or `docker-compose.yml`, but apparently it can't, so I opted for doing it via command-line as previously discussed.

jseppi:
👍

PR #411, opened 2016-07-19
Add rainbow ascii art to ultratest

toolness:
screen shot 2016-07-19 at 10 11 09 am

This embeds an [ascii logo made from a text to ascii art generator](http://patorjk.com/software/taag/#p=display&f=Big&t=UltraTest) and combines with with some rainbow color code from https://gist.github.com/danielrehn/d2e6f2129e5f853c3166.

### DIFF
--- a/meta/management/commands/ultratest.py
+++ b/meta/management/commands/ultratest.py
@@ -3,6 +3,53 @@ import sys
 import subprocess
 import djclick as click
 
+# This is based on:
+#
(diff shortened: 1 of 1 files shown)


jseppi:
Ha, awesome. Maybe don't print on verbosity level 0?

toolness:
wat. it should display always because ultratest is ULTRA!

PR #412, opened 2016-07-19
Break CSS into component files

hbillings:
This is a first pass at making the CSS more reusable.  Folder structure now looks like this:
```
style
-base
  -browsers.scss
  -reset.scss
  -utils.scss
  -variables.scss
-components
  -footer.scss
  -header.scss
  -steps.scss
  -tables.scss
-vendor
  -skeleton
    -normalize.css
    -skeleton.css
  -no-ui.css
  -tooltipster.css
-data_explorer.scss
-main.scss
```
Some notes:
- `main.scss` is just a list of imports from other files.
- We're using Skeleton's reset and grid. We maybe should be using the WDS instead, because Skeleton hasn't been updated since December 2014. But that's going to require rewriting all of the grid HTML. :-/
- Added a gulp task to copy CSS files from vendor to the `static` dir. I've never written gulp tasks before, so someone should plz check that!


toolness:
Cool!  It looks like travis is failing because of `Error: Undefined variable: "$off-white"`?

toolness:
Hmm, I think it's probably OK to stick with Skeleton for now? Since it's not really a security-related thing and Skeleton's fairly simple and doesn't seem to have any pre-existing bugs and browsers' interpretation of CSS generally doesn't change in backwards-incompatible ways over time, I'm guessing it should be OK to stick with it despite its unmaintained nature.  I'm a bit concerned about how much work might be involved in rebasing all the legacy styling on uswds. 😣

hourglass/local_settings.py
```diff
@@ -0,0 +1,28 @@
+DEBUG=True
```
toolness: Hmm I think you may have accidentally added this file?  It used to be `.gitignore`'d but we ended up removing it entirely since we now use environment variabes/`.env` files for all configuration.

toolness:
@jseppi maybe you should review the gulp changes since you're most familiar w/ that file?

hourglass/local_settings.py
toolness: Oh, and it looks like getting rid of this file will also get rid of all our codeclimate and quantifiedcode fails too!

hbillings: @toolness buh, I thought I `git rm`ed that. Will remove! Thanks for catching!

toolness:
Hmm, it looks like the `p.schedule` styling on the about page isn't working... Here's what it normally looks like:

> screen shot 2016-07-20 at 7 57 19 am

and here's what it looks like on this PR:

> screen shot 2016-07-20 at 7 57 30 am

hourglass_site/static_source/style/main.scss
```diff
@@ -1,350 +1,11 @@
...
-
-fieldset {
-  margin: 0;
-}
-
-.no-js .js-only {
-  display: none;
-}
+@import url('../vendor/skeleton/normalize.css');
+@import url('../vendor/skeleton/skeleton.css');
```
jseppi: Is there a reason to prefer these `url` imports over the sass-style imports that they are replacing (up at https://github.com/18F/calc/pull/412/files#diff-e45366889250b5e7cfa46b5e5ecb5a6eL3)?

The sass-style imports will I think inline the contents of those files into the built file, whereas the `import url(...)` imports will require additional client requests to get those referenced files. Well, except that in the `gulp('sass')` task that I cribbed, it uses `cleancss()` which does an additional step of inlining files specified in `import url(...)` statements, but this only happens on the `*.min.css` versions of the files.

toolness:
When logged in, the buttons at the top overlap funkily:

> screen shot 2016-07-20 at 8 09 00 am

(I mean, we're probably going to be changing that particular styling anyways, but since this PR is meant to be a pure refactoring I figured I'd mention it.)

The django admin still looks nice and 90's.

jseppi:
You might want to add the css copying stuff to the `gulp('watch')` task also, otherwise changes to those files won't get automatically moved. Something like:

```js
gulp.task('watch', ['sass', 'css', 'js'], () => {
  isWatching = true;
  gulp.watch(path.join(dirs.src.style, paths.sass), ['sass']);
  gulp.watch(path.join(dirs.src.style, paths.css), ['css']);
  gulp.watch(path.join(dirs.src.scripts, paths.js), ['js']);
});
```

toolness:
Other than the two visual regressions I mentioned, I'm not seeing anything funky style-wise!

jseppi:
Hrm, actually is the `gulp('css')` task necessary? My original intention was to have the plain-old, non-built, vendor'd `css` files in `static/hourglass_site/styles/vendor/` and not require any task to move them around. Then, only the `scss` (and `css` files that would get sass imported into them) would be in the `static_source/style/` directory.

jseppi:
Looks like the `gulp('clean')` task will need to be modified also, depending on how we settle on managing vendor'd css files.
As it is now, it will delete files in `static/hourglass_site/style/vendor/`, which are source-controlled, so we probably don't want that. I had intended the `clean` task to just clear out built files.

hourglass_site/static_source/style/main.scss
jseppi: We might actually just want to do away with this [line](https://github.com/18F/calc/blob/db2ecfb3e83ed84741c8039269107a0733f055a0/gulpfile.js#L107), which writes the non-minified css files built from sass, though we aren't using them in our templates anywhere. We're just using the `*.min.css` versions.

hbillings:
@toolness After looking closer at Skeleton (and recalling the pain involved with updating grid frameworks in prior jobs), I agree we're fine to keep using it. I just have a knee-jerk reaction to building new things on something that might be abandoned. 

@jseppi 
5-dog-no-idea-what-hes-doing-300x229
I originally wrote that gulp task in to copy vendor files because I noticed there were different libraries in `static/hourglass_site/style/vendor` and `static_source/style/vendor`. That said, I doubt we'll really add or modify vendor libs with any regularity, so maybe all we need to do is make a li'l note that says you need to manually put the lib in `static/` and not `static_source`. I'm happy to remove it if that makes the most sense!

hbillings:
@jseppi Would it make more sense to just move `static/style/vendor` back to `static/vendor`? (I think that's how it was before, at least?)

.codeclimate.yml
```diff
@@ -7,7 +7,7 @@ exclude_paths:
 - "hourglass_site/static/hourglass_site/js/vendor/*"
 - "hourglass_site/static/hourglass_site/style/vendor/*"
 - "hourglass_site/static_source/js/vendor/*"
-- "hourglass_site/static_source/style/vendor/*"
+- "hourglass_site/static_source/style/lib/*"
```
jseppi: whoops, should be `libs/*`

.codeclimate.yml
```diff
@@ -7,7 +7,7 @@ exclude_paths:
...
 - "hourglass_site/static_source/style/skeleton/*"
```
jseppi: I think this line can be removed since `skeleton/` is contained within `libs/` now

hourglass_site/templates/index.html
```diff
@@ -3,7 +3,9 @@
 
 {% block head %}
 
-<link rel="stylesheet" href="{% static 'hourglass_site/style/built/index.min.css' %}">
+<link rel="stylesheet" href="{% static 'hourglass_site/style/vendor/no-ui.css' %}">
+<link rel="stylesheet" href="{% static 'hourglass_site/style/vendor/tooltipster.css' %}">
```
jseppi: `style/vendor/no-ui.css` and `style/vendor/tooltipster.css` no longer exist. Instead they are in `static_source/lib/`. I think they should probably be imported in `main.scss` like skeleton and normalize are.

jseppi:
Once my last line note comment is resolved then I think this is good to go.

toolness:
👍 from me--the about page looks good now, and while the buttons are still overlapping on the header, we talked today about fixing that up later so it's all good.

hbillings:
Good catch, @jseppi!! Fixed.

jseppi:
👍 Looks good, merge away!

Issue #421, opened 2016-07-20
"What's this?" link next to "Contract Year" form element is not keyboard-accessible


toolness:
The label is nice to screenreaders via an `aria-label` attribute, but it's not very friendly to sighted keyboard-only users, because there's no way to trigger the tooltip for displaying via the enter key. Might be useful to toggle its display when enter is pressed, and make it go away when the link loses keyboard focus.

Issue #416, opened 2016-07-19
Create help widget


toolness:
The mockups have this:

> <img width="350" alt="screen shot 2016-07-19 at 4 43 00 pm" src="https://cloud.githubusercontent.com/assets/124687/16965960/71e9df14-4dd0-11e6-83a4-a28d77ed1393.png">

I'm not sure what the interaction model is there but we gotta do it!

That said, we should also keep in mind the "Don't hide basic helper text" advice from [Design Better Forms](https://uxdesign.cc/design-better-forms-96fadca0f49c#.3uh59u4y6). I assume the reason @hbillings used the widget here is because the definition of "small business" might not qualify as "basic helper text".

hbillings:
I wasn't entirely sure what the definition of a small business was, heh. Maybe it's short enough to qualify as basic helper text! I'm also not sure if there are other items on this page that need explanation to the average bear.

As far as interaction, I was just thinking of using the tooltip that's already part of CALC. Thoughts?

toolness:
Good point, I had forgotten about our use of tooltipster!  I think that's fine, though we might want to make the tooltips keyboard-accessible too (just filed #421 about this).

toolness:
In #436 I just used `form.as_p()` to display the relevant form here. We'll probably need to change it (or make a custom django form widget) to display this tooltip/widget thingy.

toolness:
It looks like we're going to try *not* having tooltips like this if at all possible, so I'm going to close this issue for now... we can re-open it later if needed.

Issue #390, opened 2016-07-14
Modularize & organize SASS


hbillings:
This is sort of halfway done, but currently we just have `main.scss`, `index.scss` and a few partial files for variables and the like. I'd like to try to separate things out by the feature they relate to rather than the page they're on. There are also things in `main`, for instance, that really won't apply to pages beyond the index.

hbillings:
Closed in #412

PR #422, opened 2016-07-20
[WIP] Add keyboard-focusable tooltips

toolness:
As per #416 and #421.

focusable-tooltip

## To Do

- [ ] Replace some/all tooltips in the data explorer with this.
- [ ] Add unit tests.
- [ ] Do we want to have the widget itself add some of the attributes upon activation? For example, we could potentially have it add the `href="javascript:void(0)"` programmatically, and perhaps we should just use a `title` attribute and set `aria-label` programmatically too, since `title` will likely gracefully degrade better without JS. I dunno.
- [ ] Rebase this once #390 lands.

### DIFF
--- a/hourglass_site/static_source/js/data-capture/focusable-tooltip.js
+++ b/hourglass_site/static_source/js/data-capture/focusable-tooltip.js
@@ -1,0 +1,21 @@
+/* global jQuery, document */
+
+const $ = jQuery;
+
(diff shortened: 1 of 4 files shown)


toolness:
Because #416 isn't a concern right now, I'm closing this PR. We can always bring it back if we decide that we need tooltips urgently.

Issue #394, opened 2016-07-15
Add file-type acceptance logic/feedback to upload widget


toolness:
As mentioned in https://github.com/18F/calc/pull/387#discussion_r70912557, it'd be nice if the upload widget used the `accept` attribute of a file input widget to ensure that an upload was of the proper type, and provided feedback to the user if it wasn't.

Issue #397, opened 2016-07-15
Consider more contrast or a border highlight in upload dragover state


jseppi:
I'm finding it a little hard to notice the transition of background color from gray to white in the upload widget's `.dragged-over` class. Perhaps a border color change or more contrast in the two colors would work a little better.

toolness:
Agreed! I didn't want to be too garish or flashy (or violate WCAG contrast ratio guidelines) when adding the drag-over state so I opted for being too subtle, but it would be great if it could be more noticeable.

PR #423, opened 2016-07-20
Validate upload file type

jseppi:
Adds validation to the upload widget via the `accept` attribute on the file `<input>`.  Note that this behavior is not fool-proof, so server-side validation will obviously still need to be done. However, it should be slightly nicer user-wise because 1) their OS file chooser will only show accepted file types (unless they manually override it) and 2) non-accepted files will not populate the widget when dropped.

Another thing I had wanted to do was add a class when a non-accepted file type was dragged over but not yet dropped. Unfortunately the list of file objects doesn't seem to get populated in the drag event, only in the drop event, so that is not possible.

TODO:
- [x] add styleguide documentation of `accept` attribute
- [x] get qunit test to work via py.test. They currently work when viewed in a browser, but not when using the runner for some reason.
- [x] figure out how to show user feedback that an invalid file type was dropped

Here is what it looks like when an invalid file type is dropped or selected:

screen shot 2016-07-20 at 2 13 04 pm

Closes https://github.com/18F/calc/issues/394

### DIFF
--- a/hourglass_site/static/hourglass_site/js/upload_tests.js
+++ b/hourglass_site/static/hourglass_site/js/upload_tests.js
@@ -89,9 +89,17 @@
     });
     upload.trigger(evt);
     assert.strictEqual(input.data('upload').file, goodFileExt);
-
   });
 
+  advancedTest("advanced upload allows any file when 'accept' not specified", function(assert) {
+    input.attr('accept', null);
+    var fakeFile = {name: "boop"};
+    var evt = jQuery.Event('drop', {
+     originalEvent: { dataTransfer: { files: [fakeFile] } }
+    });
+    upload.trigger(evt);
+    assert.strictEqual(input.data('upload').file, fakeFile);
+  });
 
   advancedTest("'changefile' event triggered on drop", function(assert) {
     var fakeFile = {name: "boop", type: "application/test"};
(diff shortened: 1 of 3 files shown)


jseppi:
@hbillings or @toolness: any input on how the user should be notified that the dropped file was an invalid type?

jseppi:
The problem with the qunit_tests run via phantom was the `for...of` loop. Apparently babelify doesn't translate those to something more supported, yet some browsers still do not support them. This was pretty hard to debug because the output from `py.test qunit_tests` doesn't really indicate where the failure is happening, just the test case.

jseppi:
Wait, according to the Babel ES2015 preset [documentation](http://babeljs.io/docs/plugins/preset-es2015/), `for...of` should be properly transpiled, so now I am again confused as to why `py.test qunit_tests` didn't work.

@toolness: any ideas?

jseppi:
Perhaps any styling changes could be made along with #397.

toolness:
It looks like the `for...of` transform in Babel uses `Symbol`, which isn't defined in PhantomJS--I guess it's one of those ES6 constructs that Babel expects us to polyfill manually?

<details>
  <summary>Traceback from py.test</summary>
```
    Failed assertion: Died on test #1     at http://localhost:8081/static/hourglass_site/tests/qunit-1.16.0.js:251
    at advancedTest (http://localhost:8081/static/hourglass_site/js/upload_tests.js:21)
    at http://localhost:8081/static/hourglass_site/js/upload_tests.js:115
    at http://localhost:8081/static/hourglass_site/js/upload_tests.js:148: Can't find variable: Symbol
ReferenceError: Can't find variable: Symbol
```
</details>

PR #429, opened 2016-07-21
Include Babel polyfill

### LINKED ISSUE #427: Include the Babel JavaScript browser polyfill
This will prevent issues such as the missing `Symbol` type referenced in https://github.com/18F/calc/pull/423#issuecomment-234066640

https://babeljs.io/docs/usage/polyfill/

jseppi:
Closes #427 

also:
- update eslint
- revert back to `for...of` loop in `upload.js` - works in Phantom now, which is a good indication that the polyfill works.
- gulpfile now looks for an env var called `IS_IN_DOCKER`, which is used to turn off the `"import/no-unresolved"` since that one won't work in our docker setup.


jseppi:
I'm not sure why CodeClimate is erroring on the polyfill import. `eslint` reports no errors on that line locally o_O

jseppi:
Hrm, I think it might be because CodeClimate doesn't do an npm install, so it might never be able to resolve npm-installed modules, kinda bringing us back to #410

jseppi:
I've gone ahead and disabled eslint import checks in CodeClimate since it seems it will never get that right for npm-installed packages.

jseppi:
Ready for review, @toolness

.codeclimate.yml
```diff
@@ -16,6 +16,9 @@ engines:
   eslint:
     enabled: true
     channel: "eslint-2"
+    checks:
+      import/no-unresolved:
+        enabled: false
```
toolness: Weird, why is eslint raising this?  It's also raising it my `docker-compose up`: 

```
gulp_1  | [17:00:26] Finished 'js:legacy' after 37 μs
gulp_1  | [17:00:26] 
gulp_1  | /calc/hourglass_site/static_source/js/data-capture/index.js
gulp_1  |   5:9  error  Unable to resolve path to module 'babel-polyfill/dist/polyfill.js'  import/no-unresolved
gulp_1  | 
gulp_1  | ✖ 1 problem (1 error, 0 warnings)
gulp_1  | 
gulp_1  | [17:00:26] Finished 'lint' after 2.12 s
```

hmmmmm

toolness:
Does this PR not incorporate #410?  I'm getting all the `import/no-resolved` errors for a bunch of modules when running `ultratest` but [only the one `import/no-unresolved` for `babel-polyfill` when running `docker-compose up`](https://github.com/18F/calc/pull/429#discussion_r71911252) ... weird.

The unit tests seem to run fine though, which is good!

.codeclimate.yml
jseppi: Something to do with the non-standard location of `node_modules` within the docker setup?

jseppi:
I just merged in develop and re-ran ultratest via `docker-compose run app python manage.py ultratest`. It reported that eslint passed.

The eslint that runs as part of the gulp build will still report an import error on that import, however. I am pretty sure it is because of the non-standard location of `node_modules/`.

jseppi:
Bleh, I guess the options are:

1. Disable that check in our `.eslintrc`. I like that check but it seems to be creating problems with the dockerized version.
1. Figure out a way to turn off the import rule just in docker. We're already doing that for `ultratest` but not for `gulp`. I think we could make a new gulp task (maybe `docker-watch`) that is just like the current `watch` task, but that disables the import rule.
1. Anyone using the dockerized version could just ignore those eslint errors from the gulp container. 

Either way, that's probably a separate issue from this PR.

jseppi:
This should be good to go, @toolness. Let me know if GTG.

toolness:
Nice! I just added a quick comment to the gulpfile describing the reason for the extra docker-specific rule (copied/pasted from ultratest).

If you think it looks good, feel free to merge!

PR #430, opened 2016-07-21
add meta management commands to coverage ignore list

jseppi:
I think we can safely ignore meta commands from testing coverage.


jseppi:
Merging since this doesn't have any code changes to review.

Issue #431, opened 2016-07-21
Figure out how to include USWDS styles


jseppi:
Some options identified so far:

- manually copy USWDS scss partials on an as-needed basis
- `npm install` USWDS's node package and figure out (probably via gulp) how to move their built css files into our assets folder.

Related #414

jseppi:
Ok, it totally looks like using their `npm` package will work, so I will get that implemented.

jseppi:
I was wrong. This is actually kind of a pain to do. We can't really just import individual components because the components have a pretty large dependency tree, requiring several USWDS 'core' files and the bourbon library. Also, alerts (and probably other components) also require inclusion of USWDS images.

I would also guess that just including the compiled USWDS stuff as vendor'd assets won't work very well because there will be probably be rules conflicts.

This might not be worth doing. cc @hbillings

jseppi:
Another issue is going to be that in SASS, `@import` paths can't have variables in them (the variables don't get interpolated). This will be an issue for our docker environment which places `node_modules` in a different path from where it normally is.

hbillings:
For reference, I think these are the bits of USWDS we'll need to use:
- Colors (already abstracted into `variables.scss`)
- Responsive breakpoints (maybe)
- Tables (modified)
- Alerts
- Buttons
- Links (https://standards.usa.gov/typography/#links)
- Forms

Of these, forms are probably the biggest and most complex piece. 

Given that we already have a grid system in place, we're keeping the current typography, and we're modifying some styles (like tables), I think it makes more sense to riff off of the WDS than copy them over. IMO, that means we can just copypasta the relevant pieces. The downsides are that a) there's margin for error, b) we won't be able to update easily. I don't think the latter is a huge deal given that we're already heavily modifying some things, and the things we're using are the more stable portions of the standards..

We _will_ need to define responsive breakpoint variables, but I don't think that will be a huge pain (there's already a `breakpoint` function in our variables that could easily be renamed `media` to reduce error.)

jseppi:
Ok, I am working on copying over the USWDS `_alerts.scss` and associated images and vars as part of #414.

PR #432, opened 2016-07-21
[WIP] install uswds package

jseppi:


jseppi:
Closing because won't be installing USWDS, but instead copying over relevant pieces of it manually, for reasons described in https://github.com/18F/calc/issues/431#issuecomment-234356669

PR #434, opened 2016-07-21
Partials should have underscore prefix

### LINKED ISSUE #433: Prefix SASS component and base file names with underscore
Not really a pressing issue, but SASS partials/components/whatever you want to call them, should have their filenames prefixed with `_`.

Example: instead of `components/footer.scss`, it should be `components/_footer.scss`.

The reason to do this is that any SASS files that are prefixed with `_` won't be written as output files when they are compiled.

Right now our SASS build outputs all kinds of files that won't actually be used (see screenshot):

screen shot 2016-07-21 at 1 45 43 pm

hbillings:
Closes #433


jseppi:
👍 looks good, merge away once lights are green

PR #435, opened 2016-07-21
Error Alert

jseppi:
This started as a Price List-specific alert, but has since been simplified to a generic error alert (and should be easily extendable to other alert types). This also adds `data_capture.scss` (which gets built out to `data_capture.min.css` from our gulp script) for future Data Capture-specific styling.

It does not have expand/collapse functionality, and we've discussed that we won't add that functionality unless it seems necessary after getting user feedback.

I think the main takeaway from this was that trying to include any of the USWDS components could be quite difficult given that we have an existing styling base already implemented from the first version of this project.


jseppi:
@hbillings: Here is what I have currently. It is very simple now.

screen shot 2016-07-25 at 9 13 40 am

I didn't end up using any USWDS components as a basis for this because none of the really fit that well with our mock-ups and would have required a lot of customization.

If we want to add expand/collapse functionality, I think I'll need a bit more guidance on how we want that interaction to work. Like, should it work like the USWDS accordion, where there the entire element is a giant button? Or perhaps should just the plus/minus icons be buttons?

hourglass_site/static_source/style/components/_price_list_alert.scss
```diff
@@ -0,0 +1,40 @@
+
+.price-list-alert {
+  background-color: $light-red;
+}
+
+.price-list-alert-heading {
```
hbillings: It might make sense to change `.price-list-alert-heading` and `.price-list-alert-text` to `h3` and `p` elements nested under `price-list-alert`.

hourglass_site/static_source/style/components/_price_list_alert.scss
```diff
@@ -0,0 +1,40 @@
+
+.price-list-alert {
```
hbillings: Maybe we could abstract this name a little into something like `alert-error` so we could potentially use it for other states?

hbillings:
This looks 💯 to me! Will merge as soon as the Travis build finishes.

jseppi:
Sweet, thanks! The builds have been taking like 30+ minutes today :/

PR #436, opened 2016-07-21
Add a data capture app

### LINKED ISSUE #418: Determine business logic for data capture workflow
Open questions:

- [ ] Should users be able to "jump" from one step to another? For example, what if they decide, after they've filled out part of the "enter contract details" step, that they want to go back and upload a slightly modified price list? Is it OK for them to have to fill out the contract details again, or do we want to save their progress on that step?
- [ ] Should users be able to go back and change their form after they've submitted it? For example, in the case where a data administrator rejects some or all of their data, it seems like they should be able to re-upload a new price list without needing to fill out the same contract details all over again.

It's also possible that some or all of these questions don't need to be answered for this sprint.

### LINKED ISSUE #425: Add a janky "accept/deny" UI for submitted price lists
This doesn't have to be anything fancy because it's going to be used by the "wizard of oz" behind the curtain during user testing sessions for now.

toolness:
This adds a `data_capture` Django app to the project and will fix #418.

Still lots of TODOs embedded in the code!

Additional TODOs not mentioned in the code:

- [x] Delete gleaned data from the user session after the user completes the submission process.
- [x] Add an initial migration.
- [ ] File separate issue to make `window.alert()`s in edge cases in `step_1.js` more user-friendly.
- [ ] File separate issues for the styling TODOs; specifically, fixing up all inline `style` attributes to be classes instead, and dealing with the `<style>` tag in `step.html`.
- [ ] File a separate issue to add alerts via django's [messages framework](https://docs.djangoproject.com/en/1.9/ref/contrib/messages/).
- [ ] File a separate issue to add a test suite for `step_1.js`.
- [ ] File a separate issue to implement a Schedule 70 importer.
- [ ] File a separate issue to verify that `S` is for "Small Business" and `O` is for "Other".
- [ ] File a separate issue to make "Tell us" (as in "Tell us you'd like your schedule added to CALC") do something useful.
- [ ] File a separate issue to figure out whether/how to show non-field errors in the invalid rows table.
- [x] Add a test suite for `data_capture.models`.
- [x] Add a test suite for `data_capture.forms`.

This PR also adds a very janky mechanism for admins to approve/unapprove price lists, so it fixes #425. Basically we use django admin, which exposes submitted price lists:

> screen shot 2016-07-25 at 6 43 29 pm

The admin checks/unchecks the "Is approved" box and submits the admin form, at which point the back-end will notice the state change and propagate effects accordingly: in the case of approval, new `Contract` models will be added to the database, and in the case of unapproval, the `Contract` models will be removed from the database.


toolness:
Phew, this is a lot of code.

In retrospect, there are a few "scope creepy" things i added that, had I held back for later PRs, would've made the PR simpler:

1. The presence of the schedule dropdown, along with a desire to abstract away schedule 70-scraping code for the sake of simplicity, prompted me to make the schedule-scraping architecture pluggable.  This isn't super complicated in theory but it did add a fair amount of code to the final PR (especially because I decided not to use `pickle` for serialization, in favor of keeping things as secure as possible).

2. I frequently forgot to manually choose "Fake Schedule" during testing, which left the schedule dropdown defaulted to the non-functional "IT Schedule 70" and resulted in my upload constantly being rejected. This made me realize that this would likely end up being a frustration for our users, once we add multiple real schedules--e.g., someone might upload a price list for Region 10 but forget to set the schedule drop-down accordingly--so I implemented a mechanism whereby the back-end attempts to see if other schedule-scraping code can make sense of the upload before telling the user that their submission failed. I probably should've held off on this.

3. I made the submitted price lists browsable and editable via django-admin, which probably wasn't needed for resolving #425.  I guess this didn't really add much code to the final PR though, mostly just time spent tweaking things.

4. I made submitted price list approval undoable.  Again, this didn't actually add much code to the final PR, though.

Other than that, I'm not sure what I could've done differently... Maybe I should've just added tests for `step_1.js` last Friday and issued a PR then?  That would've left the data capture process non-functional, but it certainly would've been a lot easier to code review.  Then I could have submitted another PR for adding step 3 & 4, and another for adding the approve/unapprove functionality. Then again, because each change builds atop the other, it might have caused churn while I waited for code reviews, so it's hard to say.

toolness:
Oh, the only reason travis hates this PR is because I haven't yet added the migrations.  I kind of want to add those *after* the code review, since it means any refactorings to the db schema won't be migrations (though I guess I could just overwrite the initial migration too, hmm).

Also, codeclimate doesn't like my use of `window.alert()` in the JS.  They're being used in some rare edge cases though, so I'm wondering if we should just let those go for now. (I assume there's some way to tell eslint that those specific lines are OK?)

jseppi:
> I assume there's some way to tell eslint that those specific lines are OK?

Yeah, you can use a comment of the form `// eslint-disable-line <rule-name>` I think the rule you're referring to is `no-alert`.

hourglass_site/static_source/js/data-capture/step_1.js
```diff
@@ -0,0 +1,76 @@
...
+
+    if (upload.isDegraded) {
+      // The upload widget is degraded; we should assume the browser has
+      // minimal HTML5 support and just let the user submit the form manually.
+    } else {
+      e.preventDefault();
+
+      const formData = new window.FormData(form);
+
+      if (upload.file) {
```
jseppi: add a comment about the need for the `delete` and `append` here

toolness:
OK HERE GOES!!!

PR #438, opened 2016-07-22
Put JS test files through gulp pipeline

jseppi:
Now we can write our JS tests in ES6 and make use of our eslint config.

I left the 'legacy' `tests.js` out of the build pipeline. That file has also has now been renamed to `hourglass_tests.js` since it is mostly testing the `hourglass` lib.


jseppi:
Ready for review, @toolness

toolness:
Loks great, thanks!

PR #439, opened 2016-07-22
fix generation of style sourcemaps so they include the sass sources

jseppi:
Previously, I wasn't including enough of the steps of the SASS->CSS pipeline within the `gulp-sourcemap` transforms. Now when an element is inspected with developer tools, you will be able to see in which `.scss` file a rule is defined.

screen shot 2016-07-22 at 10 25 02 am


toolness:
Awesome, thanks!!

Issue #386, opened 2016-07-13
Steps widget uses an unusual shade of blue


toolness:
When I originally started the Steps widget, I used plain old CSS, and I later copy-pasted it to SASS when filing #380. In doing that, though, I forgot to convert the one shade of blue we use to a color defined in `_variables.scss`.

The weird thing, though, is that the shade of blue it's using, `#0872B9`, doesn't seem to be defined anywhere in `_variables.scss`... it is the shade of blue used in the original Sketch mockups, though.

@hbillings should I add this shade of blue to `_variables.scss` or reuse one of the shades already in there?

hbillings:
I think I threw that together before I realized that by importing the WDS symbols, I'd also automatically get the colors to pick from. I probably eyedropper'd something that was the `#0071bc` shade and it picked up `#0872B9` instead -- the former is correct and the latter is just me being sloppy. Sorry!!

toolness:
Oh ok, no worries!  You mentioned that you wanted to make some tweaks to the styling of the visualization widget--want to fix this issue while you're at it?

toolness:
Fixed by #442!

PR #442, opened 2016-07-22
Simplify steps widget

hbillings:
In addition to closing #386, this replaces the SVGs with CSS `counter` elements (which are compatible to IE7). It also checks for the presence of a `current` class on the list items to set the visual style of previous/next steps.


jseppi:
Whoa, cool, I hadn't seen `counter` rules before. Looks good to me!

PR #405, opened 2016-07-18
Add compliance masonry YAML files

toolness:
This is just an initial attempt at doing something with compliance masonry.

This is based off the files in `epa-notice`:

* [`compliance/component.yaml`](https://github.com/18F/epa-notice/blob/master/compliance/component.yaml)
* [`opencontrol.yaml`](https://github.com/18F/epa-notice/blob/master/opencontrol.yaml)

## To Do

- [x] Add gemnasium, since @cmc333333 just gave us support for that today.
- [x] Consider using [bandit](https://github.com/openstack/bandit), since EPA notice does and it looks cool.

PR #448, opened 2016-07-26
add gemnasium badge to readme

jseppi:
Related: https://github.com/18F/calc/pull/405

I don't know how often gemnasium updates, but right now it says our Django version needs to be updated from 1.8.13 to 1.8.14, but we have already made that upgrade in https://github.com/18F/calc/blob/develop/requirements.txt#L3


jseppi:
From discussion in slack today, it appears our Gemnasium admins have to manually mark issues as resolved.

toolness:
NIce, thanks!

PR #452, opened 2016-07-26
Incorporate some of WDS

### LINKED ISSUE #415: Review form styling and decide what new styles we need, if any
It looks like the mockups use:

* `<input type="text">` fields
* `<input type="number">` fields
* Checkboxes
* single-value `<select>` fields (i.e., dropdowns)
* Date inputs. On the mockups, these are styled as three separate text fields for month/day/year. We could consider providing some sort of widget for easy selection, perhaps, or not. See the video mentioned below for more possibilities.

We also need to figure out how to display errors for these forms. I'm not sure if CALC (or Skeleton) currently has any styles for this. (We might want to spin this out into a separate issue.)

## Associated Research Materials

* [18F Accessibility Guide section on forms](https://pages.18f.gov/accessibility/forms/)
* Alice Bartlett's [Burn Your Select Tags](https://www.youtube.com/watch?v=CUkMCQR4TpY) talk is about how unusable select widgets are for the general population (based on gov.uk user research) and provides alternatives to them. Also includes best practices for date entry.
* WebAIM's [Usable and Accessible Form Validation and Error Recovery](http://webaim.org/techniques/formvalidation/) has some particularly useful materials on how [present errors accessibly](http://webaim.org/techniques/formvalidation/#error). Interestingly, its advice on errors might conflict with [Design Better Forms](https://uxdesign.cc/design-better-forms-96fadca0f49c#.3uh59u4y6)' advice on "specify errors inline". Not sure.

### LINKED ISSUE #417: Decide if we need new table styling
I'm not sure if we can/should reuse the data explorer's table styling or not, but this is how tables appear in the mockups for the capture tool:

> screen shot 2016-07-19 at 4 52 34 pm

### LINKED ISSUE #431: Figure out how to include USWDS styles

hbillings:
This takes some of the relevant styles of the WDS and incorporates them into the existing CALC base. Closes #431.

The best way to review this is probably run the styleguide locally, as everything is demoed there. Things this PR adds:
- Alert styles for success, warning, info
- some utils, like screen reader and media breakpoints
- Alternate button styles (related: #470, #472)
- Variables for typography (typography itself still needs more work, but that could be a separate PR as we're diverging from the standards)
- Sortable tables (closes #417)
- Form styles, specifically error/success states, labels, inline labels, and dropdowns (closes #415)

EDIT:
The USWDS files have been moved to their own folder, which required some surgery. The alerts still need surgery, but I'm going to put that off until I go to clean up the styles (which I need to do soon anyway).

There's also now a readme in the `uswds` folder explaining exactly what I did.

**Open questions**
I had to write some hacky JS to deal with sorting numbers as floats/ints instead of strings. That currently lives at the bottom of the styleguide because I wasn't sure where it should live. I'm guessing that whole chunk should just be a function that runs on page load and looks for existing tables.

<del>It literally just occurred to me that a cleaner way to do this would have been to put all of the modified WDS files in a `style/uswds` dir. That would create more of a 1:1 relationship for anyone looking to update a component later on. I'm almost certain this is what I should do, but before I rip apart everything I just did, I'd like a gut check from folks who haven't been heads-down in this for a week. ^_^</del>


styleguide/templates/styleguide.html
```diff
@@ -359,4 +359,42 @@ <h3 class="example-heading">Example (no JavaScript)</h3>
 </div>
 <script src="{% static 'styleguide/vendor/prism.js' %}"></script>
 <script src="{% static 'hourglass_site/js/built/data-capture/index.min.js' %}"></script>
+<script>
+  //TODO: Figure out where this should actually live
```
jseppi: I think it could go in its own file in `static_source/js/data-capture`, and then you could `require(...)` it in in the `index.js`.

jseppi:
Dang, sorry that the `jquery-tablesort` lib didn't even have number sorting built-in :\ Does it seem to work ok? 

I also just noticed that in its README it even suggests using StupidTable. Looks like there's an [open issue](https://github.com/joequery/Stupid-Table-Plugin/issues/154) to add StupidTable to npm, btw.

jseppi:
> put all of the modified WDS files in a style/uswds dir.

That does sound like a good idea. Maybe even with a README in that directory that gives a broad overview of where the the files came from and what the big changes were?

hbillings:
@jseppi it's all good -- the tablesorter stuff gave me a chance to break up the CSS work! `jquery-tablesort` makes it pretty straightforward to implement custom sort functions, so it's easy enough to just add attributes based on what data type the column contains and then call the right parse method. (This seems similar to what Stupid Table does anyhow!)

hbillings:
Okay, I think this is ready for review again!

hourglass_site/static_source/style/components/_alerts.scss
```diff
@@ -1,40 +1,2 @@
...
-
-  tbody tr:nth-child(odd) {
-    background-color: inherit;
-  }
-}
-
-.alert-error {
-  background-color: $light-red;
-}
+@import ('..uswds/alerts');
```
jseppi: Looks like this import needs to be changed to `@import '../uswds/alerts';`

jseppi:
I made a PR (#487) with some minor fixes/changes. Otherwise this looks pretty good to me and I think we should get it merged in before it gets too divergent from other code changes. Other modifications that need to be made can happen in their own issues/PRs.

@toolness: thoughts?

toolness:
Yeah! Once the conflicts are resolved I think we should merrrrrrrge

PR #453, opened 2016-07-26
eslint fail on warnings for ultratest

jseppi:
Use `--max-warnings 0` arg so `npm run failable-eslint` fails on warnings


toolness:
Woooot thanks!

Issue #454, opened 2016-07-27
Fix inline and temporary styling in data capture app


toolness:
When I created #436, I was focused on the back-end logic (and the way the front-end connects to it), so I wrote temporary styling for everything (my goal was to make things not look so bad as to be downright *confusing*). In particular:

- [X] There were about 10 times I used inline `style` attributes; all but one of them were fixed by #508 and others:
  <details>
    <summary>Output of `grep -r 'style=' data_capture/templates/*`</summary>
    ```
data_capture/templates//data_capture/step_1.html:<p style="color: gray">
```
  </details>

  **Update:** Oh, that inline styling is just in a debug UI for developers, so I'll let it slide.

- [x] In `data_capture/templates/data_capture/step.html`, I added an inline `<style>` tag to style the `.errorlist` and `.helptext` HTML created by Django's default form rendering. We can either use these same class names, or we can write our own custom form rendering code and use different class names/markup. **Update:** Fixed by #508.

- [x] In `hourglass_site/static_source/js/data-capture/step_1.js` we use `$.css()` to style the form to indicate that it's being submitted via ajax. We should probably use a class name instead (but remember to leave `pointer-events: none` on it, since we don't want the user submitting the same form twice).  **Update:** This is part of #548. Should mark this checkbox once it's merged.

- [x] Also in `hourglass_site/static_source/js/data-capture/step_1.js`, we use `.hide().fadeIn()` to fade-in the new form once it's been retrieved via ajax.  We *might* want to use a CSS transition or animation here instead, but it's not required. If we decide to stick with the jQuery-based animation then we should probably remove the TODO in the code. **Update:** spun this off into #577.

jseppi:
I'm working on this for the alert with price list upload errors

PR #464, opened 2016-07-27
Price list alert markup and styling

jseppi:
Part of #454 - This removes the inline styles and matches the styleguide markup for the price list errors found during upload. It also brings in `data_capture.min.css` to the `step.html` template.

screen shot 2016-07-27 at 11 14 22 am

It still shows the red box'd errors in the table cells with errors. @hbillings: how should we display these error messages, via a tooltip perhaps?


jseppi:
The actual styling will probably be updated by @hbillings based on the mock-up she posted in slack yesterday, but hopefully the markup here will make that easier to bring in via a separate PR.

toolness:
Looks good to me, thanks!

Issue #463, opened 2016-07-27
Figure out whether/how to deal with non-field errors in uploaded price list rows


toolness:
In #436 we use Django's form engine to validate individual rows from uploaded price lists--we don't actually *render* these as forms, but we do use the underlying framework for validation.

One of the properties of this form framework is that it has a concept of [non-field errors](https://docs.djangoproject.com/en/1.9/ref/forms/api/#django.forms.Form.non_field_errors), which are errors with a form (or in this case, a row in a price list) that aren't associated with a particular field.

I'm not sure if this concept applies to any of the schedules we're importing; it could be the case that any validation errors we encounter are *always* associated with a particular field.

In any case, we're not currently rendering non-field errors, and there's a TODO in the code to figure out if we need to.

If it's determined that we need to, we'll need to come up with styling for it, and implement it.

If it's determined that we don't need to, we should remove the TODO.  We might also want to add some validation logic in the `BasePriceList` class, or somewhere else, that verifies that the invalid rows from a price list gleaning don't contain any non-field errors.

jseppi:
The `TODO` comment mentioned in this issue appears to be gone, and it looks like we have `{{ form.non_field_errors }}` in all the templates that would need them. Is this good to close, @toolness?

toolness:
Sure, that sounds good to me!

PR #465, opened 2016-07-27
Error messages via Django messages

### LINKED ISSUE #459: Make failed form submission in data capture app more obvious
In #436 I didn't add any kind of alert to tell the user that their form submission was invalid; instead, the form is just re-displayed with errors on it.  We might want to display a big alert near the top of the page that says something along the lines of "Your form submission was invalid" so that users have a better context about why they're still looking at the same form they just filled out.

We could potentially do this by:

* Using Django's [messages framework](https://docs.djangoproject.com/en/1.9/ref/contrib/messages/) in our base template to display any queued messages for the user to see.  This might be tricky to integrate with the ajax form submission in step 1, though, since we don't actually do a full-page refresh there.

* Just hard-coding some logic in the form templates that contains an `{% if form.errors %}` that conditionally displays a custom alert.

jseppi:
To Do:
- [x] Figure out a strategy for dealing with messages with AJAX'd content, as in `step_1` of Data Capture
- [x] tests

Closes #459

### DIFF
--- a/data_capture/templates/data_capture/step.html
+++ b/data_capture/templates/data_capture/step.html
@@ -56,15 +56,7 @@
   </li>
 </ol>
 
-{% if messages %}
-<div class="alerts">
-  {% for message in messages %}
-    <div class="alert alert-{{ message.tags }}">
-      <p>{{ message }}</p>
-    </div>
-  {% endfor %}
-</div>
-{% endif %}
+{% include 'data_capture/messages.html' %}
 
 {% block step_body %}{% endblock %}
 
--- a/data_capture/templates/data_capture/step.html
+++ b/data_capture/templates/data_capture/step.html
@@ -56,6 +56,16 @@
   </li>
 </ol>
 
+{% if messages %}
+<div class="alerts">
+    {% for message in messages %}
+    <div class="alert alert-{{ message.tags }}">
+      <p>{{ message }}</p>
+    </div>
+    {% endfor %}
+</div>
+{% endif %}
+
 {% block step_body %}{% endblock %}
 
 </div>
--- a/data_capture/templates/data_capture/step.html
+++ b/data_capture/templates/data_capture/step.html
@@ -58,11 +58,11 @@
 
 {% if messages %}
 <div class="alerts">
-    {% for message in messages %}
+  {% for message in messages %}
     <div class="alert alert-{{ message.tags }}">
       <p>{{ message }}</p>
     </div>
-    {% endfor %}
+  {% endfor %}
 </div>
 {% endif %}
 
--- a/data_capture/views.py
+++ b/data_capture/views.py
@@ -7,6 +7,8 @@ from django.template.loader import render_to_string
 from django.http import JsonResponse, HttpResponseRedirect
 from django.core.urlresolvers import reverse
 from django.contrib.auth.decorators import login_required
+from django.contrib import messages
+from django.template.defaultfilters import pluralize
 
 from . import forms
 from .schedules import registry
@@ -29,6 +31,11 @@ def step_1(request):
             if request.is_ajax():
                 return JsonResponse({'redirect_url': redirect_url})
             return HttpResponseRedirect(redirect_url)
+        else:
+            messages.add_message(
+                request, messages.ERROR,
+                'Sorry, that file could not be processed. Try another?'
+            )
 
     ctx = {
         'step_number': 1,
@@ -99,6 +106,12 @@ def step_3(request, gleaned_data):
             del request.session['data_capture:gleaned_data']
 
             return redirect('data_capture:step_4')
+        else:
+            messages.add_message(
+                request, messages.ERROR,
+                'Oops, please correct the error{} below and resubmit.'
+                    .format(pluralize(form.errors))
+            )
 
     return render(request, 'data_capture/step_3.html', {
         'step_number': 3,
--- a/data_capture/views.py
+++ b/data_capture/views.py
@@ -36,6 +36,8 @@ def step_1(request):
                 request, messages.ERROR,
                 'Sorry, that file could not be processed. Try another?'
             )
+            if request.is_ajax():
+                return JsonResponse({'status': 'validation_error'})
 
     ctx = {
         'step_number': 1,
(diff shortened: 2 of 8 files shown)


jseppi:
@toolness: I can't figure out an elegant way to use Django messages with the AJAX'd form, but what about instead of reloading/rebinding a form partial, we do a `window.location.reload()` on form error?

jseppi:
I suppose the downside of that is that we'd lose the values of fields that the user has filled out. For step_1, that is only the dropdown.

Another method I found while researching is described here: http://blog.birdhouse.org/2014/02/01/displaying-django-user-messages-with-angular-js/. Basically, you setup a GET endpoint for retrieving JSON'd user messages, and ajax that content in wherever you want to display the messages.

jseppi:
I went ahead and modified it to do a page reload on step 1 form validation error. This also slightly simplifies the JS in that the form does not need to be rebound.

Another thing we lose with this method is the non-field errors, but maybe that is ok - ref https://github.com/18F/calc/issues/463

jseppi:
We could also render the form partial with messages, but that could lead to duplicate messages in both the partial (`step_1_form.html`) and the outside template (`step.html`) if a message is already in the session when the user gets to step 1...

toolness:
Wait, I have a different potential solution!

What if we just include the messages in the partial?  It won't exactly be *consistent* in the place on the screen where we display the messages, but in the context of ajax I think that's ok because the user's eyes will be drawn to the fade-in of the ajax-ed in partial anyways.

There's two ways I can think of doing this:

1. Make a separate `step_1_ajax_response.html` that just iterates through the messages and then includes `step_1_form.html`.  Render this template instead of `step_1_form.html` when returning ajax responses.

2. So one funky thing about the way django busts up its messages is that just iterating through them is, *I think*, a destructive action: if you iterate through them twice, the second iteration won't actually retrieve any messages, because the first iteration actually marked them all as "processed". So we could potentially just include the iteration/display of the messages in `step_1_form.html`--they will essentially be a no-op when run in a non-ajax context, as the base template would've already iterated through them before getting to the form rendering. This approach would involve fewer templates but it also might be harder to understand and maintain.

We could, of course, also have a separate `step_messages.html` if we want, so that we don't have to copy-paste the display logic of the messages in two different places.

Anyways, I'd prefer if we didn't do the `window.reload()`--I'm not a fan of gaining messages while losing field errors, especially in case our step 1 form ends up becoming more complex as the app evolves... 😣

data_capture/templates/data_capture/step.html
```diff
@@ -56,6 +56,16 @@
...
+{% if messages %}
+<div class="alerts">
+  {% for message in messages %}
+    <div class="alert alert-{{ message.tags }}">
+      <p>{{ message }}</p>
+    </div>
+  {% endfor %}
+</div>
+{% endif %}
+
```
toolness: Maybe we should actually just include this in `base.html`?  That way any future functionality could use it too--although if the `alerts` class is only in our data capture CSS then that won't work. 

Makes me wonder if data capture should just have its own `data_capture/base.html` template that trivially extends `base.html` and adds the messages (and the `data_capture` CSS and JS, I guess).  That way future functionality we add won't need to duplicate all this stuff.  But I guess perhaps that's also a task for later on when we actually need it, as for the time being it's sort of premature refactoring.

jseppi: Yeah, having our own `data_capture` base template would probably be good. Let's do that later though. Filed #476.

jseppi:
Going to take a shot at your option (1).

jseppi:
This look good to you, @toolness?

I'll try to add some tests to check for the presence of messages in error responses also.

data_capture/views.py
```diff
@@ -29,6 +31,11 @@ def step_1(request):
             if request.is_ajax():
                 return JsonResponse({'redirect_url': redirect_url})
             return HttpResponseRedirect(redirect_url)
+        else:
+            messages.add_message(
+                request, messages.ERROR,
+                'Sorry, that file could not be processed. Try another?'
```
toolness: So given the proximity of the alert to the actual error message, I'm wondering if this is actually required here?  It's definitely useful for step 3, but here it feels a little redundant:

screen shot 2016-07-29 at 2 25 14 pm

I guess the assistance of a content designer might be helpful here, as I'm not sure how best to communicate the error to the user...

jseppi: Should that message be more generic, as it is in step 3?

toolness:
Nice!

One thing I realized is that because of the fact that we're no longer replacing the `<form>` with another `<form>`, but rather a form with a list of alerts, we're not replacing older alerts as new ones come in. That means that if I submit a bad file twice in a row, I get something like this:

screen shot 2016-07-29 at 2 28 59 pm

So we might want to actually enclose the ajax injection site in a `<div>`, and wrap the partial in the same `<div>` so that the same div gets replaced, and old alerts are cleared out when the form is re-submitted.

Er, come to think of it, we could actually probably get rid of `step_1_form_ajax.html`, and simply change `step_1_form.html` to contain the following:

```
{% if request.is_ajax %}
{% include 'data_capture/messages.html' %}
{% endif %}
```

That way the messages would only get included there if we're being retrieved as part of an ajax request. (Note that this would require [`django.template.context_processors.request `](https://docs.djangoproject.com/en/1.9/ref/templates/api/#django-template-context-processors-request), or we could just insert the value of `request.is_ajax()` into the context manually when we render the template I guess.)

I guess a third alternative would just be to avoid alerts entirely in the ajax part of the form as I mentioned in https://github.com/18F/calc/pull/465/files#r72839051. 😣

data_capture/views.py
toolness: Ah yeah, that'd be a good solution too!

jseppi:
This is the most complicated simplest thing

toolness:
yeah bluuugh maybe we should just can it and go isomorphic with react...

jseppi:
Good catch on the multi-messages though.

I can't tell what would be the best between the 3 alternatives you just presented. I am leaning toward option 2 because that seems pretty slick, and we might want to use `is_ajax()` in other places, and it gets rid of the confusion of multiple templates.

jseppi:
| yeah bluuugh maybe we should just can it and go isomorphic with react...
pr to change to react incoming...

jseppi:
Implemented option 2

toolness:
NOICE

PR #466, opened 2016-07-27
remove 'tell us' CTA under schedule selection.

### LINKED ISSUE #457: Remove the "Tell us you'd like your schedule added to CALC" link for now
In #436 I added this link because it was in the mockups, but I didn't know what the link was supposed to do, so I just had it pop up an alert that says "TODO IMPLEMENT THIS". We should make it actually do something useful, or remove the call to action entirely.

toolness:
This fixes #457. Replacing the CTA with [this trello card](https://trello.com/c/njr9gJ2m/84-as-a-co-whose-schedule-isn-t-yet-supported-by-calc-i-would-like-a-way-to-contact-the-calc-team-so-they-can-add-support-for-my-sc).


jseppi:
👍

PR #275, opened 2016-06-13
[DO NOT MERGE] Proof-of-concept for uploading Schedule 70 XLS price list files via website

toolness:
**This is a proof of concept, please do not merge it.** It's really just here to provide some visibility into discovery work currently being done, and to make it easy to see what code has been changed.

For now, this POC can be used at https://xls-calc-sandbox2.apps.cloud.gov/xls/.

Currently it lets you uploaded a schedule 70 proposed/awarded price list XLS--if you don't have one, you can use the [IT Schedule 70 PPL with one Labor Category Row filled out](https://docs.google.com/spreadsheets/d/1q8XTOpbWz3e56DOiWAUU0-GbzrvxlwCc5pzoWRGChTA/edit#gid=1245479766) google spreadsheet, just choose **File -> Download as -> Microsoft Excel**:

Once uploaded, it parses the spreadsheet using [`xlrd`](https://github.com/python-excel/xlrd) and shows you the relevant columns that will be imported to the database. It also highlights any invalid fields that require changes:

Lastly, we ask for some contract details that aren't available in the uploaded Excel file:

Once the user submits the form and no errors are found, the data is added to the CALC database.

## Unaddressed Concerns

The following are *not* addressed by this proof-of-concept, and this list is not exhaustive:

* User authentication. In this POC, *anyone* can submit data, they don't require a login. Obviously, this needs changing!

* Approval flows. In our meeting with Kelly on 6/13, we found that Kelly (and potentially other admins) needs some way of examining submitted PPL data and approving it for inclusion in the CALC database.

* User-friendly interpretation of XLS data. For example, "Bachelors" is case-sensitive and we don't accept misspellings or abbreviations of it. We could be a lot more forgiving, though we should probably also provide feedback informing the user what our "best guess" for ambiguous data is.

PR #467, opened 2016-07-28
Add support for gleaning Schedule 70 price lists

### LINKED ISSUE #460: Implement schedule 70 importer for data capture app
In #436 I added an architecture for supporting gleaning data from multiple schedules' price lists, and a stubbed-out a schedule 70 gleaner, but it just raises a `ValidationError` with a TODO in it right now.

We should pull out the XLS scraping logic from #275 and use it as the basis for the real thing.  After that we can finally close #275, since it won't be needed anymore.

One question I have about this one is in regards to whether we can make the [IT Schedule 70 PPL with one Labor Category Row filled out](https://docs.google.com/spreadsheets/d/1q8XTOpbWz3e56DOiWAUU0-GbzrvxlwCc5pzoWRGChTA/edit#gid=1245479766) public by adding it to the repository for use by the test suite. I'm hoping we can, because it doesn't actually contain any proprietary information... I guess if needed we could remove all the content from the other worksheets we don't use.

toolness:
Fixes #460.

Much of the gleaning logic is cribbed from #275.

To do:

- [x] Implement all abstract `BasePriceList` methods.
- [x] Add a test to ensure that we pay attention to `FEDERAL_MIN_CONTRACT_RATE`.
- [x] Pull education level constants from `contracts.models`.


toolness:
Ok @jseppi, the remaining stuff to do on this PR is super minimal but I have to head out--can you take a look at this PR and test it out locally?

jseppi:
Will do

data_capture/schedules/s70.py
```diff
@@ -1,13 +1,171 @@
...
+        # doesn't have a SIN.
+        if not sin.strip():
+            break
+
+        cval = functools.partial(safe_cell_str_value, sheet, rownum)
+
+        cat = {}
+        cat['sin'] = sin
+        cat['labor_category'] = cval(1)
+        cat['education_level'] = cval(2)
```
jseppi: I wonder if we can make this more robust. I'm worried about just checking the first cell of each row especially since I could easily see a spreadsheet coming in where the price list 'table' runs all the way down to the `Most Favored Customer’s Discount or Discount Range (MFC) __________.` cell.

Something also fragile, though maybe less so, could be to check across the entire width of the table for values being present?

jseppi: Whoa, this is supposed to be up at line 67: `if not sin.strip()`

data_capture/schedules/s70.py
```diff
@@ -1,13 +1,171 @@
...
+    price_including_iff = forms.DecimalField(
+        label="Price offered to GSA (including IFF)"
+    )
+
+    def contract_model_education_level(self):
+        for choice in EDUCATION_CHOICES:
+            if choice[1] == self.cleaned_data['education_level']:
+                return choice[0]
+
+        raise ValueError('{} is not a valid education level'.format(
```
jseppi: Is this the right type of Error you'd expect to be raised?

This functionality also needs to be covered by tests.

toolness: Oh, good catch. Hmm, so I think that we should actually do this whole thing differently... I will make a commit soon.

data_capture/schedules/s70.py
toolness: Er, isn't that at line 67? I'm confused by what you mean...

Agreed that the gleaning code is fragile, but I was thinking we could make it more robust once we see more examples of how vendors are actually filling out the sheet?

data_capture/schedules/s70.py
toolness: Ok, so I've changed things so that the function *always* succeeds because of the way we're cleaning the form.  I also changed the education level validator into a cleaner, because:

1. cleaners can actually "clean" their field values (rather than just validate), which will be needed eventually once we start normalizing the data instead of just validating it (e.g. converting "Bacchelors" into "Bachelors" etc).

2. As a cleaner, we can move the code right next to the code for `contract_model_education_level`, where it's easier to see how the cleaning ensures that `contract_model_education_level` won't throw.

data_capture/schedules/s70.py
toolness: I guess alternatively, do you want to take a stab at making the gleaning code more robust?  Feel free to add commits and more tests!

jseppi: > Er, isn't that at line 67? I'm confused by what you mean...

Yeah, I meant to comment on line 67, but my comment showed up at line 75. What happened was I started my comment on that line (67), then I pushed a commit so the page was out of date. I refreshed the page and the comment box was still filled out, but when I submitted it, it was then pointing at the wrong line (75).
Sorry for the confusion!

jseppi: > do you want to take a stab at making the gleaning code more robust?

I'll take a stab at a simple robustness change real quick.

jseppi:
I implemented a hopefully more robust row check in #473. If you like it but want changes, let me know. Otherwise feel free to merge this PR (#467) and we can revisit improved processing stop signals when we have more samples price list spreadsheets, as you said.

toolness:
Ok, I left comments on #473... I'm going to go ahead and merge this for now, but I'll file a separate issue about making the scraping better!

jseppi:
🎉

Issue #461, opened 2016-07-27
Add a test suite for JS used by step 1 of data capture app


toolness:
In #436 I added a `step_1.js` file that does all the logic required for submitting the form via ajax (which is required for supporting drag-and-drop file uploads).  I didn't add a test suite for it, though, and it should probably have one.

PR #470, opened 2016-07-28
disable/enable submit button based on upload file validity

jseppi:
Simple disable/enable of the Step 1 form's submit button based on the `changefile` event of the form's file input. The disabling only happens when JS is enabled, so this will degrade well.

The button `disabled` state could use some styling, but that should be taken of by https://github.com/18F/calc/pull/452


toolness:
Nice! The one thing this might break on, though, are situations where JS is enabled, but the advanced features of the upload widget are *disabled* (aka "degraded") because there's not enough browser support for all its requirements (drag and drop, xhr + FormData, etc).

So you might want to surround your new JS code with a check for whether the upload widget is actually not in degraded mode?

toolness:
I guess an alternative might be to make the upload widget still fire a `changefile` event even when it's in degraded mode... It *might* already do that, but I thought that I had it basically act as though JS didn't exist if not all browser features were present to make it work in a super-awesome way.

My rationale for this was that it's easier to test "no JS at all" and "fully awesome HTML5 browser" than it is to write tests (and UI) for every individual possible HTML5 feature that a browser may or may not have, but that might have been a bad assumption to make...

jseppi:
@toolness: I think I now am taking care of all cases. You are right that `changefile` is not triggered when the upload is degraded. However, the input will still fire its `change` event, so we can listen for that. However again, when the upload is not degraded, the drag-n-drop will not trigger the `change`, but it will trigger `changefile`. So --> we listen for both events and re-enable on either.

jseppi:
Also, when the upload is fully functional, selecting a file through the normal file selection window will cause BOTH `change` and `changefile` to fire, so the handler gets called twice, but that's not a problem since the handler is idempotent.

hourglass_site/static_source/js/data-capture/step_1.js
```diff
@@ -18,13 +18,28 @@ function replaceForm(html) {
...
+  // When we have a fully functional drag-n-drop upload, 'changefile' is
+  // fired on drop, while 'change' is fired on normal file browser selection.
+  // So, we need to listen to both events and reenable the submit button
+  // on either.
+  const enableSubmit = () => $submit.prop('disabled', false);
+  $fileInput.on('changefile', enableSubmit);
+  $fileInput.on('change', enableSubmit);
+
   $(form).on('submit', (e) => {
     const $uploadInput = $('input', $upload);
```
toolness: So I think `$uploadInput` is the exact same thing as `$fileInput` now... it'd be cool if we could just use `$fileInput` but I guess we can leave that as a refactoring for later when we do #461!

toolness:
Thanks, this looks great!  Feel free to merge it up.

hourglass_site/static_source/js/data-capture/step_1.js
jseppi: Oh whoops, didn't notice that. I'll fix real quick then merge.

jseppi:
hrm, my commit message got truncated! It should have read "use $fileInput in place of $uploadInput"

PR #472, opened 2016-07-28
create layout and markup for form button row

### LINKED ISSUE #419: Create next/previous component for bottom of multi-step forms
I don't think we currently have a layout strategy for this kind of thing, but I could be wrong:

> screen shot 2016-07-19 at 5 03 27 pm

jseppi:
This adds some markup and styling for adding a row of buttons to be used at the end of a form. 

Here is what it looks like in Step 2 of the Data Capture flow:

screen shot 2016-07-28 at 3 19 36 pm

Closes #419
 
To Dos:
- [x] @hbillings review
- [x] add to styleguide after review

I didn't include styling for the `.button-previous` and `.button-cancel` classes, because that should probably take place after #452 lands.


hourglass_site/static_source/style/components/_formcontrols.scss
```diff
@@ -0,0 +1,29 @@
+
+.form-controls {
+  position: relative;
+  margin-top: 2rem;
+  overflow: auto;
```
jseppi: This is to prevent the `.form-controls` div from collapsing when the `.submit-group` (which is floated) is the only child.

hourglass_site/static_source/style/components/_formcontrols.scss
```diff
@@ -0,0 +1,29 @@
...
+
+    button, .button {
+      float: right;
+      margin-right: 0;
+    }
+  }
+
+  .submit-label {
+    display: inline-block;
+    height: 38px;
```
jseppi: This was to keep the `.submit-label` span the same height as the button it is next to

hourglass_site/static_source/style/components/_formcontrols.scss
```diff
@@ -0,0 +1,29 @@
...
+    color: $dark-gray;
+    text-align: right;
+    margin-right: 1rem;
+    max-width: 160px;
```
jseppi: I just picked this `max-width` to kinda look like the width that was in the mock-ups, but it doesn't match exactly with every page of the mock-ups.

data_capture/templates/data_capture/step_2.html
```diff
@@ -98,10 +98,18 @@
...
 
-<a href="{% url 'data_capture:step_1' %}" class="button">Previous</a>
-
-{% if gleaned_data.valid_rows %}
-<a style="float: right" href="{% url 'data_capture:step_3' %}" class="button button-primary">Next</a>
-{% endif %}
+<div class="form-controls">
+  <a href="{% url 'data_capture:step_1' %}" class="button button-previous">Previous</a>
+
+  <a href="{% url 'index' %}" class="button button-cancel">Cancel</a>
```
toolness: Thanks for adding this "cancel" button, I should've made an issue for it after not adding it in #436.

One question I have about it is whether it should actually be a POST that deletes the gleaned data from the user session--or if we should file a tech debt issue to address that later.  I suppose mostly it might be useful from a privacy perspective, e.g. if someone accidentally uploads some sensitive data and expects that cancelling the process purges the uploaded data from the server.

hourglass_site/static_source/style/components/_formcontrols.scss
hbillings: I just pushed a `.clearfix` util to my giant CSS PR that should take care of this!

hourglass_site/static_source/style/components/_formcontrols.scss
```diff
@@ -0,0 +1,29 @@
+
+.form-controls {
+  position: relative;
+  margin-top: 2rem;
```
hbillings: Generally, use `em` for spacing and `rem` for font sizes (`em` is relative to its parent; `rem` is relative to the root font size, so `em` tends to work a little better for positioning and padding things within a parent container).

data_capture/templates/data_capture/step_1_form.html
```diff
@@ -24,6 +24,13 @@
   </div>
 </div>
 
-<button type="submit" style="float: right; margin-top: 2em" class="button-primary">Next</button>
+<div class="form-controls">
+  <div class="submit-group">
+    <span class="submit-label">
```
hbillings: I *think* this should be a `label` element with a `for` attribute on it so that screen readers will know that this is describing the submit button. Or maybe we just need an [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute)on the button itself? @toolness, what do you think?

hourglass_site/static_source/style/data_capture.scss
```diff
@@ -1,2 +1,3 @@
 @import 'base/variables';
 @import 'components/alerts';
+@import 'components/formcontrols';
```
hbillings: Would it be too big a pain to rename this, maybe to something like your branch name? There are some CSS bits referred to as form controls that we might need to bring over from WDS, and they're things like range sliders and checkboxes, so I'm afraid it might be confusing later.

data_capture/templates/data_capture/step_2.html
jseppi: Yeah, it should clear out the data for sure. Filed in https://github.com/18F/calc/issues/478

data_capture/templates/data_capture/step_1_form.html
toolness: Oh good question! Will play around w/ it on screenreaders and report back.

jseppi: My thoughts were that the text we have there isn't _quite_ what I would consider a label for the submit button. It's more helper text about the next step. Though it is definitely related to the button, so maybe I'm having a weird internal semantics argument.

jseppi:
I took care of all @hbillings's line comments except for the `submit-label` accessibility question. Waiting on ideas from @toolness for that.

toolness:
Actually, want to just go ahead and merge this, and I'll file a separate tech debt milestone issue for the label?

jseppi:
Yeah, that works. I'll merge after I finish the styleguide entry and get the 👍 from @hbillings.

jseppi:
Styleguide entry is done.

toolness:
Ok, just filed #480 to address the accessibility issue. If everything else looks good, feel free to merge!

hbillings:
aww yiss! go for it!

PR #473, opened 2016-07-28
add bad row threshold checking to signal when to quit processing

jseppi:
@toolness: What do you think of this idea? To determine if a row is good or not, this now checks that each cell in the row is not empty. 

However, we might want to not stop processing after just a single empty value, so there is a threshold of bad rows that must be encountered continuously before quitting. I set that threshold to just 1 to mimic the current behavior.

I think at the very least this prevents trying to process a row that has a single value in the first column, but nothing in the rest of the row. That scenario could easily happen in the example spreadsheet if the table had entries that extended all the way down to the `Most Favored Customer’s Discount or Discount Range (MFC) __________.` cell.

I could see in the future trying to save 'bad rows' and validating them to show which cells were empty (among other errors).

### DIFF
--- a/data_capture/schedules/s70.py
+++ b/data_capture/schedules/s70.py
@@ -11,6 +11,8 @@ from contracts.management.commands.load_data import FEDERAL_MIN_CONTRACT_RATE
 
 logger = logging.getLogger(__name__)
 
+BAD_ROW_THRESHOLD = 1
+
 
 def safe_cell_str_value(sheet, rownum, colnum, coercer=None):
     val = ''
@@ -27,6 +29,14 @@ def safe_cell_str_value(sheet, rownum, colnum, coercer=None):
             pass
 
     return str(val)
+
+
+def is_likely_row(sheet, rownum, rowlen=13):
+    # check that the given row's cells are not empty
+    for col in range(0, rowlen):
+        if not safe_cell_str_value(sheet, rownum, col).strip():
+            return False
+    return True
 
 
 def glean_labor_categories_from_file(f, sheet_name='(3)Labor Categories'):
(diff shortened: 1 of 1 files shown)


toolness:
Cool!  I like the idea, my only concern about this approach is that I don't know how "required" all the fields actually are, especially since we don't even use most of them... It will be really helpful to actually see some filled-out price lists, I think.

I'm also thinking about worse-case scenarios... If we use the heuristic in this PR, then we risk throwing out rows that might actually seem valid to `Schedule70PriceList`, or might be *intended* to be valid by the vendor... those rows won't even appear in the "invalid rows" alert of step 2. In that sense the failure is sort of "silent", in a way.

On the other hand, if we use the heuristic in #467, we *do* risk interpreting `Favored Customer’s Discount or Discount Range (MFC) __________` as a row, *but* the only consequence of that is that it shows up in the "invalid rows" alert, and when the submitter sees that, they might be confused for a moment, but once they realize that the gleaner was just mis-interpreting the sheet's footer as a row, they'll be OK with throwing it out.

So I'm not sure which one I like better, but I think I'm leaning towards "be loud about what we think is invalid", so that users know the gleaner can't make sense of something, rather than ignoring it, but I'm open to other interpretations!

jseppi:
Yeah, you're right. I totally agree about not being silent on errors.

Let's stick with what you have now until we have a bigger corpus of sample price lists and/or we better understand which fields are totally required. We at the very least know that the SIN is required, right? Perhaps the easiest enhancement to your original heuristic would be to make sure both the `SIN` cell and the `price_including_iff` are present? 

If we do end up using something like I've made, especially if using the threshold, we should somehow let the user know which rows were thrown out.

toolness:
Cool, that sounds good to me!

Want to submit a PR that also checks `price_including_iff` or should I do it?

jseppi:
If you're free, go ahead and do it. I'm still slogging through the 500 issue and PR comment emails I had waiting for me this morning :)

jseppi:
Closing in favor of #479.

PR #475, opened 2016-07-29
Use slim header for data capture.

### LINKED ISSUE #383: Make a "slim" version of the site header
I realized when adding the style guide in #378 that our site's header is kind of tall:

> screen shot 2016-07-13 at 11 30 05 am

Should that header be shown in its entirety on every single page on the site, or should we show a slimmed-down one (perhaps one that just contains "CALC" and the "About this tool button") on pages other than the index/homepage?

Thoughts @hbillings ?

toolness:
This fixes #383 by enclosing the non-slim part of the header in a `header_extension` block, which allows child templates to redefine it as empty (or, alternatively, add their own custom extension content). It also shuffles around some CSS so that there's a decent bottom margin on the slim version of the header.

Here is the normal header:

> fattington-header

And here's the slim one:

> slim-header

In the future we can potentially reverse things so that the header is actually slim by default, and the index and about pages are the only ones that include the big version, but I wanted to make as few changes as possible to the legacy codebase for now.

### DIFF
--- a/hourglass_site/static_source/style/components/_header.scss
+++ b/hourglass_site/static_source/style/components/_header.scss
@@ -34,6 +34,9 @@ header {
   background-repeat: repeat;
   background-size: auto 100%;
 }
+.base-header {
+  margin-bottom: .6em;
+}
 .logo {
   display: inline-block;
   width:140px;
@@ -43,7 +46,6 @@ header {
   font-size: 2.2em;
   font-weight: 400;
   margin-bottom: .3em;
-  margin-top: .6em;
 }
 .description-secondary {
   color: rgba(255,255,255,.6);
--- a/hourglass_site/templates/base.html
+++ b/hourglass_site/templates/base.html
@@ -41,19 +41,19 @@
     <header>
       <div class="container">
         <div class="base-header">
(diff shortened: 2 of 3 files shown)


hourglass_site/templates/base.html
```diff
@@ -40,6 +40,7 @@
 
     <header>
       <div class="container">
+        <div class="base-header">
         <a href="{% url 'index' %}"><img class="logo" src="{% static 'hourglass_site/images/calc_logo.png' %}" alt="CALC: Contract Awarded Labor Category Tool"></a>
```
toolness: I didn't add a level of indentation here for the sake of making the diff easier to read, but I can add it in if you want.

jseppi: I'm +1 to indentation

jseppi:
👍 nice!

hourglass_site/static_source/style/components/_header.scss
```diff
@@ -43,7 +46,6 @@ header {
   font-size: 2.2em;
   font-weight: 400;
```
hbillings: There's a `$font-weight-normal` var coming that we'll want to use here!

hbillings:
LGTM!

toolness:
Cool, I will add the indentation and merge!

PR #477, opened 2016-07-29
Fix multi-word autocomplete

### LINKED ISSUE #458: Multi-word autocomplete doesn't behave as expected
In the data explorer (i.e. CALC 1.0), with the main search field set to its default setting of **Contains words**, @tram discovered that typing "interpretation" yields the following results:

> interpretation-query

However, attempting to make the query more specific by searching for "interpretation services" yields the following:

> interpretation-services-query

This is unexpected; one would think that the second result set would be a strict subset of the first. Instead, the results in the second don't even seem to have anything to do with the word "interpretation" at all.

The behavior doesn't seem to change at all if we switch the second search to **Exact match**.  Changing it to **Contains phrase** yields a different result set that is somewhat similar to the first result set, but not quite (it's still not a strict subset).

toolness:
This fixes #458.

At first I thought that the problem was in `convert_to_tsquery()`, so I started by writing a doctest for it, but it turned out to work fine.  I left the doctest in, though, since I think it helps improve readability of the code.

### DIFF
--- a/api/tests/test_doctests.py
+++ b/api/tests/test_doctests.py
@@ -1,4 +1,3 @@
-import unittest
 import doctest
 
 from .. import views
--- a/api/tests/test_doctests.py
+++ b/api/tests/test_doctests.py
@@ -1,0 +1,9 @@
+import unittest
+import doctest
+
+from .. import views
+
+
+def load_tests(loader, tests, ignore):
+    tests.addTests(doctest.DocTestSuite(views))
+    return tests
--- a/hourglass_site/static_source/js/common/hourglass.js
+++ b/hourglass_site/static_source/js/common/hourglass.js
@@ -136,6 +136,12 @@
   hourglass.noop = function noop() {
   };
 
+  hourglass.getLastCommaSeparatedTerm = function getLastCST(term) {
+    var pieces = term.split(/\s*,\s*/);
+
+    return pieces[pieces.length-1];
+  };
+
   // export hourglass.extend()
   hourglass.extend = extend;
 
--- a/hourglass_site/static_source/js/data-explorer/index.js
+++ b/hourglass_site/static_source/js/data-explorer/index.js
@@ -128,9 +128,7 @@
         // save inputted search terms for display later
         searchTerms = term;
 
-        // search only last comma separated term
-        var pieces = term.split(/[\s,]+/);
-        term = pieces[pieces.length-1];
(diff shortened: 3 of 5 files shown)


hourglass_site/static_source/js/common/hourglass.js
```diff
@@ -136,6 +136,12 @@
   hourglass.noop = function noop() {
   };
 
+  hourglass.getLastCommaSeparatedTerm = function getLastCST(term) {
```
toolness: Adding the function to `hourglass` was the only way I could think of to write unit tests for it... Also called the function `getLastCST` internally to obey 80 character line limit blah blah blah. 😞

api/tests/test_doctests.py
```diff
@@ -0,0 +1,8 @@
+import doctest
+
+from .. import views
```
jseppi: Unrelated mostly, but are these relative imports common in pythonland? I think I've seen places where relative imports could have been used but instead package ones are used (like this would be `from api import views`).

hourglass_site/static_source/js/data-explorer/index.js
```diff
@@ -128,9 +128,7 @@
         // save inputted search terms for display later
         searchTerms = term;
 
-        // search only last comma separated term
-        var pieces = term.split(/[\s,]+/);
-        term = pieces[pieces.length-1];
+        terms = hourglass.getLastCommaSeparatedTerm(term);
```
jseppi: Did you mean `term = ...` here instead of `terms = ...`?

toolness: OH SNAP good catch!!!!!! will fix (also i suppose this exposes the fact that I didn't even *manually* test the new code I wrote... you totally outed me!)

jseppi:
👍 Nice, just manually tested it. Works much better!

api/tests/test_doctests.py
toolness: Oh, I tend to like relatives just because I want to be able to rename a python package (e.g. `api` in this case) without having to find/replace a bunch of stuff within the package itself... DRY and all that.

I was going to say that I think that the reason it's not super common in pythonland *may* be because it wasn't added to the language until relatively recently, but I think it was actually added a decade ago or something at this point, so I dunno.

toolness:
Woot thanks!!

PR #479, opened 2016-07-29
Also check price_including_iff to determine when s70 price lists end.

toolness:
As discussed in https://github.com/18F/calc/pull/473#issuecomment-236175343, it'd be nice to check `price_including_iff` (another required field) to determine whether we've reached the end of the price list.

### DIFF
--- a/data_capture/schedules/s70.py
+++ b/data_capture/schedules/s70.py
@@ -56,10 +56,11 @@ def glean_labor_categories_from_file(f, sheet_name='(3)Labor Categories'):
         cval = functools.partial(safe_cell_str_value, sheet, rownum)
 
         sin = cval(0)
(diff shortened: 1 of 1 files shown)


jseppi:
Looks good. Might be nice to extract the if-statement condition into a separate function that can be tested on its own, but that could be done as part of Tech Debt Cleanup-a-Thon.

toolness:
Yeah, I also wish the XLS import (via `xlrd`) wasn't so slow--every test that uses it seems to add an extra 0.5 seconds of test time, which adds up fast.  I also wish we could mutate the `xlrd` sheets so that we could just generate different test sheets on-the-fly and hand them over. But like you suggested, just factoring out the if-statement condition into a separate function is probably the quickest way to test it.

PR #482, opened 2016-07-29
Add support for django debug toolbar

toolness:
This gives us the fancy [Django Debug Toolbar](https://django-debug-toolbar.readthedocs.io/en/stable/index.html) when `DEBUG` is `True`.

> screen shot 2016-07-29 at 2 12 17 pm

The one major item of note here is that we *always* show the toolbar when `DEBUG` is `True`, rather than relying on the toolbar's default configuration of consulting `INTERNAL_IPS`, which makes using the toolbar under Docker and e.g. on a tablet over one's local wifi particularly cumbersome.

If this seems too insecure even for debug mode, I can modify the behavior so that it only displays the toolbar if it's running in Docker *or* the client is listed in `INTERNAL_IPS`. This would still make it hard to use the debug toolbar via other virtualization solutions, though, and it'd also make it harder to use it from other devices on one's own wifi network.

### DIFF
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -54,6 +54,7 @@ INSTALLED_APPS = (
     'django.contrib.messages',
     'whitenoise.runserver_nostatic',
     'django.contrib.staticfiles',
+    'debug_toolbar',
 
     'hourglass_site',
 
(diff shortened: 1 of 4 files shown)


jseppi:
Aw nice!

I'm not very aware of what the additioanl security concerns are, but when we're in Debug mode I'm pretty sure we have all kinds of security issues already.

toolness:
Yeah, that was my impression too!

Does this look OK to merge then?

hourglass/settings.py
```diff
@@ -69,6 +70,7 @@
 )
 
 MIDDLEWARE_CLASSES = (
+    'hourglass.middleware.DebugOnlyDebugToolbarMiddleware',
```
jseppi: not enough "Debug" in this name, consider renaming to `DebugOnlyDebugToolbarDebugMiddleware`

jseppi:
👍

PR #484, opened 2016-07-29
Move qunit-related files into qunit_tests.

### LINKED ISSUE #393: Move QUnit tests into qunit_tests app
I just realized that newcomers will probably look for QUnit tests in the `qunit_tests` folder, but that Django app was added in #287 and is only responsible for running the tests during `manage.py test`/`py.test`.

It would probably be less confusing, and also de-clutter the `hourglass_site` directory, if we moved them into `qunit_tests`.  That said, I'm not sure how all this will mesh with #389, as the new JS pipeline might find it easier for all JS stuff to be in one place.

toolness:
This fixes #393.

I haven't moved the qunit tests in `hourglass_site/static_source/tests`, though, because I'm not sure if we want to upset the structure of `static_source` or not.  Do we want to move its individual folders into the django app folders that most closely relate to the JS in question, or do we want to leave them all where they are?


jseppi:
You can ignore that codeclimate ❌ -- we should remove csslint from our `.codeclimate.yml`

jseppi:
Removed csslint in https://github.com/18F/calc/pull/486

jseppi:
> Do we want to move its individual folders into the django app folders that most closely relate to the JS in question, or do we want to leave them all where they are?

What's more common in Django-land? We certainly could change our gulp config to deal with that if you think it makes more sense. I think there are some scss partials that exist in both hourglass_site and data_capture. Maybe we should evaluate that after @hbillings finishes #452?

toolness:
Good question! I've [asked on `#django` in Slack](https://18f.slack.com/archives/django/p1470059943000031).

toolness:
Merging `develop` so codeclimate stops complaining about CSS.

toolness:
This is going to be superseded by #501 or something similar, so I'm shutting it down!

PR #485, opened 2016-07-29
Show test coverage in Ultratest

### LINKED ISSUE #469: Ultratest - add coverage report
I think it would be nice to output the coverage report table that `py.test --cov` generates as part of ultratest.

jseppi:
Adds test coverage output to Ultratest.

screen shot 2016-07-29 at 4 12 41 pm

screen shot 2016-07-29 at 4 15 38 pm

The change in coverage will only be printed if there was a `coverage.xml` file present in your root CALC directory before running the test.

If in verbose mode (`-v 2`), then a full coverage report will also be printed to the terminal.

Note that the values shown will be slightly different than in our CI, because we don't run the Selenium Tests in Travis.

Closes #469


toolness:
Wow, I haven't seen `elementtree` used for a really long time!  Not that it's bad or anything, just a blast from the past.

This looks great! Only minor change I'd suggest is also adding `coverage.xml` to `.dockerignore`, but it's not a big deal if it gets sent to a remote docker machine, as it's not that big a file and doesn't contain any sensitive data.

Feel free to merge!

jseppi:
Ha, I didn't know which xml parsing lib is common in Python these days and I saw that `ElementTree` is part of the standard lib, so that's why I used it.

toolness:
yup it totally makes sense! if xml were a more core part of what we were doing, i'd have recommended using some kind of third party wrapper, because elementtree doesn't like the way my brain works, but for this i think it was a great choice.

PR #486, opened 2016-07-29
remove csslint from codeclimate

jseppi:
We aren't writing CSS any more, just SCSS, so CodeClimate's csslint just produces noise on vendor files mostly.


toolness:
Cool, sounds good to me!!

PR #487, opened 2016-08-01
Inspired by wds mods

jseppi:
I went ahead and made some minor changes to @hbillings original PR #452.

- Fix typo'd `@import`
- Adjust styleguide css to better align h3 headings with their border-bottom
- Extract tablesorting script from styleguide.html to its own file as part of data-capture js bundle. Also ES6ify it and bring it up to our eslint standard. No logic or other changes made.
- Fix table sort handler attachment.

Fixed now: ~~I think there is something not quite right about the sorting logic, or perhaps the way it is attached to the tables.~~

screen shot 2016-08-01 at 8 49 00 am screen shot 2016-08-01 at 8 49 05 am

~~Notice the incorrect placement of `115.99` in relation to the others. It looks like it is still using a string comparison. I'll try to fix that in another issue/PR.~~

### DIFF
--- a/hourglass_site/static_source/js/data-capture/index.js
+++ b/hourglass_site/static_source/js/data-capture/index.js
@@ -4,5 +4,7 @@
 
 require('babel-polyfill/dist/polyfill.js');
 require('jquery-tablesort');
+
+require('./tablesort');
 require('./upload');
 require('./step_1');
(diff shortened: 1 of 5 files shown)


jseppi:
Made some more changes to fix the sorting function issue.

PR #496, opened 2016-08-01
improve Step 2 valid rows table markup

### LINKED ISSUE #493: Header rows on tables should not have a hover style
Noticed in the sprint demo that the header row on tables was picking up a highlight style. That shouldn't happen.

jseppi:
fixes #493

Now the valid price list rows table in Step 2 looks like

screen shot 2016-08-01 at 11 35 09 am


toolness:
Nice, thanks!

PR #500, opened 2016-08-01
Fixes to PR #452

hbillings:
These are some fixes to the WDS styles.
- remove margin from buttons
- fix references
- change some var names


jseppi:
👍

jseppi:
Merging because I need this fix to finish #501

PR #501, opened 2016-08-01
Create a new "frontend" app containing all our front-end stuff

### LINKED ISSUE #489: Move static_source to top-level folder
For better organization, and ease of development (especially for newcomers), we're going to move all of the static "source" files (ES6 files, SCSS) into a top-level directory (and out of hourglass_site).

jseppi:
To Dos:
- [x] move the folders and files
- [x] document in README

Fixes #489


toolness:
Hmm, so this might sound weird, but I'm now wondering if it actually makes sense to make `static_source` a super-trivial Django app, and just move `qunit_tests/tests.py` into it.  Or would that be weird as hell.

toolness:
Er, so should this PR also supersede #484?

jseppi:
I think it might be weird to move the `qunit_tests` into it since those tests are testing the built files, not the source ES6 files.

Maybe put the `qunit_tests` into the `meta` app?

toolness:
Cool that sounds good!

I guess the only other option would be to call the directory `frontend` instead of `static_source`?  Hrm... I suppose that might also be more obvious to folks who just want to focus on the front-end, but I'm not sure.

toolness:
Uh like...

```
frontend/
  tests/
    test_qunit.py
  static/
    vendor/
      qunit.js
      jquery.js
    built/
  source/
    js/
    sass/
```

Maybe I'm just bikeshedding now, though. 😁

jseppi:
Hrm, yeah, that organization does seem good.

toolness:
Oh ok!

I guess we could potentially even put `selenium_tests` in there too (as e.g. `test_selenium.py`), the only argument against that is that *technically* selenium is really testing the full stack, but WHATEVS.  I'm fine with doing it either way!

jseppi:
Ok, here's a more ambitious attempt at reorganizing our frontend files. I've moved all sass/css, js, and images out of `hourglass_site` and into the new `frontend` app. Tests from the `qunit_tests` and `selenium_tests` apps have also been moved (into `frontend/tests`), and those original apps have been removed. I've also gone ahead and renamed `hourglass_site` to `data_explorer`.

See screenshot below for new structure.

screen shot 2016-08-03 at 4 18 08 pm

@toolness: How does this look to you?

toolness:
Yess this is 💥 💃 👍 !!!!!

PR #513, opened 2016-08-02
Make the footer sticky

### LINKED ISSUE #491: Make footer sticky
Our pages that don't have much in-page content look like this on bigger screens:

screen shot 2016-08-01 at 11 55 36 am

Would be nice to have that footer stick to the bottom; it's not hard to do w/ flexbox, and really old browsers that don't support flexbox will still look the same as they currently do.  I did the same in https://github.com/18F/brand/pull/92 and https://github.com/18F/brand/pull/107.

toolness:
This fixes #491.

What's weird is that I'm not experiencing any of the bugs I've previously experienced on safari and chrome, which forced me to add workarounds regarding the [minimum sizing of flex items not honored](https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored).  I didn't want to add unnecessary workaround code unless it was absolutely needed, so I've left it out for now...


toolness:
Uh, somehow this broke post-merge.  Will fix.

toolness:
Fixed!

PR #503, opened 2016-08-02
[DO NOT MERGE] User testing quick fixes for 2016/08/02

toolness:
These are a bunch of quick fixes for today's user test. We might want to cherry-pick some of them into future PRs, though.

PR #523, opened 2016-08-04
Fix #520 by manually constructing FormData

### LINKED ISSUE #520: Upload widget fails on IE11
Unfortunately, today's user test participant was on IE 11 (11.0.9600.18376 to be precise) and after dropping their file into the widget and clicking "next", nothing happened.  I ended up having to add a `data-force-degradation` to the widget and re-deploy for the interview to proceed. 😭 

When I originally wrote the upload widget, I did test on IE11, but I was less vigilant about testing some of the improvements we've made since then... Need to either fix this ASAP or add the `data-force-degradation` to #503 so it sticks for now.

toolness:
Oy.

### DIFF
--- a/frontend/source/js/data-capture/step_1.js
+++ b/frontend/source/js/data-capture/step_1.js
@@ -49,16 +49,23 @@ function bindForm() {
     } else {
       e.preventDefault();
 
-      const formData = new window.FormData(form);
+      const formData = new window.FormData();
 
-      if (upload.file) {
-        // It's possible that the user may have dragged-and-dropped a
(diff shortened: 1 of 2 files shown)


toolness:
Ok now we are iterating through the form elements and constructing the `FormData` ourselves as per @jseppi's suggestion in https://github.com/18F/calc/issues/520#issuecomment-237644641, so we have IE11 support again!

toolness:
I've merged this into #503 so we can user test it.

jseppi:
👍

PR #525, opened 2016-08-05
Fix #504 by seeking to 0 before reading from file.

### LINKED ISSUE #504: Uploading a CSV w/ s70 selected doesn't fall back to Fake Schedule
Not sure why.

toolness:
It turns out #504 was caused because the schedule 70 reader would read the upload, leaving nothing left to read by subsequent readers.


jseppi:
👍

PR #527, opened 2016-08-05
Add HIDE_DEBUG_UI setting

toolness:
This is a cherry-pick from #503. It adds a `HIDE_DEBUG_UI` setting that hides development affordances on the front-end, which can be useful when demoing or user testing a debug build.

I guess an alternative to this is simply to always demo/user test production builds.  I thought maybe this could be useful if we still wanted the nice error tracebacks, quick-reload, and other affordances of debug mode, without any visual clutter that might confuse onlookers.

This might be a dumb idea in which case we can just close without merging.


jseppi:
👍

PR #528, opened 2016-08-05
Add a test suite for step_1

### LINKED ISSUE #461: Add a test suite for JS used by step 1 of data capture app

toolness:
This will fix #461.

We're adding [sinon](http://sinonjs.org/) to `package.json` for this.

To do:

- [X] Ensure degraded upload widgets pass-through to just submitting the form (no ajax).
- [X] Ensure ajax submission works.
- [X] Ensure ajax `form_html` response works.
- [x] Ensure ajax `redirect_url` response works.
- [x] Ensure other ajax errors work.
- [ ] File a separate issue for figuring out how to `require()` a module from a different bundle for tests. Right now I'm hacking around this by setting `window.testingExports__step_1 = exports` in `step_1.js` and retrieving it in the test suite, but there is probably a better way to do this via browserify. I don't think it should block merging this PR, though.

### DIFF
--- a/frontend/source/js/data-capture/step_1.js
+++ b/frontend/source/js/data-capture/step_1.js
@@ -29,6 +29,10 @@ function bindForm() {
   $upload.uploadify();
 
   const upload = $fileInput.data('upload');
+  const self = { form, upload, $upload, $fileInput, $submit };
+
+  // This is mostly just for test suites to use.
+  $('form').data('step1-form', self);
 
   // Disable the submit button until a file is selected.
   $submit.prop('disabled', true);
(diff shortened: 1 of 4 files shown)


jseppi:
I think this is probably good for now so that we have some tests in place.

It looks like web components might work well for progressive enhancement actually, so I am hoping we can switch to more "robust" JS sometime in the future. More research needed though...

toolness:
Ok, sounds good!

Issue #519, opened 2016-08-04
/api/rates takes a long time to respond


toolness:
So we had a big spike in average response time on Tuesday:

![screen shot 2016-08-04 at 7 46 05 am](https://cloud.githubusercontent.com/assets/124687/17401142/f506a2ba-5a19-11e6-8c6f-ffbe63fe8bbf.png)

I'm not sure why this is, but looking at a single page load on my development machine with region 10 (but not schedule 70) data loaded, the obvious stand-out is the `/api/rates` request made via ajax, which takes 3.35 seconds on my system.

I tried using [`ab`](https://httpd.apache.org/docs/2.4/programs/ab.html) to benchmark and it claims we are responding to 0.41 requests per second and taking 24s per request when 10 concurrent clients are visiting `/api/rates` over 100 separate requests.

<details>
<summary>Full output of `ab -k -c 10 -n 100` on `/api/rates`</summary>
```
$ ab -k -c 10 -n 100 'http://app.calc.docker:8000/api/rates/?format=json&histogram=12&sort=current_price&query_type=match_all&min_experience=0&max_experience=45&contract-year=current&experience_range=0%2C45'
This is ApacheBench, Version 2.3 <$Revision: 1706008 $>
Copyright 1996 Adam Twiss, Zeus Technology Ltd, http://www.zeustech.net/
Licensed to The Apache Software Foundation, http://www.apache.org/

Benchmarking app.calc.docker (be patient).....done


Server Software:        WSGIServer/0.2
Server Hostname:        app.calc.docker
Server Port:            8000

Document Path:          /api/rates/?format=json&histogram=12&sort=current_price&query_type=match_all&min_experience=0&max_experience=45&contract-year=current&experience_range=0%2C45
Document Length:        70734 bytes

Concurrency Level:      10
Time taken for tests:   246.167 seconds
Complete requests:      100
Failed requests:        0
Keep-Alive requests:    0
Total transferred:      7100400 bytes
HTML transferred:       7073400 bytes
Requests per second:    0.41 [#/sec] (mean)
Time per request:       24616.731 [ms] (mean)
Time per request:       2461.673 [ms] (mean, across all concurrent requests)
Transfer rate:          28.17 [Kbytes/sec] received

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    1   0.3      1       3
Processing: 18366 24387 2282.6  24287   29333
Waiting:    18354 24139 2248.0  24136   29160
[...]

PR #529, opened 2016-08-05
Use StdDev from db instead of statistics.py

jseppi:
Starting in hopes of addressing #519 

To Dos:
- [x] cleanup/remove unused old stddev code
- [x] validate that it is faster that statistics.py
- ~~figure out how to validate StdDev aggregation~~


api/views.py
```diff
@@ -177,26 +177,31 @@ def get(self, request):
...
-                current_rates.append(rate[wage_field])
-
-        if current_rates:
-            std_dev = stdev(current_rates)
-        else:
-            std_dev = None
-
-        page_stats['first_standard_deviation'] = std_dev
+        page_stats['first_standard_deviation'] = \
+            quantize(contracts_all.aggregate(
```
jseppi: Actually we could batch all these aggregations into a single call, like 

```py
models.aggregate(Avg('price'), Max('price'), Min('price'))
```

jseppi:
@toolness Does this alleviate the performance problems you saw due to `statistics.py`?

jseppi:
For reference, this is the commit where the numpy-based standard deviation was first introduced: https://github.com/18F/calc/pull/178/commits/cb6ec4f4699519a928a95b881730398f27b50022 
Perhaps the original developer just didn't know there was an available StdDev aggregation? I couldn't find any other discussion for why numpy was used.

toolness:
NOICE

> @toolness Does this alleviate the performance problems you saw due to `statistics.py`?

Good question!  Maybe I should add some kind of `manage.py performance_test` command to make it easy to figure out?  Alternatively I could just do it manually.

jseppi:
I tried for a couple minutes to get the DjDT perf panel to work but didn't figure out within a few minutes, so I punted on it.

jseppi:
In manual testing I noticed that the standard deviations calculated were very slightly different, but only after like 8 decimal places, which seems insignificant to me.

toolness:
Nice!

To get the DjDT perf panel to work, you have to [define it in `settings.py`](https://django-debug-toolbar.readthedocs.io/en/stable/panels.html#profiling), which is a bit annoying.  I guess we could add it so it always shows up.

Anyways, then you have to go to a page that uses it directly, rather than one that uses it via an ajax request, which probably means that you need to visit `/api/rates/` in your browser (which causes django rest framework to run the query as part of the page request and render it as html, along with the debug toolbar).

jseppi:
Actually, we have tests already in place for testing the standard deviation values (in `test_api.py`), and now the values are calculated by Django/Postgres, I feel less need to validate results against the previous custom code.

jseppi:
Here are the DjDT Profiling results on `/api/rates` now:

screen shot 2016-08-12 at 4 55 36 pm

Must faster than what is shown in #519, assuming I ran the profiling correctly.

api/utils.py
```diff
@@ -1,24 +1,4 @@
 
-import math
-import statistics
```
toolness: Heh, took me a little while to figure out that [`statistics`](https://docs.python.org/3/library/statistics.html) is a standard Python module introduced in 3.4. First I thought it was a module we wrote, then I thought it was a third-party module. 😁

toolness:
Wow, this is *much* faster! The qunit test suite takes a really long time on my local machine with all the data loaded, because the `hourglass.API: #get()` test takes forever, but this PR makes it almost instantaneous. Great work!

api/utils.py
jseppi: Yeah, it looks like a nice package, but is apparently slow (at least slower than numpy and native db ops, at least).

PR #530, opened 2016-08-05
Make drag-and-drop widget's draggable state more noticeable

hbillings:
Addresses #397. Both the text and color will change, though TBH I'm not sure I love the color change. Feedback very welcome! 
screen shot 2016-08-05 at 5 45 39 pm

### DIFF
--- a/frontend/source/sass/components/_upload.scss
+++ b/frontend/source/sass/components/_upload.scss
@@ -68,18 +68,24 @@
   display: none;
 }
 
-.no-js .upload, .js .upload.degraded {
-  font-size: 1.5rem;
-  text-align: center;
-
-  .upload-chooser {
-    display: inline;
-  }
+.no-js .upload,
+.js .upload.degraded {
+  font-size: 1.3rem;
+  display: block;
+  border: none;
 
   input {
+    border: none;
+    font-size: 1.3rem;
+    color: $color-primary;
+    font-weight: $font-weight-bold;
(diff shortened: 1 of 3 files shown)


frontend/source/sass/components/_upload.scss
```diff
@@ -36,7 +40,16 @@
...
   &.dragged-over {
-    background-color: $color-white;
+    background-color: $color-green-lightest;
+    border: 2px dashed $color-green;
+    .upload-chooser {
+      * {
+        visibility: hidden;
+      }
+      &:before {
+        content: "Drop files here!"
```
toolness: Should this be "Drop file here", since they can only upload one file at a time?

toolness:
This looks great!  What do you think of changing the "Drop files here" text to "Drop file here" or "Drop your file here", since they can only upload one at a time?

Also, can you update the styleguide example to put a `<span>` around the "XLS, XLSX, or CSV format, please" like you did in the step 1 form?  Otherwise the styleguide example looks a bit broken when hovering over...

hbillings:
Good catch on the language! Will update that and the styleguide.

hbillings:
Okay, this should be ready. Plus, I added some more styling for the gracefully degraded version. 
screen shot 2016-08-08 at 10 54 37 am

toolness:
Oooo nice, thanks for the degraded styling!  Looks much better now.

Issue #291, opened 2016-06-17
djorm_pgfulltext logs RemovedInDjango19Warning


toolness:
As mentioned in #279 and #288:

> During `manage.py load_data`, our version of `djorm_pgfulltext` is using `django.db.models.get_app` and  `django.db.models.get_model`, which are removed in 1.9. Fixing this will require an upstream change to that package (or alternatively, using a different package, or waiting until Django 1.10 is out and switching to its built-in full-text indexing for postgres).

I've made some additional notes on the 1.9 incompatibilities at https://github.com/linuxlewis/djorm-ext-pgfulltext/issues/65#issuecomment-226490485.

toolness:
Just wanted to mention that I [asked on Slack about using the Django 1.10 beta in production](https://18f.slack.com/archives/infrastructure/p1467208029000536) and got the 👍 . So assuming that the full-text search features in 1.10 support our needs, we've got the go-ahead to upgrade!

toolness:
I saw a video on [Django and Postgres](https://www.youtube.com/watch?v=oeLACa7cj1Q) and it included content on the new full-text search functionality in Django 1.10 beta, which seems pretty awesome. I haven't yet investigated the current implementation of full-text search in CALC but I am bullish that we will be able to migrate to the new native functionality.

toolness:
I played around with [django full text search](https://docs.djangoproject.com/en/dev/ref/contrib/postgres/search/) in 1.10rc1 today and I think it is pretty slick.  I also looked into the full-text search in CALC a bit, and I think we can migrate over to Django's built-in text search, but it'll take a bit of work, since we do some fairly custom stuff with `djorm_pgfulltext`.

toolness:
I'm thinking that as a general migration strategy, it might actually be easiest to make a minor fork of djorm_pgfulltext that fixes those Django 1.9 incompatibilities, switch to that, and then upgrade to Django 1.9, and 1.10 after that.  Once we've migrated to 1.10, we can then migrate over to Django's native full-text search and drop our fork of djorm_pgfulltext.

toolness:
This is basically resolved by our fork, so I'm closing it.

PR #531, opened 2016-08-06
Add full text search tests, a few refactorings

toolness:
This adds a number of full text search tests and refactors things a bit by moving full text-related functionality from `api.views` into `contracts.models`.

Aside from increasing the robustness of our test suite, this is intended to collect all full-text search functionality in one place so that clients don't have to be concerned the concept of a `tsquery`. This will allow us to more easily migrate to [Django 1.10's full text search](https://docs.djangoproject.com/en/1.10/ref/contrib/postgres/search/) in the future (see #291 for more details).


jseppi:
Niiiice!

👍

toolness:
Woot thanks!

Issue #509, opened 2016-08-02
Errors look like form inputs


hbillings:
They need _some_ sort of background color, but the WDS is too heavy.

Issue #494, opened 2016-08-01
Restyle required field error markers on enter contract data page


hbillings:
When data isn't entered in a required field, the error text style should match the standards

PR #532, opened 2016-08-08
Update error styling

hbillings:
To address #494 & #509. I replaced the button with a link for now, but we will probably want a button again as per our convo about breaking this out into its own step.
screen shot 2016-08-08 at 3 08 28 pm


jseppi:
👍

Issue #533, opened 2016-08-08
Buttons are funky


hbillings:
<img width="861" alt="screen shot 2016-08-08 at 1 29 15 pm" src="https://cloud.githubusercontent.com/assets/509309/17491385/05eabd0e-5d6d-11e6-996a-aa0a1d2e3501.png">

Issue #490, opened 2016-08-01
Address typography


hbillings:
Typography needs to be declared more specifically -- right now it's cobbled together from what was there before and what the standards declare for a completely different font.

Examples:
- Raleway is getting picked up in weird places

- Line height's not right on some dropdowns:
![screen shot 2016-08-02 at 10 19 00 am](https://cloud.githubusercontent.com/assets/509309/17334116/a061cfb2-589a-11e6-8392-8c30f6a1cc2d.png)

- How many heading sizes do we need?? (actually this is an h5, a legend label, and a fieldset legend, respectively)
![screen shot 2016-08-01 at 5 32 09 pm](https://cloud.githubusercontent.com/assets/509309/17311043/ed37a574-580d-11e6-94af-5edff919738a.png)

- this leading is ugh:
![screen shot 2016-08-01 at 5 32 37 pm](https://cloud.githubusercontent.com/assets/509309/17311061/0d2b4d5e-580e-11e6-9b7d-abcda52bab83.png)

jseppi:
I also noticed yesterday that the "Typography" section of the styleguide is empty currently, so it might be good to fill that out once things have been addressed.

Ref: https://github.com/18F/calc/blob/develop/styleguide/templates/styleguide.html#L25

PR #536, opened 2016-08-08
Typography tweaks

hbillings:
Addresses #490. Should also address #533.

UPDATE: This can be merged if desired for 8/9's user interviews. The WIP part is updating the typography section of the styleguide, which @jseppi pointed out was blank (oops!). 
screen shot 2016-08-11 at 12 50 57 pm


frontend/source/sass/base/_typography.scss
```diff
@@ -1,11 +1,8 @@
 @import 'variables';
 
 h1 {
-  letter-spacing: 0;
```
toolness: 😲

toolness:
Thanks, this looks splendid!

PR #537, opened 2016-08-09
Fix sticky footer on ie11

### LINKED ISSUE #535: Sticky footer messes up page on IE11
Blugh, apparently [IE11 has tons of flexbox bugs](http://caniuse.com/#search=flexbox) and makes the sticky footer added in #513 look like this:

screen shot 2016-08-08 at 4 27 31 pm

Will need to either revert it or add all the weird workarounds that make it not look horrible.

toolness:
This fixes #535. It's basically just a port of https://github.com/18F/brand/pull/107 and it seems to get the job done.

### DIFF
--- a/frontend/source/sass/components/_page.scss
+++ b/frontend/source/sass/components/_page.scss
@@ -5,5 +5,6 @@ body {
 }
 
 main {
-  flex: 1;
+  // https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored
+  flex: 1 0 auto;
 }


toolness:
Oops, I should note that this doesn't actually make the footer *sticky* on IE11--it just guarantees that the footer appears below the main content, i.e. it appears as a non-sticky footer.

I think this is ok, though, since it means that the footer looks just like it always has on IE11, which is a massive improvement over #535, where the footer appearing behind the main content destroyed its readability.

jseppi:
👍

PR #548, opened 2016-08-12
Encapsulate upload widget in a Django Widget, rename step-1-form to ajaxform

toolness:
This adds a custom `UploadWidget` [Django widget](https://docs.djangoproject.com/en/1.10/ref/forms/widgets/) to encapsulate the upload widget added in #387.

It also renames `step1-form` to `ajaxform`, since that's what it basically is, and attempts to lay a foundation for making it reusable (as this will be required really soon).

It also adds documentation and adds/updates style guide information for both. In particular, the `ajaxform` example includes a back-end component that allows the user to easily manually test what will happen on edge cases, such as server errors.

Advantages:

* Keeps all the widget HTML code in one place, keeping things DRY and less error-prone. In fact, our QUnit tests now include server-rendered `UploadWidget` HTML, ensuring that we're testing what comes out of the server rather than what's hard-coded into our test suite.
* Makes it easier to add an upload widget to any form.
* Makes it easier to add more ajax-enabled forms that gracefully degrade.

Disadvantages:

* We're still not encapsulating *everything* in the `upload.js` module. However, this seems inevitable if we want to keep supporting browsers without JS and can only run Python on the server-side.

Other notes:

* `UploadWidget` is now moved to `frontend/upload.py`; it's so closely coupled to the front-end, and not really specific to `data_capture`, that I thought it belonged there.  Plus, this way one can just search for file names matching the string `"upload"` and find all related files. Feel free to disagree though, we can always move it back into the `data_capture` package!
* Similarly, `ajaxform`-related utilities have been moved to `frontend/ajaxform.py`.
* Some CSS for `ajaxform` has been moved into classes defined in `_ajaxform.css`, rather than being defined inline.

To do:

- [x] Update the style guide to not display the upload widget HTML source code.  Instead, it should tell the reader to use the `UploadWidget` widget and/or `ajaxform`.


frontend/templates/tests.html
```diff
@@ -18,6 +18,10 @@
   <body>
     <div id="qunit"></div>
     <div id="qunit-fixture"></div>
+    <script>
+      {% load qunit_fixture_data %}
+      var QUNIT_FIXTURE_DATA = {% qunit_fixture_data_json %};
```
toolness: It feels a bit weird to define this as a JS variable rather than raw HTML that we lookup in tests--it definitely makes things a bit less readable and harder to debug--*but* I wanted to make sure that these fixtures weren't messed with by the scripts that auto-load in the page. I guess that is a potential disadvantage to using markup APIs.

styleguide/templates/styleguide.html
```diff
@@ -5,8 +5,8 @@
 {% block title %}Style Guide{% endblock %}
 
 {% block head %}
-<link rel="stylesheet" href="{% static 'frontend/built/styles/data_explorer.min.css' %}">
-<link rel="stylesheet" href="{% static 'frontend/built/styles/data_capture.min.css' %}">
+<link rel="stylesheet" href="{% static 'frontend/built/style/data_explorer.min.css' %}">
+<link rel="stylesheet" href="{% static 'frontend/built/style/data_capture.min.css' %}">
```
toolness: Um, this was a typo introduced some time ago, not sure when...

styleguide/templates/styleguide.html
```diff
@@ -48,44 +48,26 @@
...
-  {% endexample %}
+  </div>
 
   <p>
-    When the page is first loaded, all <code>.upload</code> elements are
-    automatically activated. However, if you dynamically inject content
-    into the page via Ajax, you'll want to manually activate them via
-    code like this:
+    For a simple example of the Ajax Form component, see
+    {% pathname 'styleguide/ajaxform_example.py' %}.
```
toolness: Rather than attempting to describe exactly how to use these two components, which involves cooperation between front-end and back-end, I thought it might be better to just link to an example, to avoid documentation rot.

toolness:
Ug, this ended up being kind of big. 😞   Let me know if you want to go over this in a screen-sharing session, @jseppi, or if you want me to split it up into smaller PRs!

jseppi:
reviewing now...

styleguide/templates/styleguide.html
jseppi: Looks like that was me in https://github.com/18F/calc/commit/018078c5898687e7f7fadbc6f8f06546abac66da 😬

frontend/source/js/data-capture/ajaxform.js
```diff
@@ -41,8 +40,16 @@ function bindForm() {
   const upload = $fileInput.data('upload');
   const self = { form, upload, $upload, $fileInput, $submit };
 
+  if (!upload) {
+    throw new Error('ajaxform must contain an upload widget');
```
jseppi: I suppose that `ajaxform.js` could be written in such a way that it doesn't _need_ an upload widget, but that still deals with it (or multiples of it) if present. We don't have a usecase for that kind of behavior currently though, so this is probably ok.

toolness: Oh yeah, sorry I didn't comment on that part--I definitely agree.  I just wanted to encode all our assumptions about the way it's currently used into the code, so that e.g. folks don't try adding an ajaxform without an uploadwidget and then get confused when weird unhelpful exceptions are thrown.

Definitely agree that we should eventually support this use case though--maybe I should change the exception text or add a comment above the `throw` indicating that we want to support zero or more than one upload widgets eventually, but don't currently?

jseppi: 👍 to comment

toolness:
@jseppi just merged in your #556 so you should be able to run these tests now!

jseppi:
Hrm, I am having CSRF problems submitting the example form:

`WARNING [django.request:98] Forbidden (CSRF token missing or incorrect.): /styleguide/ajaxform`

toolness:
Uh, that's weird... there is a `<input type='hidden' name='csrfmiddlewaretoken'>` in the form, which should be copied into the `FormData` that's being submitted... and it seems to be working on my end... Hmmmmm.  Maybe log out and back in again?

toolness:
Hmm, so I have been able to reproduce the error, but by doing something very weird: if I have two CALC tabs open to `/styleguide/ajaxform` and log in on one of them, then switch to the other one (where I'm logged out), submitting the form 403's. However, that's going to happen on *any* form, not just the ajaxform / uploadwidget scenario, since there's not really anything we can do about stale CSRF tokens on old tabs...

styleguide/templates/styleguide.html
jseppi: Minor nit, but it might be nice to move this line up above the rendered example, and maybe change the wording to `For a simple Ajax Form code example, see {% pathname 'styleguide/ajaxform_example.py' %}.`

jseppi:
Ok, this seems pretty awesome to me! 👍 

I had a couple very minor comments that you can address or not. Merge when/if you address them.

I don't feel very qualified to review the Python aspects very much but they at least seem understandable to me. I think the real test will be using it for #555, but I'm not foreseeing any issues right now.

toolness:
Woot! Fixed the nits and merging now!

PR #552, opened 2016-08-12
explicitly set djdt panels and add profiling panel

jseppi:
Adds the Profiling panel to the DjDT.

We could also do something like:

```python
from debug_toolbar import settings as djdt_settings
djdt_settings.PANELS_DEFAULTS.append(
    'debug_toolbar.panels.profiling.ProfilingPanel')
```

I'm not sure if that is considered a better practice, though


toolness:
Hmm I'm not a fan of mutating `djdt_settings.PANELS_DEFAULTS`... but another option might be to add something like this to our `settings.py`:

```py
from debug_toolbar import settings as djdt_settings

DEBUG_TOOLBAR_PANELS = djdt_settings.PANELS_DEFAULTS + [
    'debug_toolbar.panels.profiling.ProfilingPanel'
]
```

Maybe?

jseppi:
Good point. Though doing it either that way or the way I proposed also only
permits adding to the front or back of the list of panels, which isn't
ideal either.


On Aug 12, 2016, at 5:24 PM, Atul Varma <notifications@github.com> wrote:

Hmm I'm not a fan of mutating djdt_settings.PANELS_DEFAULTS... but another
option might be to add something like this to our settings.py:

from debug_toolbar import settings as djdt_settings
DEBUG_TOOLBAR_PANELS = djdt_settings.PANELS_DEFAULTS + [
    'debug_toolbar.panels.profiling.ProfilingPanel'
]

Maybe?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/18F/calc/pull/552#issuecomment-239563415>, or mute the
thread

toolness:
Oh true dat, if we want to add it somewhere else then your solution is rad!  Feel free to merge.

Issue #478, opened 2016-07-29
Cancel button in Data Capture Upload flow should clear any submitted data


jseppi:
(Implement after https://github.com/18F/calc/pull/472 is merged)

Also this needs to happen in the Region 10 Upload flow after #555 is finished.

jseppi:
Perhaps Step 1 of each data capture flow also clear out the session values related to that flow?

jseppi:
Or maybe Step 1 should clear out all session keys that start with `data_capture:`?

Issue #476, opened 2016-07-29
Create base template for Data Capture pages


jseppi:
From @toolness in a [line comment](https://github.com/18F/calc/pull/465/files/7f9d2d4574e6f5eafc00bba40c2d35bf35f418a4#r72775994):
> Makes me wonder if data capture should just have its own data_capture/base.html template that trivially extends base.html and adds the messages (and the data_capture CSS and JS, I guess). That way future functionality we add won't need to duplicate all this stuff. But I guess perhaps that's also a task for later on when we actually need it, as for the time being it's sort of premature refactoring.

jseppi:
We do have `data_capture/templates/data_capture/base.html` so I think this one is done!

PR #555, opened 2016-08-15
Region 10 Bulk Upload

jseppi:
Main functionality is now complete!

Instead of taking a CSV, this flow can directly take the XLSX spreadsheet that Kelly generates from an export of his Region 10 database. Upon submission of step 2 in this flow, all the existing Region 10 data is purged from the database and the newly uploaded data is saved. To accomplish purging the old data, I added to add a new model `BulkUploadContractSource` for tracking Region 10 data provenance, and associate it with the `Contract` model.

**To Dos**
- [x] import from R10 Excel dump file
- [x] commit migrations once models are finished
- [x] clean up names (`bulk_upload`, `price_list_upload`, `price_list`, etc.)
- [x] tests
- [x] use base template, ref #476 

Other stuff to be addressed later:
- Proper cancel behavior that clears session storage (#478)
- Improve UI. I just made it up as I went along but it would be good to have other eyes on it to clean up language/styling/etc.
- In step 1, add a screenshot or something to let the user know which file we're looking for.
- On final step, add ability to download a spreadsheet or CSV of the `bad_rows` so that the admin can see which things _did not_ get added.
- Add JS (and maybe server-side code?) to disable/hide the submit button in step 2 after it is clicked. Since this is such a long process, it would be good to prevent accidental double-submissions.
- Add a spinner in step 2 so the user knows something is happening even though it is taking a while.
- Add a link somewhere (in Admin UI or in CALC header) to the Region 10 bulk upload flow (for now one has to manually visit it at `/data-capture/bulk/region-10/step/1`).
- Examine price list upload and region 10 bulk upload URL schemes.
- Set `BulkUploadContractSource` association in `load_s70` command.
- Return 403 page when non-`is_staff` user attempts to visit bulk upload views. Right now it keeps redirecting to the login flow when a non-staff user logs in with the redirect_url set to a bulk upload step.
- Add file type to `BulkUploadContractSource` so we know if it was a CSV or XLS(X)


data_capture/templates/data_capture/bulk/region_10.html
```diff
@@ -0,0 +1,23 @@
+{% extends 'base.html' %}
+{% load static %}
+
+{% block title %}Upload Region 10 Data{% endblock title %}
+
+{% block head %}
+<link rel="stylesheet" href="{% static 'frontend/built/style/data_capture.min.css' %}">
+{% endblock head %}
+
+{# TODO: We'll use the slim header more often so it should be the default #}
```
toolness: Makes sense to me--should we do that in this PR or file a separate one for it?

data_capture/templates/data_capture/bulk/region_10_form.html
```diff
@@ -0,0 +1,30 @@
...
+
+  <div class="upload">
+    <input type="file" name="{{ form.file.html_name }}"
+           id="{{ form.file.id_for_label }}" accept=".xlsx,.xls,.csv">
+    <div class="upload-chooser">
+      <label for="{{ form.file.id_for_label }}">Choose file</label>
+      <span class="js-only" aria-hidden="true">or drag and drop here.</span>
+      <span>XLS, XLSX, or CSV format, please.</span>
+    </div>
+  </div>
```
toolness: Once #548 is merged, this should hopefully just turn into `{{ form.file }}`, or something like that. Hooray for DRY!

data_capture/templates/data_capture/bulk/region_10_form.html
```diff
@@ -0,0 +1,30 @@
+<form enctype="multipart/form-data" method="post"
+  action="{% url 'data_capture:bulk_region_10' %}">
```
toolness: Once #548 is merged and the form uses a fancy `UploadWidget`, this will need a `data-ajaxform` on it.

data_capture/templates/data_capture/bulk/region_10.html
jseppi: separate one would be good, then I can merge it into here once it gets into `develop`

data_capture/forms.py
```diff
@@ -42,3 +42,18 @@ class Meta:
             'contract_start',
             'contract_end',
         ]
+
+
+class BulkRegion10Form(forms.Form):
+    '''
+    Form for bulk upload of Region 10 data export
+    '''
+    file = forms.FileField()
```
toolness: Once #548 is merged, this should probably change to something like:

```py
    file = forms.FileField(widget=UploadWidget(
        accept=['.csv'],
        extra_instructions="CSV files only, please."
    ))
```

data_capture/templates/data_capture/step.html
```diff
@@ -12,32 +12,28 @@
...
+    <li {% if step_number == 1 %}class="current"{% endif %}>
+      <label>Upload price list</label>
+    </li>
+    <li {% if step_number == 2 %}class="current"{% endif %}>
+      <label>Validate data</label>
+    </li>
+    <li {% if step_number == 3 %}class="current"{% endif %}>
+      <label>Contract details</label>
+    </li>
+    <li {% if step_number == 4 %}class="current"{% endif %}>
```
toolness: Repetition of view logic here makes me wonder if it should be factored out into a `templatetag` or something.  Not a big deal though since we've only got two of them right now, and regardless, it shouldn't be handled by this PR.

data_capture/views.py
```diff
@@ -124,3 +124,35 @@ def step_4(request):
     return render(request, 'data_capture/step_4.html', {
         'step_number': 4
     })
+
+
+# TODO: restrict to administrator role
```
toolness: Cool, you might want to use Django's [`user_passes_test`](https://docs.djangoproject.com/en/1.9/topics/auth/default/#django.contrib.auth.decorators.user_passes_test) decorator for that.

jseppi: Nice, I was going to ask you what the proper decorator would be -- thanks!

data_capture/views.py
```diff
@@ -124,3 +124,35 @@ def step_4(request):
...
+
+    if request.is_ajax():
+        return JsonResponse({
+            'form_html': render_to_string(
+                'data_capture/bulk/region_10_form.html',
+                RequestContext(request, ctx)
+            )
+        })
+
+    return render(request, 'data_capture/bulk/region_10.html')
```
toolness: Once #548 is merged, you can use the functions in the `ajaxform` module to do the redirect and template rendering, which will relieve you from having to do the `request.is_ajax()` checking.

data_capture/templates/data_capture/bulk/region_10.html
toolness: Ok, filed as #557!

toolness: Oh, um, I just realized that another thing we could do is refactor our existing data capture pages to use a base template (#476) which would also obviate this issue, at least for our data capture pages. Hrm.

jseppi: Yeah, that's true, and would be a good idea to possibly leave room to style those pages slightly differently. For example, a subheading under CALC that says "Data Capture" or something like that.

toolness: Noice! Do you want to handle that in this PR or should I file a separate one?

jseppi: I'll go ahead and do that in this PR since I'm already mucking with file and folder structure.

jseppi:
0ac27073768e7dd437fae4495528e2860fe9aa13 implements default slim header behavior, per #557. Now the styleguide will also have the slim header, which I think is fine, but thought I should note it.

jseppi:
This is getting kinda big...

jseppi:
It turns out all the memory issues and slowness of the bulk upload stuff was due to the Django Debug Toolbar. Turning it off allows even the very large real export to run through the import process in this PR. However, for some reason it still looks like the new contracts are not in the database, but I'm not sure why that is.

jseppi:
The new contracts were not showing up in the results table because of a weird bug in `Region10Loader` where it was constructing the `hourly_rate_year#` field of the `current_price` by grabbing the `contract_year` value from `line[14]` in the "csv row". However, xlrd was not returning an int, but a float in that cell, so it would look like `2.0`, for example. `hourly_rate_year2.0` is not a valid field, so `current_price` was getting set to `0` for all new contracts.

jseppi:
@toolness: any input on subfolder/sub-module names?
Here's what it looks like currently:

screen shot 2016-08-22 at 9 03 00 am

data_capture/spreadsheet_converter.py
```diff
@@ -0,0 +1,132 @@
...
+        headings = sheet.row(0)
+
+        idx_map = {}
+        for i, cell in enumerate(headings):
+            # find the val in the xl_heading_to_csv_idx_map
+            if cell.value in self.xl_heading_to_csv_idx_map:
+                idx_map[cell.value] = i
+
+        if raises and len(idx_map) != len(self.xl_heading_to_csv_idx_map):
+            raise ValueError('Missing expected column(s) in Excel sheet')
```
jseppi: @toolness: Is this the error class you'd expect for this kind of error?

data_capture/tests/test_spreadsheet_converter.py
```diff
@@ -0,0 +1,92 @@
...
+        }
+        self.assertEqual(expected, indices_map)
+
+    def test_get_heading_indices_map_raises(self):
+        converter = Region10SpreadsheetConverter(xls_file=r10_file())
+        converter.xl_heading_to_csv_idx_map = {'blah': 0}
+        with self.assertRaises(ValueError):
+            converter.get_heading_indices_map(raises=True)
+
+    def test_convert_next(self):
```
jseppi: Not sure exactly how to best do this test or `test_convert_file` -- should I hardcode the full expected results (basically 3 csv rows) and compare each to the returned values?

data_capture/tests/test_bulk_upload_views.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_staff_login_is_required(self):
+        self.login(is_staff=True)
+        res = self.client.get(self.url)
+        self.assertEqual(res.status_code, 200)
+
+    def test_non_staff_login_errors(self):
+        self.login()
+        res = self.client.get(self.url)
+        # TODO: currently this redirects to
```
jseppi: @toolness: any ideas on this response when a non-staff user attempts to access these views?

jseppi:
ready for review, @toolness

data_capture/tests/test_spreadsheet_converter.py
jseppi: I did end up hardcoding the expected results.

toolness:
Cool, screenhero'ing would be good, maybe sometime tomorrow morning?

> @toolness: any input on subfolder/sub-module names?

It looks good!  My only nit is that maybe the `templates/data_capture/bulk` folder should be renamed `bulk_upload`, to be consistent w/ the naming of the `bulk_upload.py` files?

data_capture/spreadsheet_converter.py
toolness: From a purely technical standpoint, I think that `ValueError` is fine, but I'm wondering if this is an error that we'd like to propagate back to the end-user if the file they've uploaded ends up throwing this error?  In that case we might want to subclass `ValueError`, catch it when calling this method from the view/form logic, and convert it into a Django `ValidationError` so that it makes its way to the end-user... On the other hand, if it's enough to just catch all exceptions and turn them into generic "Sorry buddy, your XLS file is messed up" errors then this is fine!

jseppi:
Sure, I should be online relatively early.

data_capture/spreadsheet_converter.py
jseppi: I was going to go with the simple option for now.

contracts/models.py
```diff
@@ -115,6 +115,25 @@ def order_by(self, *args, **kwargs):
...
+
+    PROCUREMENT_CENTER_CHOICES = (
+        (REGION_10, 'Region 10'),
+    )
+
+    created_at = models.DateTimeField(auto_now_add=True)
+    updated_at = models.DateTimeField(auto_now=True)
+    submitter = models.ForeignKey(User, null=True, blank=True)
+    has_been_loaded = models.BooleanField(default=False)
+    original_file = models.BinaryField()
```
toolness: As discussed on our screenhero session, consider adding a mime type field here eventually (could be in a later PR).

data_capture/spreadsheet_converter.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def get_metadata(self):
+        '''
+        Returns a dict containing metadata about the related xls_file
+        '''
+        book = xlrd.open_workbook(file_contents=self.xls_file.read())
+        sheet = book.sheet_by_index(self.sheet_index)
+        self.xls_file.seek(0)
+        return {
+            'num_rows': sheet.nrows - 1  # subtract 1 for the header row
```
toolness: Would it be useful to have a named constant or something for the number of header rows?

data_capture/spreadsheet_converter.py
```diff
@@ -0,0 +1,138 @@
...
+                    cell_value = date.strftime('%m/%d/%Y')
+
+                # Save the string value into the expected CSV col
+                # index of the row
+                row[csv_col_idx] = str(cell_value)
+
+            yield row
+
+        self.xls_file.seek(0)
+        return
```
toolness: Is this explicit `return` needed?

data_capture/templates/data_capture/bulk_upload/region_10_step_2.html
```diff
@@ -0,0 +1,32 @@
...
+      <a href="{% url 'data_capture:bulk_region_10_step_1' %}" class="button button-previous">Previous</a>
+
+      <a href="{% url 'index' %}" class="button button-cancel">Cancel</a>
+
+      {% if total %}
+      <div class="submit-group">
+        <span class="submit-label">
+          Load data and view results
+        </span>
+        {# TODO: It would be nice to jave JS that disables this button once clicked #}
```
toolness: minor typo, `jave` -> `have`

data_capture/templates/data_capture/price_list/step_1.html
```diff
@@ -7,11 +7,11 @@
 {% endblock %}
 
 {% block step_body %}
-{% include 'data_capture/step_1_form.html' %}
+{% include 'data_capture/price_list/step_1_form.html' %}
 
 {% if show_debug_ui %}
 {% load static %}
-<p style="color: gray">
+<p class="dev-note">
```
toolness: Thanks for fixing that inline CSS!

contracts/models.py
jseppi: Added it to the list in the root PR description. I'll turn those all into issues once this PR is merged.

data_capture/tests/common.py
```diff
@@ -25,3 +32,51 @@ def uploaded_csv_file(content=None):
...
+)
+class StepTestCase(TestCase):
+    def login(self, is_staff=False):
+        user = User.objects.create_user(
+            username='foo',
+            password='bar'
+        )
+        if is_staff:
+            user.is_staff = True
+            user.save()
```
toolness: Huh, I *think* that you can actually just pass `is_staff` as another keyword argument to `create_user`?  No big deal either way.

data_capture/spreadsheet_converter.py
jseppi: nope, will remove. I had mistakenly thought it was for the generator, but just checked and it is not.

data_capture/tests/common.py
jseppi: Nope, you cannot. It's a Django bug that I think is fixed in 1.10 if I am reading correctly: https://code.djangoproject.com/ticket/25009

toolness:
This looks good to me!! Feel free to merge!

data_capture/spreadsheet_converter.py
jseppi: I had considered that, but this won't work at all if there is more than 1 header row (due to how `get_heading_indices_map` looks at the `0`th row to determine headers).

data_capture/tests/test_bulk_upload_views.py
```diff
@@ -0,0 +1,184 @@
+import json
+from .common import StepTestCase, R10_XLSX_PATH
```
toolness: Oh, I just noticed this... I'm not sure where I got this convention from--whether it was PEP8 or somewhere else--but I tend to break up my `import`s into two groups, separated by a blank line: the first group is all first and third-party libraries, preferably with all the first-party ones first, while the second group is all the app-specific modules.

So in this example I'd put a blank line after line 1, and remove the blank line at line 3.

jseppi: Oh yeah, I usually try to follow that also. Must have just been sloppy on this one. Feel free to change in your current PR.

toolness: Cool cool I fixed it in #606 I think.

PR #556, opened 2016-08-15
Fix for Blobs in Phantom 1.9.x and 2.1.x

jseppi:
I checked both natively and within our docker-compose setup.

@toolness: Please verify this works, and merge if so :)


toolness:
Woot travis likes it and so does my docker setup!

PR #559, opened 2016-08-15
move djdt middleware to end of middleware stack

jseppi:
This is to fix an issue we encountered after #552. The issue was that form csrf tokens were not being rendered because the DjDT ProfilingPanel was preventing the CsrfViewMiddleware's `process_view` from running.


jseppi:
Merging! @toolness 👍 'd it in Slack.

PR #561, opened 2016-08-16
Implement SplitDateField and use it in step 3

### LINKED ISSUE #534: Date fields should each be 3 separate fields
As mentioned in #415, the [Date input](https://standards.usa.gov/form-controls/#date-input) section of the USWDS claims that three text fields are the easiest way for users to enter most dates:

> screen shot 2016-08-08 at 3 32 32 pm

Currently our date fields are just standard text input fields, and we've observed at least one user enter the incorrect format in testing sessions.

It seems like Django's [`MultiValueField`](https://docs.djangoproject.com/en/1.9/ref/forms/fields/#django.forms.MultiValueField) might come in handy here.

toolness:
This is an ~~experimental~~ attempt to fix #534 by using Django's [`MultiValueField`](https://docs.djangoproject.com/en/1.9/ref/forms/fields/#django.forms.MultiValueField) and [`MultiWidget`](https://docs.djangoproject.com/en/1.9/ref/forms/widgets/#django.forms.MultiWidget):

> screen shot 2016-08-18 at 7 48 31 am

The implementation feels kind of weird, mostly because of the way Django's architecture forces a separation between form inputs and their labels. This makes it difficult to render a single multi-value field/widget as its own "component".

Another weirdness is that our date widget is meant to be placed in a `<fieldset>`, with the main "label" (e.g. "contract start date") in a `<legend>` rather than a `<label>`, which Django's form rendering framework *really* doesn't like.  We'll probably need to ensure that users don't use this widget with standard form-rendering methods like `as_p()`, `as_table()`, and so on, which is a bummer.

An alternative is to have each sub-field (month, day, year) be its own actual form field, and combine them in the form's `clean()` method, but this could get rather annoying if there are lots of dates in a form.

Also, if we use lots of forms with dates, we can always add a custom template filter that does the form rendering for us, e.g. `{{ form|our_cool_form_renderer }}` (this is how Django crispy forms does things).

To do:

- [x] Add tests. There's lots of weird branching here and Django's documentation doesn't fully elucidate what kinds of edge cases might occur in a multi-value field/widget and why.
- [x] Replace the date widget example in the style guide.
- [x] ~~It might also be nice to add a "disable form validation" button to the styleguide pages that runs the JS mentioned at [novalidate.com](http://novalidate.com/), to make it easy to simulate what will happen on browsers that don't have HTML5 form validation.~~ Spun off to #568.
- [x] Replace the date fields in the data capture form with this new field/widget. Unfortunately I don't see any way to do this that doesn't involve replacing our `form.as_p()` with a manual displaying of every single widget on the form, which is annoying.

### DIFF
--- a/data_capture/templates/data_capture/step_3.html
+++ b/data_capture/templates/data_capture/step_3.html
@@ -8,7 +8,44 @@
 <form method="post" action="{% url 'data_capture:step_3' %}">
 {% csrf_token %}
 
-{{ form.as_p }}
+{{ form.non_field_errors }}
+
+{{ form.contract_number.errors }}
+{{ form.contract_number.label_tag }}
+{{ form.contract_number }}
+<span class="helptext">{{ form.contract_number.help_text }}</span>
+
+{{ form.vendor_name.errors }}
+{{ form.vendor_name.label_tag }}
+{{ form.vendor_name }}
+
+{{ form.is_small_business.errors }}
+{{ form.is_small_business.label_tag }}
+{{ form.is_small_business }}
+
+{{ form.contractor_site.errors }}
+{{ form.contractor_site.label_tag }}
+{{ form.contractor_site }}
+
+{{ form.contract_year.errors }}
+{{ form.contract_year.label_tag }}
+{{ form.contract_year }}
+
+<p>
+  <fieldset>
+    <legend>{{ form.contract_start.label }}</legend>
+    {{ form.contract_start.errors }}
+    {{ form.contract_start }}
+  </fieldset>
+</p>
+
+<p>
+  <fieldset>
+    <legend>{{ form.contract_end.label }}</legend>
+    {{ form.contract_end.errors }}
+    {{ form.contract_end }}
+  </fieldset>
+</p>
 
 <div class="form-button-row clearfix">
   <a href="{% url 'data_capture:step_2' %}" class="button button-previous">Previous</a>
--- a/styleguide/templates/styleguide.html
+++ b/styleguide/templates/styleguide.html
@@ -366,27 +366,24 @@
   </select>
   {% endexample %}
 
-  {% example %}
-  <fieldset>
-    <legend>Date of birth</legend>
-    <span class="form-hint" id="dobHint">For example: 04 28 1986</span>
+  {% guide_section "Dates" %}
 
-    <div class="date-mm-dd-yy">
-      <div class="form-group form-group-month">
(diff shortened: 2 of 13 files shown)


data_capture/templates/data_capture/step_3.html
```diff
@@ -8,7 +8,44 @@
...
+{{ form.contractor_site.errors }}
+{{ form.contractor_site.label_tag }}
+{{ form.contractor_site }}
+
+{{ form.contract_year.errors }}
+{{ form.contract_year.label_tag }}
+{{ form.contract_year }}
+
+<p>
+  <fieldset>
```
jseppi: Is it semantic to have a `<fieldset>` inside of a `<p>`? Is `<fieldset>` block-level?

styleguide/templates/styleguide.html
```diff
@@ -366,27 +366,24 @@ <h3 class="example-heading">Example (no JavaScript)</h3>
...
-        <label for="date_of_birth_2">Day</label>
-        <input class="input-inline" aria-describedby="dobHint" id="date_of_birth_2" name="date_of_birth_2" pattern="0?[1-9]|1[0-9]|2[0-9]|3[01]" type="number" min="1" max="31" value="">
-      </div>
-      <div class="form-group form-group-year">
-        <label for="date_of_birth_3">Year</label>
-        <input class="input-inline" aria-describedby="dobHint" id="date_of_birth_3" name="date_of_birth_3" pattern="[0-9]{4}" type="number" min="1900" max="2000" value="">
-      </div>
+  {% guide_section "Dates" %}
+
+  <p>
```
jseppi: Missing a closing `</p>` on this one

jseppi:
Could this be spun-off so that other 18F Django projects could use it since it is implementing a component of the USWDS?

jseppi:
I think this looks good. It seems like a lot of code for a relatively simple widget but I guess that's how it is in Django.

toolness:
Heh heh, yeah.  Admittedly, a fair chunk of it is actually the stuff I added to the style guide--I'd kinda like to refactor things at some point so it's a lot easier to add a styleguide entry/example with a back-end component like we're currently doing for `ajaxform` and `date`.

Also, I knew when writing the original `{{ form.as_p }}` that we'd probably eventually have to migrate away from it, since it's so horribly inflexible.  So that was kind of inevitable. :)

toolness:
Ok, merging now!

PR #563, opened 2016-08-16
fix typo in styleguide ajax form example.

toolness:
Doh. I mistakenly wrote `.label` in the template when I should have written `.label_tag`.  Oof.


jseppi:
👍

PR #564, opened 2016-08-16
Make eventual migration to Django 1.9 easier

toolness:
**Note: This PR does not actually migrate us to Django 1.9.** It just migrates to new ways of doing things that were introduced in Django 1.8 or earlier, but which are removed in Django 1.10 (and thus raise warnings in 1.9), and/or which simply break in Django 1.9.

Stuff changed includes, but is not limited to:

* Using [`settings.TEMPLATES`](https://docs.djangoproject.com/en/1.10/ref/templates/upgrading/).
* Not [passing `RequestContext` instances](https://docs.djangoproject.com/en/1.10/releases/1.8/#dictionary-and-context-instance-arguments-of-rendering-functions) directly to template-rendering functions.
* Using lists instead of `django.conf.urls.patterns()`.

Stuff *not* changed includes, but is not limited to:

* It looks like Django 1.9 changes `Location` headers to not include the server's origin (e.g. `http://testserver`) anymore, which messes up some of our tests. We can just change our tests when we actually upgrade to 1.9.
* `manage.py test data_explorer` fails because of #291, which is obviously non-trivial to fix, but it also raises some other weird errors that I'm not going to deal with right now.

Notes to my future self:

* I had to set `PYTHONWARNINGS=error:render` to e.g. convert any warnings that started with the word "render" into errors, so that I could get a useful traceback instead of a useless line number.


jseppi:
Your "Stuff not changed" items are only a problem if we actually moved to Django 1.9, right?

👍  if so!

toolness:
Yup that is correct!  Cool, merging now!

PR #566, opened 2016-08-17
Use ContractSource to identify Contract models/rows

jseppi:
@toolness This is the start of an idea I had to add a ContractSource model for Contract provenance. What do you think of this?


data_capture/models/region_10.py
```diff
@@ -1,82 +1,91 @@
...
+    updated_at = models.DateTimeField(auto_now=True)
     submitter = models.ForeignKey(User)
-    serialized_gleaned_data = models.TextField(
-        help_text=(
-            'The JSON-serialized data from the upload, including '
-            'information about any rows that failed validation.'
-        )
-    )
+    original_file = models.BinaryField()  # TODO: Not sure if this is necessary
+    procurement_center = models.TextField()  # ie "Region 10"
```
jseppi: This is how we would identify which Contract models came from Region 10 in order to blow them away and replace with new data from a Region 10 bulk upload.

toolness: Oh, it took me a while to realize this change is being made to `region_10.py` and not the existing `SubmittedPriceList` model. Heh.

Yeah, I think that is a good idea, as long as there aren't any privacy concerns raised by it. I don't think there are--I was thinking of storing the original XLS file for submitted price lists in the `SubmittedPriceList` model at first, but I was worried that there might be other sensitive PII in it, that we don't glean, that might not be a good idea to store persistently. But since the region 10 bulk upload files are just CSVs containing exactly the data we need, it seems to make sense to just save it?

jseppi: Cool! Actually I'm planning on making this upload work with the original XLSX that Kelly produces from his Access DB, since it is actually some (unnecessary) work to turn that into a CSV.

toolness: Ah cool! Well as long as that XLSX doesn't have a worksheet in it that lists the names and addresses of all his employees or something, I think that's great! 😁

jseppi:
closing this PR, will implement back in the original PR

PR #567, opened 2016-08-17
Add support for sending email

toolness:
This adds [dj-email-url](https://github.com/migonzalvar/dj-email-url), and also cherry-picks the [`manage.py sendtestemail`](https://docs.djangoproject.com/en/1.10/ref/django-admin/#sendtestemail) command from Django 1.9 (we'll remove it once we actually upgrade to 1.9).

It also adds documentation about configuring email to the readme.

(Context: I asked around on Slack and it looks like we use Mandrill for email sending, which is super easy to use and works via SMTP, so this PR should give us what we need.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,11 @@ string), the boolean is true; otherwise, it's false.
 * `DATABASE_URL` is the URL for the database, as per the
   [DJ-Database-URL schema][].
 
+* `EMAIL_URL` is the URL for the service to use when sending
+  email, as per the [dj-email-url schema][]. When `DEBUG` is true,
+  this defaults to `console:`. The setting can easily be manually
+  tested via the `manage.py sendtestemail` command.
+
 * `ENABLE_SEO_INDEXING` is a boolean value that indicates whether to
   indicate to search engines that they can index the site.
 
@@ -405,6 +410,7 @@ for other than small business.
 [SASS]: http://sass-lang.com/
 [`deploy.md`]: https://github.com/18F/calc/blob/master/deploy.md
 [DJ-Database-URL schema]: https://github.com/kennethreitz/dj-database-url#url-schema
+[dj-email-url schema]: https://github.com/migonzalvar/dj-email-url#supported-backends
 [pytest]: https://pytest.org/latest/usage.html
 [docker-machine-cloud]: https://docs.docker.com/machine/get-started-cloud/
 [Django Debug Toolbar]: https://github.com/jazzband/django-debug-toolbar/
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -11,6 +11,7 @@ https://docs.djangoproject.com/en/1.7/ref/settings/
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 import os
 import dj_database_url
+import dj_email_url
 from dotenv import load_dotenv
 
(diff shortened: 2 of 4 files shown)


hourglass/settings.py
```diff
@@ -34,6 +35,19 @@
...
+email_config = dj_email_url.config()
+
+EMAIL_FILE_PATH = email_config['EMAIL_FILE_PATH']
+EMAIL_HOST_USER = email_config['EMAIL_HOST_USER']
+EMAIL_HOST_PASSWORD = email_config['EMAIL_HOST_PASSWORD']
+EMAIL_HOST = email_config['EMAIL_HOST']
+EMAIL_PORT = email_config['EMAIL_PORT']
+EMAIL_BACKEND = email_config['EMAIL_BACKEND']
+EMAIL_USE_TLS = email_config['EMAIL_USE_TLS']
+EMAIL_USE_SSL = email_config['EMAIL_USE_SSL']
```
toolness: As documented in the instructions for dj-email-url, an alternative to this is to just do `vars().update(email_config)`, but I opted to be explicit. We could do it the less verbose way if folks want, though.

README.md
```diff
@@ -217,6 +217,11 @@ string), the boolean is true; otherwise, it's false.
 * `DATABASE_URL` is the URL for the database, as per the
   [DJ-Database-URL schema][].
 
+* `EMAIL_URL` is the URL for the service to use when sending
+  email, as per the [dj-email-url schema][]. When `DEBUG` is true,
```
toolness: I'm not a big fan of the inconsistent capitalization of dj-email-url vs. DJ-Database-URL, but that is how the authors of those packages actually capitalize their packages' names in their readmes.

meta/management/commands/sendtestemail.py
```diff
@@ -0,0 +1,45 @@
...
+import socket
+
+from django.core.mail import mail_admins, mail_managers, send_mail
+from django.core.management.base import BaseCommand
+from django.utils import timezone
+
+
+class Command(BaseCommand):
+    help = "Sends a test email to the email addresses specified as arguments."
+    missing_args_message = "You must specify some email recipients, or pass the --managers or --admin options."  # NOQA
```
jseppi: are all the `# NOQA` comments to disable line length warnings?

hourglass/settings.py
jseppi: Is it guaranteed that all those keys will be in `email_config` or is there a chance of a getting a `KeyError`?

jseppi:
👍  Looks good, just some minor questions.

meta/management/commands/sendtestemail.py
toolness: Yup. Weird that Django doesn't conform to PEP 8... Also because the file is verbatim from Django's source code, and temporary to boot, I didn't want to manually line wrap the stuff myself.

hourglass/settings.py
toolness: Yup, the snippet was taken directly from the [usage docs](https://github.com/migonzalvar/dj-email-url#usage).  (It's also evident that no `KeyError`s are being raised when the email url is set to `console:`, which is the current default setting, so that's reassuring.)

toolness:
Wooot merging now!

PR #571, opened 2016-08-18
Fix #499 by making 'is small business' radios.

### LINKED ISSUE #499: Make "is small business" a radio button
I guess the options in the radio button should be something like:

* This is a small business.
* This is something other than a small business.

... or something?  Thoughts on the wording @jseppi @hbillings ?  I guess we could ask on `#g-content` or the writing lab too.

toolness:
This fixes #499:

> screen shot 2016-08-18 at 7 41 41 am

As discussed earlier, this UI requires the user to explicitly choose one of the radios, rather than defaulting to "not a small business", which is what happens when we use a single checkbox.

~~Currently this PR is against #561, but it doesn't actually *require* any of the functionality in that PR; I'm just doing it b/c there will be slightly annoying merge conflicts otherwise, since both PRs modify the `Step3Form`. (Due to the fact that github PRs can now change the base branch they're being merged into, though, it'll be easy to first merge #561 and later merge this one, if we want, which is nice.)~~

@hbillings does this look OK?

### DIFF
--- a/frontend/source/sass/components/_forms.scss
+++ b/frontend/source/sass/components/_forms.scss
@@ -33,9 +33,9 @@ label input[type="radio"] {
   width: auto; // WDS override for width: 100%;
 }
 
-/* A <ul> with an id attribute is a Django-rendered radio widget. */
+/* A <ul> holds a Django-rendered list of radio buttons. */
 
-form ul[id] {
+form ul {
   list-style-type: none;
 }
 
--- a/frontend/source/sass/components/_forms.scss
+++ b/frontend/source/sass/components/_forms.scss
@@ -33,6 +33,12 @@ label input[type="radio"] {
   width: auto; // WDS override for width: 100%;
 }
 
+/* A <ul> with an id attribute is a Django-rendered radio widget. */
+
+form ul[id] {
+  list-style-type: none;
+}
+
 /* TODO: Django's default form rendering uses a <ul class="errorlist">
    for errors, and <span class="helptext"> for help text. We should either
    style them properly, or render our own markup. The following are some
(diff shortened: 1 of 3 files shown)


frontend/source/sass/components/_forms.scss
```diff
@@ -33,6 +33,12 @@ label input[type="radio"] {
   width: auto; // WDS override for width: 100%;
 }
 
+/* A <ul> with an id attribute is a Django-rendered radio widget. */
+
+form ul[id] {
```
toolness: This is admittedly a bit of a weird selector. I think it might be possible to tell Django to add a specific `class` to the markup it generates, but then there's the danger of someone forgetting to add that class to a field object...

jseppi: Hrm, that does seem weird. Could also make a custom widget, right? http://stackoverflow.com/questions/13704592/how-do-i-set-a-class-for-the-ul-of-a-typedchoicefield

Though then we'd have to remember to use the custom widget.

Or maybe just make any `form ul` have `list-style-type: none;` and add another class later if we end up needing to have a bulleted list within a form?

hbillings: I'm fine with generally removing the styling from `form ul`s -- seems like it'd have to be an odd use case that actually required bullet points within a form. We could also potentially use sibling selectors and use the `label` as the fuzzy-match point of reference, like `label[for*="small_business"] + ul`. That has the benefit of being a li'l more readable to future teams (including future us).

toolness:
Sounds good, just made the radios be styled by `form ul`!

jseppi:
:+1:

PR #572, opened 2016-08-18
Only expose QUnit tests when DEBUG is enabled

toolness:
I thought we had an issue filed for this but apparently we don't, or I can't find it.

This disables the `/tests/` ~~and `/styleguide/`~~ endpoint on production, purely for the sake of reducing the attack surface of the app.

~~I'm not sure how I feel about disabling the styleguide here, actually.  It's nice to be able to easily point to it as an example of a style guide (of sorts).  I guess we could add another environment setting that allows it to be exposed even when `DEBUG` is `False`, which we can enable on dev/staging but not prod...~~

~~Note that this PR is currently against #564 because that PR changes `urlpatterns` to be standard arrays. We can always re-target this PR to `develop` if that PR is merged before this one though.~~

### DIFF
--- a/frontend/tests/test_qunit.py
+++ b/frontend/tests/test_qunit.py
@@ -1,8 +1,9 @@
 import os
 import subprocess
 
-from django.test import LiveServerTestCase
+from django.test import LiveServerTestCase, override_settings
 
+from hourglass.urls import urlpatterns, tests_url
 from .utils import build_static_assets
 
 
@@ -10,6 +11,10 @@ RUNNER_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                            'vendor', 'runner.js')
 
 
+urlpatterns += [tests_url]
+
+
+@override_settings(ROOT_URLCONF=__name__)
 class QunitTests(LiveServerTestCase):
 
     def test_qunit(self):
--- a/hourglass/urls.py
+++ b/hourglass/urls.py
@@ -21,18 +21,22 @@ urlpatterns = [
     url(r'^data-capture/',
         include('data_capture.urls', namespace='data_capture')),
     url(r'^admin/', include(admin.site.urls)),
-    url(r'^tests/$', TemplateView.as_view(template_name='tests.html')),
-    url(r'^styleguide/', include('styleguide.urls', namespace='styleguide')),
     url(r'^robots.txt$', robots_txt),
     url(r'^auth/', include('uaa_client.urls', namespace='uaa_client')),
 ]
 
+tests_url = url(r'^tests/$', TemplateView.as_view(template_name='tests.html'))
+
+styleguide_url = url(r'^styleguide/', include('styleguide.urls',
+                                              namespace='styleguide'))
+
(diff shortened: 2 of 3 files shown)


frontend/tests/test_qunit.py
```diff
@@ -1,15 +1,20 @@
...
 
 
 RUNNER_PATH = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                            'vendor', 'runner.js')
 
 
+urlpatterns += [tests_url]
+
+
+@override_settings(ROOT_URLCONF=__name__)
```
toolness: I didn't know about [`ROOT_URLCONF`](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-ROOT_URLCONF) until just now. Pretty handy!

jseppi: I'm having trouble understanding what this override is doing...care to explain?

jseppi:
Cool! I definitely like disabling the `/tests/` endpoint since that just seems strange to me to have on prod. I do agree that it could be nice to leave the styleguide live though, but don't really want to add another env setting to do it. Maybe we should just leave the styleguide in?

frontend/tests/test_qunit.py
toolness: Oh, so basically, it looks like tests are being run with `DEBUG=False`, which actually appears to be a best practice based on conversation in our Django slack channel.  But since this PR moves the tests and styleguide endpoints inside the `if settings.DEBUG:` conditional in `urls.py`, it means that our tests can't actually make requests against those endpoints.

Simply overriding `DEBUG=True` doesn't work, because the `urls.py` module has already been loaded into the python process from earlier tests (and though module reloading is technically possible in Python, it's generally not advised).

So instead we're creating a new `urlpatterns`.  There's a *bit* of magic happening here, because `ROOT_URLCONF` is supposed to point to a module that contains an attribute called `urlpatterns`.  We're defining `urlpatterns` in the test module, and to keep things DRY we're setting `ROOT_URLCONF` to `__name__`, which is basically the fully-qualified dotted name of the current module (i.e. `frontend.tests.test_qunit` in this case). Thus when this test suite runs, it will configure the app's URLs based on `frontend.tests.test_qunit.urlpatterns` (which exposes our `/tests/` endpoint) instead of `hourglass.urls.urlpatterns`.

Personally I think this would be a bit easier to understand if the setting were called `ROOT_URLPATTERNS` and the value were set to e.g. `hourglass.urls.urlpatterns'`, since I'm not really sure what a `urlconf` is. I think it's basically "a module that contains an attribute called `urlpatterns`" but it just seems to add to the jargon one needs to learn to use Django.

jseppi: whoa o_O

Thanks for the detailed explanation!

toolness:
Oh, that's not a bad idea either!  Will add that to the PR.

toolness:
Done!  Now only `/tests/` is disabled on production.

jseppi:
👍

PR #574, opened 2016-08-19
Fix typo in ajaxform.js.

toolness:
Ack! Just spotted a big typo.


jseppi:
👍  Maybe safer to use a different name for the variable `form`? Like `formEl` or something like that?

toolness:
Good idea!  Will revisit renaming after #581, which could mean even more changes...

PR #575, opened 2016-08-19
Enable aria alerts for ajaxform.

### LINKED ISSUE #481: Ajaxform updates aren't noticeable to screen readers
We ajax-in the result of step 1 of data capture, but screen reader users will probably be befuddled ~~because we haven't surrounded the injection site with an `aria-live` region. Will probably want to test this out on a screenreader and make sure it's not confusing~~.

Update: See #575.

toolness:
This is an attempt to fix #481.

Unfortunately I was hoping just adding `aria-live` somewhere would be enough, but it's not, really. The [web accessibility tutorial on user notifications - listing errors](https://www.w3.org/WAI/tutorials/forms/notifications/#listing-errors) spells everything out in detail, but the gist of it is that we need a `<div role="alert">` at the top of the page that we put stuff in when new stuff is loaded into the page. Blech.

This PR attempts to accomplish this by copying any elements with `role="alert"` and pasting them into a screenreader-only `<div role="alert">` at the top of the page.

### DIFF
--- a/data_capture/templates/data_capture/messages.html
+++ b/data_capture/templates/data_capture/messages.html
@@ -1,5 +1,5 @@
 {% if messages %}
-<div class="alerts">
+<div class="alerts" role="alert">
   {% for message in messages %}
     <div class="alert alert-{{ message.tags }}">
       <p>{{ message }}</p>
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -2,6 +2,8 @@
 
 const $ = jQuery;
 
+let $ariaAlerts = null;
+
 let delegate = {
   redirect(url) {
     window.location = url;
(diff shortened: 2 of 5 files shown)


data_capture/templates/data_capture/messages.html
```diff
@@ -1,5 +1,5 @@
 {% if messages %}
-<div class="alerts">
+<div class="alerts" role="alert">
```
toolness: I'm wondering if we want to just move this template into the `frontend` django app, since, like `ajaxform` itself, it seems pretty app-agnostic.

toolness:
Hmm, so if we end up going w/ WebComponents (see #581), this could potentially be a lot cleaner.

toolness:
Blerg, I hate this solution and we're going to be doing web components anyways, so I'm going to close this.

PR #578, opened 2016-08-19
Remove steps widget from the a11y tree.

### LINKED ISSUE #413: Current state of steps widget is not perceivable by screen readers
Because information about what step the user is currently on is entirely conveyed through styling, there's no way for screen reader users (or, uh, users on `links`) to know what step they're currently on.

As described in #381, here's one solution:

> We could alleviate this through the use of [invisible content just for screen reader users](http://webaim.org/techniques/css/invisiblecontent/) (this is `.sr-only` in Bootstrap, I'm not sure if we already have such a class in CALC). For example, `<aside class="sr-only">You are currently on this step</aside>` inside a list item would make it obvious to screen reader users the step they're currently on.

toolness:
Ok, this is *one* approach to fix #413--by simply removing the steps widget from the accessibility tree entirely.

I think there's a decent argument for this: the fact that it's repeated on every step could get really annoying to vision-impaired users.  And we're always reiterating the current step via a `<h1>` or `<h2>` header on the page, so the user won't be lost if we hide the widget from them.  Hiding the widget also ensures that they won't be confused by it (which will currently be the case, since there's no way for them to tell which step they're currently on--it just sounds like a list of steps).

If this doesn't seem like a good idea to folks, we can go with the solution proposed in #413, whereby we add a `sr-only` "You have completed this step" / "You are currently on this step" / "You haven't reached this step yet" to every single list item.  But this will likely require writing a template helper or something, since it'll get extremely repetitive to write this out for every single step widget we use; as I mentioned earlier, I also have doubts as to the usefulness of even doing this.  It'd be nice to at least do some user testing before embarking on that path.

### DIFF
--- a/styleguide/templates/styleguide.html
+++ b/styleguide/templates/styleguide.html
@@ -119,11 +119,21 @@
   </p>
 
   <p>
-    Use <code>li.steps-done</code> to identify a step as being completed.
+    Use <code>li.current</code> to identify a step as being the current
+    one.
+  </p>
+
+  <p>
+    We recommend adding <code>aria-hidden="true"</code> to the list, as
+    there isn't an easy way to communicate the information to vision-impaired
+    users without being overly verbose (thus annoying the user) or confusing.
+    However, <em>do</em> be sure to repeat the name of the current step
+    elsewhere in a header, so that the user knows what step they're currently
+    on.
   </p>
 
   {% example %}
-  <ol class="steps">
+  <ol class="steps" aria-hidden="true">
     <li>
       <label>Upload price list</label>
     </li>
--- a/data_capture/templates/data_capture/step.html
+++ b/data_capture/templates/data_capture/step.html
@@ -1,7 +1,7 @@
 {% extends 'base.html' %}
 {% load static %}
 
-{% block title %}Add price data to CALC / {% block subtitle %}{% endblock %}{% endblock %}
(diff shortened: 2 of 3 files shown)


styleguide/templates/styleguide.html
```diff
@@ -119,11 +119,21 @@ <h3 class="example-heading">Example (no JavaScript)</h3>
   </p>
 
   <p>
-    Use <code>li.steps-done</code> to identify a step as being completed.
+    Use <code>li.current</code> to identify a step as being the current
```
toolness: Ooops, doc rot!  This text wasn't changed when we merged #442.

jseppi:
I'm not well versed in a11y concerns, but your argument to hide it seems reasonable. Perhaps run it by #g-accessibility?

hbillings:
I like the idea of hiding it. I agree that the way it's structured is repetitive to screen readers.

Could we give sight-impaired users a similar wayfinding experience by adding some `sr-only` text to the heading? I.e., "Step 2 of 5: Upload your spreadsheet"?

toolness:
Ah great idea! I'll add that to the PR.

toolness:
Ok, the page title now looks like this:

```html
<title>CALC / Add price data to CALC / Step 1 of 4: Upload price list</title>
```

And the step number is also in a `sr-only` in the first `<h1>` of each step.

toolness:
Blerg, I should probably just wait until #583 is merged before resolving the merge conflicts here.

toolness:
This PR is superseded by #734.

PR #579, opened 2016-08-19
Friendlier ajax error copy.

### LINKED ISSUE #456: Review and polish ajax error edge cases in data capture app
In #436 I used `window.alert()` in `hourglass_site/static_source/js/data-capture/step_1.js` to deal with two edge cases when submitting step 1 of data capture via ajax:

* **If the server returns a success (200-ish) response that's unexpected.** Currently this alert says `Invalid server response: ${data}`, where `${data}` is the raw response from the server. Given that this is almost more of an assertion failure than something that users will actually encounter in the wild, I *suspect* it might be ok for the message to be fairly technical in nature, but I'm not sure. I guess we might also want to report some sort of custom google analytics/new relic event or something, too, since users should never really see this.

* **If the ajax request fails.** In jQuery-land, I believe this may be due to broken internet connection, request timeout, or non-200-ish responses like 404 or 500. Currently this alert says `An error occurred when submitting your data.`. At the very least, even if we continue to use `window.alert()` here, we should probably check the content of this alert to make sure it's not confusing to users, since they're more likely to see it if e.g. their internet connection is spotty.

We might also want to use this as an opportunity to define an ajax error handling strategy for the data capture app in general. For example, we could use jQuery's [`ajaxError()`](http://api.jquery.com/ajaxerror/) to always handle any ajax errors in the same way, etc.

toolness:
This is an attempt to fix #456 by using friendlier copy when network or other unexpected errors occur. I [asked on `#g-content`](https://18f.slack.com/archives/g-content/p1471615533000559) in Slack and they recommended this:

> screen shot 2016-08-19 at 10 36 11 am

Because these errors happen so infrequently (um, hopefully), I think it's ok to keep using a standard `window.alert()` for now, but let me know if you disagree!

### DIFF
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -2,6 +2,9 @@
 
 const $ = jQuery;
 
+const MISC_ERROR = 'Sorry, we’re having trouble. ' +
+                   'Please try again later or refresh your browser.';
+
 let delegate = {
   redirect(url) {
     window.location = url;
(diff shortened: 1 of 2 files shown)


jseppi:
👍

Issue #581, opened 2016-08-20
Consider converting data-capture widgets into Web Components


toolness:
Ok, I've watched Thomas Wilburn's [Custom Elements in Production](http://2016.cascadiafest.org/speakers/thomas-wilburn/) talk that @jseppi recommended and I've played around with making some trivial components using the [webcomponents.js](https://github.com/WebComponents/webcomponentsjs) polyfill.

Overall, I think that this is a pretty straightforward tool that can potentially make our code cleaner, more predictable, and better encapsulated.

Advantages:

* The biggest advantage I see over our current jQuery-driven markup API approach is that web components are automatically discovered and "upgraded" by the browser. This means, for example, that if we load some new content in via Ajax, we can count on our custom elements being automatically upgraded, rather than having to manually upgrade them ourselves (which is currently the case). This is really nice!

* Another bonus of auto-upgrading is that it makes it easier to document our components. No more "oh, and if you load in content via ajax, make sure you call this one function" in our docs.

* Our code is *probably* going to be less jQuery spaghetti-ish, but I'm not sure.

* Depending on how we do things, our code might be less error-prone and progressively enhance better, but I'm not sure.

Disadvantages/Concerns:

* According to Wilburn's talk, Web Component polyfills can reliably work in IE10+... I'm not sure how many of our users will be using something older than that.  Based on the user interviews we've done so far, it *seems* like none of them will, but we should make sure of that.  Fortunately our progressive enhancement strategy should help us here, since all our widgets already work fine on browsers without JS enabled.

* It looks like `webcomponents-lite.js`, which is the only one we *really* need, is about 40k minified, and around 15k gzipped, which isn't bad.  Granted, that also includes HTML imports (which we probably don't need), and I think it might even include Weak Map support (which should already be provided by our babel polyfill) so we can probably slim that down even more if needed. I think that Wilburn doesn't even use this particular polyfill--he uses [document-register-element](https://github.com/WebReflection/document-register-element).

* I'm a bit concerned about the whole "upgrade timing" thing, but it *seems* as though elements are upgraded if we wait for `DOMContentLoaded` or `load`?  I guess we should read up on this and make sure that our code won't explode on some browsers due to upgrade timing issues... Or at least that, if upgrade timing issues do occur, the core "gracefully degraded" functionality of our widgets still works.

* We'll have to think carefully about how we want to reconcile our Web Components with progressive enhancement. For instance, I could imagine our upgrade widget looking like this, maybe:

  ```html
[...]

PR #582, opened 2016-08-20
Convert data capture widgets to use Web Components

toolness:
This is an exploration of implementing #581.

Here's how a basic Ajax Form + Upload Widget looks like now:

```html
<form is="ajax-form" enctype="multipart/form-data" method="post">
  <upload-widget>
    <input is="upload-input" type="file" accept=".csv" id="id_file" name="file" />
    <div class="upload-chooser">
      <label for="id_file">Choose file</label>
      <span>CSV only, please.</span>
    </div>
  </upload-widget>
  <button type="submit" class="button-primary">Submit</button>
</form>
```

So not a lot has changed with respect to the actual markup, really. Mostly because we do want to preserve progressive enhancement.

## Notes

* I made *two* components to encapsulate the upload widget: `<upload-widget>` and, within it, a `<input is="upload-input" type="file">`.  These are really tightly coupled; one can't really exist without the other.  There might be better ways to do this but right now this is the one I've gone with to preserve our progressive enhancement strategy.

  That said, the two components *do* have a somewhat different separation of concerns, at least. The widget is responsible for presenting user feedback and receiving dropped files, while the input is responsible for keeping track of what file is currently picked and emitting events whenever that changes (or when an invalid file is picked).

* The ajax form is no longer tightly coupled to the upload widget--nor does it require one. Instead of expecting one, it just checks every input element it has to see if its `isUpgraded` property is true.  If it is, it appends the element's `upgradedValue` property to its `FormData` instance instead of using other introspective means (e.g. looking at `value`, etc).

* An undocumented limitation of ajaxform was that there could actually only be one of them on a page at a time.  This PR removes that restriction.

* Previously, the upload widget would check to see if a `data-force-degradation` attribute was on it to see if it should degrade itself.  However, it now uses [jQuery's `.closest()`](https://api.jquery.com/closest/) to see if any ancestors up the DOM tree have it.  Among other things, this:

    * Makes writing tests easier.

    * Makes it easier for us to just plop a `<html data-force-degradation>` onto the top of the page to put the whole thing into "safe mode" (which could come in handy for folks with unusual browsers who may be having issues with our site, if we expose it in a smart way).

* The ajaxform component now supports a `data-force-degradation` attribute in the same way as the upload widget.

## To Do

- [x] Fix `upload_tests.js`.
- [x] Fix `ajaxform_tests.js`.
- [x] Investigate whether our upgrade timing code is still necessary after we converted the components to use `attachedCallback` instead of `createdCallback`. (Alas, the answer is nope.)
- [x] Figure out how to bring back the "only enable submit button when a file has been picked" button. Probably want to lean on HTML5 form validation here. **Update:** See #584 and #585 for two different solutions.
- [x] Add an ajaxform test to make sure it works fine with non-upgraded upload widgets.
- [x] Move any new npm dependencies out of the `Dockerfile` and into `package.json` (I'm adding them to the `Dockerfile` for now to make iteration quicker).
- [x] Web component-ify the ajaxform component.
- [x] Link to Modernizr license in `feature-detection.js`.
- [x] Figure out licensing issues for `CustomEvent` polyfill.
- [x] Refactor code so that functions are methods of classes rather than just hangin' out.
- [x] Directly require/import dependencies in the tests rather than weirdly exporting them to the window object. This will likely involve removing the data capture bundle from `tests.html` and adding the babel polyfill to the tests bundle.
- [x] Thoroughly test on IE11!


frontend/source/js/data-capture/custom-event.js
```diff
@@ -0,0 +1,32 @@
+/* global document, window */
+
+// The following code is based off:
+// https://github.com/d4tocchini/customevent-polyfill
```
toolness: The original source didn't use anything like `Object.assign` to assign defaults, so I just copy-pasted it here to make it do that for us...

frontend/source/js/tests/ajaxform_tests.js
```diff
@@ -34,75 +37,95 @@ function createBlob(content) {
...
-  assert.ok(!s.$submit.prop('disabled'), 'submit button is enabled');
-});
+  div.one('uploadwidgetready', e => {
+    s.uploadwidget = $('input', e.target).data('upload');
+    update();
+  });
+  div.one('ajaxformready', e => {
+    s.ajaxform = $(e.target).data('ajaxform');
+    update();
+  });
```
toolness: Blergh, waiting for multiple web components to upgrade themselves is a hassle.  This could probably be done better with promises or something.

frontend/templatetags/qunit_fixture_data.py
```diff
@@ -22,9 +22,9 @@ class AjaxformTestsForm(forms.Form):
     def render(self):
         return ''.join([
             '<form enctype="multipart/form-data" method="post"',
-            ' data-ajaxform',
+            ' is="ajax-form"',
             ' action="/post-stuff">',
-            '  %s' % str(self.as_p()),
+            '  %s' % str(self.as_ul()),
```
toolness: It turned out `as_p()` was actually resulting in markup that browsers didn't like--I think because they don't like `<div>`s in their `<p>`s, so they were actually creating a really weird DOM out of our HTML that resulted in malformed upload widgets.  We might want to document this somewhere, actually. :disappointed:

data_capture/templates/data_capture/step_1_form.html
```diff
@@ -1,5 +1,5 @@
 <form enctype="multipart/form-data" method="post"
-      data-ajaxform
+      is="ajax-form"
```
toolness: I would call this `ajaxform` to keep the spelling consistent, but as I understand it, web components need to have a dash somewhere in their name, so `ajax-form` it is.

frontend/source/js/tests/ajaxform_tests.js
```diff
@@ -34,75 +37,95 @@ function createBlob(content) {
...
     isDegraded: false,
     fooValue: 'bar',
   }, extraOptions || {});
-  const $form = $('<div></div>')
+
+  const iframe = document.createElement('iframe');
+
+  document.body.appendChild(iframe);
+
+  const $form = $(iframe.contentDocument.body)
```
toolness: So this is kind of funny.  Basically, I was in a conundrum because I wanted to be able to modify the HTML of our form via a DOM API without actually upgrading the form (which would potentially mutate the DOM, based on what the `createdCallback` did; this is actually the case with our upload widget, for example, which adds the `or drag and drop here` text).  So I figured that making an empty iframe, where the web component wasn't registered, might do the trick--and it did! :sweat_smile: 

I also tried `document.createDocumentFragment()` and adding the web component HTML to that instead, but the web component was still upgraded there, alas.

frontend/source/js/data-capture/upload.js
```diff
@@ -1,42 +1,101 @@
 /* global jQuery, window, document */
+/* eslint no-underscore-dangle: ["error", { "allowAfterThis": true }] */
```
toolness: Erm, it feels weird to not at least allow dangling underscores after `this`, e.g. `this._myPrivateProperty`.  I'm not sure how else we're supposed to do private instance variables unless we force ourselves to not use `class` or prototype-based inheritance, which feels like an arbitrary restriction... Can we move this to the `.eslintrc`?

data_capture/templates/data_capture/step_1_form.html
jseppi: could also prefix all of our components with `calc-`

frontend/source/js/data-capture/upload.js
jseppi: Fine with me

jseppi:
During the Cascadia talk, someone asked about how to message between web components while keeping them as decoupled as possible. The presenter said to use custom events. So I guess one thing we could do with the submit button is have it be a component itself that listens for some kind of `'form:isvalid'` event. Maybe.

data_capture/templates/data_capture/step_1_form.html
toolness: oh good idea, that would also make it easier for folks unfamiliar w/ web components to understand that it's probably a custom thing we made, rather than a weird html tag they've never seen before.

toolness:
Ok, I was attempting to re-add the grayed-out-submit-button functionality but realized there are actually multiple things to consider with it.  I've spun off #584 and #585, which both merge against this PR, to illustrate two possible solutions.

toolness:
Ack!!!  I totally didn't realize that I should've been using `attachedCallback` instead of `createdCallback`.  From the [spec](https://w3c.github.io/webcomponents/spec/custom/#custom-elements-upgrades-examples):

> Note that upgrades only apply to elements in the document tree. (Formally, elements that are connected.) An element that is not inserted into a document will stay un-upgraded.

Having everything in `createdCallback` actually makes it so that just calling `document.createElement('upload-widget')` throws an error because it's looking across its children to make sure its DOM is well formed and stuff.  It's also looking up at its ancestors to see if any of them have the `data-force-degradation` attribute.  So almost all our work should be done in the `attachedCallback` rather than the `createdCallback`.

frontend/source/js/tests/ajaxform_tests.js
toolness: Actually, scratch all of that, this is entirely due to my inability to distinguish between `createdCallback` and `attachedCallback`.  See https://github.com/18F/calc/pull/582#issuecomment-241567022 for more.

toolness:
Ok this is ready for review @jseppi !

frontend/source/js/data-capture/custom-event.js
jseppi: Hrm, that repo doesn't have a license, so we might need to find another one to use since I don't think we can commit it to our repo, which does a very open license.

frontend/source/js/data-capture/feature-detection.js
```diff
@@ -0,0 +1,29 @@
+/* global $, window, document */
+
+// The following feature detectors are ultimately pulled from Modernizr.
```
jseppi: Should link to the Modernizr license and specify that the code here is covered by that license: https://github.com/Modernizr/Modernizr/blob/master/LICENSE

frontend/source/js/data-capture/upload.js
```diff
@@ -1,42 +1,128 @@
...
+class UploadInput extends window.HTMLInputElement {
+  attachedCallback() {
+    this.isUpgraded = false;
+    this._upgradedValue = null;
+    if (this.getAttribute('type') !== 'file') {
+      throw new Error('<input is="upload-input"> must have type "file".');
+    }
+    if (this.hasAttribute('multiple')) {
+      throw new Error('<input is="upload-input"> does not currently ' +
+                      'support the "multiple" attribute.');
```
jseppi: haha pep8 continuation identation in JS!

frontend/source/js/tests/ajaxform_tests.js
```diff
@@ -34,75 +37,114 @@ function createBlob(content) {
...
-    .html(makeFormHtml(extraOptions))
-    .appendTo('body')
-    .hide();
-  return ajaxform.bindForm();
+function addForm(extraOptions, cb) {
+  const div = $('<div></div>').appendTo('body').hide();
+
+  $parentDiv = div;
+
+  Promise.all([
```
jseppi: Might want to make sure tests have the Babel Polyfill (`require('babel-polyfill/dist/polyfill.js');`) No version of IE supports `Promise` so this will cause problems if we want to run the tests in IE.

frontend/source/js/data-capture/ajaxform.js
```diff
@@ -122,18 +94,28 @@ function bindForm() {
...
   return delegate;
 };
 exports.MISC_ERROR = MISC_ERROR;
-exports.getForm = getForm;
 exports.bindForm = bindForm;
+exports.populateFormData = populateFormData;
 
 window.testingExports__ajaxform = exports;
+
+class AjaxForm extends window.HTMLFormElement {
```
jseppi: Could you make `bindForm` and `populateFormData` methods of the `AjaxForm` class instead of having them as external functions?

frontend/source/js/data-capture/ajaxform.js
```diff
@@ -122,18 +94,28 @@ function bindForm() {
...
 exports.setDelegate = newDelegate => {
   delegate = newDelegate;
   return delegate;
 };
 exports.MISC_ERROR = MISC_ERROR;
-exports.getForm = getForm;
 exports.bindForm = bindForm;
+exports.populateFormData = populateFormData;
 
 window.testingExports__ajaxform = exports;
```
jseppi: I think you brought this up elsewhere, perhaps in another issue or PR, but could you just `import` or `require` these modules in the test scripts instead of using this `testingExports` pattern?

jseppi:
Ok, this looks pretty good. My biggest concerns are about licenses of borrowed code. I think we need to pretty careful with that stuff since in our license we are saying all the code in this repo is public domain.

I think organizationally I would like to have all the functions related to each component encapsulated within that component's class, but perhaps you've already investigated doing that and determined it wasn't going to work.

frontend/source/js/data-capture/custom-event.js
toolness: Ummm well the code seems really public-domainy... 😅 

I think I actually found it from this repo:

https://github.com/krambuhl/custom-event-polyfill/blob/master/custom-event-polyfill.js

That page mentions the code was also taken from MDN, but the polyfill on MDN was since removed and I'm not sure what it looked like (searching through wiki history on MDN is painful)...

I'd say I could rewrite the polyfill from scratch, but it'd look just like what I committed.  Saying the polyfill is copyrighted seems *almost* like saying an API is copyrighted, and I thought some legal case said you can't copyright APIs?  I have no idea.

What if we just removed the comment? 😁

frontend/source/js/tests/ajaxform_tests.js
toolness: Oh, I thought that polyfill was automatically included in our bundles or something?

toolness:
Cool!  Actually the main reason I didn't put the methods in classes was because I wanted to keep the diff as small as possible, but that's clearly a lost cause and not very helpful anyways.  Will do the refactoring!

frontend/source/js/tests/ajaxform_tests.js
jseppi: It is in the data_capture bundle, but not in the test bundle. Though the test page does include the data_capture bundle, so, yeah, we're safe on this one actually.

frontend/source/js/data-capture/custom-event.js
jseppi: Pretty sure code is by default under copyright of the author, and it is up to the author to assign a license if they so choose.

Here's an MIT licensed one: https://github.com/krambuhl/custom-event-polyfill Will it work? If so, we could just `npm install` this one also.

frontend/source/js/data-capture/ajaxform.js
toolness: Oh sure. I was tempted to do that but was worried this PR was already getting too big, but now I will bust that up.

frontend/source/js/tests/ajaxform_tests.js
toolness: Ah ok--but actually, once I start importing/requiring the modules directly in the data capture tests, I'm going to remove the data capture bundle from the test page, so this will likely become an issue before this PR is merged. Thanks for the heads-up!

frontend/source/js/data-capture/custom-event.js
toolness: Hmm, so even though the MIT licensed one *says* it's MIT-licensed, its [code references the exact same unlicensed repo](https://github.com/krambuhl/custom-event-polyfill/blob/master/custom-event-polyfill.js#L4) as ours does. :persevere:

toolness: Oh!  I found out that they just moved the polyfill on MDN to the [page for the `CustomEvent()` constructor](https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill) (rather than the page for the `CustomEvent` class).  It's almost identical to the polyfill we're already using, so I made some tweaks and credited MDN.

toolness:
Ok I think this is good to go now!

frontend/source/js/tests/index.js
```diff
@@ -1,2 +1,4 @@
+require('../data-capture');
+
```
jseppi: might want to include the babel polyfill also?

`require('babel-polyfill/dist/polyfill.js');`

frontend/source/js/data-capture/custom-event.js
```diff
@@ -0,0 +1,33 @@
+/* global document, window */
+
+// The following polyfill is based off the code at:
+//
+// https://developer.mozilla.org/en-US/docs/Web/API/CustomEvent/CustomEvent#Polyfill
+//
+// The polyfill is by Mozilla Contributors and licensed under CC-BY-SA 2.5:
+//
+// https://creativecommons.org/licenses/by-sa/2.5/
```
jseppi: FYI, if you use JSDoc style `@license` comments, uglify will keep them include in the minified scripts. I don't think it is necessary but it seems to be a nice practice.

ref: http://usejsdoc.org/tags-license.html

jseppi:
👍 I really like the new class and exports refactoring!!!

Looks great! Just had a couple minor comments.

frontend/source/js/tests/index.js
toolness: Nope the data capture bundle already includes them!  I figured this was actually better practice anyways--that way we're sort of testing `data-capture`'s inclusion of the babel polyfill too!

jseppi: Ha, whoops, didn't realize that bundle was included :)
Sweet!

toolness:
Woot thanks! Just added the `@license` directive in 66fea6ab54aecc7f92debbdc9dd5c2036dfd1321 and will merge once Travis is happy.

PR #583, opened 2016-08-22
Rejigger form field order

hbillings:
^this is so true

This PR regroups the form fields as follows:
1. Contract #, schedule, vendor
2. Small biz, worksite, contract start, contract end, contract year
3. Upload form
4. Verification
5. Submit

I haven't made the visual tweaks we talked about for the forms, mostly because I just want to get this sucker merged. I'll open other tix for those if they don't already exist.

I also haven't broken apart the verification steps yet (error state on one step, verified info on the next).

I feel like there ought to be a way to change the schedule on the upload page still. (At the very least we should display the schedule you selected previously.)


data_capture/forms.py
```diff
@@ -1,38 +1,23 @@
 from django import forms
 
-from .models import SubmittedPriceList
+from .models import NewPriceList, SubmittedPriceList
```
jseppi: Instead of creating a new database model and using the `form.ModelForm` base class for the forms, I wonder if it might be better to just use plain ole fields and the `form.Form` base class. 

I think that using a database model means that modifications will be slightly harder in the future, since we'd also have to migrate the models and associated data. Speaking of which, if we do end up keeping the `NewPriceList` model, I think a migration needs to be included in this PR.

@toolness probably has better insight into this than I do though.

data_capture/views.py
```diff
@@ -19,97 +20,122 @@ def add_generic_form_error(request, form):
...
     if request.method == 'GET':
         form = forms.Step1Form()
     elif request.method == 'POST':
-        form = forms.Step1Form(request.POST, request.FILES)
+        form = forms.Step1Form(request.POST)
+        if form.is_valid():
+            # TODO: Check for session and delete it if found
+            request.session['data_capture:contract_number'] = request.POST['contract_number']
+            request.session['data_capture:vendor_name'] = request.POST['vendor_name']
+            request.session['data_capture:schedule'] = request.POST['schedule']
```
jseppi: Maybe to keep `session` tidier, all of these values from the `POST` could be put into a single `dict`, and then that could be saved into `session` instead (might have to JSON serialize the `dict` if it doesn't do it automatically).

data_capture/views.py
```diff
@@ -19,97 +20,122 @@ def add_generic_form_error(request, form):
...
+def step_5(request, gleaned_data):
+    if gleaned_data.valid_rows:
+        current_price_list = SubmittedPriceList.objects.get(id=request.session['price_list_id'])
+        current_price_list.serialized_gleaned_data = json.dumps(
+            request.session.get('data_capture:gleaned_data')
+        )
+        current_price_list.save()
+        gleaned_data.add_to_price_list(current_price_list)
+
+        del request.session['data_capture:gleaned_data']
```
jseppi: Probably also want to `del` the other saved session variables somewhere around this point.

jseppi:
Looks like there are a few `pep8` issues as reported by CodeClimate. You can check for these locally by running `python manage.py ultratest --lintonly` (or `docker-compose run app python manage.py ultratest --lintonly` if you are using the docker setup).

I made some inline comments. My biggest concern is around creating a new model. I tend to be conservative about model creation/modification, probably because of some weird fear of the unknown with regards to the Django ORM and the DB. Would definitely like to hear (see) @toolness's thoughts on that.

Also, this will need the most fun part to write: tests! Though there were some organizational changes with tests made in the R10 PR, so it might make sense to merge `develop` into this before you tackle that, because nobody likes dealing with merge conflicts.

data_capture/forms.py
hbillings: @jseppi Oh, d'oh! Great idea. I'll make that modification.

hbillings:
I can totally rewrite the code without the model -- that's just an artifact of me copying things that worked and modifying them without knowing best practice!

I'll merge and then do the tests. I'll probably have some questions about tests, 'cause that's new ground for me.

data_capture/views/price_list_upload.py
```diff
@@ -1,63 +1,125 @@
...
+        'form': form
+    })
+
+
+@login_required
+def step_3(request):
+    if request.method == 'GET':
+        form = forms.Step3Form()
+    elif request.method == 'POST':
+        posted_data = dict(request.POST, schedule=request.session['data_capture']['schedule'])
```
hbillings: Code Climate is complaining about this being too long for PEP8 compatibility. I'm not sure how to break it into two lines -- I'd love a hint!

jseppi: I hate the line length thing in pep8, especially given the preferred 4-space-wide tabs and wordy syntax.

I think this should do it
```py
            posted_data = dict(
                request.POST,
                schedule=request.session['data_capture']['schedule'])
```

data_capture/views/price_list_upload.py
```diff
@@ -1,63 +1,127 @@
...
         request,
         context={
-            'step_number': 1,
+            'step_number': 3,
             'form': form
         },
-        template_name='data_capture/price_list/step_1.html',
-        ajax_template_name='data_capture/price_list/step_1_form.html',
+        template_name='data_capture/price_list/step_3.html',
+        ajax_template_name='data_capture/upload_form.html',
```
jseppi: I think this path be `data_capture/price_list/upload_form.html` since the folders have been moved around a bit.

data_capture/views/price_list_upload.py
```diff
@@ -1,104 +1,169 @@
...
+                print(form.errors)
 
-    return render(request, 'data_capture/price_list/step_2.html', {
-        'step_number': 2,
+    return render(request, 'data_capture/price_list/step_4.html', {
+        'step_number': 4,
         'gleaned_data': gleaned_data,
+        'price_list': request.session['data_capture'],
         'is_preferred_schedule': isinstance(gleaned_data, preferred_schedule),
         'preferred_schedule': preferred_schedule,
```
jseppi: The `preferred_schedule` seems to be blank when the `step_4.html` template is rendered.

screen shot 2016-09-05 at 1 27 56 pm

data_capture/models.py
```diff
@@ -5,16 +5,14 @@
 
 
 class SubmittedPriceList(models.Model):
-    # This is the equivalent of Contract.idv_piid.
     contract_number = models.CharField(
         max_length=128,
-        help_text='This should be the full contract number, e.g. GS-XXX-XXXX.'
```
jseppi: I may have missed this elsewhere, but was the `help_text` for this not useful?

jseppi:
I checked this out locally and ran through it, and it all seemed to work as intended! Though I did find that the `preferred_schedule` was not being printed in `step_4`, as I mentioned in an inline-comment.

Some other To Dos:

- [ ] Fix `pep8` issues so that codeclimate posts a ✔️ 
- [ ] Merge in `develop` and fix conflicts
- [ ] Either add in use of the `@handle_cancel` decorator, or file an issue for it to be done

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -1,14 +1,131 @@
...
+
+  {% with total=rows|length %}
+  <h3>{{ total }} row{{ total|pluralize:" has,s have" }} errors</h3>
+
+  <p>
+    The row{{ total|pluralize }} below appear{{ total|pluralize:"s," }} to be
+    invalid and <strong>will be discarded</strong> when you upload your
+    price list.
+    If you'd like, you may correct {{ total|pluralize:"this row,these rows" }}
+    in your original spreadsheet and <a href="{% url 'data_capture:step_1' %}">try uploading it again</a>.
```
jseppi: The `href` here needs to be updated to the right view name.

data_capture/views/price_list_upload.py
toolness: Oof, this regression has actually existed for a while, so I don't think it needs to be the purview of this PR to fix it (unless it's a super easy fix and @hbillings really wants to).  Filed it separately as #663!

toolness:
Regarding the `@handle_cancel` decorator, I recommend filing a separate issue and dealing with it later--this PR is already pretty hefty! 😁

data_capture/views/price_list_upload.py
```diff
@@ -10,98 +12,165 @@
...
+        return redirect('data_capture:step_1')
 
-        if d is None:
-            return redirect('data_capture:step_1')
+    if request.method == 'GET':
+        form = forms.Step2Form()
+    elif request.method == 'POST':
+        form = forms.Step2Form(request.POST)
+        if form.is_valid():
+            request.session['data_capture:price_list']['is_small_business'] = \
```
jseppi: small thing: might be a little easier to read this section if the main session dict is saved in a local variable and then used for setting all the sub-keys.
```py
price_list_sess = request.session['data_capture:price_list']
price_list_sess['is_small_business'] = form.cleaned_data['is_small_business']
# ... all the others ...
```

Actually, I think you could also/instead do something like:

```py
request.session['data_capture:price_list'] = {
  'is_small_business': form.cleaned_data['is_small_business'],
  # ... all the others ...
  'contract_end': form.cleaned_data['contract_end'].strftime("%Y-%m-%d")
}

File: data_capture/views/price_list_upload.py [hunk 10]
```diff
@@ -10,98 +12,165 @@
...
+            request.session['data_capture:price_list']['contract_year'] = \
+                form.cleaned_data['contract_year']
+            request.session['data_capture:price_list']['contract_start'] = \
+                form.cleaned_data['contract_start'].strftime("%Y-%m-%d")
+            request.session['data_capture:price_list']['contract_end'] = \
+                form.cleaned_data['contract_end'].strftime("%Y-%m-%d")
+
+            # Changing the value of a subkey doesn't cause the session to save,
+            # so do it manually
+            request.session.modified = True
```
jseppi: oof, I didn't know about that -- must have been a pain to figure out.

data_capture/models.py
```diff
@@ -11,13 +11,12 @@ class SubmittedPriceList(models.Model):
     # This is the equivalent of Contract.idv_piid.
     contract_number = models.CharField(
         max_length=128,
-        help_text='This should be the full contract number, e.g. GS-XXX-XXXX.'
```
toolness: Is there a particular reason we moved this into `forms.py`?  One advantage of keeping it in the model is that it ends up showing up in the django admin too... But one downside is that the migration system keeps flagging it as a migration whenever it changes, which is annoying.

data_capture/tests/test_price_list_views.py
```diff
@@ -57,151 +182,122 @@ def test_valid_post_sets_session_data(self):
...
         'education': 'Bachelors',
         'price': '15.00',
         'service': 'Project Manager',
         'sin': '132-40',
         'years_experience': '7'
+    }, {
+        'education': 'School of hard knocks',
+        'price': '0',
+        'service': 'Oil change & tune up',
+        'sin': 'absolved',
```
toolness: lolol

data_capture/tests/test_price_list_views.py
```diff
@@ -57,151 +182,122 @@ def test_valid_post_sets_session_data(self):
...
-        self.login()
-        res, json_data = self.ajax_post({'schedule': FAKE_SCHEDULE})
-        assert '<!DOCTYPE html>' not in json_data['form_html']
-        self.assertRegexpMatches(json_data['form_html'],
-                                 r'This field is required')
-        self.assertHasMessage(
-            res,
-            'error',
-            'Oops, please correct the error below and try again.'
-        )
```
toolness: Were `test_invalid_post_returns_html` and `test_invalid_post_via_xhr_returns_json` relocated anywhere?  Just want to make sure the logic is still being tested...

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -1,14 +1,89 @@
...
+  <form method="post">
+  {% csrf_token %}
+    <input type="hidden" name="contract_number" value="{{ price_list.contract_number }}">
+    <input type="hidden" name="vendor_name" value="{{ price_list.vendor_name }}">
+    <input type="hidden" name="schedule" value="{{ price_list.schedule }}">
+    <input type="hidden" name="is_small_business" value="{{ price_list.is_small_business }}">
+    <input type="hidden" name="contractor_site" value="{{ price_list.contractor_site }}">
+    <input type="hidden" name="contract_year" value="{{ price_list.contract_year }}">
+    <input type="hidden" name="contract_start" value="{{ price_list.contract_start }}">
+    <input type="hidden" name="contract_end" value="{{ price_list.contract_end }}">
```
toolness: Hmm, I'm not a big fan of the repetition here... Is `price_list` a `Form` instance?  If so, I'd have thought that e.g. `{{ price_list.contract_end }}` would actually render the whole `<input>` tag...

toolness: Hmm, I assume the reason you're using hidden fields, rather than just grabbing the data out of the session, is so that you can use a `ModelForm` to actually build the price list instance?

data_capture/models.py
hbillings: I'mma guess it's copypasta error from me trying to figure out how to create new forms. Moving back!

data_capture/tests/test_price_list_views.py
hbillings: I appear to have lost those. 😬  Will update!

data_capture/templates/data_capture/price_list/step_4.html
toolness: I guess I have some concerns about the serialization/deserialization of this data working properly between the `GET` and the `POST` of step 4, but we can probably just address it in a future PR!

toolness:
I noticed that the help text for the contract number ("This should be the full contract number, e.g. GS-XXX-XXXX.") is no longer showing:

> screen shot 2016-09-08 at 1 03 03 pm

If that's intentional, no worries, but otherwise I guess we should address it?  Could be filed as a separate issue too.

toolness:
So I successfully submitted a price list, which is awesome!

But then I used my browser's back button to try to go back to step 2 and it raised an error.

<details>
  <summary>Full traceback for `/data-capture/step/2`</summary>
```
Environment:


Request Method: GET
Request URL: http://localhost:8000/data-capture/step/2

Django Version: 1.8.14
Python Version: 3.4.5
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.humanize',
 'whitenoise.runserver_nostatic',
 'django.contrib.staticfiles',
 'debug_toolbar',
 'django_rq',
 'data_explorer',
 'contracts',
 'data_capture',
 'api',
 'djorm_pgfulltext',
 'rest_framework',
 'corsheaders',
 'uaa_client',
 'styleguide',
 'meta',
 'frontend',
 'fake_uaa_provider')
Installed Middleware:
('hourglass.middleware.ComplianceMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'hourglass.middleware.DebugOnlyDebugToolbarMiddleware')


Traceback:
File "/usr/local/lib/python3.4/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/calc/data_capture/decorators.py" in wrapper
  46.                 return function(request, *args, **kwargs)
File "/usr/local/lib/python3.4/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  22.                 return view_func(request, *args, **kwargs)
File "/calc/data_capture/views/price_list_upload.py" in step_2
  58.     if 'contract_number' not in request.session['data_capture:price_list']:
File "/usr/local/lib/python3.4/site-packages/django/contrib/sessions/backends/base.py" in __getitem__
  48.         return self._session[key]

Exception Type: KeyError at /data-capture/step/2
Exception Value: 'data_capture:price_list'
```
</details>

data_capture/templates/data_capture/price_list/step_4.html
toolness: I also tinkered with different values in step 2 and everything seems to be ultimately passed to the model fine, so 👍 !

data_capture/views/price_list_upload.py
hbillings: So apparently `update` will allow a structure like your second example, but it will add keys instead of overwriting the entire dict! Just pushed a change to do that instead of this.

data_capture/templates/data_capture/price_list/step_4.html
hbillings: No, `price_list

hbillings: `price_list` here is just shorthand for `request.session['data_capture:price_list']`. I'm not sure how to safely make that into a form instance before it's posted, but totally willing to be pointed in the right direction!

hbillings:
Re: help text, I threw `{{ form.contract_number.help_text }}` in the view but that doesn't actually seem to render the text unless `help_text` is specified in both `models.py` and `price_list.py`. Not sure if that's expected behavior or not?

I'll look into the other error!

data_capture/templates/data_capture/price_list/step_4.html
toolness: Ohh got it!  No worries, I think we should just leave it as-is and possibly address it in a future PR.

toolness:
Oh, regarding the `help_text`, it looks like that's because `Step1Form` isn't a `ModelForm`--it's just a regular `Form` that has a field called `contract_number` in it, so it has no way of connecting that to a model field and introspecting it to find the help text. Blech.

There's some weird ways to work around that but don't worry about it for now, we can fix in a future PR!

data_capture/templates/data_capture/price_list/upload_form.html
```diff
@@ -15,6 +15,9 @@
 {{ form.file }}
 
 <div class="form-button-row clearfix">
+<a href="{% url 'data_capture:step_2' %}" class="button button-previous">Previous</a>
+
+  <a href="{% url 'index' %}" class="button button-cancel">Cancel</a>
```
jseppi: As long as the `@handle_cancel` decorator is on the associated view, you should be able to change this `<a>` to `<button type="submit" name="cancel" class="button-secondary">Cancel</button>`

The decorator _should_ work with our ajaxform javascript (the tests say it does).

data_capture/views/price_list_upload.py
```diff
@@ -10,98 +11,164 @@
...
+        form = forms.Step2Form(request.POST)
+        if form.is_valid():
+            session_data = request.session['data_capture:price_list']
+            session_data.update({
+                'is_small_business': form.cleaned_data['is_small_business'],
+                'contractor_site': form.cleaned_data['contractor_site'],
+                'contract_year': form.cleaned_data['contract_year'],
+                'contract_start': ymd(form.cleaned_data['contract_start']),
+                'contract_end': ymd(form.cleaned_data['contract_end'])
+            })
```
toolness: Huh, is there a reason this can't just be `session_data.update(form.cleaned_data)`?  Or are you intentionally being super explicit here?

hbillings: TBH, I was copying the pattern that was there originally, heh. See image at the top of this PR X-D

toolness:
WOOOOOO 🎉  🎉 🎉  YAY!!!

PR #584, opened 2016-08-22
Add custom HTML5 form validation messaging for upload widget

toolness:
This adds HTML5 form validation support to the upload widget so that users with supported browsers will see the following if they click "submit" without uploading a required file:

> screen shot 2016-08-22 at 5 03 20 pm

I think this is a more humane solution than graying out the submit button (#585) because it tells the user *why* they can't submit the form.  This could be particularly useful for larger forms, where there could be many reasons why they can't submit the form.

Aside from that, it's also technically much simpler than #585.

Of course, server-side form validation still works too, so users on older browsers will just get a slightly degraded experience.

To do:

- [x] Add tests.


toolness:
Ok we're going with this approach!

PR #585, opened 2016-08-22
Gray out submit button if ajax form is invalid

toolness:
This is basically a broader version of what our ajax form currently does--it disables the submit button on the form (i.e., grays it out and makes it unclickable) if *any* field is invalid, by consulting the HTML5 form validation API:

> screen shot 2016-08-22 at 5 53 43 pm

There's some reasons I don't like this approach:

* It's not actually very user-friendly, because we're not giving the user any feedback on *why* the submit button is disabled. This could get particularly bad for very large forms. In fact, because we're making the submit button unclickable, it renders a lot of valuable user feedback mediated via the HTML5 form validation API unreachable.

* I am not a fan of the technical implementation. Right now it actually works by polling the form every 250ms; the alternative would be to re-check the form's validity every time some kind of event occurs that might result in a valid form, but this list of events would be hard to maintain, and forgetting one could result in a lot of user frustration. Blerg.

For an alternate solution that I find both more humane and technically simpler, see #584.

To do:

- [ ] Add tests.
- [ ] Make it work w/ multiple kinds of submit buttons (e.g. `<input>` and `<button>`).

### DIFF
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -8,6 +8,8 @@ import { dispatchBubbly } from './custom-event';
 
 const $ = jQuery;
 
+const VALIDITY_CHECK_INTERVAL = 250;
+
 const MISC_ERROR = 'Sorry, we’re having trouble. ' +
                    'Please try again later or refresh your browser.';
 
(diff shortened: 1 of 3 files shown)


toolness:
We're not going this way--closing this PR.

PR #598, opened 2016-08-23
Fix staging deploy

jseppi:
Right now our staging deployment is down. These are some relatively minor fixes to get it back up and running.

The first thing I had to fix was the `email_config` value retrieval, which was causing a `KeyError` to get raised.

After the `email_config` was fixed, I was getting migration issues while trying a manual deploy. I ran `makemigrations`, which generated `data_capture/migrations/0002_auto_20160823_2147.py` in this PR. I am not clear on why this migration is necessary since it seems to me like the `data_capture/migrations/0001_initial.py` migration already is creating the fields as specified in the `0002` migration.


jseppi:
@toolness if you think the migration is correct, then go ahead and merge this when ready. It should hopefully fix the auto-deployment.

hourglass/settings.py
```diff
@@ -37,17 +37,10 @@
...
-EMAIL_HOST_USER = email_config['EMAIL_HOST_USER']
-EMAIL_HOST_PASSWORD = email_config['EMAIL_HOST_PASSWORD']
-EMAIL_HOST = email_config['EMAIL_HOST']
-EMAIL_PORT = email_config['EMAIL_PORT']
-EMAIL_BACKEND = email_config['EMAIL_BACKEND']
-EMAIL_USE_TLS = email_config['EMAIL_USE_TLS']
-EMAIL_USE_SSL = email_config['EMAIL_USE_SSL']
+# Sets a number of settings values, as described at
+# https://github.com/migonzalvar/dj-email-url
+vars().update(email_config)
```
toolness: Um, so I just realized that the reason that this was throwing a `KeyError` was probably because `EMAIL_URL` isn't defined on staging/production (it's only defaulted to `console:` when in debug mode), so it was basically the app's very unfriendly way of telling us to define `EMAIL_URL` :)  I think it's fine to make email non-required for now, which is essentially what your fix does, but once we add real transactional emails to the app, we should probably have `settings.py` throw a helpful error if `EMAIL_URL` isn't defined!

toolness:
Oh regarding the migrations, I think it's probably due to #561, which removes the `help_text` from `contract_start` and `contract_end`.  It's pretty dumb because `help_text` clearly doesn't affect the database itself, yet apparently Django saw fit to [include them in the initial migration](https://github.com/18F/calc/blob/develop/data_capture/migrations/0001_initial.py#L27-L28).

It looks like has been [asked on Stack Overflow](http://stackoverflow.com/questions/26503826/why-does-django-generate-migrations-for-help-text-and-verbose-name-changes) and the general TL;DR is that because custom third-party fields might use *any* keyword arguments to shape the database schema, Django's migrations framework by default considers *all* of them relevant to migrations, just to be super safe I guess.

I guess this means that the (irrelevant) changes made by this migration will eventually make their way into a migration, but I doubt that these changes will somehow fix our broken staging build... That said, they should also be fairly harmless, so feel free to merge them in!

toolness:
All in all these changes look fine, feel free to merge!

PR #599, opened 2016-08-24
Fix staging deploy: load smaller sample datasets on deploy

jseppi:
So after #598 the deploy was still failing, and I tracked it down to the data loads now taking longer than the cloud.gov timeout permits (>180 seconds). This PR removes the `load_data` and `load_s70` commands from the deploy script (`cf.sh`) so that staging deploys will succeed again. However, it means that we'll have to manually load data into the staging instance, either through `cf-ssh` or by running "one-off" tasks via a `manifest.yml` as described at https://docs.cloud.gov/getting-started/one-off-tasks/


jseppi:
@toolness any thoughts? :\

jseppi:
:\ is because dealing with the R10 stuff has opened a can of worms/weaknesses.

toolness:
Oh blerg that sucks!

Is there a reason that all this loading is taking longer than it was before?  Is the refactored bulk loading code from #555 somehow taking longer to run than the old code?

I'm also wondering if in the long run we want to just rename one of our in-repo CSV files to `sample-data.csv` and use it only for development and so that our staging deployment has some data to play with?

jseppi:
I am not quite sure what the final straw broke the timeout's back was. Everything from `pip install` to `load_s70` needs to happen within that timeout window, so it could be a number of things.

I didn't think I was adding any additional overhead to the existing `load_data` command, but maybe I can do a local test versus an old commit's version to see if there is a noticeable time difference.

toolness:
Ah ok no worries either way.  I don't suppose there's a way to increase the timeout to 5 minutes or anything like that?

jseppi:
It is set to the max time of 3 minutes already unfortunately.

toolness:
Oh ok.

I guess I'm just wondering if we want to have like, 10 rows of sample data be auto-installed into the staging deployment, just so it's easy to test out the data explorer functionality, rather than having to deal with the whole `cf-ssh` thing or whatever?  Since it seems like we're going to be migrating away from loading stuff into prod via in-repo CSVs anyways...

jseppi:
Hrm...actually I may have misread what is happening from the cf logs. It might not be a timeout issue. Arg...

jseppi:
No, I still think it is a timeout problem.

I've implemented your suggestion and created `XYZ_sample.csv` files (I just did `head -n 101` on each of the current CSVs), and I load those during deploy instead of the large data files. This does seem to fix the deploy for now, and gives us some data so we can test the data-explorer interface.

cf.sh
```diff
@@ -3,8 +3,8 @@ echo "------ Starting APP ------"
 if [ $CF_INSTANCE_INDEX = "0" ]; then
```
toolness: Is this the line that basically determines whether we're on staging vs. prod?  I just want to make sure the sample data isn't loaded when deploying to prod, is all...  Otherwise this PR looks great!

jseppi: I don't think that is what that line does. I'm not actually sure what it is for to be honest.

The prod deployment has not really been worked on at all yet in `develop`. We don't have a manifest for it or anything.

toolness: Oh, ok.  I guess since it's already on our radar to change the way we deploy to prod, there's no need to address it here then... feel free to merge!

PR #602, opened 2016-08-24
Make it possible to sort and "mute" individual price list rows

toolness:
This adds a new "submitted price list row view" admin view that allows the data administrator to view all *unapproved* price list rows and "mute" them if they're out of the ordinary:

> screen shot 2016-08-24 at 5 42 24 pm

Unfortunately, actually *approving* the price lists is a different step and can be accomplished from the "submitted price list view":

> screen shot 2016-08-24 at 5 42 57 pm

I'm not a huge fan of this but it's what seems to be easily doable via django admin customization, alas.

Other changes:

* The admin detail view for submitted price lists still exists, but the "is approved" field is no longer an editable checkbox--the *only* way that a price list can be approved is via the "Approve selected price lists" bulk action from the screenshot above.  The reasons for this are annoyingly technical, but hopefully it's not that big a deal since we expect data administrators to approve price lists in bulk anyways.

* When visiting an admin detail view for *approved* price lists, all fields are read-only. This has been done to ensure that the data administrator doesn't think that they're editing live CALC data.

To do:

- [x] Add tests for the `is_muted` functionality, at least. Would be nice to add tests for the django admin functionality too but I'm not sure how easy that will be.
- [x] Add a DB migration.
- [X] Ask Kelly if he thinks this is total crap. (He thinks it's fine.)
- [x] Remove the "Delete selected rows" action from price list admin view, or at least move it to the end, to make it harder to accidentally choose.
- [x] Add a date created field to price lists and show it in the change list view so that admins can make sure they're not approving lists that have just been added in the past few seconds or something.
- [ ] Include a [short comment](https://github.com/18F/calc/pull/602/files/62e7393656f8a379052cb7a2afa0c1a028763b2b#r76493796) on why we are removing the "delete selected entries" bulk action.

### DIFF
--- a/data_capture/admin.py
+++ b/data_capture/admin.py
@@ -24,10 +24,7 @@ class SubmittedPriceListRowInline(admin.TabularInline):
         'is_muted',
     )
 
-    # Because of the fact that inline formsets are saved
-    # *after* their parent model, it's safest to not allow
-    # any of our fields to be editable.
-    readonly_fields = fields
+    readonly_fields = ()
 
     formfield_overrides = {
         models.TextField: {'widget': forms.TextInput}
@@ -36,11 +33,37 @@ class SubmittedPriceListRowInline(admin.TabularInline):
     def has_add_permission(self, request):
         return False
 
+    def get_readonly_fields(self, request, obj=None):
+        if obj and obj.is_approved:
+            return self.fields
+        return self.readonly_fields
+
+
+def approve(modeladmin, request, queryset):
+    for obj in queryset.filter(is_approved=False):
+        obj.approve()
+
+
+approve.short_description = (
+    'Approve selected price lists (add their data to CALC)'
+)
+
+def unapprove(modeladmin, request, queryset):
+    for obj in queryset.filter(is_approved=True):
+        obj.unapprove()
+
+
+unapprove.short_description = (
+    'Unapprove selected price lists (remove their data from CALC)'
+)
+
 
 @admin.register(SubmittedPriceList)
 class SubmittedPriceListAdmin(admin.ModelAdmin):
     list_display = ('contract_number', 'vendor_name', 'submitter',
                     'is_approved')
+
+    actions = [approve, unapprove]
 
     fields = (
         'contract_number',
@@ -58,6 +81,7 @@ class SubmittedPriceListAdmin(admin.ModelAdmin):
 
     readonly_fields = (
         'schedule_title',
+        'is_approved',
         'current_status'
     )
 
@@ -70,14 +94,16 @@ class SubmittedPriceListAdmin(admin.ModelAdmin):
             return mark_safe(
                 "<span style=\"color: green\">"
                 "This price list has been approved, so its data is now "
-                "in CALC. Uncheck the <strong>Is approved</strong> box to "
-                "remove its data from CALC.</span>"
+                "in CALC. To unapprove it, you will need to use the "
+                "'Unapprove selected price lists' action from the "
+                "<a href=\"..\">list view</a>."
             )
         return mark_safe(
             "<span style=\"color: red\">"
             "This price list is not currently approved, so its data is "
-            "not in CALC. Check the <strong>Is approved</strong> box to "
-            "add its data to CALC."
+            "not in CALC. To approve it, you will need to use the "
+            "'Approve selected price lists' action from the "
+            "<a href=\"..\">list view</a>."
         )
 
     def schedule_title(self, instance):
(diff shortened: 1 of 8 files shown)


toolness:
I think the tests are probably gonna fail here because I haven't yet added a db migration.

toolness:
Jacobian and CM mentioned in slack that sometimes processes are killed off once their main (request) thread is finished, so it's probably a good idea to just go for celery/rq/channels rather than spinning off a separate thread.

jseppi:
:P Wrong PR?

toolness:
Ack! Lol yup, sorry.

toolness:
@jseppi ok I think this is good for review!

data_capture/tests/common.py
```diff
@@ -52,20 +52,35 @@ def r10_file(content=None, name='r10.xlsx'):
...
+            user.is_staff = True
+            user.is_superuser = True
+            user.save()
+        return user
+
+    def login(self, username='foo', is_staff=False, is_superuser=False):
+        user = self.create_user(username=username, password='bar',
+                                is_staff=is_staff,
+                                is_superuser=is_superuser)
+        assert self.client.login(username=username, password='bar')
```
toolness: Blergh, I don't like any of this code.  I wish that [`force_login`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.Client.force_login) was available in 1.8.

data_capture/tests/test_models.py
```diff
@@ -1,45 +1,30 @@
...
-        )
-        final_kwargs.update(kwargs)
-        return SubmittedPriceList(**final_kwargs)
-
-
-@override_settings(DATA_CAPTURE_SCHEDULES=[FAKE_SCHEDULE])
-class ModelsTests(ModelTestCase):
-    def setUp(self):
-        self.user = User.objects.create_user(username='foo')
-        registry._init()
```
toolness: Weird, `ModelsTests.setUp()` and `ModelsTests.create_price_list()` were basically duplicated from `ModelTestCase` for no reason here. Got rid of the duplicates.

data_capture/tests/test_models.py
```diff
@@ -1,45 +1,30 @@
...
+    def setup_user(self):
+        self.user = self.create_user(username='foo')
```
toolness: I might be trying a bit too hard to reuse functionality through inheritance here. :persevere:

data_capture/tests/common.py
jseppi: Hrm, I'm a little confused on when we should use `is_staff` vs `is_superuser`. In the R10 flow I used `is_staff` as the requirement for whether a user could access the flow. Are we thinking that will be a difference class of users from those who are `is_superuser`?

data_capture/admin.py
```diff
@@ -40,32 +123,85 @@ def current_status(self, instance):
...
+            "<a href=\"..\">list view</a>."
         )
 
     def schedule_title(self, instance):
         return registry.get_class(instance.schedule).title
 
     schedule_title.short_description = 'Schedule'
 
+    # http://stackoverflow.com/a/25813184/2422398
+    def get_actions(self, request):
```
jseppi: Maybe include a short comment here on why this (removing the delete option) is being done

data_capture/tests/test_admin.py
```diff
@@ -0,0 +1,99 @@
...
+        self.price_list.save()
+        self.row = self.create_row(price_list=self.price_list)
+        self.row.save()
+
+    def setup_user(self):
+        self.user = self.login(is_superuser=True)
+
+
+@override_settings(
+    # Strangely, disabling DEBUG mode silences some errors in Django
```
jseppi: o_O

data_capture/tests/common.py
toolness: Oh, so generally `is_staff` gives someone access to view the admin index page, but at that point they need the [proper permissions](https://docs.djangoproject.com/en/1.10/topics/auth/default/#default-permissions) to access any of its actual views/functionality.

So rather than adding each one of those individually, for testing it was easier to just give `is_superuser` to the admin user rather than `is_staff` and the proper permissions.

That said, in the future we might want to revisit the permissions required for folks to access the R10 upload (#555), such that e.g. we test for both `is_staff` and, say, `contracts.add_bulkuploadcontractsource`.

data_capture/tests/test_admin.py
toolness: Yeah, pretty weird--without `DEBUG`, it seemed as though methods raising errors would just result in certain forms being omitted from the final template rendering, rather than propagating to a 500.

jseppi:
👍 Looks good after that comment is added!

toolness:
@jseppi ok, I actually ended up factoring out a new `UndeletableModelAdmin` class since I didn't want to put identical comments in two different places. :)  Can you review commit 14d3e9a05209d6ba647335344b17f9d89776954b real quick?

jseppi:
Neat, looks good 👍

PR #604, opened 2016-08-24
Use Gunicorn instead of Waitress

### LINKED ISSUE #601: Use gunicorn instead of waitress
Per @cmc333333's suggestion.

Ref: https://github.com/18F/atf-eregs/issues/200

jseppi:
Super simple change. @cmc333333 said that `gunicorn` is preferred because `waitress` potentially fails in strange ways (and might be related to current production outages).

I've also removed some files (`initdb.sh` and `Procfile`) that had references to waitress, but are unused as far as I can tell.

Please review, @toolness 

Closes #601


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ app:
     - DDM_HOST_USER=calc_user
     - PYTHONUNBUFFERED=yup
     - DATABASE_URL=postgres://calc_user@db/calc
-  command: sh -c "python manage.py collectstatic --noinput ; waitress-serve --port=${DOCKER_EXPOSED_PORT} hourglass.wsgi:application"
+  command: sh -c "python manage.py collectstatic --noinput ; gunicorn -b :${DOCKER_EXPOSED_PORT} hourglass.wsgi:application"
```
jseppi: This syntax works for specifying the port and binding to all addresses. I think that's what you want here in this docker-compose, but please check, @toolness.

toolness: Thanks, just tested it out and it works great!

toolness:
Woot thanks!

I think that `Procfile` might have been used for deploying to heroku a long time ago or something?  Not really sure.  Also confused about `initdb.sh`.  Either way, they both seem unused and undocumented so let's get rid of em!

This looks rad feel free to merge!

PR #606, opened 2016-08-25
Process bulk uploads asynchronously using rq/django-rq

### LINKED ISSUE #565: Add created, last modified fields to price list model
This is really easy to add and will help make it easy for data admins to sort chronologically if needed, and also for forensics and stuff like that.

### LINKED ISSUE #600: R10 Upload - Step 3 request times out on large data loads (on cloud.gov)
When loading a large file, the request for Step 3 takes too long and cloud.gov aborts it. We apparently are unable to change the request timeout to be longer.

I think the most robust solution would be to use `celery` and spin the long-running tasks off to its processing queue. This will require also having `redis` (or similar) running in our deployment. This is also probably not trivial to get setup.

toolness:
This is an attempt to fix #600 by running the bulk upload processing in a ~~separate thread~~ [`rq`](http://python-rq.org/) worker and emailing users once the upload is complete.  If the processing was successful, the user receives an email like this:

```
This message is in regards to Region 10 CALC data upload #4,
uploaded by atul.varma@gsa.gov on Aug. 25, 2016, 12:59 p.m..


The upload was successful!

3 rows of rate information have been loaded and
previous Region 10 data has been removed from the CALC database.
```

If the upload failed for some reason, the user receives an email like this:

```
This message is in regards to Region 10 CALC data upload #6,
uploaded by atul.varma@gsa.gov on Aug. 25, 2016, 1:01 p.m..


Unfortunately, an error occurred while importing the data. The
traceback below may help the development team debug the problem.

  File &quot;/calc/data_capture/views/bulk_upload.py&quot;, line 146, in run
    kablooey
NameError: name &#39;kablooey&#39; is not defined
```

(Yeah, I'm not sure how to not make it HTML-escape those quotes.  Bleh.)

## Technical Notes

~~Note that while spinning off separate threads to do resource-intensive tasks is traditionally not recommended, I'd argue this is a special case because it's performed quite rarely by a very small number of administrative users.  Normally, using a task queue like celery or [`rq`](http://python-rq.org/) is preferred because it allows us to manage lots of jobs made by many users simultaneously, which would normally result in server resource exhaustion if the jobs were instantly spun off into separate threads.  But because only one or two people will be performing this task once in a blue moon, and because this is the only possible kind of asynchronous task that currently exists, I don't think it should be a big deal.~~

Originally I thought we could potentially just spin off a separate thread to do this, but it looks like we need to use `rq`, as sometimes processes are killed off once their main (request) thread is finished.

More details around the rationale for using `rq` are at https://github.com/18F/calc/issues/600#issuecomment-242473020.

## Other Notes

* It looks like `django-rq` supports [synchronous mode](https://github.com/ui/django-rq#synchronous-mode) for debugging and testing, but what's weird is that it still requires redis, and there's no easy way to configure this aspect of it via `REDIS_URL`... So for the purposes of dev/prod parity, and the fact that it's not hard to set up a worker as part of our docker-compose setup, I'm just ignoring it for now.

* This PR also fixes #565.

## To Do

- [x] Add/fix tests (they're broken right now).
- [x] Determine if this is an OK solution in terms of usability.
- [x] Determine if this is an OK solution technically.
- [x] At this point email will be required for deploying, so we should make `settings.py` raise a helpful error if `EMAIL_URL` isn't defined.
- [x] Make this work w/ cloud.gov (in other words, merge #613)
- [x] Move the `pip install` stuff from the `Dockerfile` into `requirements.txt`.


jseppi:
Do you think it would be good to prevent other bulk uploads from starting before the current one is complete? I'm not completely sure how to do that. Maybe a a database field somewhere that gets flipped to `True` during processing and `False` once finished?

toolness:
Hmm, do we think that's likely to happen?  If Kelly's the only one uploading, and he's only doing it on rare occasions, I'm not sure if it's that big a deal...  I guess if we really want that kind of functionality, though, it might be worth looking into using celery or `rq` instead, since one of those might give us that kind of serialization for free...

jseppi:
I don't think it is very likely to happen. What "bad stuff" would happen if another price list was submitted before the first finished? Would the first finish and the second would just load over it? Actually, it'd probably be a race, depending on submission time and speed of processing.
I guess as long as it doesn't break the app then not having the lock is ok...

toolness:
Yeah, I think it'd be a race, but the good thing is that both jobs would be running as atomic transactions, so whichever "wins" will win completely, rather than the db being in some kind of bizarre state where both transactions or present or something. I think.

toolness:
Oops, meant to post this here earlier today:

> Jacobian and CM mentioned in slack that sometimes processes are killed off once their main (request) thread is finished, so it's probably a good idea to just go for celery/rq/channels rather than spinning off a separate thread.

toolness:
> The cloud.gov redis service is using redis version 2.8 (https://docs.cloud.gov/apps/managed-services/). However it looks like there is no official docker image for that version :\ Hopefully there aren't noticeable differences between 2.8 and 3.0.

Ah, that sucks!  Well, according to redis' docs, they work really hard to ensure as much backwards compatibility as possible... So I'm guessing we should be OK, although the fact that the *major* version number changed is a bit unsettling...

toolness:
YASS THE TESTS PASS 🎉 🎉 💃 💥 💯

data_capture/views/bulk_upload.py
```diff
@@ -101,28 +102,87 @@ def region_10_step_3(request):
...
     Contract._fts_manager.update_search_field()
 
     # Update the upload_source
     upload_source.has_been_loaded = True
     upload_source.save()
 
+    return len(contracts), len(bad_rows)
+
+
+def process_bulk_upload_and_send_email(upload_source_id):
```
jseppi: Might be nice organizationally to move tasks into another file, like `tasks.py` or something?

docker-services.yml
```diff
@@ -0,0 +1,11 @@
...
+  build: .
+  environment:
+    - DDM_HOST_USER=calc_user
+    - PYTHONUNBUFFERED=yup
+    - DATABASE_URL=postgres://calc_user@db/calc
+    - REDIS_URL=redis://redis:6379/0
+    # This *must* be at least 2 in order for the fake authentication
+    # server to work, as it makes the app connect to itself (which will
+    # cause the app to hang if only one worker exists).
+    - WEB_CONCURRENCY=2
```
jseppi: What does this do?

data_capture/views/bulk_upload.py
toolness: Yas good idea!

docker-services.yml
toolness: It basically tells `gunicorn` to run with at least 2 threads (it's actually recommended that it be set to the number of cores on the host system, but this is just for development so whatevs).

Anyways, the problem with the default of 1 for development is that the app will hang if any of its views ever make an HTTP request against another view--which is exactly what happens with our fake UAA, since the "fake cloud.gov" is on the app server itself.

toolness:
Just factored out the job-related stuff into a `jobs.py`.  (I didn't call it `tasks` b/c `rq` calls them `jobs`).

toolness:
Ok @jseppi I think this is good to go!

jseppi:
👍 woooo! Looks great!

jseppi:
Oh, I guess merge in #622 to this one before merging into `develop` (if you think #622 is good to go).

toolness:
Woot merging now!

PR #613, opened 2016-08-25
load REDIS_URL into env from VCAP_SERVICES

jseppi:
`calc-redis` is up and running in our staging env.

This PR adds code that grabs the `calc-redis` info from `VCAP_SERVICES` and formats it into a url. That url is set into `env['REDIS_URL']`.

This PR is against #606. Ready for review, @toolness


hourglass/settings_utils.py
```diff
@@ -34,3 +35,28 @@ def get_whitelisted_ips(env=os.environ):
...
+
+    for entry in vcap.get('redis28-swarm', []):
+        if entry['name'] == name:
+            f = furl.furl()
+            creds = entry['credentials']
+            f.host = creds['hostname']
+            f.port = creds['port']
+            f.password = creds['password']
+            url = 'redis://{}'.format(f.url)
+            env['REDIS_URL'] = url
```
jseppi: Could also make this side-effect free and instead return the URL, then set it "manually" in `settings.py`

hourglass/tests.py
```diff
@@ -115,6 +117,44 @@ def test_credentials_are_loaded(self):
...
+                    'password': 'the_password',
+                    'port': '1234'
+                }
+            }]
+        })
+
+        load_redis_url_from_vcap_services('redis-service', env=env)
+        self.assertTrue('REDIS_URL' in env)
+        self.assertEqual(env['REDIS_URL'],
+                         'redis://:the_password@the_host:1234')
```
jseppi: There is no username field, which I guess is valid, but it looks funny to me. I have not verified that a URL generated from this code actually works to connect to the redis instance.

hourglass/settings_utils.py
```diff
@@ -1,5 +1,6 @@
 import os
 import json
+import furl
```
toolness: Erm, is it possible to do this without `furl`?  I realize it's a bit more verbose to use the standard [`urllib.parse.urlunparse`](https://docs.python.org/3.0/library/urllib.parse.html#urllib.parse.urlunparse) but I think python folks would actually be more familiar w/ the "batteries included" function than reading up the docs on a third-party module (which was first released two months ago and could contain bugs or go unmaintained etc).  I guess I am also wary of adding unnecessary deps to python projects because unlike node/npm, Python's dependency tree is flat, which increases the likelihood of version conflicts and stuff.

Anyways, it could just be a knee-jerk reaction on my part, feel free to disagree!

toolness: Other than this I think this PR is 👍 !

jseppi: Sure, I can remove `furl`. I saw it being used by https://github.com/jmcarp/py-cfenv, which is used by the OpenFEC project, so that's how I even found out about it.

jseppi:
Ok, I removed the `furl` dependency, but I also didn't use `urllib`. I'm not just constructing the URL with a simple string `.format`.

I perhaps mistakenly merged in the `async-bulk-upload` branch to clear up the merge conflicts, but now I think the test suite won't pass on travis. I did locally run the `hourglass` tests against my changes and they all passed though.

jseppi:
Oh but you just got your tests passing! Pulled and merged again!

hourglass/settings_utils.py
```diff
@@ -34,3 +34,28 @@ def get_whitelisted_ips(env=os.environ):
...
+    vcap = json.loads(env['VCAP_SERVICES'])
+
+    for entry in vcap.get('redis28-swarm', []):
+        if entry['name'] == name:
+            creds = entry['credentials']
+            url = 'redis://:{password}@{hostname}:{port}'.format(
+                password=creds['password'],
+                hostname=creds['hostname'],
+                port=creds['port']
+            )
```
toolness: Oh! This is fine, but I think you might even be able to get away with `.format(**creds)`.  Maybe.

toolness:
This looks good!  I *think* the tests should work w/ the latest changes to `async-bulk-upload`.  Feel free to merge.

hourglass/settings_utils.py
jseppi: Hrm, that does work, and looks like _real Python code_ but I kinda like the explicitness of specifying the parts given that we don't actually have control over what is in `creds` and I think throwing a `KeyError` there is a good idea in case something changes with the redis service.

toolness: Ah good idea! Rad.

PR #526, opened 2016-08-05
Low-impact docker/requirements fixes

toolness:
A few quick fixes:

* Fix #510 so the Dockerfile builds from scratch again
* Fix #515 so `manage.py dbshell` works
* Fix #517 to get rid of the unused `nose` and `django-nose` python packages

PR #615, opened 2016-08-26
Fix Dockerfile to work again based on /usr/bin/python change blah.

toolness:
Blergh, in #526 we had to make a weird change to the dockerfile because the official Python Docker Hub image made `/usr/bin/python` inaccessible.

However, they just undid this in https://github.com/docker-library/python/pull/137, which made our weird change *break* our Dockerfile.  Awesome.

This undoes our weird change so that everything works again.

Blah.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.4
+FROM python:3.4.5
 
 ENV PHANTOMJS_VERSION 1.9.7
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.4
+FROM python:3.4.5
```
toolness: Aside from being more explicit (and using the same python version as we specify in `runtime.txt`) this will force our developers' machines to re-pull the latest python image, rather than using a stale one that may not contain https://github.com/docker-library/python/pull/137.

jseppi:
👍

PR #619, opened 2016-08-26
add file_mime_type to BulkUploadContractSource

### LINKED ISSUE #595: Add mime/file type to BulkUploadContractSource
Otherwise we don't know which type of file is stored in its binary field.

jseppi:
Closes #595

ready for review, @toolness


contracts/migrations/0016_bulkuploadcontractsource_file_mime_type.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('contracts', '0015_auto_20160819_1924'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='bulkuploadcontractsource',
+            name='file_mime_type',
+            field=models.TextField(default='text/csv'),
```
jseppi: Django made me submit a default for the migration. We won't actually have any to migrate since we'll be starting from a clean db, so I just picked `'text/csv'`.

contracts/models.py
```diff
@@ -130,6 +130,7 @@ class BulkUploadContractSource(models.Model):
     submitter = models.ForeignKey(User, null=True, blank=True)
     has_been_loaded = models.BooleanField(default=False)
     original_file = models.BinaryField()
+    file_mime_type = models.TextField()
```
jseppi: Used `TextField()` since mimetypes can be quite long so I didn't want to guess at a `max_length` for a `CharField`

contracts/migrations/0016_bulkuploadcontractsource_file_mime_type.py
jseppi: Hrm, I guess I could retcon the original migration to add the `file_mime_type` there since nothing will ever _actually_ be migrated.

toolness: Naw I wouldn't worry about it.

toolness:
This looks good to me, thanks!

PR #620, opened 2016-08-26
Change Schedule70Loader header_rows to 1

### LINKED ISSUE #573: First Schedule 70 row is skipped in load_s70 management command
~~I've not yet verified but I just noticed this from looking at the code.~~ Confirmed that the first row is indeed being skipped in `manage.py load_s70`.

The Schedule 70 CSV in the repo has one line of headers: https://github.com/18F/calc/blob/8dfd883c1bb888cae0817f703dd632fb0d891e37/contracts/docs/s70/s70_data.csv

But the Schedule70Loader class assumes 2 rows of headings and skips them: https://github.com/18F/calc/blob/8dfd883c1bb888cae0817f703dd632fb0d891e37/contracts/management/commands/load_s70.py#L78

jseppi:
Change header_rows to 1, fixes #573


jseppi:
Oh weird, I just noticed the S70 test fixtures (in `contracts/tests/data/`) actually _do_ have two rows of headers, but even after this change the tests still pass. This is because the second row of headers in the text fixture CSVs kinda "pass through" the loader and are silently skipped.

jseppi:
But the actual S70 data in `contracts/docs/s70/s70_data.csv` does not have two rows of headers and the first row certainly was being skipped.

I guess we should remove the second row of headers in the fixtures?

toolness:
Ummmmm weird.  Should we leave a comment above the `header_rows` definition which mentions that it's ok for it to be less than it actually is, since weird rows are thrown out?

jseppi:
They're only thrown out when `strict=False` in the `load_file` and `parse` methods (which is the default).

toolness:
Oh ok.  Cool, feel free to merge this!

PR #622, opened 2016-08-26
Add redis service instance to manifests

jseppi:
Open against `async-bulk-upload`.

### DIFF
--- a/manifests/manifest-sandbox.yml
+++ b/manifests/manifest-sandbox.yml
@@ -16,3 +16,11 @@ applications:
   timeout: 180
   env:
     EMAIL_URL: 'console:'
+- name: calc-rqworker
+  no-route: true
+  instances: 1
+  memory: 256M
+  buildpack: python_buildpack
+  command: python manage.py rqworker default
+  stack: cflinuxfs2
+  timeout: 180
--- a/manifests/manifest-sandbox.yml
+++ b/manifests/manifest-sandbox.yml
@@ -2,6 +2,7 @@
 services:
 - db-calc-sandbox
 - calc-env
+- calc-redis
 applications:
 - name: calc-sandbox
   instances: 1
@@ -13,3 +14,5 @@ applications:
   command: bash cf.sh
   stack: cflinuxfs2
   timeout: 180
+  env:
+    EMAIL_URL: 'console:'
--- a/manifests/manifest-sandbox.yml
+++ b/manifests/manifest-sandbox.yml
@@ -24,3 +24,5 @@ applications:
   command: python manage.py rqworker default
   stack: cflinuxfs2
   timeout: 180
+  env:
+    EMAIL_URL: 'console:'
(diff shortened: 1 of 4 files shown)


manifests/manifest-sandbox.yml
```diff
@@ -13,3 +14,5 @@ applications:
   command: bash cf.sh
   stack: cflinuxfs2
   timeout: 180
+  env:
+    EMAIL_URL: 'console:'
```
jseppi: I think this should work so that the sandbox doesn't require email access.

jseppi:
This should be good to go, @toolness. I verified that queue processing works in my sandbox.

Some notes:

1. I set `EMAIL_URL` in both apps in the sandbox manifest. For staging, we will need to set up `EMAIL_URL` in the UPS config, but we need to first get the MailChimp stuff setup.
1. Messages logged within the method called by the `rq` worker DO NOT show up in the logs for the web app. For example, they will not show up in `cf logs --recent calc-sandbox`. They will show up in the logs for the worker process, like `cf logs --recent calc-rqworker`.
1. I can't seem to deploy both apps in a single push, via `cf push -f manifests/manifest-staging.yml`. The web app gets deployed, but then the deployment of the worker app fails. I can however deploy the worker separately via `cf push -f manifests/manifest-staging.yml calc-rqworker`. I _think_ it is a timing issue. I'll ask in Slack to see if anyone can help with that.

manifests/manifest-sandbox.yml
toolness: I assume it's specified the same way all [cloud.gov-based deployments configure environment variables](https://github.com/18F/calc/blob/develop/deploy.md#environment-variables)?

toolness:
> Messages logged within the method called by the rq worker DO NOT show up in the logs for the web app.

Maybe document this in `deploy.md`?

manifests/manifest-sandbox.yml
jseppi: I meant that just in this sanbox config I had to specify it in both apps since they both rely on EMAIL_URL. For staging and prod we should definitely specify it via the linked UPS.

toolness: Oh I see. Um, so does this mean they show up in the actual OS environment on cloud.gov sandbox deploys, rather than through UPS?  If they're specified in both, does UPS supersede environment?

Basically I just want to make sure that it's possible to actually set up email sending on sandbox accounts if one really wants to.

jseppi: Yeah, they'll be OS-level environment variables in the sandbox deploy. The ordering of what supercedes what is based on the code in `settings.py`. I'm pretty sure the UPS value would supercede the environment variable.

That all said, I set `EMAIL_URL` here just to be explicit so that when we did get email sending access, it would be obvious that we should delete these `env:` values (and commit the change) and set the value in UPS instead. We can do that now though if you prefer.

toolness:
NOICE

PR #629, opened 2016-08-30
Canceling data_capture flows

### LINKED ISSUE #478: Cancel button in Data Capture Upload flow should clear any submitted data

jseppi:
This PR adds a decorator called `@handle_cancel` for canceling Data Capture flows. The decorator checks if the request is a `POST` and if the `POST` data contains a `cancel` key. If so, then the decorator clears session of all `data_capture:` keys, and redirects the user to the app's index page (both the prefix and the redirect page name can be passed as args to the decorator if desired).

Form templates have been updated so that all the `<a>Cancel</a>` elements have been changed to `<button type="submit" name="cancel">...</button>` elements. The `name` attribute is important, because this is what the decorator looks for in the `POST` data.

I also did some minor style fixes to `.form-button-row` so that the change from `<a>` to `<button>` looked right, and I replaced instances of the class `button-cancel` with `button-secondary` since that class already exists with the color we want.

Closes #478

To Do:
- [x] functionality
- [x] tests
- [x] add decorator to applicable view methods
- [x] styling tweaks for making cancel buttons (and updates to Styleguide as necessary)


data_capture/decorators.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def handle_cancel(redirect_name='index', key_prefix='data_capture:'):
+    '''
+    Decorator to handle cancel behavior in Data Capture flows.
+    The associated request's POST data is checked for a 'cancel' key,
+    and, if found, all session keys that start with `key_prefix`
+    are deleted and the request is redirected to `redirect_name`.
+    '''
+    def decorator(function):
```
jseppi: Does this all seem reasonable, @toolness?

jseppi:
Ugh, I can't seem to get my test methods to do what I think they should be doing.

data_capture/decorators.py
```diff
@@ -0,0 +1,58 @@
...
+    no_args = False
+    if callable(redirect_name):
+        function = redirect_name
+        no_args = True
+
+    def decorator(function):
+        @wraps(function)
+        def wrapper(request, *args, **kwargs):
+            if request.method != 'POST':
+                raise
```
jseppi: If I raise a specific exception type here, I'm not able to catch it in the tests with `with self.assertRaises(ValueError):` for example. I don't know why this is.

data_capture/tests/test_decorators.py
```diff
@@ -0,0 +1,70 @@
...
+            self.client.post('/empty_redirect_name/')
+
+    def test_removes_keys_from_session(self):
+        session = self.client.session
+        session['data_capture:key_a'] = 'value_a'
+        session['data_capture:key_b'] = 'value_b'
+        session['something_else'] = 'boop'
+
+        res = self.client.post('/test_view/')
+        self.assertEqual(302, res.status_code)
```
jseppi: These tests don't work :\

jseppi:
ready for review @toolness

jseppi:
The problem I was having with tests was that I was trying to raise exceptions in the decorator, but those exceptions are not really catchable in a straightforward way. It seems that throwing decorator exceptions is not a great practice. Instead of checking arguments, I think we're covered by normal Django behaviors. For example, f a view name isn't found, then Django will through an exception somewhere else when it tries to do the reverse lookup, so I think that's ok.

data_capture/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+
+                # if AJAX request, then send JSON response
+                # that has a 'redirect_url' property
+                if request.is_ajax():
+                    return JsonResponse(
+                        {'redirect_url': reverse(redirect_name)}
+                    )
+
+                # redirect to the view named redirect_name
+                return redirect(redirect_name)
```
toolness: Would it be easier/posible to just use [`ajaxform.redirect`](https://github.com/18F/calc/blob/develop/frontend/ajaxform.py#L13) here instead?

data_capture/decorators.py
```diff
@@ -0,0 +1,53 @@
...
+        @wraps(function)
+        def wrapper(request, *args, **kwargs):
+            if request.method == 'POST' and 'cancel' in request.POST:
+                if key_prefix:
+                    # .keys() returns an iterator, so we use list()
+                    # to get an actual list of keys
+                    session_keys = list(request.session.keys())
+                    for k in session_keys:
+                        if k.startswith(key_prefix):
+                            del request.session[k]
```
toolness: Hmm, why do we need a `list` and not an iterator? Oh, is it because we can't delete items from an iterator while it's being iterated through or whatever?

I wonder if it'd be a bit more concise to use a [list comprehension](http://www.secnetix.de/olli/Python/list_comprehensions.hawk) here, e.g.:

```
for [k for k in request.session.keys() if k.startswith(key_prefix)]:
    del request.session[k]
```

toolness:
This looks good, feel free to merge it!

data_capture/decorators.py
jseppi: Hrm, it would certainly work though I think it creates a somewhat weird dependency. But I guess there is already an implicit dependency in the `JsonResponse` format being used here.

data_capture/decorators.py
jseppi: That's correct for why I needed to use `list(...)`

Would using a list comprehension obviate the need to use `list(...)`, or would it have to be:

```py
for [k for k in list(request.session.keys()) if k.startswith(key_prefix)]:
    del request.session[k]
```
?

I personally find the comprehension a little harder to read in this case, but if that is the more accepted way of doing it then I can go with it.

data_capture/decorators.py
jseppi: @toolness: created PR https://github.com/18F/calc/pull/641 to factor out the JSON object creation in `ajaxform.py`. How does that look?

toolness: Sure, it looks good!

data_capture/decorators.py
toolness: Nah, I am just a list comprehension junkie--it is definitely not a preferred thing in the python community as a whole.

That said, I *think* that using the list comprehension would obviate the need to use `list()` because of the fact that it's surrounded by `[]`, which forces it into a list.  (You could just write `(x for x in blah)` and then it's a *generator expression* rather than a list comprehension.)

I do think that inlining the list comprehension directly in the `for` loop isn't particularly readable, though.  I would generally write something like this:

```py
keys_to_delete = [k for k in request.session.keys() if k.startswith(key_prefix)]
for key in keys_to_delete:
    del request.session[key]
```

Doing it your way is fine too, I just find it mentally easier to separate tasks like this into a "first create a list of things you want to delete, then delete them" sort of algorithm, as it always obviates the whole "argh mutating a thing you're iterating through is problematic" issue.

jseppi:
womp womp, travis failed with error `Cannot download "https://github.com/sass/node-sass/releases/download/v3.8.0/linux-x64-47_binding.node":`

Resubmitted the job

jseppi:
Lol, and now codeclimate is apparently stalled. Happy Wednesday!

jseppi:
Um, something is wrong with codeclimate. It's been running for nearly hour yet hasn't even finished checking out the code according to its status page. Given that it had a green light before the merge of #641, and #641 had its own green light, I am just going to merge this PR.

screen shot 2016-08-31 at 3 46 43 pm

PR #630, opened 2016-08-30
Only call attachedCallback() once on web components.

### LINKED ISSUE #626: Web components need to work when attachedCallback() is called more than once
I was just looking through the Web Components specification and noticed that it said the `attachedCallback()` (or rather, `connectedCallback()`, as it's called in the latest spec) needs to tolerate being called more than once--I think because it's possible for a custom element to to be appended to the document structure, detached, and then appended again.

The current `attachedCallback()` methods in our web components don't do this.

toolness:
Fixes #626.

Blerg, I thought for a while about how to test this and the only thing I could think of is to attach an event listener for the components' ready events, remove them from the DOM and re-attach them and wait a certain amount of time to make sure the ready events don't fire again.  But I don't really like that.

### DIFF
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -31,6 +31,10 @@ exports.MISC_ERROR = MISC_ERROR;
 
 class AjaxForm extends window.HTMLFormElement {
   attachedCallback() {
+    if ('isDegraded' in this) {
+      return;
+    }
+
     this.isDegraded = !supports.formData() ||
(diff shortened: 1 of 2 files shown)


jseppi:
looks like eslint doesn't like the lines with `return finishInitialization();`  The `finishInitialization` function doesnt' return anything, so perhaps you just need to separate those lines into:

```js
finishInitialization();
return;
```

toolness:
Blerg, I don't like this PR. I'm going to close it and maybe re-open a different one later.

PR #631, opened 2016-08-30
Degrade upload widget if not in ajaxform

### LINKED ISSUE #628: <upload-widget> should degrade itself if not in an ajaxform
Currently, if an upgraded `<upload-widget>` isn't in an ajaxform, the end-user will be frustrated if they drag a file to the widget and submit the form.  We may want to just log a console warning and auto-degrade the widget if this is the case, so that at least the page still works.

toolness:
Fixes #628.

Note that `<upload-widget>` actually only cares if it's in a `<form>` that is not an ajaxform.  If it's not in a `<form>` *at all*, then it actually doesn't care, which allows us to still test it and stuff without having to put it in a form.

### DIFF
--- a/frontend/source/js/data-capture/upload.js
+++ b/frontend/source/js/data-capture/upload.js
@@ -100,6 +100,21 @@ class UploadWidget extends window.HTMLElement {
     event.preventDefault();
   }
 
+  _checkForAjaxFormParent() {
+    if (this.uploadInput.form &&
+        this.uploadInput.form.getAttribute('is') !== 'ajax-form') {
+      if (window.console && window.console.log) {
+        window.console.log(
+          'Warning: <upload-widget> must have a ' +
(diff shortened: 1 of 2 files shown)


frontend/source/js/data-capture/upload.js
```diff
@@ -100,6 +100,21 @@ class UploadWidget extends window.HTMLElement {
     event.preventDefault();
   }
 
+  _checkForAjaxFormParent() {
+    if (this.uploadInput.form &&
+        this.uploadInput.form.getAttribute('is') !== 'ajax-form') {
+      if (window.console && window.console.log) {
+        window.console.log(
```
toolness: I guess in a future PR we could make a `logging` module that only logs stuff if `window.console` exists? We could even make it stubbable so that tests could verify that console messages are output and stuff.  I dunno.

jseppi:
👍

PR #633, opened 2016-08-31
A few styleguide enhancements

toolness:
When @hbillings was working on #583, she ran into a number of problems understanding how our ajax form and upload widget work.  This PR adds a few enhancements to the styleguide that will hopefully alleviate the frustration of future developers:

* When source files are mentioned by name, we hyperlink them to the syntax-highlighted GitHub source.
* When Python classes/modules/etc are mentioned by name, we use introspection to determine which file they came from, and hyperlink them to the GitHub source.
* Added a comment to the example ajax form source pointing out the `is="ajax-form"` attribute, as it's super easy to miss.

Note that I link to GitHub instead of adding some kind of syntax-highlighting static file server partly out of expediency, but also party for security purposes--I don't want to somehow accidentally expose a hole in our app allowing attackers arbitrary read access to the whole file system.  I guess in the future maybe we could add such a server that's active only when `DEBUG` is true, and otherwise link only to GitHub.


jseppi:
Nice! 👍

toolness:
Oops.  Um, I just added a kind of hefty addition in 3a2c0b7db4413e51fcb105e438bdd28ace99e535 before seeing that you'd already reviewed this... I think I will file everything you reviewed as a separate PR and then insta-merge it, and then file 3a2c0b7db4413e51fcb105e438bdd28ace99e535 as a separate PR.

PR #632, opened 2016-08-30
Add /healthcheck/ endpoint

toolness:
This fixes #625 by adding a `/healthcheck/` endpoint that checks to make sure the DB is synchronized and that redis/`rq` is working.

PR #634, opened 2016-08-31
Fix flake8 error.

toolness:
Ack, #632 actually contained a linting error in it, but I don't think it was raised by our CI because we don't use QuantifiedCode anymore.

This fixes the linting error--but should we also perhaps run flake8 on travis now?  To maintain DRY, I guess we could just run `ultratest`, even though it means that our eslint would be running twice I guess--once by CodeClimate and once on travis.  Hrm.


jseppi:
Good catch! Yeah, we should definitely add `flake8` to travis. I think `ultratest --lintonly` is the way to go.

jseppi:
Documented flake8 need in https://github.com/18F/calc/issues/636

PR #635, opened 2016-08-31
Add an <alerts-widget>

### LINKED ISSUE #481: Ajaxform updates aren't noticeable to screen readers

toolness:
This is an alternative attempt to fix #481.

Whereas #575 was a really complex attempt to fix the issue, this solution is far more simple. The following is suggested by  [WebAIM's Accessibility of Rich Internet Applications guide](http://webaim.org/techniques/aria/#keyboard):

> Additionally, it is often necessary to set focus to page elements. For instance, a form validation error message might be displayed as text (not a link or form element) within a page using scripting. Visual users can immediately see the error message. However, a screen reader user may not know that the new message is present. In order to set focus to the error message so it can be read by a screen reader, the message must be able to receive focus - something it cannot typically do unless it is a link or form control.

This basically implements the above approach through the use of an extremely simple `<alerts-widget>` web component that just sets `tabindex="-1"` on itself and focuses itself on attachment.  That is literally all it does, and it seems to work great on voiceover and NVDA.

Another nice benefit of this approach is that because the alerts appear just above the form fields, the user only has to press tab to instantly focus on the first field.  (I believe that without this focusing, the focus became undefined when loading an ajax-form, as the submit button which previously had focus was replaced by the new content, which led to weird keyboard accessibility issues.)

Note, though, that this *does* also mean that alerts which appear on non-ajax-loaded content (i.e., a normal page) will be auto-focused.  I think this is probably OK though, because they're still the first thing we want screen readers to announce when the users visit a new page.

In the future we may consider changing the `:focus` styling of the web component.  Currently it puts a big fuzzy blue outline around the alerts on a Mac, which is kind of overkill, especially since the alerts themselves already have very visible borders (though those borders don't indicate keyboard focus).

## To Do

- [x] Add tests.

- [x] Document the widget in the style guide.

### DIFF
--- a/data_capture/templates/data_capture/messages.html
+++ b/data_capture/templates/data_capture/messages.html
@@ -1,9 +1,9 @@
 {% if messages %}
-<div class="alerts">
+<alerts-widget>
   {% for message in messages %}
     <div class="alert alert-{{ message.tags }}">
       <p>{{ message }}</p>
     </div>
   {% endfor %}
-</div>
+</alerts-widget>
 {% endif %}
--- a/frontend/source/js/data-capture/alerts.js
+++ b/frontend/source/js/data-capture/alerts.js
@@ -1,5 +1,7 @@
 /* global window, document */
 
+import { dispatchBubbly } from './custom-event';
+
 /**
  * AlertsWidget represents an <alerts-widget> web component.
  *
@@ -13,6 +15,7 @@ class AlertsWidget extends window.HTMLInputElement {
   attachedCallback() {
     this.setAttribute('tabindex', '-1');
     this.focus();
+    dispatchBubbly(this, 'alertswidgetready');
   }
 }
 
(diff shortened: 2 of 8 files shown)


data_capture/templates/data_capture/messages.html
```diff
@@ -1,9 +1,9 @@
 {% if messages %}
-<div class="alerts">
+<alerts-widget>
```
toolness: I did a search through the source code and it *seemed* as though there aren't any rules involving `.alerts`, so I just got rid of the class name.

jseppi:
I definitely like the technical approach of this one better. Does the accessibility guild have any guidance?

toolness:
[Just asked in #g-accessibility](https://18f.slack.com/archives/g-accessibility/p1473353184000016)...

toolness:
Bristow says it's cool!  Should I go ahead and add tests?

jseppi:
woot! yeah!

toolness:

jseppi:
Great! Seems to work well 👍

PR #638, opened 2016-08-31
Mention, link to web components in style guide.

toolness:
Ok, this is a bit funkadelic. It basically allows us to link to the source code of web components in our style guide.

As with the changes in #633, this is done both to help onboard new/rusty devs, but also to prevent doc rot: an alert is displayed if any of the web components don't exist, or if metadata about their location isn't found.

The only annoying part of this is that it requires us to add a `SOURCE_FILENAME` property to each web component's prototype that is set to `__filename` (the actual value of which is filled-in by browserify).

It'd be nice to make the QUnit test suite open the style guide in an iframe and verify that no errors are thrown, but I'd prefer to file that as a separate bug. For now just seeing a big alert when visiting `/styleguide/` if anything's broken should be enough.

### DIFF
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -108,6 +108,8 @@ class AjaxForm extends window.HTMLFormElement {
   }
 }
 
+AjaxForm.prototype.SOURCE_FILENAME = __filename;
+
 document.registerElement('ajax-form', {
   extends: 'form',
   prototype: AjaxForm.prototype,
--- a/styleguide/static/styleguide/styleguide.css
+++ b/styleguide/static/styleguide/styleguide.css
@@ -58,3 +58,11 @@ html .styleguide-example pre {
   margin: 0;
   border: none;
 }
+
+code > a {
+  text-decoration: none;
+}
+
+code > a:hover {
+  text-decoration: underline;
+}
--- a/styleguide/templatetags/styleguide.py
+++ b/styleguide/templatetags/styleguide.py
@@ -79,6 +79,16 @@ class WebComponentHTMLParser(HTMLParser):
 
 @register.simple_tag
 def webcomponent(html):
+    '''
+    Link to the source code of a web component, e.g. <foo> or
+    <input is="bar">.
+
+    This actually renders markup that client-side code will resolve into
+    a definitive link via introspecting the JS runtime environment. The
+    front-end will raise an error if the web component or its source code
+    can't be found, to help prevent documentation rot.
+    '''
+
     parser = WebComponentHTMLParser()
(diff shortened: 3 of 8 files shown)


frontend/source/js/data-capture/ajaxform.js
```diff
@@ -108,6 +108,8 @@ class AjaxForm extends window.HTMLFormElement {
   }
 }
 
+AjaxForm.prototype.SOURCE_FILENAME = __filename;
```
toolness: This basically tells the style guide what file the web component's source code exists in.

styleguide/static/styleguide/styleguide.css
```diff
@@ -58,3 +58,11 @@ html .styleguide-example pre {
...
   border: none;
 }
+
+code > a {
+  text-decoration: none;
+}
+
+code > a:hover {
+  text-decoration: underline;
+}
```
toolness: I changed the CSS for hyperlinked code because the combination of off-white background + border + blue text + underline was very busy visually.

styleguide/templatetags/styleguide.py
```diff
@@ -60,13 +61,37 @@ def github_url_for_path(path):
     GitHub URL to its syntax-highlighted source code.
     '''
 
-    return '{}/blob/{}/{}'.format(
+    return '{}/tree/{}/{}'.format(
```
toolness: It turns out the `/tree/` endpoint will redirect to `/blob/` if the thing being referenced is a file, but not necessarily the other way around.

toolness:
I should also add that this actually encourages us to document our web components better, since it makes the links to their source code a lot more prominent.  I'd like to add docs in a separate PR.

jseppi:
This is pretty neat!
I'm a bit wary of having to remember to do the `SOURCE_FILENAME` thing. I wonder if we could make a simple gulp plugin that checks if the source file contains a web component, and if so, automatically adds `SOURCE_FILENAME` to either that file, or maybe to a global collection of them.

toolness:
Thanks!

One nice thing about the way it currently works is that the styleguide will pop up a big alert that tells you to add `SOURCE_FILENAME` if a web component is mentioned on a styleguide page but no `SOURCE_FILENAME` is defined on the component's prototype.

A gulp plugin would be cool, but I'm not sure how we'd make it--I think it'd take a bit of research...

Another option might be to make a helper function that wraps `document.registerElement` and requires that `__filename` be passed in as one of its arguments...

jseppi:
Yeah, I think that the `alert` is good for now.

toolness:
Cool, should I go ahead and merge this?

jseppi:
👍

PR #639, opened 2016-08-31
Strengthen references to CSS files in style guide and link to them

toolness:
Currently there are two references to `data-capture.min.css` in the styleguide, but there's no information on where that file is or where it comes from.  It's also a "weak" reference in that it's just plain text that will become confusing and out of date if we ever get rid of `data-capture.min.css`.  So instead, I've changed them both to link directly to the SCSS source files that contain the respective component(s).


jseppi:
Good idea! 👍

PR #640, opened 2016-08-31
Send email to submitter when Price List is un/approved by admin

jseppi:
This PR adds email notices when a SubmittedPriceList is approved or unapproved by a CALC administrator. It also refactors all email sending logic into a single module, `email.py`, and moves all related tests to `test_email.py`.

I've also added a new setting (that should come from env/UPS) called `SYSTEM_EMAIL_ADDRESS` from which all system emails are sent. This address (whatever it ends up being) will most likely need to be registered in the MailChimp account.

To Do:
- [x] implement it
- [x] make emails better: include details such as contract number, company name, date submitted, and approver email
- [x] test it
- ~~send all emails asynchronously using `rq`~~ Let's do this in another PR.
- [x] add setting for global `EMAIL_FROM` ~~(or maybe do this in another PR?)~~
- [x] add tests that the email sending code is called from `approve()` and `unapprove()` admin functions.


jseppi:
I wonder if it makes sense to mark all email sending methods as `@job` so that they can be asynchronously processed. Thoughts, @toolness?

toolness:
That's fine with me--one of the things I like about that `@job` decorator is that it actually returns the same function, it just adds a `.delay()` method to it so that it can be easily turned into a job.  So it gives us a lot of freedom.

That said, the one thing that gives me pause about jobs is that `rq` seems to use pickling or some similar form of serialization whereby it's basically storing a reference to the dotted import name of the job.  That means that if we have a job at `foo.bar.baz` and put it into `rq`, then upgrade the app and suddenly things are refactored such that the job is now located at `foo.boop.baz`, I think our `rq` worker will barf.  I'm not sure though.

Anyways, it's just a thing to keep in mind, I'm fine with making the email a jorb!

jseppi:
> I think our rq worker will barf. I'm not sure though.

Hrm, I wonder if restarting the worker on every deploy would mitigate that.

> I'm fine with making the email a jorb!

Sweet! My plan is actually to move all the email sending logic into separate functions (as I've already started in my last commit). That way they are independently testable, and I guess we could just mark them as `@job`s there.

toolness:
> Hrm, I wonder if restarting the worker on every deploy would mitigate that.

I don't think doing that by itself would do it, as the references to the old source code location of the jobs (e.g. `foo.bar.baz`) is in redis, rather than in the queue worker itself, so once the worker spins up again and grabs a job from redis, it will try looking up `foo.bar.baz` and explode.

That said, we could also clear redis every deploy, but then we might lose jobs...

Anyhow, this is also all speculation on my part, so I could be wrong.  I was confused at first that `rq` didn't require me to have a "registry" of jobs somewhere until I decided that the only way it could just magically work w/o a registry would be if it stored the actual dotted name to the job as part of the job itself.  But my assumption could be wrong.

jseppi:
ready for review, @toolness

jseppi:
Whoops, actually I should add tests that the email sending code is called in the `approve()` and `unapprove()` admin methods. Hold up, @toolness.

data_capture/tests/test_admin.py
```diff
@@ -85,6 +86,9 @@ def test_approve_works(self, mock):
         )
         self.price_list.refresh_from_db()
         self.assertTrue(self.price_list.is_approved)
+        self.assertEqual(len(mail.outbox), 1)
```
jseppi: Not actually sure what the best way to check this functionality is. Could instead mock `data_capture.email` and check that the appropriate methods are called?

jseppi:
Ok, now ready for review, @toolness.

data_capture/tests/test_email.py
```diff
@@ -0,0 +1,69 @@
...
+    def test_price_list_approved(self):
+        price_list = self.create_price_list(is_approved=True)
+
+        result = email.price_list_approved(price_list)
+        self.assertTrue(result.was_successful)
+        message = mail.outbox[0]
+        self.assertEqual(message.recipients(), [self.user.email])
+        self.assertEqual(message.subject, 'CALC Price List Approved')
+        self.assertEqual(message.from_email, settings.SYSTEM_EMAIL_ADDRESS)
+        self.assertEqual(result.context['price_list'], price_list)
```
jseppi: The actual email contents are not being tested in any of these. I suppose we could create email fixtures and compare the generated content against the fixtures.

data_capture/email.py
```diff
@@ -0,0 +1,98 @@
...
+        self.was_successful = was_successful
+        self.context = context
+
+
+def price_list_approved(price_list):
+    ctx = {
+        'price_list': price_list
+    }
+    if not price_list.is_approved:
+        raise ValueError('price_list.is_approved must be True')
```
toolness: This (and the other similar error in the unapproved method) might be better specified as an `AssertionError`, since we're sort of asserting a precondition here?  I'm not sure I guess... `ValueError` is probably fine too yo.

data_capture/tests/test_admin.py
toolness: Hmm I am confused about what the probalo is.  I think that checking `mail.outbox` is cool.  I have always been wary of mocks but am especially so after watching Sandi Metz's [https://www.youtube.com/watch?v=URSWYvyc42M](Magic Tricks of Testing) talk.

data_capture/tests/test_email.py
toolness: NAaaah it's too fragile!

jseppi: Yeah, I don't like mocks in general. It would be cool if we could spy on the actual email module's methods, but I don't know if that's doable.

README.md
```diff
@@ -231,6 +231,9 @@ string), the boolean is true; otherwise, it's false.
   this defaults to `console:`. The setting can easily be manually
   tested via the `manage.py sendtestemail` command.
 
+* `SYSTEM_EMAIL_ADDRESS` is the email from-address to use in all system
+  generated emails. When `DEBUG` is true, this defaults to `dev@calc.gsa.gov`.
```
toolness: I wonder if it should be something less officious-sounding like `dev@localhost` just to make sure that it doesn't confuse anyone if the email actually gets sent out?

I guess if we want to be really fancy we could dynamically construct the debug email using [`socket.gethostname()`](https://docs.python.org/2/library/socket.html#socket.gethostname) but that is probably going overboard.

toolness:
Noice, this looks great to me!  Feel free to merge.

FWIW, I'm a bit conflicted about turning every email send into a job queue thing. On the plus side, It will make the website more responsive to our users.  But on the downside, if redis/`rq`happens to be  down for some reason, emails won't get sent at all... Hmmmmmm.

jseppi:
Yeah, I have the same concerns about the async processing stuff. I'll change the email to `dev@localhost` or something -- that way MailChimp will definitely not accept the sender if it does accidentally make it in.

data_capture/tests/test_email.py
toolness: Hmm good question--[it looks like `unittest.mock` lets you do it](http://wesmckinney.com/blog/spying-with-python-mocks/), albeit after a bit of configurating.

data_capture/email.py
jseppi: Good point, changed!

data_capture/tests/test_email.py
jseppi: Good find! Changed to test using the suggestion in that article - works great!

jseppi: I just realized I was commenting about the mock stuff here, but the OP was about fixtures. Soz!

data_capture/tests/test_admin.py
```diff
@@ -76,31 +75,33 @@ def test_unapprove_ignores_unapproved_price_lists(self, mock):
...
+            msg_mock.assert_called_once_with(
+                "fake request",
+                messages.INFO,
+                '1 price list(s) have been unapproved and removed from CALC.'
+            )
+            em_monkey.assert_called_once_with(self.price_list)
+
         self.price_list.refresh_from_db()
         self.assertFalse(self.price_list.is_approved)
-        self.assertEqual(len(mail.outbox), 1)
```
jseppi: here is where the spy methods are useful. We're already testing the email method logic in `test_email.py` so it doesn't really make sense to (incompletely) double test it here IMO.

toolness: Yup totes--in fact that Sandi Metz talk specifically recommends *not* testing it, because then it just makes it harder to maintain + refactor your code in the future and stuff.

PR #641, opened 2016-08-31
factor out ajax_redirect to a method in ajaxform

jseppi:
Factored out an `ajax_redirect` method in `ajaxform.py` that returns a JSON response with the correct JSON object for use by `ajaxform.js`. Added a call to that method in the `handle_cancel` decorator instead of duplicating the JSON object creation logic.

Opened against #629.


toolness:
This looks good!

PR #644, opened 2016-09-01
Add excel table CSS styles

toolness:
This adds some excel table styles that we can use to show users accessible "screenshots" of what the price list we need them to upload looks like.

It's intended for use in the following section of @hbillings's [mockups](https://gsa.invisionapp.com/d/main/#/console/8513772/182332256/preview):

> screen shot 2016-09-01 at 8 18 39 am

And here's what the new styles look like in the styleguide:

> screen shot 2016-09-01 at 8 20 26 am

## Notes

* I don't think overriding the base table styles, as I'm doing in this PR, is very maintainable, as it starts specificity wars and such.  We might want to consider namespacing our default table styles instead of using them to define what every single table everywhere on our site should look like.

* I'm not distinguishing between `<thead>` and `<tbody>` in the excel tables; everything is assumed to be in a `<tbody>`, basically, which I think makes sense because an Excel spreadsheet is really just a grid of cells without a defined head. But also, reverting our default `<thead>` styling is a *huge pain*.  😣 

* I'm not a big fan of the `excel-wrapper` class here but it's the only way I can see of making really large tables scrollable, and also adding the padded border for the whole thing.

### DIFF
--- a/frontend/source/sass/components/_exceltables.scss
+++ b/frontend/source/sass/components/_exceltables.scss
@@ -1,0 +1,46 @@
+.excel-wrapper {
+  overflow: scroll;
+  padding: 1em;
+  border: 2px solid lightgray;
+}
+
+table.excel {
+  font-family: Arial, Helvetica, serif;
+  width: auto;
+  margin: 0;
+
+  tbody tr {
+    &:hover > * {
+      background: unset !important;
+    }
+
+    &:nth-child(odd) {
+      background: unset;
+    }
+  }
+
+  td {
+    min-width: 10em;
+    padding: 0;
+    border: 1px solid lightgray;
+  }
+}
+
+table.excel.excel-schedule-70 {
+  tr:first-child {
+    background: lightgray;
+
+    td {
+      vertical-align: bottom;
+    }
+  }
+
+  tr:not(:first-child) td:not(:first-child) {
+    font-style: italic;
+  }
+
+  td {
+    border-color: black;
+    text-align: center;
+  }
+}
--- a/frontend/source/sass/components/_exceltables.scss
+++ b/frontend/source/sass/components/_exceltables.scss
@@ -1,7 +1,8 @@
+@import '../base/variables';
+
 .excel-wrapper {
   overflow: scroll;
-  padding: 1em;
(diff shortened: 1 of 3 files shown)


frontend/source/sass/components/_exceltables.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+table.excel {
+  font-family: Arial, Helvetica, serif;
+  width: auto;
+  margin: 0;
+
+  tbody tr {
+    &:hover > * {
+      background: unset !important;
```
toolness: I didn't even know [`unset`](https://developer.mozilla.org/en-US/docs/Web/CSS/unset) existed!  Aside from its functionality, I also like its name, because it really drives home the notion that we're waging specificity wars here.

hbillings:
Okay, I love the idea of making these into tables instead of having them as screenshots! I do share your concerns about the table design currently being too opinionated. some thoughts:

- We should create a `hoverable` class for tables that should use highlighting.
- The `table` CSS isn't really in line with the way WDS does it -- it's all kind of all over the place. We should default to WDS table styles and namespace the rest, like you suggested. Maybe an easy way to start on this would be to add a `prices` or somesuch class to the existing table styles.
- re: `excel-wrapper`: We can drop the padded border -- that was more useful when the thing was an image, I think. If it's just a table, let's just make it look like a table. 

Could we:
- bump the text down in size a hair?
- add some padding to the cells?

toolness:
Yeah nice!

Happy to make those changes... Regarding defaulting to WDS table styles & namespacing the rest, as well as the `hoverable` class, should we do that as a separate PR since it'll involve poking at a lot of the other CSS and HTML across the rest of the site?

frontend/source/sass/components/_exceltables.scss
```diff
@@ -0,0 +1,47 @@
+@import '../base/variables';
+
+.excel-wrapper {
+  overflow: scroll;
+  font-size: $small-font-size;
```
toolness: Note that this line doesn't appear to do anything right now due to #671, but once we fix that, this CSS directive will kick in, and the font in the table will appear smaller.

Issue #636, opened 2016-08-31
Add flake8 to travis build


jseppi:
Because we are no longer using QuantifiedCode.

Probably via `ultratest --lintonly`, though as @toolness pointed out, this will make our eslint run twice, once by CodeClimate and once by `ultratest`.

toolness:
Another option, if we really want, is to get rid of `--lintonly` and instead allow `ultratest` to accept any number of positional arguments that specify the "slug" of the test to run.  For example:

* `manage.py ultratest flake8` would run only flake8.
* `manage.py ultratest flake8 py.test` would run flake8 and py.test.

And so on.

jseppi:
Oooo, that's pretty cool!

PR #647, opened 2016-09-01
Remove unused imports, run flake8 during travis

toolness:
This adds the `ultratest` functionality described in https://github.com/18F/calc/issues/636#issuecomment-243761792 and runs `python manage.py ultratest flake8` during travis, fixing #636:

screen shot 2016-09-01 at 1 48 45 pm

It also removes some unused imports from `decorators.py`, as `flake8` was actually failing 😄

### DIFF
--- a/meta/management/commands/ultratest.py
+++ b/meta/management/commands/ultratest.py
@@ -3,6 +3,7 @@ import tempfile
 import subprocess
 import djclick as click
 import xml.etree.ElementTree as etree
+from django.core.management.base import CommandError
 
 from docker_django_management import IS_RUNNING_IN_DOCKER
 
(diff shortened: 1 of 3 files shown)


meta/management/commands/ultratest.py
```diff
@@ -3,6 +3,7 @@
 import subprocess
 import djclick as click
 import xml.etree.ElementTree as etree
+from django.core.management.base import CommandError
```
toolness: I originally figured I should use `click.ClickException`, but it resulted in a big long grody traceback while `CommandError` did not, so I ended up going with the latter.

jseppi:
Sweet, works great. Is there some way you could add a message about specifying the test name and what the available options are when someone runs `ultratest --help`?

jseppi:
> It also removes some unused imports from decorators.py, as flake8 was actually failing

😞 oops

toolness:
Yeah, I would like to do that, though it's a bit hard right now because of the fact that the test types are defined within the actual command function, so it's hard to do introspection that way.  I will bust that up in a separate PR!

toolness:
Just filed #649!

Issue #646, opened 2016-09-01
Change no-js strategy to be more fault tolerant


toolness:
So I was reading Aaron Gustafson's [Interaction is an Enhancement](http://alistapart.com/article/interaction-is-an-enhancement) article and he noted that the vast majority of users who are running a site without JS don't actually have JS disabled, but rather have run into a variety of edge-cases where the browser's JS crashes or fails to load entirely.

This means that the first script tag on our pages, which follows the Modernizr approach of removing the `no-js` class from `<html>` and adding the `js` class, might not actually be best for progressive enhancement, as we haven't actually yet determined if the rest of the JS on the page has loaded or not yet.

This seems like a bit of a conundrum, since we don't want to show users a "flash of no-js content" before all the JS has been loaded (that's why we remove the `no-js` class as soon as possible).

Anyways, just wanted to document this here as a thing to think about.  Probably not super urgent.

toolness:
Oh, this was fixed by #759.

PR #654, opened 2016-09-01
Implement safe mode opt-in if/when JS errors occur

toolness:
Ok, this is an attempt to address #646 by offering to opt-in users to no-JS mode if/when JS errors occur.

## How it works

1. Early in every page's HTML, we have an inline `<script>` that sets up an `onerror` handler.  If it's ever triggered, we display the following alert at the bottom of the current page's viewport (using `display: fixed`):

  screen shot 2016-09-01 at 3 36 49 pm

2. If the user clicks the **Activate Safe Mode** button, we store a flag in the user's session to indicate that they've disabled JS, and reload the page.

3. All page loads now omit `<script>` tags, effectively behaving as though the user has disabled JS in their browser.  A UI at the bottom of the page allows them to opt back out of safe mode.

## Notes

* We don't necessarily have to disable *all* the JS on a page.  Some well-tested libraries that have a very limited scope, no dependencies, and known fault tolerance characteristics, such as [prismjs](http://prismjs.com/) in the style guide, might be worth loading at all times.

* We should be doing logging and metrics not only on what errors occur, but also on how the user responds to the UI we present them with.

* Consider renaming "safe mode" to "compatibility mode" or even not giving the mode a name, to reduce cognitive load.

* We might want to run the copy by the writing lab, or just tweak it ourselves, to make sure it'll make sense to non-technical users.  (It's for this reason that I decided to call the functionality "safe mode", a term users might recognize from Windows and Mac OS, rather than "disable JS". But I'm sure more improvements can be made!) **Update:** Filed as https://github.com/18F/writing-lab/issues/197.

## To do

- [x] Add tests.
- [x] While in safe mode, the background of the "deactivate safe mode" UI is still red, which we may want to change to stand out less (and not be indicative of some kind of error state).
- [x] We might want to selectively disable only JS that we know won't break the site: i.e., all JS added to `data_capture` and `styleguide`--but not `data_explorer`, as that won't work without JS.
- [x] Similarly, we should also only offer to opt users into safe mode in those pages that we know will actually work with JS disabled.
- [x] Rename the files from `enable_js` to `safe_mode` since that's what we're telling users it's called. Also, nouns are good.


frontend/source/sass/components/_enablejs.scss
```diff
@@ -0,0 +1,22 @@
...
+  }
+
+  button {
+    margin: 0 0 0 1em;
+    float: right;
+  }
+}
+
+aside.js.js-enable {
+  background-color: rgba(0, 0, 0, 0.8);
```
toolness: Should probably convert this to use a declared color variable.

frontend/templates/frontend/safe_mode/script_tag.html
```diff
@@ -0,0 +1,27 @@
+<script>
+var globalErrorCount = 0;
+
+function showEnableSafeModeUI() {
+  var el = document.getElementById('safe-mode-enable');
+
+  if (el && el.hasAttribute('style')) {
+    el.removeAttribute('style');
+    el.focus();
```
toolness: Hmm, so while this is good for accessibility, I'm concerned about its effect on "false positives", e.g. when an `onerror` event occurs that doesn't actually impact user experience.  It means, for example, that if the user is in the middle of typing in a field, their keystrokes will stop going there, at which point they may actually start thinking the page *is* broken when it's not.

toolness:
I wanted the safe mode JS to be as conservative as possible (so it'd work on any browser), and also as portable to other projects as possible (so it wouldn't rely on an ES6 build toolchain).  I also wanted it to load quickly, so it's an inline script instead of loaded via a `src` attribute.

All this made writing a QUnit test suite a bit challenging and unusual.

I think this is ready for review @jseppi.  The writing lab is still working on better copy but I think we should merge this for now--it's reasonably friendly as-is, and hopefully users will see it very rarely anyhow!

data_explorer/templates/base.html
```diff
@@ -10,7 +10,10 @@
     <script src="{% static 'frontend/js/vendor/aight.v2.min.js' %}"></script>
     <script src="{% static 'frontend/js/vendor/history.min.js' %}"></script>
     <![endif]-->
+    {% include 'frontend/safe_mode/script_tag.html' %}
+    {% if not is_safe_mode_enabled %}
     <script>(function(e,t,n){var r=e.querySelectorAll("html")[0];r.className=r.className.replace(/(^|\s)no-js(\s|$)/,"$1js$2")})(document,window,0);</script>
+    {% endif %}
```
toolness: I'm going to get rid of this Modernizr bit in a future PR.  The only thing that relies on it right now is the upload widget, I believe, and it violates the [Apply no style before its time](http://adaptivewebdesign.info/1st-edition/read/chapter-4.html) maxim of progressive enhancement (in particular, it could be the case that this JS executes and applies the `js` style to the page, yet the data capture bundle fails to be retrieved for some reason, thus resulting in a broken page).

jseppi:
I really like this!

I do have some concerns about having the safe-mode stuff on the data_explorer pages, however, given that the front page of CALC won't work at all without JS, as you noted. I know the main JS is not being removed when `is_safe_mode_enabled`, but Google Analytics, DAP, and ethnio are, which are probably very unlikely to be the cause of uncaught/page-breaking errors.

My concern, which I believe is something you perhaps brought up in a hangout, is that if the front page JS has an error and the user enables safe mode...then nothing gets "fixed" because the script causing the error is not (and cannot) be removed. I think that would be confusing to the user, perhaps even more so than just broken functionality.

toolness:
Oh, that actually shouldn't happen, I don't think, because the safe mode UI isn't being included on the data explorer pages, so the user will never see the "activate safe mode" requests on them?

Unless you mean that the code causing the error is actually something being thrown from the common bundle, which is always included in data capture pages regardless of the safe mode state?  While that's true, I'm not actually sure if that common bundle code will ever actually get called, because I think that bundle just contains `hourglass.js`, which is a bunch of utility functions (which will never get called b/c the main data capture bundle isn't being loaded).

Does that make sense?

I guess if we want, I could further refactor things so that the common script tags between the explorer and the data capture apps are only disabled when safe mode is enabled *and* the page being rendered is in the data capture app?

frontend/templates/frontend/safe_mode/script.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable */
```
jseppi: I wonder if we should use the `"airbnb-base/legacy"` config for non-ES6 script files. I think it would just involve putting another `.eslintrc` file in this directory.
Not a big deal of course, this is a super simple script and a bit of a special case.
Eh, don't worry about it. :)

jseppi:
Oooo, I missed that the UI wasn't on the data explorer pages, my bad.

So, could a user enable safe mode in Data Capture, then go to Data Explorer and not be able to disable it? It wouldn't impact their experience, but we'd miss out on capturing analytics for hopefully an extreme minority of users.

toolness:
That's correct!  Yeah, I can revert the disabling of analytics for safe mode if you want, actually--I just wanted to make it "as close to disabling JS as possible" I guess.

jseppi:
I will defer to your judgement 🙇  I hope and expect that this situation is so rare that it won't be a real issue. And if it is (though I'm not sure how we'd find out it is), we could change what is disabled later.

toolness:
Woot sounds good!  I guess I will just leave it as-is for now and we can revisit later.

PR #655, opened 2016-09-01
Add support for periodic tasks and send admin approval reminder email

### LINKED ISSUE #648: Send Email to CALC Admin periodically
CALC should send a reminder email to the CALC Admin every [NUM_DAYS] if there are  [count > THRESHOLD] submitted price lists that have not been approved.

jseppi:
This PR adds support for using [rq-scheduler](https://github.com/ui/rq-scheduler/) (via [django_rq](https://github.com/ui/django-rq)) to set up periodic tasks. It also adds a periodic task (every Monday) to send an email to admins (Django superusers) if the number of unapproved price lists is >= 10.

Closes #648 

To Do:
- [x] figure out how to add scheduled tasks
- [x] schedule weekly reminder email for calc admin
- [x] test reminder email
- [x] test custom AppConfig
- [x] test cron-scheduled job somehow
- [x] figure out who to send emails to. ans: all superusers (who have an email address)
- [x] figure out what to do about the `RemovedInDjango19Warning`


jseppi:
Hrm, getting this warning on start:

```sh
/calc/data_capture/models.py:7: RemovedInDjango19Warning: Model class data_capture.models.SubmittedPriceList doesn't declare an explicit app_label and either isn't in an application in INSTALLED_APPS or else was imported before its application was loaded. This will no longer be supported in Django 1.9.
```

I'm not sure of the implications of using a custom `AppConfig`. @toolness ?

jseppi:
Hrm, maybe a better way than the custom `AppConfig` would be a management command that grabs the scheduler and queues up the periodic tasks. I think that could work...

jseppi:
I'm still getting that warning mentioned in https://github.com/18F/calc/pull/655#issuecomment-244418698 :\

hourglass/settings.py
```diff
@@ -270,6 +270,15 @@
 
 LOGIN_REDIRECT_URL = '/'
 
+# Add the custom rq_scheduler version of the data_capture cap
+# when IS_RQ_SCHEDULER is in the env
+IS_RQ_SCHEDULER = 'IS_RQ_SCHEDULER' in os.environ
+if IS_RQ_SCHEDULER:
+    # First remove the 'regular' data_capture app
+    INSTALLED_APPS = tuple(x for x in INSTALLED_APPS
```
jseppi: Have to recreate the`tuple` of `INSTALLED_APPS` without `'data_capture'` since I don't think you can remove an item from a `tuple`

jseppi:
I think I can test the functionality of custom `AppConfig`'s `ready()` method by grabbing an instance of the scheduler and making sure it has a single cron job

jseppi:
woot, think I got the testing down now

ready for review @toolness

jseppi:
Got the warning figured out and cleared. It was because I was importing `periodic_tasks`, which itself imports `SubmittedPriceList` before the app was "ready". I just moved the import into the custom `AppConfig`'s `ready()` method.

data_capture/apps.py
```diff
@@ -0,0 +1,33 @@
+import logging
+from django.apps import AppConfig
+
+import django_rq
+
+logger = logging.getLogger(__name__)
+
+
+class DataCaptureSchedulerApp(AppConfig):
```
toolness: Maybe we should add a docstring here mentioning that this app config should only be used by the scheduler process, which is (notably) a singleton across the entire application?  (Otherwise some of the logic in `ready()`, such as cancelling any jobs already in the scheduler, don't make sense.)

Another thing that might make this clearer may be to create a second config called `DefaultApp` that represents the configuration the process is usually using.  (In fact, I thought it wasn't possible to mix the old and the `AppConfig` style of doing things without resorting to `default_app_config`, but I guess I'm wrong!)

OTOH, maybe it's obvious and I'm just not used to scheduler stuff. :)

hourglass/settings.py
toolness: I'm curious if this might be easier / more maintainable if we defined a `DATA_CAPTURE_APP_CONFIG` constant that is part of `INSTALLED_APPS`, and defaulted to `data_capture` but was changed to `data_capture.apps.DataCaptureSchedulerApp` if `IS_RQ_SCHEDULER` is true?

Or even better, if we actually make a `data_capture.apps.DefaultApp`, we could have `DATA_CAPTURE_APP_CONFIG` be just the class name, and insert `'data_capture.apps.{}'.format(DATA_CAPTURE_APP_CONFIG)`  into `INSTALLED_APPS`...

data_capture/apps.py
```diff
@@ -0,0 +1,33 @@
...
+    name = 'data_capture'
+    verbose_name = 'CALC Data Capture RQ Scheduler'
+
+    # every Monday at noon, but scheduler uses UTC so it will be 5AM Pacific
+    admin_reminder_cron = '* 12 * * MON'
```
toolness: Blerg, I find cron syntax really confusing/hard to remember, but after looking at `rq-scheduler` it doesn't seem like there's a more humane solution available.

jseppi: Heh, yeah, I have to look it up every time I see it :\

data_capture/apps.py
```diff
@@ -0,0 +1,48 @@
...
+        scheduler = django_rq.get_scheduler(self.rq_queue_name)
+
+        # Cancel any jobs already in the scheduler. Jobs can be left dangling
+        # from any restart of this custom AppConfig.
+        for job in scheduler.get_jobs():
+            scheduler.cancel(job)
+
+        # Add cron-type job to send a reminder email to the CALC admins
+        # about approving price lists
+        logger.info('Adding send_admin_approval_reminder_email job on cron \
```
jseppi: I'm not seeing any of these logger messages printed out in docker-compose logs. Am I not using logging correctly?

toolness: Unfortunately, logging in Django/python is really confusing.

The main thing to keep in mind is the name you've given your logger; in this case, it's set to `__name__` in line 6, which means it's probably `data_capture.apps`.

This means that you need to edit your `LOGGING['loggers']` in `settings.py` to have an entry for `data_capture.apps`.

I guess it might make more sense to call the logger something like `rq_scheduler` or something, since that's what's doing the logging?  But in that case it should only be defined in `ready()` or as a class member of `DataCaptureSchedulerApp` I guess.

jseppi:
I wonder if there is a way we could mock the current time so that we could run a test that made sure the periodic_job processing happened at the scheduled cron time.

Quick search led me to http://stackoverflow.com/questions/1042900/django-unit-testing-with-date-time-based-objects

toolness:
If that's a chore, I suspect it's probably enough just to introspect into `rq`/`rq_scheduler` and make sure that there's a task in the queue after app startup?  Or is that already being done in tests?

jseppi:
Yes, the tests are already checking that the correct job is getting added and that the cron string is set to the specific every Monday schedule: https://github.com/18F/calc/pull/655/files#diff-5efc5ebf4321b31650a4759d35d001ffR23

toolness:
Noice! I think that's plenty enuf--if we add a test that mocks time and ensures the job actually gets called, we'll really be testing integration with `rq_scheduler`, I suppose, which doesn't seem that necessary unless you think they'll start interpreting cron strings differently in the future or something...

jseppi:
Heh, I should hope they don't start doing that :)

toolness:
Coool, this looks good to me!

PR #656, opened 2016-09-02
Document web components.

toolness:
This just adds some documentation for our web components. I don't want to cause doc rot down the road, so I've kept this fairly minimal for now, and only focus on what's most important and not totally obvious from reading the code.

### DIFF
--- a/frontend/source/js/tests/ajaxform_tests.js
+++ b/frontend/source/js/tests/ajaxform_tests.js
@@ -172,6 +172,14 @@ advancedTest('submit triggers ajax w/ form data', (assert, s) => {
   $(s.ajaxform).submit();
 });
 
+advancedTest('submitted XHR has X-Requested-With header', (assert, s) => {
+  s.setFile(createBlob('hello there'));
+  $(s.ajaxform).submit();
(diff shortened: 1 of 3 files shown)


frontend/source/js/tests/ajaxform_tests.js
```diff
@@ -172,6 +172,14 @@ advancedTest('submit triggers ajax w/ form data', (assert, s) => {
   $(s.ajaxform).submit();
 });
 
+advancedTest('submitted XHR has X-Requested-With header', (assert, s) => {
+  s.setFile(createBlob('hello there'));
+  $(s.ajaxform).submit();
+
+  const req = server.requests[0];
+  assert.equal(req.requestHeaders['X-Requested-With'], 'XMLHttpRequest');
+});
```
toolness: Just realized while writing the docs that we should really test this.  It's being added by jQuery, not us, but it's crucial to the client-server interaction so we ought to test it, which will be especially important if we ever decide to decouple the component from jQuery or anything.

jseppi:
Looks great 👍

PR #657, opened 2016-09-02
Add <expandable-area> web component.

### LINKED ISSUE #645: Create expander component
@hbillings's latest [mockps](https://gsa.invisionapp.com/d/main/#/console/8513772/182332256/preview) imply the need for some kind of expander component:

> screen shot 2016-09-01 at 9 03 36 am

For accessibility, we can probably use [`aria-expanded`](http://www.3needs.org/en/testing/code/aria-expanded.html) here, it seems to have decent support.

For progressive enhancement, we should probably default the expander to be in an expanded state, and only expand it if JS is available.

toolness:
This fixes #645 by adding a ~~`<collapsible-area>`~~ `<expandable-area>` web component:

> screen shot 2016-09-02 at 10 12 11 am

Implementation is inspired by Aaron Gustafson's [Progressive Enhancement with JavaScript](http://adaptivewebdesign.info/1st-edition/read/chapter-4.html), specifically his three maxims:

1. Make sure all content is accessible and all necessary tasks can be completed without JavaScript turned on.
2. Use JavaScript to generate any additional markup it needs.
3. Apply no style before its time.

The CSS is inspired by techniques illustrated in Heydon Pickering's [Effortless Style](http://slides.com/heydon/effortless-style#/) presentation.

## To do

- [x] Add tests.
- [x] Rename to `<expandable-area>`, since the component starts out collapsed.

### DIFF
--- a/frontend/source/js/data-capture/collapsible-area.js
+++ b/frontend/source/js/data-capture/collapsible-area.js
@@ -4,6 +4,8 @@ import 'document-register-element';
 
 import * as supports from './feature-detection';
 
+import { dispatchBubbly } from './custom-event';
+
 const KEY_SPACE = 32;
 const KEY_ENTER = 13;
 
@@ -20,9 +22,8 @@ class CollapsibleArea extends window.HTMLElement {
 
     this.expander = this.firstElementChild;
 
-    this.isUpgraded = this.expander &&
-                      'classList' in this &&
-                      !supports.isForciblyDegraded(this);
+    this.isUpgraded = Boolean(this.expander &&
+                              !supports.isForciblyDegraded(this));
 
     if (this.isUpgraded) {
       this.expander.setAttribute('aria-expanded', 'false');
@@ -34,6 +35,7 @@ class CollapsibleArea extends window.HTMLElement {
           this.toggle();
         }
       }, true);
+      dispatchBubbly(this, 'collapsibleareaready');
     }
   }
 
@@ -53,3 +55,5 @@ CollapsibleArea.prototype.SOURCE_FILENAME = __filename;
 document.registerElement('collapsible-area', {
   prototype: CollapsibleArea.prototype,
 });
+
+module.exports = CollapsibleArea;
--- a/frontend/source/js/data-capture/collapsible-area.js
+++ b/frontend/source/js/data-capture/collapsible-area.js
@@ -1,0 +1,55 @@
+/* global window, document */
+
+import 'document-register-element';
+
+import * as supports from './feature-detection';
+
+const KEY_SPACE = 32;
+const KEY_ENTER = 13;
+
+/**
+ * CollapsibleArea represents a <collapsible-area> web component.
+ */
+
+class CollapsibleArea extends window.HTMLElement {
+  attachedCallback() {
+    if ('isUpgraded' in this) {
+      // We've already been attached.
+      return;
+    }
+
+    this.expander = this.firstElementChild;
+
+    this.isUpgraded = this.expander &&
+                      'classList' in this &&
+                      !supports.isForciblyDegraded(this);
+
+    if (this.isUpgraded) {
+      this.expander.setAttribute('aria-expanded', 'false');
+      this.expander.setAttribute('role', 'button');
+      this.expander.setAttribute('tabindex', '0');
+      this.expander.addEventListener('click', this.toggle.bind(this), false);
+      this.expander.addEventListener('keyup', e => {
+        if (e.keyCode === KEY_SPACE || e.keyCode === KEY_ENTER) {
+          this.toggle();
+        }
+      }, true);
+    }
+  }
+
+  toggle() {
+    if (!this.isUpgraded) {
+      return;
+    }
+
+    const newVal = this.expander.getAttribute('aria-expanded') === 'true'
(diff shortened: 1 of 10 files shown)


frontend/source/js/tests/collapsible-area_tests.js
```diff
@@ -0,0 +1,84 @@
...
+function makeArea() {
+  const ca = document.createElement('collapsible-area');
+  const h1 = document.createElement('h1');
+  ca.appendChild(h1);
+  ca.attachedCallback();
+
+  return { ca, h1 };
+}
+
+QUnit.test('is upgraded whenc containing at least one element', assert => {
```
jseppi: typo on `whenc`

styleguide/templates/styleguide.html
```diff
@@ -405,6 +405,44 @@ <h3 class="example-heading">Example</h3>
...
 
+  {% guide_section "Collapsible Area" %}
+
+  <p>
+    The {% webcomponent '<collapsible-area>' %} web component makes it possible
+    to progressively enhance an area to be collapsible. The first child
+    element is assumed to be the expander (which is dynamically given the
+    <a href="https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role">button role</a>) while all of its siblings are
+    the collapsible content.
+  </p>
```
jseppi: Looks like you can specify `aria-expanded` to set the default at expanded or collapsed, is that right? If so, might be good to document.

toolness: Oh, hrm, you actually can't--right now if you do that, the attribute will just be overwritten to `false` when the web component upgrades.

I can add some logic that checks for the existence of the attribute and doesn't touch it if so, though... But if we do that, we should also emphasize that the `aria-expanded` attribute needs to be put on the first child element of the component, *not* the component itself.  If we want it to go on the component itself, it might be better to support a different markup API like e.g. `<collapsible-area initial-state="collapsed">`.

toolness: I guess I would recommend just forcing this to be collapsed for now, and adding functionality to start it out as un-collapsed later if we find a use case for it?

jseppi: I am fine with it starting collapsed. Maybe it should be called `expandable-area` in that case?

toolness: Hmm, but if we allow the default state to be configurable later, does it matter?  Also, it's always expanded when JS is disabled, which I guess is why I called it `collapsible`... since it's collapsible only if JS is enabled... hmm, I guess I am fine either way really, mostly it's just more work to rename it to `expandable-area` :P  But I'll do it if it's really confusing as `collapsible-area`!

frontend/source/js/data-capture/collapsible-area.js
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+class CollapsibleArea extends window.HTMLElement {
+  attachedCallback() {
+    if ('isUpgraded' in this) {
+      // We've already been attached.
+      return;
+    }
+
+    this.expander = this.firstElementChild;
```
toolness: Just FYI, this will always evaluate to `undefined` on IE6-8 because `firstElementChild` wasn't added to IE until version 9.  However, IMO this is fine because the area will just not be progressively enhanced, rather than broken.

jseppi: Yeah, that seems fine. Maybe add a note about it?

styleguide/templates/styleguide.html
jseppi: I don't feel _too_ strongly about it, but I would think that `collapsible` means it starts expanded and could then be collapsed through user interaction.

frontend/source/js/data-capture/collapsible-area.js
toolness: Okydoky.

styleguide/templates/styleguide.html
toolness: Okydoke, I will rename!

toolness:
Ok, did the rename--I think this is ready for review!

frontend/source/js/data-capture/expandable-area.js
```diff
@@ -0,0 +1,62 @@
...
+      this.expander.setAttribute('aria-expanded', 'false');
+      this.expander.setAttribute('role', 'button');
+      this.expander.setAttribute('tabindex', '0');
+      this.expander.onclick = this.toggle.bind(this);
+      this.expander.onkeyup = e => {
+        if (e.keyCode === KEY_SPACE || e.keyCode === KEY_ENTER) {
+          this.toggle();
+        }
+      };
+      dispatchBubbly(this, 'expandableareaready');
```
jseppi: Would it be good to use (fake) namespaced events, such as `'expandablearea:ready`? It make them slightly easier to read.

jseppi:
👍 Looks great!

frontend/source/js/data-capture/expandable-area.js
toolness: Good idea, I've filed this as #673 so we can take care of all the components in one fell swoop (and keep things consistent too).

PR #659, opened 2016-09-05
[WIP] Classes and styling for numeric cells in tables

### LINKED ISSUE #652: Right-align and monospace prices in approval table
For scannability

jseppi:
Work-in-progress

Closes #652


jseppi:
Actually the more I think about this the more I think we should be "manually" constructing the table of price list rows instead of iterating through the model fields. Constructing it in HTML will give us a lot more control over things like number currency styling as well as classes per column (like to specify min/max-widths).

PR #660, opened 2016-09-05
Price List table row templates and styling

### LINKED ISSUE #652: Right-align and monospace prices in approval table

jseppi:
Work-in-progress. I've gone with the approach of making the price list row tables explicitly defined instead of just iterating through their rows and fields. By being explicit we can better control the resulting HTML, though this comes at the expense of having to make the templates and keep up with any model changes that might happen. I tried exploring more "automatic" methods, but none of them really pleased me.

Closes #652 

Opened against #583.

To Dos:
- [x] tests - if the `to_table` and `to_error_table` methods look good, then I'll add tests for them.
- [x] styling review by @hbillings 

Some screenshots:

screen shot 2016-09-06 at 10 24 17 am

screen shot 2016-09-06 at 10 24 59 am

One thing that gets a little funky is if a `td class="number"` has an error because the bad value is styled like a number even if it isn't one, such as in the following screenshot:

screen shot 2016-09-06 at 10 26 20 am


data_capture/schedules/base.py
```diff
@@ -39,6 +39,22 @@ def serialize(self):
 
         raise NotImplementedError()
 
+    def to_table(self):
```
jseppi: Rather than adding branch-logic in the templates, I opted to instead add methods on the PriceList classes for rendering `to_table` and `to_error_table`. I think this works well, but happy to hear opinions/alternatives.

frontend/source/sass/base/_variables.scss
```diff
@@ -65,6 +65,7 @@ $base-line-height:    1.4; // WDS uses 1.5, but our base font is .2rem smaller
 $heading-line-height: 1.3;
 
 $font-family-main: 'Lato', 'Helvetica', 'Arial', sans-serif;
+$font-family-mono: 'Courier New', monospace;
```
jseppi: Added a new font-family value for monospace fonts. Not sure which monospace fonts are preferred though. @hbillings?

frontend/source/sass/components/_forms.scss
```diff
@@ -52,6 +52,8 @@ form ul {
   margin-bottom: 0;
   margin-left: -10px;
   padding-left: 10px;
+  text-align: left;
+  font-family: $font-family-main;
```
jseppi: blech, I added these two rules because otherwise the `.number` class on the parent `td` element would cause the `.errorlist` items to be right-aligned and in the monotype font.

frontend/source/sass/components/_tables.scss
```diff
@@ -106,6 +106,28 @@ td.column-exclude a {
...
+}
+
+td.number {
+  font-family: $font-family-mono;
+  text-align: right;
+}
+
+td > .currency-symbol {
+  color: $color-gray-medium;
+  width: 50%;
```
jseppi: Did this so that there would be some space between the number and the `currency-symbol` but without the `currency-symbol` being all the way to the left. Not super sure about this. @hbillings?

frontend/source/sass/components/_tables.scss
```diff
@@ -106,6 +106,28 @@ td.column-exclude a {
...
+  float: left;
+}
+
+.price-list-table {
+  th, td {
+    max-width: 1em;
```
jseppi: Added the `max-width` rule to make the table columns render to more equally-spaced widths. Table styling stinks.

frontend/source/sass/base/_variables.scss
toolness: On `#g-content` I noticed someone linked to [Input](http://input.fontbureau.com/info/), but I'm not sure if they were recommending it for the content writers to use, or for our websites.

Also did some quick [digging on Slack](https://18f.slack.com/archives/ffd-uswds/p1434129481001453) and found Maya mentioning the use of [Source Code Pro](http://blog.typekit.com/2012/09/24/source-code-pro/), which is apparently in the same family as Source Sans Pro (which USWDS uses along w/ Merriwether).

data_capture/schedules/fake_schedule.py
```diff
@@ -68,6 +69,14 @@ def add_to_price_list(self, price_list):
     def serialize(self):
         return self.rows
 
+    def to_table(self):
+        return render_to_string('data_capture/price_list/tables/fake.html',
+                                {'rows': self.valid_rows})
+
+    def to_error_table(self):
+        return render_to_string('data_capture/price_list/tables/fake.html',
```
toolness: Maybe call the filename `fake_schedule.html` just to be consistent w/ the naming of the Python file?

data_capture/schedules/base.py
toolness: NOICE

frontend/source/sass/base/_variables.scss
jseppi: You mean just specifying them and hoping the users have those fonts on their machines? Including additional webfonts seems like a not great idea for such a small amount of styled text.

toolness: Er, sorry, yeah, that makes sense!

Though that said, I *think* that browsers load webfonts on-demand, so if a page doesn't use any monospaced text, the browser won't retrieve it.  I am not sure though.

toolness: It looks like GitHub on mac uses `Consolas, "Liberation Mono", Menlo, Courier, monospace`, if that's useful.

jseppi: hehe, nice! I'll use that list :)

jseppi:
Added tests, though they are pretty basic -- just testing that the methods return string values.

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -19,42 +19,19 @@
...
-  </table>
+    {% with total=rows|length %}
+      <h3>{{ total }} row{{ total|pluralize:" has,s have" }} errors</h3>
+
+      <p>
+        The row{{ total|pluralize }} below appear{{ total|pluralize:"s," }} to be
+        invalid and <strong>will be discarded</strong> when you upload your
+        price list.
+        If you'd like, you may correct {{ total|pluralize:"this row,these rows" }}
+        in your original spreadsheet and <a href="{% url 'data_capture:step_1' %}">try uploading it again</a>.
```
hbillings: This should probably actually go to step 3 once #583 is merged.

frontend/source/sass/components/_tables.scss
hbillings: I'm sorta wondering if we want to render this with a `td.number::after { content: '$'; }` sort of a rule. That would remove any sort of weird responsive issues that setting a width like this might create. My only concern is screen readers, but I'm not sure they'd really parse this as `n` dollars -- @atul, any thoughts on that?

jseppi: Will using `content` keep things aligning well? As it is now, the `$`s and actual numbers are aligned with each other (as shown in the  screenshots), as opposed to looking like:

```
$ 123.45
  $ 2.50
```

toolness: Hmm, I think you just notified someone who is not me 😁   That's a good question--I had actually assumed that screen readers wouldn't process CSS `content:` rules at all, but [apparently I am mistaken](http://www.456bereastreet.com/archive/201205/css_generated_content_and_screen_readers/), which is cool!  While screen readers may or may not actually parse it as "`n` dollars", they should at least read out the actual dollar sign symbol, and combined with the fact that users can always read the table heading to get a better context that the number they're looking at is a price, I think we're probably good!

data_capture/templates/data_capture/price_list/step_4.html
jseppi: Oops, this PR is against the branch for #583. I'll fix real quick.

frontend/source/sass/components/_tables.scss
jseppi: I just tried this locally and here is what it looks like, which illustrates what I meant about the currency symbols lining up:

screen shot 2016-09-07 at 4 47 27 pm

also I had to use `td.number::before { content: '$'; }` instead of `::after`

hbillings: @jseppi I think you should still be able to float the `content` left (though you'd want to use `::before` and not `::after`, derp)! You just won't need to specify a width, AFAIK.

This is pretty nitpicky, though, and we can always wait and see if having the element in the table becomes a responsive issue later.

jseppi: Yeah, tried `float:left` on it, but that moves it too far to the left in my opinion, unless I add something like `width: 50%` also, but that kinda brings me very nearly back what I originally had. I just don't know what's best and I don't know what to do and ~barf~

hbillings: There is no emoji for "I both empathize with and find amusement in the pain that is tables." Just pass on it -- I may poke around with it after you merge to see if I can get it working the way I think it should work, but at this point I think it's not worth spending more time on!

jseppi: 🙇

Issue #651, opened 2016-09-01
Form Button Row - Swap cancel for next button


hbillings:
Cancel is too prominent and Next is too easy to miss.
Consider:
- Put Next where Cancel is currently
- Put label for Next on right side
- Bold Next
- Try `primary-alt-blue`
- Turn Cancel into a link and place it somewhere with less prominence

jseppi:
I assume this is in regards to the Form Button Row buttons?

hbillings:
@seppi yeah, I coulda been clearer about that, sorry!

jseppi:
> Turn Cancel into a link and place it somewhere with less prominence

It doesn't appear the WDS have a "link-style" button class _à la_ `.btn-link` in Bootstrap (ref: http://v4-alpha.getbootstrap.com/components/buttons/#button-tags). We have to use a `<button>` element for cancel since it issues a `POST` request to clear session.

toolness:
Fixed by #740!

PR #661, opened 2016-09-05
add class for button-link

jseppi:
Adds a new `.button-link` class for buttons to make them look like links. This has precedent in other frameworks, like Bootstrap.

It will be useful for the Cancel button styling described in https://github.com/18F/calc/issues/651

Screenshot from Style Guide:

screen shot 2016-09-05 at 3 38 08 pm


frontend/source/sass/components/_buttons.scss
```diff
@@ -86,3 +86,34 @@ input.button-secondary[type="button"] {
...
 }
+
+button.button-link,
+input.button-link[type="submit"],
+input.button-link[type="reset"],
+input.button-link[type="button"] {
+
+  border: none;
+  padding-top: 0;
+  padding-bottom: 0;
```
jseppi: I left the left and padding alone, but I could also see just setting `padding: 0;` here.

toolness:
Huh, it feels a bit odd to make something that looks like a link but changes server state (and potentially even loses user data), but I suppose it's no less weird than making a link that looks like a button. 😁

toolness:
Would it be helpful to use [`appearance: none`](https://css-tricks.com/almanac/properties/a/appearance/) here for browsers that support it, just to make sure other weird native platform CSS properties don't bleed through?

jseppi:
Neat, didn't know about `appearance`!

jseppi:
Added `appearance` rules. Will wait for 👍 from @hbillings.

hbillings:
LGTM!

PR #662, opened 2016-09-05
Is staff decorator

### LINKED ISSUE #642: Better decorator for is_staff
Should check if user is logged in and if `user.is_staff=True`. If not logged in, redirect to login flow, then back to the originally requested view. Once (or if already) logged in, check if is_staff. If not, return not authorized error+page, otherwise let through to the view.

It can go in `decorators.py` once #629 is merged.

jseppi:
Closes #642 

This PR implements a better decorator for requiring an authenticated `is_staff` user. 

If not authenticated, then the user is redirect to login. If the user is authenticated and `is_staff=True`, then the user is permitted to access the view. If the user is authenticated but `is_staff=False`, then a `403` error is returned.


toolness:
Nice! This looks good.

What does our default 403 page look like?

Maybe file a separate issue to create a human-friendly 403 page if the current one is bad?

jseppi:
Yeah, it looks bad. It has no styling or content or theme or anything.

PR #669, opened 2016-09-06
Custom 403 template

### LINKED ISSUE #664: Make human-friendly 403 page
Ref #662

jseppi:
- Adds a template for 403 error pages. Closes #664.
- Fixes up 404 and 500 error page template code formatting.


data_explorer/templates/403.html
```diff
@@ -0,0 +1,14 @@
...
+{% load staticfiles %}
+
+{% block head %}
+{% endblock %}
+
+{% block body %}
+<div class="container">
+  <h1>Permission Denied</h1>
+
+  <p>We're sorry, but your account does not have permissions to view the requested page.</p>
```
toolness: Should we use `permissions` (plural) or `permission` (singular)?  From a technical standpoint, I think that plural is most accurate, but end-users might find the singular more understandable, I'm not sure.

toolness: Also, should we give the user some kind of recourse if they think that they *should* be able to access the page, e.g. "if you think this is a mistake, email `SOMEBODY`?"

If we're concerned about revealing email addresses to just anybody, I suppose we could check to see if `user.is_staff` is true, and if so, at least we actually know that we can trust the user, and thereby give them more specific information on how to get help... Anyways, we could always address that in a separate PR too.

jseppi: Yeah, when I first made this page I did include a line about emailing, but then realized we don't really have a general email for that, other than the `calc@gsa.gov`. I think we should leave it out for now :\

data_explorer/templates/403.html
```diff
@@ -0,0 +1,14 @@
+{% extends "base.html" %}
+{% load staticfiles %}
+
+{% block head %}
+{% endblock %}
+
+{% block body %}
+<div class="container">
+  <h1>Permission Denied</h1>
```
toolness: Hmm, if we want to be consistent with the other error pages, perhaps the text here should be `InCALCulable!` 😁

toolness: Which I always say in my mind like this:

toolness:
This looks good to me! 👍

PR #674, opened 2016-09-06
Never show django login

### LINKED ISSUE #670: Default Django Admin Login page is shown for logged-in, non-staff user
Just found some bizarre behavior that allows a user to get to the default Django Admin login page (the one with username and password).

Steps to reproduce:
1. Login as a non-staff user
2. Visit `/admin`
3. You will be redirected to `/admin/login/?next=/admin/`, which will show the default Admin login page.

This should be disabled because we never want password login since we are using cloud.gov's UAA provider for all authentication.

jseppi:
In this PR we replace the use of `login_required` with our own `staff_login_required` decorator to protect the admin site login. The reason for doing this is logged in #670: basically the `login_required` decorator still allows a logged-in, but `is_staff=False`, user to get redirected to the default Django admin login screen with username and password fields, which we never want to happen. Our own `staff_login_required` decorator, however, will instead raise a `PermissionDenied` error when a non-staff logged-in user attempts to visit the decorated view, which is want we want.

This also adds tests to make sure the intended behavior is happening.

Closes #670


hourglass/tests.py
```diff
@@ -189,3 +190,46 @@ def test_returns_whitelisted_ips_list(self):
...
+        )
+        user.is_staff = False
+        user.save()
+        logged_in = self.client.login(
+            username=user.username, password='foo')
+        self.assertTrue(logged_in)
+        res = self.client.get('/admin/')
+        self.assertEqual(res.status_code, 302)
+        res = self.client.get(res['Location'])
+        self.assertEqual(res.status_code, 403)
```
toolness: This is fine--but just FYI, there's also a [`follow` argument to `client.get()`](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.Client.get) that allows for this to happen a bit more transparently.

.bandit
```diff
@@ -1,2 +1,2 @@
 [bandit]
-exclude: node_modules,frontend/tests,api/tests,meta/management/commands,docker_django_management.py,data_capture/tests
+exclude: node_modules,frontend/tests,api/tests,meta/management/commands,docker_django_management.py,data_capture/tests,hourglass/tests.py
```
toolness: Huh, I'm curious what bandit dislikes about `hourglass/tests.py`...

toolness:
This looks good, feel free to merge!

.bandit
jseppi: It didn't like the password arguments to `create_user`

PR #675, opened 2016-09-07
Add "Data Administrator" role for Django admin

### LINKED ISSUE #668: Modify permissions/role model
- Only TTS Infra and devs should be superusers.
- CALC Admin account should have proper permissions only to the necessary models

cc @toolness

toolness:
This fixes #668 by adding a `Data Administrators` role, which is created by a new `manage.py initgroups` command (based on [hive-django's `initgroups.py`](https://github.com/toolness/hive-django/blob/master/directory/management/commands/initgroups.py)).

As of this PR, authorization is set up as follows (this is taken from the `README.md` in this PR):

> * **Non-staff users** can upload individual price lists for approval.
> * **Staff users** can bulk upload price list data.
> * **Staff users** in the **Data Administrators** group can create and edit
>   users and assign them to groups. They can also review submitted price lists
>   and approve/unapprove them.
> * **Superusers** can do anything, but only infrastructure/operational
>   engineers should be given this capability. An initial superuser can be
>   created via `manage.py createsuperuser`.

To do:

- [x] Add actual roles.
- [x] Make sure `manage.py initgroups` is run on new deploys.
- [x] Document role information in readme.
- [x] Add tests for the admin user detail view

We should also make issues to address the following in future PRs:

- [ ] Only Data Administrators should be able to upload region 10 data, probably (currently anyone who `is_staff` can).
- [ ] add another role for CO's, which gives them the ability to upload price lists (currently any logged-in user can).

### DIFF
--- a/uaa_client/authentication.py
+++ b/uaa_client/authentication.py
@@ -3,6 +3,7 @@ import requests
 import jwt
 from django.core.urlresolvers import reverse
 from django.contrib.auth.models import User
+from django.contrib.auth.backends import ModelBackend
 from django.conf import settings
 
 logger = logging.getLogger('uaa_client')
@@ -53,7 +54,14 @@ def get_user_by_email(email):
         return None
 
 
-class UaaBackend:
+class UaaBackend(ModelBackend):
+    '''
+    Custom auth backend for Cloud Foundry / cloud.gov User Account and
+    Authentication (UAA) servers.
+
+    This inherits from ModelBackend so that the superclass can provide
+    all authorization methods (e.g. `has_perm()`).
+    '''
 
     def authenticate(self, uaa_oauth2_code=None, request=None, **kwargs):
         if uaa_oauth2_code is None or request is None:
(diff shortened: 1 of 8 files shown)


toolness:
Ok I think this is ready @jseppi !

uaa_client/authentication.py
```diff
@@ -66,9 +74,3 @@ def authenticate(self, uaa_oauth2_code=None, request=None, **kwargs):
         user_info = jwt.decode(access_token, verify=False)
 
         return get_user_by_email(user_info['email'])
-
-    def get_user(self, user_id):
-        try:
-            return User.objects.get(pk=user_id)
-        except User.DoesNotExist:
-            return None
```
toolness: This is the exact same `get_user` as `ModelBackend.get_user`, so I removed it.

jseppi:
👍 looks good to me!

I agree with your additional issues (and think we should get them done before launch).

PR #679, opened 2016-09-07
Add a Django debug toolbar panel exposing scheduled jobs

toolness:
This builds on #655 by adding a **Scheduled Jobs** panel to the debug toolbar:

> screen shot 2016-09-07 at 12 56 57 pm

It also fixes some string formatting issues.  Basically, we want to avoid using `\` to continue strings that span multiple lines, e.g.:

```py
logger.info("hello \
             there")
```

This is because the above code will log:

```
hello              there
```

(Also, my sublime text syntax highlighter has a hissy fit.)

Instead, we can take advantage of the fact that Python automatically concatenates adjacent string literals, e.g.:

```py
logger.info("hello "
            "there")
```


jseppi:
👍 Super cool! Just tried it out and it looks great!

data_capture/apps.py
```diff
@@ -42,7 +46,7 @@ def ready(self):
 
         # Add cron-type job to send a reminder email to the CALC admins
         # about approving price lists
-        logger.info('Adding send_admin_approval_reminder_email job on cron \
-          schedule "{}"'.format(self.admin_reminder_cron))
+        logger.info('Adding send_admin_approval_reminder_email job on cron '
+                    'schedule "{}"'.format(self.admin_reminder_cron))
```
jseppi: ah, whoops. I should have caught that :(

PR #680, opened 2016-09-07
add redis to 'stack'

jseppi:
extremely minor change, just adding redis to the `stack` section of `.about.yml`


jseppi:
Merging without review. It's just a simple doc change that has no code impact.

PR #681, opened 2016-09-07
Split dev and prod dependency requirements files

### LINKED ISSUE #678: Split testing dependencies into a separate requirements file, do not include in prod/staging
Per ATO discussion (https://github.com/18F/Infrastructure/issues/649#issuecomment-245316653) we should split out all testing dependencies into another requirements file (such as `requirements-test.txt`). `requirements.txt` should then only contain packages necessary for production.

We'll need to modify our Travis build to install both from `requirements.txt` and `requirements-test.txt` (so that it can run tests) and we'll need to ensure that the cloud.gov deploys only install from `requirements.txt`.

jseppi:
This PR adds `requirements-dev.txt` for specifying dev dependencies separately from `requirements.txt`, which should now only contain production dependencies. Something cool I learned is that you can refer to other requirements files within the contents of a requirements file (see https://github.com/18F/calc/compare/split-prod-test-deps?expand=1#diff-f4c265073ea4274a80eca85440875e56R2).

The Travis config has been modified to install from `requirements-dev.txt` since testing dependencies are needed to run tests on the CI server.

Closes #678


toolness:
NIce, this looks great!

Issue #682, opened 2016-09-07
Move email sending logic behind a feature flag?


jseppi:
Regarding the [decision to postpone email notifications](https://github.com/18F/Infrastructure/issues/649#issuecomment-245325401) until a later release, and thus an ATO modification, after the initial ATO (hopefully) gets signed, how should we go about removing email sending functionality?

Note that right now no emails are being sent because we haven't actually integrated with a mailing service.

Options:
- Comment-out email-sending logic
- Hide email sending calls behind a feature flag that is Falsy in staging/production
- Just set the `EMAIL_URL` env var to `'console:'` in staging (and later in prod), which will log all email contents to stdout, but not actually send emails.

I think I like the `console:` option best. 

In any case, we'll have to modify some of our view templates to not mention that an email is coming as well.

@toolness - thoughts?

toolness:
Yeah, I think I like the `console:` option!

It looks like dj-email-url also supports [`dummy:`](https://docs.djangoproject.com/en/1.10/topics/email/#dummy-backend) too, in case there are objections/difficulties involved in spewing emails to the console in prod.

jseppi:
Ah neat. Do you think we should have logic in the templates that checks if `EMAIL_URL is 'console:'` (or `'dummy:'`)?

toolness:
Hmm why would we want that?  So we don't tell users that they're going to be emailed?

jseppi:
Exactly. If the email sending logic is disabled/hidden then we shouldn't tell the user that they will receive an email once the price list is approved (like in https://github.com/18F/calc/blob/01dcd77e7014a753206ad0527dffd8d4c3c324ca/data_capture/templates/data_capture/bulk_upload/region_10_step_3.html#L8)

toolness:
Hmmmmm I am not sure actually... when devs set the email URL to `console:` I think they want everything to work just like it normally would if the email is actually sent out...  I guess we could detect for `dummy:` though, and suppress if that's the case?  Or perhaps have something like the following in `settings.py`:

```py
SEND_TRANSACTIONAL_EMAILS = os.environ['EMAIL_URL'] != 'dummy:'
```

And then just check `SEND_TRANSACTIONAL_EMAILS` in templates and whatnot?

toolness:
blerg

jseppi:
Right, but we need to disable all the email functionality for the ATO/first release.

[...]

PR #684, opened 2016-09-07
Hide template content about emails when EMAIL_URL is dummy

jseppi:
Adds `settings.SEND_TRANSACTIONAL_EMAILS`. It is `True` when `EMAIL_URL` is `'dummy:'`. When it is true, template contents about emails being sent will be hidden (right now this is just in step 3 of the Region 10 bulk upload flow). The reason for doing this is because we will not be sending any emails in the initial release, as discussed in #682.

We will need to make sure `EMAIL_URL` is set to `dummy:` in the production user-provided service (ref https://github.com/18F/calc/issues/692)

Closes https://github.com/18F/calc/issues/682


toolness:
NIce, feel free to merge!

PR #687, opened 2016-09-08
Disable DjDT profiling panel by default

toolness:
This just disables the profiling panel by default, i.e.:

> screen shot 2016-09-08 at 8 02 27 am

Developers can just mark the checkbox and reload the page to enable it.  I think that this is good because we don't actually need profiling most of the time, and it tends to slow down our pages a lot.

### DIFF
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -297,6 +297,10 @@ if not UAA_CLIENT_SECRET:
 DEBUG_TOOLBAR_PATCH_SETTINGS = False
 
 DEBUG_TOOLBAR_CONFIG = {
+    'DISABLE_PANELS': set([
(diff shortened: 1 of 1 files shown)


hourglass/settings.py
```diff
@@ -297,6 +297,10 @@
 DEBUG_TOOLBAR_PATCH_SETTINGS = False
 
 DEBUG_TOOLBAR_CONFIG = {
+    'DISABLE_PANELS': set([
+        'debug_toolbar.panels.redirects.RedirectsPanel',
```
toolness: This is the default value of `DISABLE_PANELS` so I re-included it in our override.

jseppi:
Nice 👍

PR #688, opened 2016-09-08
Add a Documentation DjDT panel

toolness:
This adds a **Documentation** panel at the top of the panel list:

> screen shot 2016-09-08 at 8 43 19 am

Specifically, it makes our style guide and test suite actually discoverable, so that should hopefully help new devs.

(I originally tried adding links to the style guide and tests in the top-right nav area of the base template, but I didn't like how much visual noise it added, so I resorted to a DjDT panel. Also, there's no way to embed links directly in the toolbar itself, which would've been preferable, so I resorted to a Documentation panel.)

### DIFF
--- a/data_capture/templates/data_capture/panels/docs.html
+++ b/data_capture/templates/data_capture/panels/docs.html
@@ -1,0 +1,20 @@
+<h4>Welcome to CALC!</h4>
+
+<p>
+  Please make sure you read the
+  <a href="https://github.com/18F/calc#readme">README</a> thoroughly. It
+  should answer most of your questions.
+</p>
+
+<p>
+  Also check out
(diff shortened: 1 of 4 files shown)


data_capture/templates/data_capture/panels/docs.html
```diff
@@ -0,0 +1,20 @@
...
+  Also check out
+  <a href="https://github.com/18F/calc/blob/develop/CONTRIBUTING.md">CONTRIBUTING</a>
+  if you'd like to improve this tool.
+</p>
+
+<h4>Useful Links</h4>
+
+<ul>
+  <li><a href="{% url 'styleguide:index' %}">Style guide</a></li>
+  <li><a href="{% url 'tests' %}">Test suite</a></li>
```
jseppi: Maybe "Front end test suite" for clarity?

jseppi:
👍

PR #694, opened 2016-09-08
Implement Contract Officers role and protect Price List Upload views

### LINKED ISSUE #686: Add a Django role for Contract Officers
As mentioned in #675, currently, any logged-in user can upload price lists, but we should add a role for Contract Officers and ensure that only they can do so, to enforce POLA.

jseppi:
Adds a new decorator to test that user has Contract Officer permissions, and decorates price list upload views with the new decorator.

closes #686 

To Do:
- [x] add Contract Officers role and permissions
- [x] restrict Price List Upload views based on relevant permissions
- [x] tests
  - [x] for decorator
  - [x] that views are protected
  - [x] for new role in `initgroups`
- [x] merge `develop`
- [x] modify to have this decorator do login redirection when a user is not-authenticated, and throw `PermissionDenied` if the user is authenticated but doesn't have the right perms.


data_capture/management/commands/initgroups.py
```diff
@@ -11,6 +11,11 @@
     'change_submittedpricelistrow',
```
jseppi: does 'change_xyz' also imply 'add_xyz' in Django permissions?

toolness: Nope, I believe the two are completely independent of  one another.

data_capture/management/commands/initgroups.py
```diff
@@ -11,6 +11,11 @@
...
 )
 
+ROLES['Contract Officers'] = (
+    'add_submittedpricelist',
+    'add_submittedpricelistrow',
+)
```
toolness: We might want to consider having the data administrator inherit these permissions too (e.g. via `ROLES['Data Administrators'].update(ROLES['Contract Officers'])`) so that kelly can upload a price list if he really wants to?  I'm not actually sure, I guess I'd considered Data Administrators to be kind of like "one step removed from superusers" in that they should be able to do *almost* anything on the site.

jseppi: Yup, that's kinda why I was curious if 'change' implied 'add'

data_capture/decorators.py
```diff
@@ -7,6 +7,12 @@
 from frontend import ajaxform
 
 
+contract_officer_perms_required = decorators.permission_required(
+    ['data_capture.add_submittedpricelist',
+     'data_capture.add_submittedpricelistrow'],
```
toolness: I wonder if we should actually pull these directly from `initgroups.py` to maintain `DRY`?  It is kind of annoying that one of them is namespaced to `data_capture` and the other isn't, though...

jseppi: Hrm, I wonder if that namespacing is necessary here. I haven't done anything with Django permissions before so I'm using this as a learning opportunity :)

data_capture/management/commands/initgroups.py
```diff
@@ -11,6 +11,14 @@
...
+
+# Data Administrators should also have any perms that Contract Officers do
+ROLES['Data Administrators'].update(ROLES['Contract Officers'])
```
jseppi: whoops, the roles are a `tuple` so `.update` won't work :(

data_capture/decorators.py
jseppi: Ah cool, the namespacing here is not required.

data_capture/management/commands/initgroups.py
toolness: Oh, my bad.  Maybe we should make the values a python [`set`](https://docs.python.org/3/tutorial/datastructures.html#sets) so we can easily unionize them?

data_capture/decorators.py
jseppi: Urrrggggg, I was wrong about that.

data_capture/decorators.py
```diff
@@ -5,11 +5,11 @@
...
 
 from frontend import ajaxform
-
+from .management.commands.initgroups import ROLES
 
 contract_officer_perms_required = decorators.permission_required(
-    ['add_submittedpricelist',
-     'add_submittedpricelistrow'],
+    # need to add app namespace to each permission
+    ["data_capture.{}".format(r) for r in ROLES['Contract Officers']],
```
jseppi: blerg, this feels weird to me

toolness: Yeah, I don't understand why some Django methods want the namespaced version while others don't... I wonder if it's easier/cleaner to modify `initgroups` to use namespaced permissions?  It'd also make `initgroups` more explicit, too... blerg

jseppi:
ready for review, @toolness 

One note is that the new decorator `@contract_officer_perms_required` doesn't redirect to login, but instead raises a `PermissionDenied` error when the user has insufficient perms. This means that if we want a view to redirect to login, the new decorator needs to be used in conjunction with `@login_required`. I'm pretty sure it would be relatively easy to change it so that it first redirects if the user is not authenticated, but let me know if you think that should be done.

jseppi:
@toolness: if you think that making this decorator do the login redirect is useful, I don't mind if you want to commit to this branch since I'll be out this afternoon.

jseppi:
Merged in `develop` again, no problemo

toolness:
So the one thing that makes me think we should have this decorator do the login redirect is that such behavior is implied by Django's own [`user_passes_test`](https://docs.djangoproject.com/en/1.10/topics/auth/default/#django.contrib.auth.decorators.user_passes_test), so I'm thinking that folks will likely assume this will be the case with your decorator too.

I'm cool w/ building on this PR, but I'm going to be leaving in a few hours myself and right now #710 is my priority since it's a regression...  I'll hack on this if I have time though!

jseppi:
Ok, cool, that makes sense

jseppi:
Ok, ready for review, @toolness!

data_capture/decorators.py
```diff
@@ -80,3 +81,33 @@ def check_if_staff(user):
...
+def contract_officer_perms_required(function=None, login_url=None):
+    '''
+    Decorator to check that a user accessing a view has the permissions
+    of a Contract Officer.
+    If the user is not-authenticated, then they will first be redirected to
+    login. If the user is authenticated but does not have the correct
+    permissions, a PermissionDenied exception will be raised.
+    '''
+    # need to add app namespace to each permission
+    perms = ["data_capture.{}".format(r) for r in ROLES['Contract Officers']]
```
jseppi: I still don't like this weirdness very much.

toolness: Yeah it's harrible.  Lemme see if I can figure out how to have the roles in `importorgs` be namespaced so that you don't have to do this crap.

toolness: Just issued #726 as a solution!

data_capture/email.py
```diff
@@ -9,9 +9,9 @@ class EmailResult():
     '''
     Simple class to hold result data from email sending functions
     '''
-    def __init__(self, was_successful, context={}):
+    def __init__(self, was_successful, context=None):
```
toolness: O shoot, I shoulda caught that earlier when it was first committed!

data_capture/tests/test_price_list_views.py
```diff
@@ -18,7 +20,19 @@ def set_fake_gleaned_data(self, rows):
         session.save()
 
     def setUp(self):
+        super().setUp()
         registry._init()
+        call_command('initgroups', stdout=io.StringIO())
```
toolness: Huh, I wonder if this adds much overhead to how fast the tests run.

data_capture/tests/test_price_list_views.py
```diff
@@ -33,6 +47,9 @@ class Step1Tests(PriceListStepTestCase):
     def test_login_is_required(self):
         self.assertRedirectsToLogin(self.url)
 
+    def test_group_is_required(self):
+        self.assertDeniedForInvalidPerms(self.url)
```
toolness: I wonder if we should just lift these up to the `PriceListStepTestCase` superclass since all the steps require the same permissions?  Hmmm.

toolness:
This looks good to me!

I'm not really sure if we're *over*-testing it or not, though...  I guess some of these tests are integration-y while others are unit-y.  Always hard to figure out what the right trade-off is I suppose.

jseppi:
Conflicts --- noooo!!!!

data_capture/email.py
jseppi: QuantifiedCode did, oddly enough. I removed QuantifiedCode but for some reason still got an email with identified issues yesterday o_O

toolness: LOL, it's like it's saying "don't get rid of me, I can still help!  Even though I went out of business..."

data_capture/tests/test_price_list_views.py
```diff
@@ -17,8 +19,31 @@ def set_fake_gleaned_data(self, rows):
...
+    def login(self, groups=None, **kwargs):
+        if groups is None:
+            groups = ['Contract Officers']
+        return super().login(groups=groups, **kwargs)
+
+    def test_login_is_required(self):
+        if hasattr(self, 'url'):
+            self.assertRedirectsToLogin(self.url)
+
+    def test_group_is_required(self):
```
jseppi: moved `test_group_is_required` and `test_login_is_required` to the parent class. Look good to you @toolness?

toolness: Yeah, noice!  The test count didn't change when you did this, right?

jseppi: It went up by one since I guess it tries to run it from the base class, which is why I had to add the `hasattr` check.

toolness: Oh, heh!  One way to get around the increased test count might be to raise a [`unittest.SkipTest`](https://docs.python.org/2/library/unittest.html#unittest.SkipTest) if it fails.

Also, you might want to consider just setting `url = None` in the base test case class--that way the property becomes sort of self-documenting, and you can just test against `if self.url is None` rather than the `hasattr`, which in Python-land... well, I guess I prefer "real" tests to `hasattr`'s but that's just me, not sure if it's Pythonic or not.

jseppi:
let me know if 👍 now, @toolness

data_capture/tests/test_price_list_views.py
```diff
@@ -35,12 +37,11 @@ def login(self, groups=None, **kwargs):
         return super().login(groups=groups, **kwargs)
 
     def test_login_is_required(self):
-        if hasattr(self, 'url'):
+        if self.url:
             self.assertRedirectsToLogin(self.url)
```
toolness: This is fine, but another alternative would be:

```python
if self.url:
    raise unittest.SkipTest()
self.assertRedirectsToLogin(self.url)
```

That way the test count doesn't increase on the base class, but it's up to you!

toolness:
Yassss I made one comment but it's up to you if you want to change it, 👍 🎉 💃 🇺🇸 🍷 !

PR #703, opened 2016-09-08
Fix rem import issue

### LINKED ISSUE #671: Fix `rem()` in `variables`
Currently there's a circular import: variables imports the rem  func, but the rem func needs to import variables. So nothin' works.

hbillings:
Closes #671


toolness:
Coool thanks!

PR #704, opened 2016-09-08
Use font family variables

### LINKED ISSUE #445: Create font family variables
We're hardcoding these right now. Let's factor them into variables with reasonable fallback fonts.

hbillings:
Closes #445. We were almost using these everywhere as is, but I just did a final sanity check to make sure. Only found one place that needed updating!

### DIFF
--- a/frontend/source/sass/components/_exceltables.scss
+++ b/frontend/source/sass/components/_exceltables.scss
@@ -6,7 +6,7 @@
 }
 
 table.excel {
-  font-family: Arial, Helvetica, serif;
+  font-family: $font-family-main;
   width: auto;
   margin: 0;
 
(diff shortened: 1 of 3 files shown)


frontend/source/sass/components/_exceltables.scss
```diff
@@ -6,7 +6,7 @@
 }
 
 table.excel {
-  font-family: Arial, Helvetica, serif;
+  font-family: $font-family-main;
```
toolness: Oh, the use of Arial was actually intentional on my part actually--since we're explicitly trying to make the excel tables look like Excel, I thought it'd be best if the `font-family` were actually Arial, rather than inheriting whatever our main font is.  What do you think?

hbillings: Ah! well then, ignore this.

PR #707, opened 2016-09-09
Use a ModelForm in step 1 of price list upload

toolness:
This improves [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) by using a `ModelForm` in step 1 of price list upload.

@hbillings does this look ok and make sense to you?  As with `Step2Form`, we're using a `ModelForm` to make things easy and DRY for us, but not actually saving the model until step 4.

### DIFF
--- a/data_capture/forms/__init__.py
+++ b/data_capture/forms/__init__.py
@@ -1,6 +1,6 @@
 
-from .price_list import Step1Form, Step2Form, Step3Form, Step4Form
+from .price_list import Step1Form, Step2Form, Step3Form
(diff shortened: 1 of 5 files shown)


hbillings:
I can't remember precisely what the deal was, but I had `Step1Form` as a `ModelForm` originally, and @jseppi suggested I use a regular `Form` instead. I'm sure you're doing things totally differently than I was, but if he's cool with it, I'm cool with it!

jseppi:
Yeah, sorry, I think I was misunderstanding how `ModelForms` can/should be used. My bad!

PR #708, opened 2016-09-09
Actually run doctests during py.test

toolness:
Um, so when I added `contracts.tests.test_doctests`, I was actually using `manage.py test` and not `py.test`... I had assumed that `py.test` was a superset of `manage.py test`, but I was wrong--it appears `py.test` (and therefore Travis and `ultratest`) was not finding and running our doctests. 😞 

This fixes things so that `py.test` will [scour all our source for doctests](http://doc.pytest.org/en/latest/doctest.html) and run them.  It also relieves us of having to explicitly find and add doctests to our test suites, though that also means that `manage.py test` won't run them, but whatevs, we're really using `py.test`.


pytest.ini
```diff
@@ -2,3 +2,4 @@
 DJANGO_SETTINGS_MODULE=hourglass.settings
 python_classes=*Test Test*
 python_files=tests.py test_*.py
+addopts = --doctest-modules
```
jseppi: cool, didn't know about `addopts`, I guess we could also add `--cov` to `addopts` since we always want it (during ultratest and travis tests).

toolness: Can we either do that in a separate PR, or can you add that as a commit on top of this one?  I've got my hands full w/ other PRs right now, sorry 😣

jseppi: yeah, I'll add here

toolness:
NOICE 👍

jseppi:
Urghghghg, getting some weird travis errors still in this one. Will try to work it out once I finish merging `develop` into #686 

```sh
==================================== ERRORS ====================================
____________ ERROR collecting node_modules/node-gyp/gyp/gyp_main.py ____________
node_modules/node-gyp/gyp/gyp_main.py:13: in <module>
    import gyp
E   ImportError: No module named 'gyp'
____________ ERROR collecting node_modules/node-gyp/gyp/gyptest.py _____________
../../../virtualenv/python3.4.4/lib/python3.4/site-packages/py/_path/local.py:650: in pyimport
    __import__(modname)
E     File "/home/travis/build/18F/calc/node_modules/node-gyp/gyp/gyptest.py", line 199
E       print test
E                ^
E   SyntaxError: Missing parentheses in call to 'print'
```

toolness:
I just added a line that might've fixed things!

jseppi:
Woot, nice!
Stupid `node_modules`

PR #709, opened 2016-09-09
Use a separate redis DB for tests

### LINKED ISSUE #706: Region10UploadStep3Tests fail if "docker-compose up" is already running
If I'm already running `docker-compose up` in one terminal and I run the tests in another terminal, `Region10UploadStep3Tests` fail.

<details>
  <summary>Full test output</summary>
```
======================================================================
FAIL: test_old_contracts_are_deleted (data_capture.tests.test_bulk_upload_views.Region10UploadStep3Tests)
----------------------------------------------------------------------
  File "/calc/data_capture/tests/test_bulk_upload_views.py", line 186, in test_old_contracts_are_deleted
    self.assertEqual(len(contracts), 3)
AssertionError: 0 != 3

======================================================================
FAIL: test_post_is_ok_and_contracts_are_created_properly (data_capture.tests.test_bulk_upload_views.Region10UploadStep3Tests)
----------------------------------------------------------------------
  File "/calc/data_capture/tests/test_bulk_upload_views.py", line 159, in test_post_is_ok_and_contracts_are_created_properly
    self.assertEqual(len(contracts), 3)
AssertionError: 0 != 3

----------------------------------------------------------------------
Ran 127 tests in 43.825s

FAILED (failures=2)
```
</details>

I *think* this may be happening because the `rq_worker` running in `docker-compose up`, which is connected to my dev (not test) database, is seeing the jobs added by the test suite and processing them before the test suite itself has a chance to grab them.

I think we can probably fix this if we use a different [redis database number](http://www.rediscookbook.org/multiple_databases.html) when running tests. That way we can mimic the convention that Django uses for relational DBs when testing--using a separate test database so that dev/prod data isn't touched.

toolness:
This fixes #706.


jseppi:
hrm, the failing tests look to be selenium-related

toolness:
Uh, I thought we disabled selenium tests in travis?

toolness:
See #330...

jseppi:
eep, well, we had, but then we (I) moved all the frontend tests to `frontend/`, so this line is wrong now: https://github.com/18F/calc/pull/331/files#diff-354f30a63fb0907d4ad57269548329e3R17
 >.<

jseppi:
fix incoming

toolness:
ah cool thanks!!!!!!!!!!!!!!!

AWEOGJaowegjopaegwj i hate selenium

jseppi:
fixed in https://github.com/18F/calc/pull/712

toolness:
Woot, travis likes it now, @jseppi !

jseppi:
Schweeeeet!

PR #712, opened 2016-09-09
re-ignore selenium tests in travis

jseppi:
We were ignoring selenium tests as of #331 during Travis builds because they often time out, per #330. However, when all the frontend-related code and tests were moved to `frontend/` I neglected to update the `--ignore` argument to `py.test` in our `.travis.yml`. This PR re-ignores selenium tests on Travis.


toolness:
Woooot thanks!

PR #713, opened 2016-09-09
Fix #705, #711, simplify price list form logic.

### LINKED ISSUE #705: Step 2 of price list upload fails if user leaves date field blank
I think we need to make the fields required or something?

<details>
  <summary>Full stack trace</summary>
```
Environment:


Request Method: POST
Request URL: http://localhost:8000/data-capture/step/2

Django Version: 1.8.14
Python Version: 3.4.5
Installed Applications:
('django.contrib.admin',
 'django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.humanize',
 'whitenoise.runserver_nostatic',
 'django.contrib.staticfiles',
 'debug_toolbar',
 'django_rq',
 'data_explorer',
 'contracts',
 'data_capture.apps.DefaultDataCaptureApp',
 'api',
 'djorm_pgfulltext',
 'rest_framework',
 'corsheaders',
 'uaa_client',
 'styleguide',
 'meta',
 'frontend',
 'fake_uaa_provider')
Installed Middleware:
('hourglass.middleware.ComplianceMiddleware',
 'whitenoise.middleware.WhiteNoiseMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'hourglass.middleware.DebugOnlyDebugToolbarMiddleware')


Traceback:
File "/usr/local/lib/python3.4/site-packages/django/core/handlers/base.py" in get_response
  132.                     response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/calc/data_capture/decorators.py" in wrapper
  46.                 return function(request, *args, **kwargs)
File "/usr/local/lib/python3.4/site-packages/django/contrib/auth/decorators.py" in _wrapped_view
  22.                 return view_func(request, *args, **kwargs)
File "/calc/data_capture/views/price_list_upload.py" in step_2
  73.                 'contract_start': ymd(form.cleaned_data['contract_start']),
File "/calc/data_capture/views/price_list_upload.py" in ymd
  57.         return date_input.strftime("%Y-%m-%d")

Exception Type: AttributeError at /data-capture/step/2
Exception Value: 'NoneType' object has no attribute 'strftime'
```
</details>

### LINKED ISSUE #711: Uploading a price list with no valid rows redirects user to step 3 w/o any feedback
Even though the template for step 4 includes logic to tell the user that their data has no valid rows, the actual `step_4` function redirects the user to step 3 if there are no valid rows, so the user never actually sees the information.

I'll try to fix this along w/ my refactoring!

toolness:
**Note: This builds upon #707.**

This simplifies our price list form logic by storing the raw POST data from earlier steps in the session, rather than storing the same data in the session using a potentially error-prone custom serialization mechanism.

This also allows us to "resurrect" the actual `ModelForm`s for earlier steps when the user submits their data in step 4 and call their `save()` methods, saving us from having to assign data to the new model ourselves.

This new way of doing things also happens to fix #705.

I also fixed #711 because it was easy to fix along the way.

@hbillings can you take a look at this?  Does it make sense?

### DIFF
--- a/data_capture/forms/__init__.py
+++ b/data_capture/forms/__init__.py
@@ -1,6 +1,6 @@
 
-from .price_list import Step1Form, Step2Form, Step3Form, Step4Form
+from .price_list import Step1Form, Step2Form, Step3Form
 from .bulk_upload import Region10BulkUploadForm
 
-__all__ = ['Step1Form', 'Step2Form', 'Step3Form', 'Step4Form',
+__all__ = ['Step1Form', 'Step2Form', 'Step3Form',
            'Region10BulkUploadForm']
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -71,33 +71,3 @@ class Step3Form(forms.Form):
             cleaned_data['gleaned_data'] = gleaned_data
 
         return cleaned_data
-
-
-class Step4Form(forms.ModelForm):
-    class Meta:
-        model = SubmittedPriceList
-        fields = [
-            'contract_number',
-            'vendor_name',
-            'schedule',
-            'is_small_business',
-            'contractor_site',
-            'contract_start',
-            'contract_end',
-            'contract_year',
-        ]
-
-        # TODO: It'd be nice if we could actually get rid of all these
-        # hidden inputs, since all this data is already stored in our request
-        # session.
-
-        widgets = {
-            'contract_number': forms.HiddenInput(),
-            'vendor_name': forms.HiddenInput(),
-            'schedule': forms.HiddenInput(),
-            'is_small_business': forms.HiddenInput(),
-            'contractor_site': forms.HiddenInput(),
-            'contract_year': forms.HiddenInput(),
-            'contract_start': forms.HiddenInput(),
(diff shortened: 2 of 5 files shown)


toolness:
Oh, I just realized that another benefit of keeping the raw POST data from earlier steps is that we can easily pre-fill forms if the user decides to go back to earlier steps, which we don't currently do. Filed #717 to address this later.

hbillings:
Ah!! I wasn't entirely sure what you meant a few days ago when you suggested just saving the `POST` data -- it makes sense now! Definitely seems a less brittle way to do things.

toolness:
Woot!  Um, should I go ahead and merge it?

PR #719, opened 2016-09-09
Explain use of font-family setting in excel table css.

toolness:
I figured we should explain ourselves here after #704!


jseppi:
👍

PR #724, opened 2016-09-11
Improve price list views tests

toolness:
This simplifies `data_capture.views.price_list_upload` a bit and improves its test coverage to 100%.

To see the test coverage for this file, you can run:

```
docker-compose run app py.test -k price_list_views --cov-report term-missing | grep data_capture/views/price_list_upload.py
```

Notes:

* I made `step_5` not require any kind of session information: all we're doing here is presenting the user with some information, so it's OK if they randomly decide to visit that URL.  It also moves the deletion of the price list from the session to step 4, which IMO makes the whole transaction a bit more atomic--and anyways, a GET request to `step/5` shouldn't change server state.

* Steps 1-3 will only accept `GET` or `POST` requests now.  Previously, other HTTP methods could potentially cause 500s because our branching logic didn't anticipate them.

### DIFF
--- a/data_capture/tests/test_price_list_views.py
+++ b/data_capture/tests/test_price_list_views.py
@@ -17,6 +17,11 @@ class PriceListStepTestCase(StepTestCase):
             registry.serialize(pricelist)
         session.save()
 
+    def delete_price_list_from_session(self):
+        session = self.client.session
+        del session['data_capture:price_list']
+        session.save()
+
     def setUp(self):
         registry._init()
 
@@ -94,6 +99,12 @@ class Step2Tests(PriceListStepTestCase):
         res = self.client.get(self.url)
         self.assertEqual(res.status_code, 200)
 
+    def test_redirects_to_step_1_if_not_completed(self):
+        self.login()
+        self.delete_price_list_from_session()
+        res = self.client.get(self.url)
+        self.assertRedirects(res, Step1Tests.url)
+
     def test_valid_post_updates_session_data(self):
         self.login()
         self.client.post(self.url, self.valid_form)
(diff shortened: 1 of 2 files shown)


data_capture/views/price_list_upload.py
```diff
@@ -24,16 +25,13 @@ def wrapper(request):
...
 def step_1(request):
     if request.method == 'GET':
         form = forms.Step1Form()
-    elif request.method == 'POST':
+    else:
         form = forms.Step1Form(request.POST)
         if form.is_valid():
-            # Clear out request.session if it previously existed
-            if 'data_capture:price_list' in request.session:
-                del request.session['data_capture:price_list']
```
toolness: There's no reason to explicitly delete this session key because we're overwriting it anyways on the next line.

data_capture/tests/test_price_list_views.py
```diff
@@ -94,6 +99,12 @@ def test_get_is_ok(self):
         res = self.client.get(self.url)
         self.assertEqual(res.status_code, 200)
 
+    def test_redirects_to_step_1_if_not_completed(self):
+        self.login()
+        self.delete_price_list_from_session()
+        res = self.client.get(self.url)
+        self.assertRedirects(res, Step1Tests.url)
```
jseppi: cool, didn't realize there was an `assertRedirects` method

toolness: Yeah, there's actually [lots of really useful assertions](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#assertions) in Django's custom `unittest.TestCase` subclass.  Just perusing through them now, there's actually a lot I didn't know about, and would probably have been useful writing some of our earlier tests!

toolness:
@hbillings does this look OK to you?

PR #725, opened 2016-09-11
Disable uglify when DEBUG is false

toolness:
This disables uglifying our bundles when `DEBUG` is false, which helps a lot with [bundle build times](https://github.com/18F/calc/pull/597) during development.

Among other things, it "translates" the value of `DEBUG` into node's preferred `NODE_ENV` setting at the very beginning of `gulpfile.js`, so we can use it in the future for other useful things (e.g., pruning debugging code from bundles in production builds).


jseppi:
👍

PR #726, opened 2016-09-12
Use namespaced permissions for initgroups cmd.

toolness:
This should make #694 less harrible.

Had to dig in to [`django/contrib/auth/backends.py`](https://github.com/django/django/blob/master/django/contrib/auth/backends.py) to figure out how the heck it was making up those namespaced permission names in the first place.  I wish this was better documented in Django-land.


data_capture/management/commands/initgroups.py
```diff
@@ -25,26 +25,28 @@ class Command(BaseCommand):
...
             self.stdout.write("  Group does not exist, creating it.")
             group = Group(name=groupname)
             group.save()
+        splitnames = [perm.split('.') for perm in perms]
         group.permissions = [
-            Permission.objects.get(codename=codename)
-            for codename in codenames
+            Permission.objects.get(codename=codename,
+                                   content_type__app_label=app_label)
+            for app_label, codename in splitnames
```
jseppi: ha, it took me a while to mentally parse this code because I didn't realize it was a comprehension at first :P

jseppi:
Woot! 👍

PR #729, opened 2016-09-12
In price list upload, allow users to go back to previous steps w/o filling everything in again

### LINKED ISSUE #717: Going back to previous steps in price list upload shows empty forms
To reproduce:

1. Fill out step 1 of price list upload and click "next".
2. Click "previous" in step 2.

You're taken back to step 1 with none of the fields filled out.

Ideally, we should pre-fill the fields with whatever the user last filled them out as, so that they can easily tweak values instead of having to fill everything out again.

toolness:
~~This is a PR against #724.~~

This fixes #717 by making it so users can choose "previous" in steps 2 and 3, and not have to re-enter everything in the previous step.


data_capture/tests/test_price_list_views.py
```diff
@@ -99,6 +99,11 @@ def test_get_is_ok(self):
         res = self.client.get(self.url)
         self.assertEqual(res.status_code, 200)
 
+    def test_going_back_to_previous_step_prefills_its_form(self):
+        self.login()
+        res = self.client.get(Step1Tests.url)
+        self.assertContains(res, 'GS-123-4567')
+
```
toolness: It was easier to add these tests in the step *after* the actual view being tested, because their sessions were already set up appropriately...  Kind of unintuitive though.

data_capture/views/price_list_upload.py
```diff
@@ -24,11 +24,35 @@ def wrapper(request):
...
+        >>> get_nested_item(d, ('foo', 'blarg'))
+    '''
+
+    key = keys[0]
+    if key not in obj:
+        return default
+    if len(keys) > 1:
+        return get_nested_item(obj[key], keys[1:], default)
+    return obj[key]
+
```
toolness: I'm not sure if a method like this already exists in the python standard library but brief googling didn't seem to indicate so.  Also, this might be better off in a separate module...

jseppi: Yeah, maybe a utils module or something, but for now we don't have any uses for it so keeping it here is fine until we do.

toolness: I guess we could also move it to `common.py`, even though it's not actually used by anything outside of `price_list_upload.py` at present.

data_capture/tests/test_price_list_views.py
```diff
@@ -99,6 +99,11 @@ def test_get_is_ok(self):
         res = self.client.get(self.url)
         self.assertEqual(res.status_code, 200)
 
+    def test_going_back_to_previous_step_prefills_its_form(self):
+        self.login()
+        res = self.client.get(Step1Tests.url)
+        self.assertContains(res, 'GS-123-4567')
```
jseppi: Is just testing this one value enough?

toolness: Hmm, I think so--we're relying on Django forms to do most of the heavy lifting for us, so here we're just testing to make sure that at least one of the pieces of information has been passed back.  I *could* add tests for other fields but I'm worried that might make our tests more fragile than they ought to be, e.g. if we decide to move fields between steps and stuff.

jseppi: I wonder then if there is a way to make sure there is a populated form of the expected class instead of checking a specific value. I'll defer to you though if you think that what is already here is ok :)

toolness: Hmm, I guess we could check the [template context](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#django.test.Response.context) for the `form` key and make sure it's populated...  I guess either seems OK to me?  I'd be more worried if `GS-123-4567` was something that might inadvertently be in the rendered HTML that *wasn't* actually part of the form, which would be a false positive or whatever, but that particular string seems fairly unique.

Hmm, that said, I guess I could see us using such a fake string in our example help text or something in the future, so maybe checking the template context is better.  I can bust up that change!

jseppi:
Hrm, codeclimate seems to have failed but not because of something in this PR. I guess after you merge in `develop` it will re-run and hopefully fix itself.

jseppi:
👍 Looks good to me

PR #730, opened 2016-09-12
Move final Bulk Upload POST and processing to step 2 view handler

### LINKED ISSUE #721: Bulk Upload: Step 3 405 just shows a blank page
I let my session time out between steps 2 and 3, hit the button to continue in step 2, was redirected to the login page, logged in, and then was redirected to Step 3. But I got a 405 and a completely blank page. Seems like we could handle that error better?

jseppi:
Closes #721


jseppi:
ready for review, @toolness

data_capture/views/bulk_upload.py
```diff
@@ -54,41 +55,44 @@ def region_10_step_1(request):
...
-        'step_number': 2,
-        'file_metadata': file_metadata,
-    })
+        return render(
+            request,
+            'data_capture/bulk_upload/region_10_step_2.html', {
+                'step_number': 2,
+                'file_metadata': file_metadata,
+            })
+    elif request.method == 'POST':
```
toolness: This can actually just be a `POST` since you've got the `@require_http_methods` decorator--in particular it'll increase our test coverage, as py.test won't think that we aren't testing the branch where the HTTP method is something *other* than `GET` or `POST`.  Um, unless py.test is actually smart enough to know that'll never happen because of the decorator, but I don't think it is...

Anyways, it's up to you really.  I do like the readability of explicitly testing to see if it's a `POST` so I'm conflicted myself.

toolness:
Noice, I like it!

data_capture/views/bulk_upload.py
jseppi: I like the explicitness also. Maybe I could move the `return` statement from `GET` branch down to be the implicit return value.

jseppi:
Ok, moved the `POST` branch logic to be the implicit code path. Let me know if 👍.

toolness:
Nice, it looks good!

It seems like a lot of the sample django code I've seen does the opposite and makes `GET` the implicit one but I think either is fine.  Feel free to merge!

PR #734, opened 2016-09-13
Make step-based workflows more accessibility-friendly, consolidate step logic

### LINKED ISSUE #413: Current state of steps widget is not perceivable by screen readers

toolness:
This supersedes #578 and fixes #413, as it was easier to just start a new PR instead of deal with the merge conflicts.  I'm also taking a slightly different approach that will make it easier to consolidate logic across all step-based workflows (i.e. region 10 bulk upload and price list upload).

Basically I'm doing this through the creation of a new `Steps` class, which encapsulates all the steps in a particular workflow.  Here's the docstring (which contains doctests that are verified during CI 🎉 ):

```
    The `Steps` class makes it easier to consolidate the logic of
    multi-step workflows.

    The `Steps` constructor takes a format string representing what
    the template path for each step looks like:

        >>> steps = Steps('data_capture/tests/my_step_{}.html')

    The `@step` view decorator can be used to "register" steps in the
    workflow.  Each step's function name is expected to end with
    a number, and the steps are expected to be defined in
    ascending order, e.g.:

        >>> @steps.step
        ... def step_1(request, step): pass

        >>> @steps.step
        ... def step_2(request, step): pass

    (What's that extra `step` argument?  We'll get to that in a moment!)

    Breaking this ordering, e.g. by skipping numbers, is not allowed:

        >>> @steps.step
        ... def step_5(request, step): pass
        ...
        ValueError: Expected "step_5" to end with the number 3

    At this point, our example workflow has two steps:

        >>> steps.num_steps
        2

    URL patterns for our steps are automatically defined, too, and can
    be included via Django's standard `include()` function. The names
    of each view will be identical to their view function's name:

        >>> steps.urls
        [<RegexURLPattern step_1 ^1$>, <RegexURLPattern step_2 ^2$>]

    Now, what's that `step` argument passed into each view function?

    It's a `StepRenderer` instance bound to the particular step that the
    view represents, and it provides some tools that make it easy to
    render steps.

    `StepRenderer` instances can also be retrieved manually if needed:

        >>> step1 = steps.get_step_renderer(1)

    A `StepRenderer` instance can be used to easily access
    commonly-used view logic, e.g.:

        >>> step1.number
        1

        >>> step1.steps.num_steps
        2

        >>> step1.description
        'step 1 of 2'

    Tools for templates and their contexts are also available.

    The `context()` function builds a context dictionary that always
    contains a `step` key referring to the `StepRenderer` instance for
    that step:

        >>> ctx = step1.context({'foo': 'bar'})
        >>> ctx['foo']
        'bar'
        >>> ctx['step']
        <StepRenderer for step 1 of 2>

    The `template_name` property always refers to the template for the step:

        >>> step1.template_name
        'data_capture/tests/my_step_1.html'

    And `render`/`render_to_string` ties everything together:

        >>> step1.render_to_string({'foo': 'bar'})
        'Hello from step 1 of 2! foo is bar.'
```

The API for the `Steps`/`StepRenderer` classes are inspired by Django's [`Paginator` / `Page`](https://docs.djangoproject.com/en/1.9/topics/pagination/) classes.

If these new classes seem useful, we can build on them more in future PRs to e.g. automatically render the steps widget for us, among other things.  I tried curbing my inner [architecture astronaut](http://www.joelonsoftware.com/articles/fog0000000018.html) to make this PR as small and non-disruptive as possible, but I'm not sure if I succeeded... Any feedback is appreciated!

## To Do

- [x] Add unit tests.
- [x] Add documentation (specifically, docstrings for `Steps` and `StepRenderer`).
- [x] Apply the `Steps` class to Region 10 bulk upload once #694 is merged. (it's currently just applied to price list upload.)

### DIFF
--- a/data_capture/views/price_list_upload.py
+++ b/data_capture/views/price_list_upload.py
@@ -7,11 +7,11 @@ from django.http import HttpResponseBadRequest
 from .. import forms
 from ..decorators import handle_cancel
 from ..schedules import registry
-from .common import add_generic_form_error, StepBuilder
+from .common import add_generic_form_error, Steps
 from frontend import ajaxform
 
 
-steps = StepBuilder(
+steps = Steps(
     template_format='data_capture/price_list/step_{}.html',
 )
 
--- a/data_capture/views/price_list_upload.py
+++ b/data_capture/views/price_list_upload.py
@@ -1,5 +1,4 @@
 import json
-from functools import wraps
 from django.views.decorators.http import require_http_methods
 from django.shortcuts import redirect
 from django.contrib.auth.decorators import login_required
@@ -17,22 +16,10 @@ steps = StepBuilder(
 )
 
 
-def gleaned_data_required(f):
-    @wraps(f)
-    def wrapper(request):
-        try:
-            d = request.session['data_capture:price_list']['gleaned_data']
-        except:
-            return redirect('data_capture:step_3')
-
-        return f(request, registry.deserialize(d))
-    return wrapper
-
-
 @steps.step
 @login_required
 @require_http_methods(["GET", "POST"])
-def step_1(request):
+def step_1(request, step):
     if request.method == 'GET':
         form = forms.Step1Form()
     else:
@@ -46,7 +33,7 @@ def step_1(request):
         else:
             add_generic_form_error(request, form)
 
-    return steps.render(1, request, {
+    return step.render(request, {
         'form': form,
     })
 
@@ -55,7 +42,7 @@ def step_1(request):
 @handle_cancel
 @login_required
 @require_http_methods(["GET", "POST"])
-def step_2(request):
+def step_2(request, step):
     # Redirect back to step 1 if we don't have data
     if 'step_1_POST' not in request.session.get('data_capture:price_list',
                                                 {}):
@@ -77,7 +64,7 @@ def step_2(request):
         else:
             add_generic_form_error(request, form)
 
-    return steps.render(2, request, {
+    return step.render(request, {
         'form': form
     })
 
@@ -86,7 +73,7 @@ def step_2(request):
 @handle_cancel
 @login_required
 @require_http_methods(["GET", "POST"])
-def step_3(request):
+def step_3(request, step):
     if 'step_2_POST' not in request.session.get('data_capture:price_list',
                                                 {}):
         return redirect('data_capture:step_2')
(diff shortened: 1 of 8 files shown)


data_capture/views/price_list_upload.py
```diff
@@ -106,19 +99,27 @@ def step_3(request):
...
-@gleaned_data_required
 @handle_cancel
-def step_4(request, gleaned_data):
+def step_4(request, step):
+    gleaned_data = None
+    try:
+        price_list = request.session['data_capture:price_list']
+        gleaned_data = price_list['gleaned_data']
+    except KeyError:
+        return redirect('data_capture:step_3')
```
toolness: This can be made a lot cleaner via `get_nested_item`, which is introduced in #729 (which hasn't yet been merged).

jseppi:
Was going through backlog/issues and wanted to bring up the card for [moving Upload Errors to Separate Step](https://trello.com/c/x9ZJf3Mr/133-move-data-upload-errors-to-a-separate-step-instead-of-the-current-combined-step) here. A quick glance of this PR looks like it is using the function name to determine the step number, but in the [mocks](https://gsa.invisionapp.com/d/main#/projects/8513772) related to the linked Trello card, the price list errors and the price list preview are both in Step 4 (and both show Step 4 highlighted in the steps widget).

toolness:
Ahh yeah, I was thinking about that too.  Hmm.  I'm curious how that will visualize on the steps widget too, though?  Will both appear to be in "step 4" but at different URLs or something?

toolness:
Oh sorry, you answered that question already!  Nevermind...

I guess we could have the decorator take an *optional* `step_number` so that it pulls it out of the view name by default, but otherwise specific info can be passed-in to override?  I guess that we'd also have to provide information about the specific template to use too or something, blah.

toolness:
Ok, so there's some different ways we can go here.

I don't think that we should explicitly add support for multiple sub-steps to `Steps`, because that would add a lot of complexity to the class, as well as documentation and tests and whatnot. However, I *do* think we should make it *possible* to have multiple sub-steps.

Given the current API for `Steps`/`StepRenderer`, there are some different ways this can be done:

1. Just have one actual view function but vary what it does based on a querystring parameter or something in the session.

2. Make the first substep be the "official" step, decorated as usual by the `@steps.step`.  For the rest of the substeps, make views *without* the `@steps.step` decorator.  Register them in `urls.py` and make them retrieve the proper `StepRenderer` via `steps.get_step_renderer()`. For example:

  ```python
@steps.step
def step_4(request, step):
      # Do the standard thing for step 4 here...

def step_4_optional_substep(request):
      step = steps.get_step_renderer(4)
      # Do stuff for our custom sub-step here.
      # Also make sure to register this view function in urls.py,
      # since our Steps instance doesn't "know" about it!
```

Thoughts?

jseppi:
This is super cool!

Regarding your sub-step options I agree we should probably refrain from implementing full substep-support.

I'm having some internal trouble deciding which option I like better though (especially without actually trying either of them in practice yet). I think I am leaning toward option 2 because I like the view function separation, and also I have a general unfounded aversion to query params.

toolness:
Woot sounds good!

I can think of one concrete downside to querystring args: there's not any easy way to tell Django what query args a particular view expects, and as a result, it means that we could be using the wrong querystring arg/values in our templates, and we wouldn't find it out until we did manual testing. OTOH, if the views are actual Django views registered with the URL mapper, and we use `{% url %}` to generate them in our templates, then Django will help ensure that the links we're generating are legit, and (hopefully) our unit tests will fail if they're not.

Should I go ahead and merge this?

jseppi:
👍 merge away!

Being able to specify the url for the sub-step might actually be somewhat nice so we could use eg `/step/3/errors` to give the "sub-step" a more meaningful URL (even though probably nobody would care).

PR #738, opened 2016-09-13
Restrict R10 Bulk Upload views and Price List Upload views based on permissions

### LINKED ISSUE #685: Only data administrators should be able to upload bulk data
As mentioned in #675, currently anyone who `is_staff` can upload bulk data, but we should limit the ability to Data Administrators specifically to enforce the principle of least authority.

jseppi:
This PR restricts the Region 10 Bulk Upload flow views to users with the `contracts.add_bulkuploadcontractsource` permission, and the Price List Upload views to users with the `data_capture.add_submittedpricelist` permission.

~~I accomplished this by genericizing `@contract_officer_perms_required` into a new `@role_permissions_required(role)` decorator. `@contract_officer_perms_required` has been replaced and removed.~~

This is done using Django auth's `@login_required` and `@permission_required(...)` decorators. The custom decorators that I had implemented have now been removed, per discussion further down in this PR.

To Do:
- [x] modify and test according to suggestion in https://github.com/18F/calc/pull/738#issuecomment-247035815

Closes #685


jseppi:
ready for review, @toolness

toolness:
This looks good!  Except that I just resolved the merge conflicts in my other PRs because of `@contract_officer_perms_required` and now I'm gonna have to resolve them again because of `@role_permissions_required` 😭 😢

toolness:
Hmm, it's interesting to me that data administrators no longer actually need to have `is_staff`.  In a practical sense I guess they will since it's needed to actually access the Django admin panel...

Another thing I just realized is that my intent in mapping granular permissions to roles was so that individual parts of site functionality would test against those granular permissions, rather than the roles--that is, for price list upload we should really just be checking to see if the user has `data_capture.add_submittedpricelist` rather than if they're specifically a contract officer, and for R10 bulk upload we should check to see if the user has `contracts.add_bulkuploadcontractsource` (which the data administrator role doesn't actually have yet!).

Can you add `contracts.add_bulkuploadcontractsource` to the Data Administrator permissions list?  That way at least the list of permissions still reflects the reality of what they're capable of.

jseppi:
I added `contracts.add_bulkuploadcontractsource` to the `Data Administrators` perms.

Regarding the checking of permissions rather than roles, I get what you're saying about checking individual specific permissions rather than roles. It does not really make complete sense to be checking for eg `'auth.add_user'` in the R10 upload views when really just `'contracts.add_bulkuploadcontractsource'` is needed, which is (part of what is) being done with how I've implemented the decorator in this PR.

Hrm. I do find the way I've done here more readable since devs can just see which role is intended to access the views. I suppose the potential danger in how I've done this is that it could be too strict at some point in the future. For example, say we want to create a new group of users that can do bulk uploads but can't add new users themselves (so not quite `Data Administrators`), we'd have to add a new role and modify all the R10 upload views to check against that role.

It would be a pretty easy change for me to scrap this and just use Django's built-in `@permission_required` decorator (which their [docs](https://docs.djangoproject.com/en/1.10/topics/auth/default/#the-permission-required-decorator) advise using in conjunction with `@login_required` if you want to first check for `is_authenticated`.

toolness:
Cool!  As you said, since we don't *really* need the permission-based thing right this minute, I think the way we've got things going is fine, so it's really up to you if you want to replace with `@permission_required`.  Though if we don't address it, we might want to file a low-priority issue linking to this discussion, just so future folks know our thought process.

I totally get what you're saying about the `@role_permissions_required` being more readable, though.  One alternative is to add the following to `initgroups` (or some similar place):

```python
R10_BULK_UPLOAD_PERMISSION = `contracts.add_bulkuploadcontractsource`

ROLES['Data Administrator'] = set([
    R10_BULK_UPLOAD_PERMISSION,
    # And so on...
])
```

Then the R10 bulk upload views could just import and reuse `R10_BULK_UPLOAD_PERMISSION`, instead of explicitly referring to `contracts.add_bulkuploadcontractsource`, which might improve readability.  It also gives us room to convert the permission into a [custom permission](https://docs.djangoproject.com/en/1.10/topics/auth/customizing/#custom-permissions) in the future, if we find a need to.

jseppi:
Ok, I am going to modify according your suggestion there. I do like leaning more on the built-in decorators and decreasing our code surface area.

jseppi:
ok, ready for review again, @toolness

jseppi:
errr, false alarm, should adjust the tests to test for the specific permission, not the group now.

jseppi:
alright @toolness: tests adjusted and `develop` merged in

data_capture/tests/common.py
```diff
@@ -91,11 +91,19 @@ def create_user(self, username, password=None, is_staff=False,
...
                                 is_staff=is_staff,
                                 is_superuser=is_superuser,
                                 groups=groups)
+        if permissions:
+            splitnames = [perm.split('.') for perm in permissions]
+            user.user_permissions = [
+                Permission.objects.get(codename=codename,
+                                       content_type__app_label=app_label)
+                for app_label, codename in splitnames
+            ]
```
toolness: This is very similar/identical to code in `initgroups.py`--I wonder if we should consolidate it?  Not that big a deal I guess, since it's only repeated twice, so it's up to you.

data_capture/decorators.py
```diff
@@ -57,7 +56,17 @@ def wrapper(request, *args, **kwargs):
 def staff_login_required(function=None,
                          redirect_field_name=REDIRECT_FIELD_NAME,
                          login_url=None):
```
toolness: Wait, do we even need `staff_login_required` anymore?  As far as I can tell, you removed our only uses of it in this PR (aside from tests)...

jseppi: It is used in `hourglass.urls` to wrap the admin site login. So actually it should probably be moved into `hourglass/`.

jseppi: ok, moved `staff_login_required` and associated tests to `hourglass/`, though now some (not much) of the `data_capture.tests.common.BaseTestCase` stuff is repeated a in `hourglass.tests` (and `hourglass.tests` is even longer than it was).

data_capture/tests/common.py
jseppi: ✔️

toolness:
Awesome, this looks great!  Feel free to merge.

Issue #538, opened 2016-08-09
Upload widget text should properly specify file formats


jseppi:
<img width="683" alt="screen shot 2016-08-09 at 2 55 14 pm" src="https://cloud.githubusercontent.com/assets/697848/17529457/4a65dfde-5e41-11e6-830b-0cbf06a6841d.png">

For Schedule 70 import, as the parser is currently written, uploading a CSV won't work. The text in the upload widget is therefore misleading because only the specific S70 Excel price list will actually work.

toolness:
Now that the user chooses their schedule in a step prior to the one where they upload a spreadsheet, we can dynamically set this value based on the schedule!  For example, since the "Fake Schedule" for development only allows CSV, we can generate the widget to only accept (and mention) CSV, while for Schedule 70 we can have it only accept and mention XLS/XLSX.

PR #741, opened 2016-09-13
Pass schedule into price_list.Step3Form constructor

toolness:
`price_list.Step3Form` used to have `schedule` as a user-facing field but we removed it in [the great form rejiggering of 2016](https://github.com/18F/calc/pull/583).  However, the field was still kept around in the actual form class.  This removes it as a field and instead expects the schedule to be passed-in to the form constructor.

Aside from making the code a bit more understandable, this will also help with #538, as it'll be easier to dynamically set the upload widget properties (accepted file types and help text) based on the schedule.

### DIFF
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -48,6 +48,14 @@ class Step3Form(forms.Form):
     file = forms.FileField(widget=UploadWidget())
 
     def __init__(self, *args, **kwargs):
+        '''
+        This constructor requires `schedule` to be passed in as a
+        keyword argument. It should be a string referring to the
+        fully-qualified class name for an entry in the schedule
+        registry, e.g.
+        "data_capture.schedules.fake_schedule.FakeSchedulePriceList".
+        '''
+
         self.schedule = kwargs.pop('schedule')
         super().__init__(*args, **kwargs)
 
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -45,22 +45,19 @@ class Step2Form(forms.ModelForm):
 
 
 class Step3Form(forms.Form):
-    # TODO: We should figure out a way of getting rid of this field, since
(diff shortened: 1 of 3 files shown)


data_capture/forms/price_list.py
```diff
@@ -45,22 +45,19 @@ class Meta:
...
-    # we're not actually asking the user for it anymore.
-
-    schedule = forms.ChoiceField(
-        choices=registry.get_choices
-    )
-
     file = forms.FileField(widget=UploadWidget())
 
+    def __init__(self, *args, **kwargs):
+        self.schedule = kwargs.pop('schedule')
```
jseppi: neat, I didn't know this pattern

jseppi: hrm, actually, any reason to get it from `kwargs` instead of being more explicit about including it in the parameter list to `__init__()`? It looks like you need to specify it or things will break, right? Or maybe that isn't doable with how forms are used (also I'm not entirely clear on the python's `*args` and `**kwargs` stuff).

jseppi:
Somewhat minor comment, but other than that, 👍

data_capture/forms/price_list.py
toolness: Hehe, it is courtesy of [stack overflow](http://stackoverflow.com/a/1697770) :)

My original implementation actually made `schedule` optional, until I realized that it'd be required to customize the upload widget help text and stuff (and would also mean less branching in the code).

Unfortunately I don't know of a way to explicitly include it in the parameter list other than by copying it out from `Form`, so that the "normal" conventions of constructing forms still work (e.g. passing the POST data as the first arg).  But I'm a bit worried about doing that b/c of future changes to the `Form` constructor in future versions of Django or whatever, blech.  I could add a comment or docstring mentioning the requirement of `schedule` I guess?

jseppi: > add a comment or docstring

👍

Issue #737, opened 2016-09-13
Display chosen schedule, contract # and vendor name on step 2


hbillings:

toolness:
You'll probably want to "resurrect" the actual `Step1Form` here and pull from its `cleaned_data` for displaying that data... Um... I think I will also change `Step1Form`'s `cleaned_data` to actually convert the schedule into a `BasePriceList` subclass (the raw `POST` data actually refers to the dotted Python class naming a `BasePriceList` subclass)...

toolness:
Ok, as of #742 you can now rez `Step1Form` and grab its `cleaned_data['schedule_class']` to get the actual `BasePriceList` subclass associated with the preferred schedule (just grab its `title` property to get the human-friendly schedule name).

toolness:
Er, might want to actually wait until #748 lands before starting this, as it adds a new `get_step_form_from_session()` helper function that can make this particularly easy.

PR #742, opened 2016-09-14
Display preferred schedule during fallback

### LINKED ISSUE #663: During schedule fallback, preferred schedule is not displayed
This was mentioned in https://github.com/18F/calc/pull/583#discussion_r77549567 but it actually has been a bug for a while:

> screen shot 2016-09-05 at 1 27 56 pm

How to reproduce:

1. Select Schedule 70 as the schedule.
2. Upload `fake_schedule_example.csv`.

toolness:
This fixes #663.

It also makes `Step1Form` take care of automatically converting the schedule the user chose into a `BasePriceList` subclass, which should help with #737.

I also discovered that even though the fallback logic was working, the actual `schedule` property of the submitted price list was still being set to the user's preferred schedule (rather than the one we fell back to), so I fixed that too.

### DIFF
--- a/data_capture/templates/data_capture/price_list/step_4.html
+++ b/data_capture/templates/data_capture/price_list/step_4.html
@@ -8,7 +8,7 @@
 {% if not is_preferred_schedule %}
 <p>
   <strong>Note:</strong> You specified that the price list you were
-  uploading was for <em>{{ preferred_schedule.title }}</em>, but it
+  uploading was for <em>{{ preferred_schedule_title }}</em>, but it
   appears to be for <em>{{ gleaned_data.title }}</em>. If this is
   incorrect, please upload a different file.
 </p>
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -19,6 +19,13 @@ class Step1Form(forms.ModelForm):
             'vendor_name',
         ]
 
+    def clean(self):
+        cleaned_data = super().clean()
+        schedule = cleaned_data.get('schedule')
+        if schedule:
+            cleaned_data['schedule_class'] = registry.get_class(schedule)
+        return cleaned_data
+
 
 class Step2Form(forms.ModelForm):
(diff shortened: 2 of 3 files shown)


data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -8,7 +8,7 @@
 {% if not is_preferred_schedule %}
 <p>
   <strong>Note:</strong> You specified that the price list you were
-  uploading was for <em>{{ preferred_schedule.title }}</em>, but it
+  uploading was for <em>{{ preferred_schedule_title }}</em>, but it
```
toolness: Um, so unfortunately I have **NO CLUE** why this was appearing to be blank, but when I explicitly set it to a string rather than pulling the `title` from the `Class` instance, it seemed to work.  Bizarre.

jseppi: Whoa, that is weird o_O

toolness: Yeah, what's even weirder is that originally that was somehow capitalizing the string weird, so it'd say "It schedule 70" instead of "IT Schedule 70".  Inspecting the template context via DjDT indicated that the right string was in the template context, but the capitalized version would show up in the generated HTML.  Then after a few reloads or something, the problem went away... Realllly weird. 😅

jseppi: Now it is really bugging me.
Was somehow like a Django `model` attribute/field colliding with it?

toolness: Hmm, the `preferred_schedule` isn't actually a descendant of the Django `Model` class though, it's just a class object representing a subclass of `BasePriceList`, which is our own custom abstract base class that we created from scratch.  I guess we could do a `git bisect` or whatever to go back in history to the point that this broke (and I'm not sure when that was).

jseppi:
weirdness aside, 👍

Issue #700, opened 2016-09-08
Make Cancel buttons submit POST requests in price list upload


toolness:
#583~~, which at the time of this writing is ABOUT TO BE MERGED,~~ only implements the *server* side of handling cancelling--the actual cancel buttons are *links* that go back to step 1 rather than POST requests.  We need to turn them into POST requests and also make sure they work w/ the step that's an ajaxform.

jseppi:
Related: https://github.com/18F/calc/issues/665

toolness:
So I tried doing this but failed.

Issues/notes/concerns:

* We need to remember to put a [`formnovalidate` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Attributes) on all our cancel buttons so that any HTML5 form validation is ignored.  Otherwise the user won't be able to click cancel unless they've filled out all the fields properly, which is ridiculous.
* It looks like pressing enter on a text input in a form will submit the first submit button in the DOM.  With our page layout, that actually means that the *cancel* button is the default submit button, which is probably not what the user wants!  We might be able to get around this by using a flexbox layout with `flex-direction: row-reverse`, so that the OK button appears first in source/DOM order but to the right of the cancel button visually. Or we might be able to use `float` in a funky way.
* Currently our ajaxform component *always* includes the `cancel` name when submitting the form, because it's just iterating through all the form elements and adding their name/value pairs to `FormData`.  This means that even the  *real* submit button acts like a cancel button!  And unfortunately it also seems as though [`submit` events](https://developer.mozilla.org/en-US/docs/Web/Events/submit) don't contain information about *which* submit button was clicked, so it may be challenging to figure out, though [stack overflow offers some possible solutions](http://stackoverflow.com/questions/2298635/can-i-determine-which-submit-button-was-used-in-javascript).

Blerg.  Given all these issues, I'm wondering if we want to just consider server-side deletion of multi-step form state a "nice to have". If we do, then we can actually leave the cancel button as a simple link, but also convert it into a web component that submits an ajax request on click/enter/space...  Blerg, even that might be an annoying amount of work.

jseppi:
>.< 

Thanks for trying and documenting all these issues.

I wonder if we could put the cancel submit button into its own `<form>` that `POST`s to an action to clear `data_capture:` session keys instead of trying to have one form with two submits.

toolness:
[...]

PR #740, opened 2016-09-13
Button row redux

hbillings:
Closes #718 and #651.
screen shot 2016-09-13 at 4 51 06 pm

PR #743, opened 2016-09-14
Support multiple submit buttons in ajaxform

toolness:
This fixes the main remaining issue mentioned in https://github.com/18F/calc/issues/700#issuecomment-246176550 by having `ajaxform` check to see if a button/input with `type="submit"` is the currently focused element on the page before adding its name and value to a `FormData` instance.

The functionality can be manually tested through the ajax form in the style guide. I tried it on IE11, Chrome, Firefox, and Edge, and it works on all of them!  🎉

### DIFF
--- a/styleguide/templates/styleguide_ajaxform_example.html
+++ b/styleguide/templates/styleguide_ajaxform_example.html
@@ -24,5 +24,7 @@
     {{ form.file }}
   {% endwith %}
 
-  <button type="submit" class="button-primary">Submit</button>
+  <button type="submit" class="button-primary" style="display: inline-block">Submit</button>
+
+  <button type="submit" class="button-secondary" name="cancel" formnovalidate style="display: inline-block">Cancel</button>
 </form>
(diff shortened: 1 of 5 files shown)


toolness:
Oh, this doesn't actually change the ajaxform in our price list upload form to have a cancel button--I figured I'd wait for #740 to land to do that, so as to avoid merge conflicts.

styleguide/templates/styleguide_ajaxform_example.html
```diff
@@ -24,5 +24,7 @@
     {{ form.file }}
   {% endwith %}
 
-  <button type="submit" class="button-primary">Submit</button>
+  <button type="submit" class="button-primary" style="display: inline-block">Submit</button>
+
+  <button type="submit" class="button-secondary" name="cancel" formnovalidate style="display: inline-block">Cancel</button>
```
toolness: Kind of weird that I have to manually set these to `display: inline-block`, but once #740 lands, we can change them to use the new form button row styling.

jseppi: Maybe leave out the `style` attributes so that there is less to deal with when merging #740?

jseppi:
This seems 👍

toolness:
WooooooooT!

Issue #727, opened 2016-09-12
Add "I'm not sure what to upload" content to price list upload form


toolness:
We need to add this:

> <img width="938" alt="screen shot 2016-09-01 at 9 03 36 am" src="https://cloud.githubusercontent.com/assets/124687/18168311/0ba95d90-7023-11e6-9e82-1dd7cee9fcf3.png">

This will involve using an [`<expandable-area>`](https://github.com/18F/calc/pull/657) and we should probably delegate the actual expanded content to the `BasePriceList` subclass that's being used.

It'll probably be easy to take care of #538 at the same time.

PR #748, opened 2016-09-14
Allow schedules to provide upload examples and help text

### LINKED ISSUE #538: Upload widget text should properly specify file formats

toolness:
This fixes #538 by adding a `upload_widget_extra_instructions` property to `BasePriceList`.

It also lays a foundation for fixing #727 by adding an `upload_example_template` property and `render_upload_example()` method.  It doesn't actually provide the actual *content* for Schedule 70, though--we'll address that in a future PR.  (And it's fine to merge this now b/c the "I'm not sure what to upload" expander only appears if there's upload example content to fill it with.)

While I didn't add upload example content for schedule 70 (this is addressed in #754), I did add some for the fake schedule, which is only ever shown to developers:

> screen shot 2016-09-15 at 9 58 13 am

Aside from that, it also makes a few refactorings to the price list views, by adding a `get_step_form_from_session()` helper function, which simplifies some things.

It also adds a new test suite for `BasePriceList` since we never had one.

### DIFF
--- a/data_capture/templates/data_capture/price_list/step_3.html
+++ b/data_capture/templates/data_capture/price_list/step_3.html
@@ -7,13 +7,13 @@
 {% endblock %}
 
 {% block step_body %}
+{% if upload_example %}
+<expandable-area>
+  <p><strong>I don't know what to upload.</strong></p>
+  <div>
+    {{ upload_example }}
+  </div>
+</expandable-area>
+{% endif %}
 {% include 'data_capture/price_list/upload_form.html' %}
 {% endblock %}
-{% if show_debug_ui %}
-{% load static %}
-<p class="dev-note">
-  <strong>Developers:</strong>
-  Download <a href="{% static 'data_capture/fake_schedule_example.csv' %}"><code>fake_schedule_example.csv</code></a>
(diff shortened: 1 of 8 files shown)


data_capture/templates/data_capture/price_list/step_3.html
```diff
@@ -7,13 +7,13 @@
...
 {% include 'data_capture/price_list/upload_form.html' %}
 {% endblock %}
-{% if show_debug_ui %}
-{% load static %}
-<p class="dev-note">
-  <strong>Developers:</strong>
-  Download <a href="{% static 'data_capture/fake_schedule_example.csv' %}"><code>fake_schedule_example.csv</code></a>
-  and choose "Fake Schedule" from the drop-down to get started.
-</p>
-{% endif %}
```
toolness: Moved this content to be the "upload example content" for the fake schedule.

PR #750, opened 2016-09-14
Strip non-numeric chars from S70 price cells during gleaning

jseppi:
This PR adds a simple function called `strip_non_numeric` that is used as a "coercer" function in all the price-type cells when gleaning data from an uploaded S70 price list.

In many of the sample S70 price lists we have received, due to both human data input errors and Excel weirdness (I think between PC and Mac versions), often cells that are formatted as currency are parsed `xlrd` as text and will include the currency symbol, commas, and sometimes even padding-spaces.

To Do:
  - [x] tests


jseppi:
Ready for review, @toolness

jseppi:
I added a new XLSX file for testing that that price cells with weird formatted are properly gleaned. Right now it is in `static/data_capture/`, but perhaps we should instead start a `tests/fixtures/` directory because I foresee more of these test files being necessary.

toolness:
Hmm, since these XLSX files do get kinda big, and since we're likely going to change them over time as the official price lists change, I wonder if we want to use [Git Large File Storage](https://git-lfs.github.com/) for storing them?

jseppi:
Yeah, I'm not a huge fan of putting additional binary files in the repo either :\ I only have a passing familiarity with git-lfs though. I wonder if it works easily with Travis.

The price list I've added in this PR is only 280kb, and I'd like to keep the fixture files relatively constrained to the functionality they are testing, but you're right in that if we have a bunch them and change them often, the repo size will balloon.

data_capture/tests/test_s70.py
```diff
@@ -28,6 +28,16 @@ def uploaded_xlsx_file(content=None):
...
 
 
+class CoercerTests(TestCase):
+    def test_strip_non_numeric_works(self):
+        s = '$ 5,600.72  '
+        self.assertEqual('5600.72', s70.strip_non_numeric(s))
+
+    def test_strip_non_numeric_works_for_non_string_input(self):
+        s = 1234
+        self.assertEqual('1234', s70.strip_non_numeric(s))
```
toolness: I think these would actually be really great as doctests, since they'd help folks reading `s70.py` understand what `strip_non_numeric` is actually *doing*, without having to find its unit tests.  But it's up to you if you want to actually convert them over, since these tests are fine too.

(If you want to run only a doctest, you can do it w/ e.g. `py.test data_capture/schedules/s70.py`.)

toolness:
Uhhh, I didn't mean to submit my last comment as my review synopsis, lol.  

Basically this looks 👍 to me!  Regarding git LFS, we could either re-do this PR and add the new XLS file via LFS, or we could just merge this one, taking the 200K repo bloat, and do LFS stuff later.  Up to you!

jseppi:
Cool, opened #755 for adding git-lfs

Issue #753, opened 2016-09-15
For schedule 70, offer a download link to the PPL template


toolness:
@tram mentioned we should do this. We currently do have a PPL template with one row filled out for purely testing purposes--I think this download would basically be that file, but without the one row filled out.

toolness:
I think a good place to put this would be under the "I don't know what to upload" expander content for schedule 70, which is being added in #754.

toolness:
This isn't hard to add--what should the copy say, though?  Something below the "I don't know what to upload" screenshot like:

> If you're not using a spreadsheet like this, please download the [Schedule 70 Price List Template](#).

?

jseppi:
maybe 
> ... please download and fill out ...
?

toolness:
Radical!

Is there a newer XLSX file available than the one Kelly gave us at the beginning of this project?  If not, I guess I'll just take our "PPL with one row added", remove the row, and use that as the template...

jseppi:
I don't _think_ so. Maybe @tram would know.

tram:
Nope! Still the same one.

PR #754, opened 2016-09-15
Add schedule 70 upload example table

### LINKED ISSUE #727: Add "I'm not sure what to upload" content to price list upload form

toolness:
~~**Note: This is a PR against #748, *not* `develop`.**~~

This fixes #727.

> screen shot 2016-09-15 at 9 49 31 am

### DIFF
--- a/data_capture/schedules/s70.py
+++ b/data_capture/schedules/s70.py
@@ -10,6 +10,41 @@ from contracts.models import EDUCATION_CHOICES
 from contracts.loaders.region_10 import FEDERAL_MIN_CONTRACT_RATE
 
 
+DEFAULT_SHEET_NAME = '(3)Labor Categories'
+
+EXAMPLE_SHEET_ROWS = [
+    [
+        r'SIN(s) PROPOSED',
+        r'SERVICE PROPOSED (e.g. Job Title/Task)',
+        r'MINIMUM EDUCATION/ CERTIFICATION LEVEL',
+        r'MINIMUM YEARS OF EXPERIENCE',
+        r'COMMERCIAL LIST PRICE (CPL) OR MARKET PRICES',
+        r'UNIT OF ISSUE (e.g. Hour, Task, Sq ft)',
+        r'MOST FAVORED CUSTOMER (MFC)',
+        r'BEST  DISCOUNT OFFERED TO MFC (%)',
+        r'MFC PRICE',
+        r'GSA(%) DISCOUNT (exclusive of the .75% IFF)',
+        r'PRICE OFFERED TO GSA (excluding IFF)',
+        r'PRICE OFFERED TO GSA (including IFF)',
+        r'QUANTITY/ VOLUME DISCOUNT',
+    ],
+    [
+        r'132-51',
+        r'Project Manager',
+        r'Bachelors',
+        r'5',
+        r'$125',
+        r'Hour',
+        r'All Commercial Customers',
+        r'7.00%',
+        r'$123.99',
+        r'10.00%',
+        r'$110.99',
+        r'$115.99',
+        r'15.00%',
+    ]
+]
+
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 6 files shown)


data_capture/schedules/s70.py
```diff
@@ -10,6 +10,41 @@
...
+        r'Hour',
+        r'All Commercial Customers',
+        r'7.00%',
+        r'$123.99',
+        r'10.00%',
+        r'$110.99',
+        r'$115.99',
+        r'15.00%',
+    ]
+]
```
toolness: I thought that defining this as a data structure in `s70.py` would increase the chances of actually keeping it up-to-date, rather than shunting it out to a HTML template that we probably won't look at very often.

toolness: Also, I guess an even nicer approach might have been to pull the info directly out of the sample XLSX file that's already in our repo, but that seemed complicated and stuff, so I didn't.  (Our gleaner pulls out the rows but not the column names, so I couldn't just reuse that.)

data_capture/schedules/s70.py
```diff
@@ -10,6 +10,41 @@
...
+DEFAULT_SHEET_NAME = '(3)Labor Categories'
+
+EXAMPLE_SHEET_ROWS = [
+    [
+        r'SIN(s) PROPOSED',
+        r'SERVICE PROPOSED (e.g. Job Title/Task)',
+        r'MINIMUM EDUCATION/ CERTIFICATION LEVEL',
+        r'MINIMUM YEARS OF EXPERIENCE',
+        r'COMMERCIAL LIST PRICE (CPL) OR MARKET PRICES',
+        r'UNIT OF ISSUE (e.g. Hour, Task, Sq ft)',
```
jseppi: Hrm, I wonder if we should only be keeping rows for which the unit of issue is "hour." Definitely a separate issue from this PR, though.

data_capture/templates/data_capture/price_list/upload_examples/schedule_70.html
```diff
@@ -0,0 +1,27 @@
+<p>
+  For Schedule 70, this is your completed Price Proposal Template.
+  We'll only read the <strong>{{ sheet_name }}</strong> tab. It
+  looks like this:
```
jseppi: maybe "It should look like this:"

data_capture/tests/test_s70.py
```diff
@@ -200,3 +200,8 @@ def test_to_error_table_works(self):
         table_html = s.to_error_table()
         self.assertIsNotNone(table_html)
         self.assertTrue(isinstance(table_html, str))
+
+    def test_render_upload_example_works(self):
+        html = s70.Schedule70PriceList.render_upload_example()
+        self.assertTrue('looks like this' in html)
+        self.assertTrue('Bachelors' in html)
```
jseppi: These tests seem a little weird to me, but so do the `test_to_table_works` and `test_to_error_table_works` right above it. I don't know how to better the functionality in any of these, so I guess what we have is good enough.

data_capture/schedules/s70.py
jseppi: I wonder if we should provide a S70 price list template download for all users so that if they have a price list that is way off, they could format it into what we are expecting. I think @tram mentioned that could potentially be helpful.

data_capture/tests/test_s70.py
toolness: Yeah, testing views is weird because of how brittle they can get.  This is mostly just a smoke test to make sure that it doesn't throw an exception.

data_capture/schedules/s70.py
toolness: Yep, totes, I filed that as #753 :)

PR #756, opened 2016-09-15
use defusedxml to open excel workbooks securely

### LINKED ISSUE #749: Use defusedxml to guard against XML vulnerabilities in XLSX uploads
Ref and example usage of `defusedxml`: http://xlrd.readthedocs.io/en/latest/vulnerabilities.html

jseppi:
Closes #749. Implemented according to guidance at http://xlrd.readthedocs.io/en/latest/vulnerabilities.html

I'm not quite sure how to test this. My first thought was to use bandit to make sure the plain `xlrd.open_workbook` is not used anywhere, but I don't quite understand if either of the methods for registering a bandit plugin would work for us: https://github.com/openstack/bandit#extending-bandit

Also, maybe the method described in the xlrd docs is a little overkill and all that needs to happen is a call to `defusedxml.defuse_stdlib()` in `data_capture/__init__.py`. @toolness, thoughts/ideas?

FYI `defuse_stdlib()` is an "experimental" feature to monkeypatch all stdlib modules: https://pypi.python.org/pypi/defusedxml


toolness:
Nice!!

Um... so if someone uploads a XLSX file that `defusedxml` doesn't like... what will end-users see?

toolness:
Oh blergh. When it comes to security, I am especially unsettled by words like "untested" and "experimental"... 😣

jseppi:
uhhh good question :) 
They'll see whatever the page displays when an exception is raised during gleaning, for example:

screen shot 2016-09-15 at 3 31 56 pm

jseppi:
> I am especially unsettled by words like "untested" and "experimental"...

Yeaaaaahhh...

Uh, I'm not sure what else to do. Maybe monkeypatching `xlrd` directly to use `defusedxml.ElementTree.parse`. I think this is one place where the monkeypatching could be done: https://github.com/python-excel/xlrd/blob/b718ba60773c6ba7fb8c3fb151f2f090bcde359f/xlrd/xlsx.py#L25

toolness:
Ok, this was fixed by #833!

PR #751, opened 2016-09-15
Restyle form inputs for step 1 & 2

hbillings:
Closes #739, #736 

Really, the only difference for step 1 is graying the help text. Here's what step 2 looks like:
screen shot 2016-09-14 at 10 35 53 pm

I ended up just writing the HTML for the radio button form directly in the template. Ugh. I investigated making a custom radio button widget, but TBH after 40 minutes of wrestling with it, I gave up for the more expedient approach. (I'm sure I was making it more difficult than it needed to be!)

PR #744, opened 2016-09-14
Add pop-out link to styleguide example forms

toolness:
Some of our example forms in the style guide--the ones that can actually be submitted, which are useful for manual testing--are also embedded as "standalone" views with their own URLs, which is particularly useful for manual testing.  This just adds a "pop-out" link that makes those views discoverable:

> popout

The SVG happens to be taken from Font Awesome but we can really just use whatevs.

PR #757, opened 2016-09-15
Add a USWDS radio widget

toolness:

This adds a USWDS-style radio widget.

> screen shot 2016-09-15 at 2 38 59 pm

I wanted to add the pop-out link to the styleguide form as per #744 but it looks like #751 hasn't merged those changes in, so I left out the pop-out link.

Note that this just adds the widget and an example use of it in the style guide--it doesn't actually change #751 to use it.

### DIFF
--- a/frontend/radio.py
+++ b/frontend/radio.py
@@ -1,0 +1,33 @@
+from django import forms
+from django.utils.html import format_html
+
+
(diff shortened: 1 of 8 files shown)


toolness:
Thanks! @hbillings if this looks good to you, feel free to merge it into your PR.

PR #758, opened 2016-09-15
[WIP] add simple coercer for education level

### LINKED ISSUE #755: Make test fixtures directory and use git-lfs for storage
We should create a directory of test fixtures for our price list parsing, especially as we start adding more complex coercions. Unfortunately XLSX files are binary, so they are not great to store in a git repo. Instead we should also use and (document in our README) [Git Large File Storage](https://git-lfs.github.com/), as suggested by @toolness in https://github.com/18F/calc/pull/750#issuecomment-247346391

jseppi:
Attempt at adding a simple coercer function for the education level field. The new `extract_min_education` function just tries to find valid values in the given text by stripping punctuation and lowercasing them both. See the [doctests](https://github.com/18F/calc/pull/758/files#diff-494ded7f44a6451dbc147d388e2bab59R27) for examples of what will get returned.

This PR also adds usage of `git-lfs`, so once it is merged we will all need to install the `git-lfs` extension. Closes #755.

```txt
-- Results before this coercer --

Total rows attempted: 294
Total valid rows: 62 (21.09%)
Total invalid rows: 232 (78.91%)

Num files attempted: 54
Num files not parsed at all: 31

Error counts by field:
  education_level: 231
  min_years_experience: 109
  price_including_iff: 36
  labor_category: 32

-- Results after adding this coercer --

Total rows attempted: 294
Total valid rows: 105 (35.71%)
Total invalid rows: 189 (64.29%)

Num files attempted: 54
Num files not parsed at all: 31

Error counts by field:
  education_level: 142
  min_years_experience: 109
  price_including_iff: 36
  labor_category: 32
```

To Do:
- [ ] fixture tests
- [x] tokenize on whitespace boundaries so that zzzBacherlorsYYY will not be a match


jseppi:
A possible bit of weirdness due to basically just checking for substrings is that if a key word is misspelled in the middle of itself, it won't work, but if it has some extra characters at the beginning or end, it it.

Ex:
'BaXYZchelors' won't work, but 'zzzBacherlorzzz' will.

This behavior would probably be relatively simple to correct though if I make this code slightly more complex and split on whitespace. A wrinkle to that is that one of the desired matches is "High School", which has a space in it, but that shouldn't be too bad to take into account.

jseppi:
Welp, I screwd up this branch with trying to add git-lfs, so I'm closing this in favor of #764

PR #759, opened 2016-09-16
Get rid of html.js / html.no-js classes

toolness:
As mentioned in https://github.com/18F/calc/pull/654#r79152047, the `js` and `no-js` classes applied to `<html>` a la Modernizr actually violate the [Apply no style before its time](http://adaptivewebdesign.info/1st-edition/read/chapter-4.html) maxim of progressive enhancement.  In particular, it could be the case that this JS executes and applies the `js` style to the page, yet the data capture bundle fails to be retrieved for some reason, thus resulting in a broken page.

Beyond that, though, removing these classes, and our upload widget's reliance on them, actually *simplifies* our CSS and reduces the possibility of our site appearing to be broken.

Fortunately, removing the need for these classes was also easy, since the only component that relies on them is the upload widget.

### DIFF
--- a/frontend/source/js/data-capture/upload.js
+++ b/frontend/source/js/data-capture/upload.js
@@ -199,7 +199,6 @@ class UploadWidget extends window.HTMLElement {
 
     if (!this._checkForAjaxFormParent() || !HAS_BROWSER_SUPPORT ||
         supports.isForciblyDegraded(this)) {
-      $el.addClass('degraded');
       this.isDegraded = true;
       return finishInitialization();
     }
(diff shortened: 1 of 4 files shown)


frontend/source/js/data-capture/upload.js
```diff
@@ -209,6 +208,8 @@ class UploadWidget extends window.HTMLElement {
 
     // The content of the upload widget will change when the user chooses
     // a file, so let's make sure screen readers let users know about it.
+    // Note that this is also used by CSS rules to determine whether or
+    // not the widget has actually been upgraded by JS.
     $el.attr('aria-live', 'polite');
```
jseppi: I'm a little hesitant about the overloading of this attribute to determine if the widget is upgraded, mostly because it seems like it could get lost/missed by a future developer especially in the styling rules.

jseppi:
Minor comment that you can address or not, otherwise 👍

frontend/source/js/data-capture/upload.js
toolness: I think that I'm OK with this because of the fact that `aria-live` implies dynamic functionality, though... So in some sense it's not really "overloading" even.  Hmmm.  Gustafson's book also explicitly recommends using these kinds of attributes to aid in CSS styling when appropriate, I think, so I'll leave this one be for now, but we can revisit it later if we want!

PR #760, opened 2016-09-16
Add tests for radio.py

toolness:

This adds a test suite for `frontend/radio.py`.

It also slightly changes the implementation of `radio.py` to raise an error if the radio doesn't have an `id`: because of the fact that USWDS-style radios have labels as sibling elements instead of parents, the radio *needs* an `id` or else the browser won't be able to associate the label with it.  So now the radio widget raises an exception if it isn't given an `id`, instead of staying silent about it and generating broken markup.


toolness:
This is ready for review @hbillings, feel free to merge it into your PR if you think it looks okay!

PR #762, opened 2016-09-16
[WIP] Try finding data in variant S70 sheet names and table formats

jseppi:
Here is an attempt at being able to take in more price lists variants. I noticed that many of the sample workbooks that @tram found us were in a different format than the one we have been developing against. Fortunately, the the price list table structure all the necessary columns, but 1) in a different sheet name, 2) with some additional columns interspersed, and 3) on a different starting rownum.

To Do:
- [ ] use column heading instead of column index to find which column holds which field (similar to the R10 importer) -- implemented in #767
- [ ] instead of hardcoding the `rownum` to find the table, instead try to find the row containing headers -- implemented in #765
- [ ] tests
- [ ] fix upload_example language


jseppi:
Closing this because we decided to not support variant S70 price list templates.

PR #763, opened 2016-09-16
Improvements to #751 (form input restyling)

toolness:

This addresses @jseppi's comments at https://github.com/18F/calc/pull/751#issuecomment-247641856 and a few other things.

It resolves the help text margins by moving the help text back below the `<input>` field, which is where it was before the great rejiggering of #583. Here the bottom margin on `<p>` elements works to distance the help text from the next field, rather than from the input:

> screen shot 2016-09-16 at 1 46 45 pm

But I'm not sure if the move of the help text to be above the input in #583 was a deliberate decision, or just an accidental effect of the rejiggering.  Regardless, it's really easy to undo.

### DIFF
--- a/data_capture/models.py
+++ b/data_capture/models.py
@@ -5,6 +5,12 @@ from contracts.models import Contract, EDUCATION_CHOICES
 
 
 class SubmittedPriceList(models.Model):
+    CONTRACTOR_SITE_CHOICES = [
+        ('Customer', 'Customer/Offsite'),
+        ('Contractor', 'Contractor/Onsite'),
+        ('Both', 'Both'),
+    ]
+
     created_at = models.DateTimeField(auto_now_add=True)
     updated_at = models.DateTimeField(auto_now=True)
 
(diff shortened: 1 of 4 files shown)


jseppi:
This looks good to me, but I'd like @hbillings to weigh in on moving the help desk, so I think we should wait for her return.

data_capture/models.py
```diff
@@ -20,11 +26,7 @@ class SubmittedPriceList(models.Model):
     is_small_business = models.BooleanField()
     contractor_site = models.CharField(
         verbose_name='Worksite',
-        choices=[
-            ('Customer', 'Customer/Offsite'),
-            ('Contractor', 'Contractor/Onsite'),
-            ('Both', 'Both'),
-        ],
```
hbillings: Excellent idea!

hbillings:
I'm good with either way as long as it's consistent across the board! The standards show help text above the inputs, but don't explicitly offer guidance. Right now the contract start/end date help text is above the boxes, but I've been thinking below might help break it up more. I think the errors might also currently be below, but I need to check.

PR #764, opened 2016-09-16
Coerce education take 2

jseppi:
Adds a coercer function for the education level field. The new `extract_min_education` function just tries to find valid values in the given text by stripping punctuation and lowercasing them both. See the [doctests](https://github.com/18F/calc/pull/764/files#diff-494ded7f44a6451dbc147d388e2bab59R58) for examples of what will get returned.

I've also created a `tests/fixtures/` directory for storing price list (and other) fixtures, ref #755.

To Do: 
- [x] use mocks instead of fixture files


data_capture/schedules/coercers.py
```diff
@@ -94,13 +94,18 @@ def extract_min_education(text):
...
+    # generate a list of tuples of the form
+    # (education choice, stripped-lowered-split education choice)
     desired = [(label, strip_and_lower(label).split())
                for _, label in EDUCATION_CHOICES]
 
+    # for each stripped-lowered-split education choice
     for label, stripped_and_split_label in desired:
+        # find and return the first one that matches an equally-sized sublist
+        # of the stripped_and_split_text
         for sublist in gen_sublists(stripped_and_split_text,
```
jseppi: This part looks a little funky because it was hard to come up with decent variable names.

jseppi:
@toolness: ready for review

data_capture/schedules/coercers.py
```diff
@@ -0,0 +1,114 @@
...
+    '''
+    >>> list(gen_sublists([1, 2, 3, 4, 5], 2))
+    [[1, 2], [2, 3], [3, 4], [4, 5]]
+
+    >>> list(gen_sublists(['a'], 3))
+    []
+
+    >> list(gen_sublists([1, 2, 3]), 3)
+    [1, 2, 3]
+    '''
```
toolness: It's not particularly obvious what this function is doing from the examples, could you provide a docstring?

data_capture/tests/test_s70.py
```diff
@@ -101,6 +101,29 @@ def test_text_formatted_prices_are_gleaned(self):
...
+            'commercial_list_price': '125.0',
+            'unit_of_issue': 'Hour',
+            'most_favored_customer': 'All Commercial Customers',
+            'best_discount': '0.07',
+            'mfc_price': '123.99',
+            'gsa_discount': '0.1',
+            'price_excluding_iff': '110.99',
+            'price_including_iff': '115.99',
+            'volume_discount': '0.15',
+        }])
```
toolness: Neat, this seems a bit verbose for a test that's just making sure that the minimum education is gleaned properly... I wonder if we could figure out how to use this [`model_mommy`](http://model-mommy.readthedocs.io/en/latest/basic_usage.html) thing that the data explorer's tests use to make them a bit less verbose at some point.  This is fine though!

toolness: Hmm, what if you changed this to only look at the `education_level` key? That way the test would be less likely to break if we changed the field names or something... Though I guess we would still have to fix up `price_list_with_min_education_text.xlsx`... Blerg.

I wonder if we could use [`xlwt`](http://xlwt.readthedocs.io/en/latest/) to make little modifications to just one excel file so we don't have to store tons of different excel files in the repository, and so that it'd be easy to move to new price list formats if older ones were deprecated or something. I dunno.

toolness:
Is the new `.xlsx` file just like our other excel files but with a different minimum education field?  If so, I dunno if it really justifies adding 200K to the repository...  Blergh.  I wonder if we should commit this w/o the xls file and accompanying test, and just assume that `extract_min_education` is being called properly?

data_capture/schedules/coercers.py
jseppi: 👍 
Ha, and actually that third doctest was 1) incorrect and 2) not actually being invoked because I only had `>>` instead of `>>>`. Fixed now.

jseppi:
Hrm, or maybe figure out how to Mock the excel instead of having file fixtures?

toolness:
Yeah mock the excel!!!

jseppi:
woot!

toolness:
Ok, we should make sure we **Squash and merge** this one to ensure that we don't have an extra 250K of binary data in our commit history.

data_capture/tests/test_s70.py
```diff
@@ -105,7 +134,7 @@ def test_rows_are_returned(self):
 
     def test_text_formatted_prices_are_gleaned(self):
         file_path = path(
-            'static', 'data_capture', 'price_list_with_text_prices.xlsx')
+            'tests', 'fixtures', 's70', 'price_list_with_text_prices.xlsx')
```
jseppi: Should probably remove this fixture and use `FakeWorkbook` here as well. Could be done in a separate PR though.

toolness: Filed as #785!

PR #765, opened 2016-09-16
Find Labor Categories start row index instead of hardcoding it

jseppi:
Small change to find the start row of the labor categories table instead of hardcoding `rownum = 3`. The new `find_header_row` works pretty naively by just returning the row index for the row that contains the first expected column heading (`"SIN(s) PROPOSED"`) in column 0.

To Do:
- [x] add test ~~and fixture~~ (used MagicMocks instead)

### DIFF
--- a/data_capture/schedules/s70.py
+++ b/data_capture/schedules/s70.py
@@ -83,6 +83,19 @@ def safe_cell_str_value(sheet, rownum, colnum, coercer=None):
     return str(val)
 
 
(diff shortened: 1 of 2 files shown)


jseppi:
Instead of a fixture maybe I can figure out how to use `MagicMock` finally o_O

jseppi:
ready for review, @toolness

toolness:
Super cool, thanks!

PR #767, opened 2016-09-19
Find S70 price list field columns by heading instead of by index 

jseppi:
This PR changes the internal workings of `glean_labor_categories_from_file()` to use price list table column headings instead of hardcoding the column indices when constructing the `dict`s that are used to make `Schedule70PriceListRows`.

To Do:
- [x] tests


data_capture/schedules/s70.py
```diff
@@ -83,6 +99,26 @@ def safe_cell_str_value(sheet, rownum, colnum, coercer=None):
...
 
+def generate_column_index_map(heading_row, field_title_map=None):
+    def normalize(s):
+        return re.sub(r'\s+', '', str(s).lower())
+
+    def find_col(col_name):
+        for idx, cell in enumerate(heading_row):
+            if normalize(cell.value) == normalize(col_name):
+                return idx
+        raise ValueError('Heading "{}" not found'.format(heading_row))
```
jseppi: Not exactly sure if raising an exception is the best option

jseppi:
The weakness to this approach is that the headings have to match (nearly) completely. I made the heading-matching logic a little fuzzy by lowercasing and removing whitespace, but they still need to match all the weird other text, like in
image

jseppi:
plz review @toolness

data_capture/schedules/s70.py
toolness: Will this eventually get turned into a `ValidationError` and surfaced to end-users who upload sheets?

data_capture/schedules/s70.py
```diff
@@ -47,6 +48,22 @@
...
+    'min_years_experience': 'MINIMUM YEARS OF EXPERIENCE',
+    'commercial_list_price': 'COMMERCIAL LIST PRICE (CPL) OR MARKET PRICES',
+    'unit_of_issue': 'UNIT OF ISSUE (e.g. Hour, Task, Sq ft)',
+    'most_favored_customer': 'MOST FAVORED CUSTOMER (MFC)',
+    'best_discount': 'BEST  DISCOUNT OFFERED TO MFC (%)',
+    'mfc_price': 'MFC PRICE',
+    'gsa_discount': 'GSA(%) DISCOUNT (exclusive of the .75% IFF)',
+    'price_excluding_iff': 'PRICE OFFERED TO GSA (excluding IFF)',
+    'price_including_iff': 'PRICE OFFERED TO GSA (including IFF)',
+    'volume_discount': 'QUANTITY/ VOLUME DISCOUNT',
```
toolness: I wonder if we should allow these to be [`Regular Expression Objects`](https://docs.python.org/3.6/library/re.html#regular-expression-objects) for fancier searching.  Probably for a later PR tho!

toolness:
This seems supercool to me.  Maybe file a separate issue for surfacing the "column not found" type error to the end-user, or at least logging it in such a way that we/admins can know why a sheet wasn't able to be gleaned?

data_capture/schedules/s70.py
jseppi: Good idea. Done!

jseppi:
Woot, merging now. Made minor change of `ValueError` -> `ValidationError`.

PR #769, opened 2016-09-19
Fix step 4 form-button-row markup

### LINKED ISSUE #768: Price List Upload Step 4 submit-label layout weirdness
The `.submit-label` span in Step 4 of the Price List Upload seems to be in the wrong place.

screen shot 2016-09-19 at 4 42 21 pm

jseppi:
Fixes #768

Also moves the `cancel` button to outside of the `{% if %}` block so that the price list submission can still be canceled when no valid rows are present in the selected file.


toolness:
Noice!

PR #771, opened 2016-09-20
[WIP] Capture and use Escalation Rate in S70 Price List Upload

### LINKED ISSUE #770: Capture Escalation Rate in S70 Price List Upload flow
We have a trello card for this feature, but opening here for better visibility/discussion.

The Price List Upload flow needs to capture escalation rate (I think it is optional), and then the backend needs to calculate "out-year" pricing (ie, years 2-5 typically, based on the chosen contract year) for each labor category in the price list.

First we need to decide where to put the input for capturing this number. Probably easiest place to put it without modifying the flow too much is at the bottom of Step 2. @hbillings: thoughts?

jseppi:
work-in-progress

Closes #770

To Do:
- [x] migration
- [x] form field
- [ ] calculate and save the out-year prices based on the captured escalation rate
- [ ] tests


data_capture/migrations/0004_auto_20160920_1944.py
```diff
@@ -0,0 +1,25 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='submittedpricelist',
+            name='escalation_rate',
+            field=models.IntegerField(null=True, blank=True, validators=[django.core.validators.MinValueValidator(1), django.core.validators.MaxValueValidator(100)]),
+        ),
+        migrations.AlterField(
+            model_name='submittedpricelist',
+            name='contractor_site',
+            field=models.CharField(choices=[('Customer', 'Customer/Offsite'), ('Contractor', 'Contractor/Onsite'), ('Both', 'Both')], verbose_name='Worksite', max_length=128),
```
jseppi: Apparently a migration containing these slight changes to `contractor_site` had not yet been committed. The change was just in the labels to include the `'/Offsite'` and `'/Onsite'` suffixes.

data_capture/forms/price_list.py
```diff
@@ -57,6 +57,15 @@ class Step2Form(forms.ModelForm):
         widget=UswdsRadioSelect,
     )
 
+    escalation_rate = forms.IntegerField(
```
jseppi: It seems weird to me that I have to specify a `forms.IntegerField` here but the default widget created from the model's `IntegerField` doesn't include the attributes for min and max value on the rendered `<input type="number">`.

jseppi:
Hrm, actually I used `IntegerField` for the escalation rate, but actually I'm unsure if they are going to be full integers.

@tram: any idea what valid escalation rates look like? Also what's the lowest/highest values they could be?

jseppi:
Going to close this due to changes in #794. Will open new PR in its stead.

PR #773, opened 2016-09-20
improve cloud deployment YAML config.

toolness:
Blerg.

The cloud deploy doesn't copy the `.env` file over, but the downside is that it means that we need to separately configure every container, which is a bit annoying, especially for environment variables like `DEBUG` that are shared across containers.


jseppi:
👍

Issue #772, opened 2016-09-20
collectstatic is failing because collect8.svg is missing


jseppi:
```sh
whitenoise.storage.MissingFileError: The file 'frontend/built/img/correct8.svg' could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0x7fed6952e9b0>.
The CSS file 'frontend/built/style/main.min.css' references a file which could not be found: frontend/built/img/correct8.svg
```

Looks like the missing svg is referenced in `uswds/_inputs.scss`:

```scss
input[type="checkbox"]:checked + label::before {
  background-image: url('../img/correct8.png');
  background-image: url('../img/correct8.svg');
  background-position: 50%;
  background-repeat: no-repeat;
}
```

jseppi:
@hbillings: the suspect rules appear to have been added in https://github.com/18F/calc/commit/9e6017877aa885bb1db7ddf7213a62446cde353f

As a fix, should we deleted the `background-image` rules, or should we include the referenced `correct8.png` and `correct8.svg` images?

@toolness: any ideas for testing that collectstatic won't fail? I wonder if we could just add it to the `script:` section of `.travis.yml`

jseppi:
My bad, `collectstatic` is not failing, but something in `Whitenoise` is. Digging into it with @toolness...

jseppi:
Ok, wrong again, `collectstatic` _is_ failing, but only when `DEBUG` is not in the env, that's why we weren't seeing it locally.

toolness:
@hbillings can you add the fix for this to the `test-compressed-collectstatic` branch so it's part of #776?

PR #774, opened 2016-09-20
[wip] add collectstatic to travis script

jseppi:
Add `python manage.py collectstatic --noinput` to our Travis script so that we can test that Whitenoise will not fail on stuff like missing files as in #772.


jseppi:
Well, it is failing now, which is the desired outcome, but turns out it is not failing for the right reason.

jseppi:
Closing because this is not a great way to test what we need to test

toolness:
Eh, actually best practices recommend that `DEBUG` *not* be set to true when running tests.  Folks were actually just talking in #django on Slack about it. 😟 

So this might actually be a good idea in the long-run...

jseppi:
That does sound reasonable, but the way I was doing it here was definitely not the place to do it. Another issue?

toolness:
Sure sounds good!

PR #775, opened 2016-09-20
Show step 1 deets in step 2

### LINKED ISSUE #737: Display chosen schedule, contract # and vendor name on step 2

hbillings:
Closes #737. Note that this doesn't display the schedule, in part because I'm no longer sure if we want to stick with displaying the schedule, or if we think we might want to shift to using SIN numbers a la the micropurchase.
screen shot 2016-09-20 at 3 50 39 pm


toolness:
NOICE

PR #782, opened 2016-09-21
Limit width on labor category table cells

hbillings:
Sets max-width and also sets the overflow to "scroll".
screen shot 2016-09-21 at 11 10 33 am


hbillings:
Ugh, this doesn't work on FF.
screen shot 2016-09-21 at 11 26 33 am

jseppi:
Maybe `word-break: break-all;` will work?

hbillings:
So `word-break: break-all` _does_ work, but it ends up doing crap like this: 
screen shot 2016-09-21 at 11 37 26 am
blargh argh argh

jseppi:
`word-break: break-on-outrageously-long-unbroken-strings` ?

hbillings:
Sooooo...probably the easiest way to do this is to wrap each title in a `div` and set the `overflow: scroll` on the div because FF won't respect overflows on table cells. ugh ugh ugh why

hbillings:
Hmm...alternately, I could attempt to insert spaces after commas, since the d3 code already manipulates the text...

jseppi:
Given that this behavior is only happening due to one ~~whacky~~ poorly-formatted labor category, and that the current fix will only not work completely in Firefox (but does work in Chrome, Safari, and IE), I think that we should keep what you currently have in this PR. That is, using `max-width` and `overflow: scroll` to deal with the very long unbroken labor category string.

jseppi:
I should add: 👍 from me, merge when ready, but please log an issue about it not quite working in Firefox

toolness:
Did you try `overflow-wrap: break-word`?  According to http://stackoverflow.com/a/15137272 it looks like it might be friendlier than the screenshot shown at https://github.com/18F/calc/pull/782#issuecomment-248669534, in that it will try to break words at spaces if possible.

toolness:
Yass I think `overflow-wrap: break-word` might help!  Check out this jsbin:

http://jsbin.com/koxevabide/edit?html,output

hbillings:
@toolness The problem is that the cell that's causing the jankiness has almost no spaces. 😞 this is the content: ```Interpretation Services Class 4: Afrikan,Akan,Albanian,Amharic,Arabic,Armenian, Azerbaijani,Bengali,Burmese,Cambodian, Cantonese,Dzongkha,Farsi,Fukienese,Gaelic,Gujarati,Hebrew,Hindi,Hmong,Icelandic,Indonesian,Japanese,Javanese,Kannada,Kazakh,Khalkha Mong```

toolness:
Yup yup. If you see http://pode.hksr.us/toolness/overflow-wrap and choose `word-break: break-all` you will see that it breaks *all* words where-ever it wants, whereas if you choose `overflow-wrap: break-word` it will only break *really long* words.  So the latter is better for our needs because it will wrap that super-long `Interpretation Services Class 4: ...` thing but not e.g. `Sign Language Interpreter Medical/Legal`.

hbillings:
Oh, derp, I misunderstood! This looks awesome; will amend!

jseppi:
🎆

PR #786, opened 2016-09-21
add <uswds-date> custom element that focuses on next widget on "/" key

toolness:
During user testing today someone tried typing `4/28/1986` into a date widget, but was confused when it didn't work.  This PR adds a `<uswds-date>` custom element which detects if the user presses `/`, `-`, or `.` and, if so, advances to the next field if applicable.

To do:

- [x] Add tests.
- [x] Add documentation.

### DIFF
--- a/frontend/templates/frontend/date.html
+++ b/frontend/templates/frontend/date.html
@@ -1,6 +1,6 @@
 <span class="form-hint" id="{{ hint_id }}">For example: 04 28 1986</span>
 
-<div class="date-mm-dd-yy">
+<uswds-date class="date-mm-dd-yy">
   <div class="form-group form-group-month">
     <label for="{{ month.id }}">Month</label>
 
@@ -14,4 +14,4 @@
     <label for="{{ year.id }}">Year</label>
     <input class="input-inline" aria-describedby="{{ hint_id }}" id="{{ year.id }}" name="{{ year.name }}" pattern="[0-9]{4}" type="number" min="1900" max="9999" value="{{ year.value }}">
   </div>
-</div>
+</uswds-date>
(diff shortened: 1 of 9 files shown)


toolness:
@jseppi ready for review!

frontend/templates/frontend/date.html
```diff
@@ -1,6 +1,6 @@
 <span class="form-hint" id="{{ hint_id }}">For example: 04 28 1986</span>
 
-<div class="date-mm-dd-yy">
+<uswds-date class="date-mm-dd-yy">
```
jseppi: I don't feel too strongly about it, but I think it would be better to use the `is`-attribute instead of a custom tag name for attaching the custom element behavior. As you pointed out, @toolness, using `is` will also obviate having `_date.scss` in this PR.

styleguide/templates/styleguide.html
```diff
@@ -484,6 +484,12 @@ <h3 class="example-heading">
     {% pyobjname 'frontend.date.SplitDateField' %} is a custom Django form
     field and widget that offers three separate text fields for users to enter
     dates, as per the <a href="https://standards.usa.gov/form-controls/#date-input">USWDS section on date input</a>.
+  </p>
+  <p>
+    We use a {% webcomponent '<uswds-date>' %} to provide some optional
+    dynamic functionality, such as automatically advancing focus to the
+    next input field in the date whenever <code>/</code> is pressed.
```
jseppi: I think it would be good to include the other key presses here as well (`-` and `.`)

jseppi:
nice!
👍

frontend/templates/frontend/date.html
toolness: Ok!  Um, I am going to not do that because I don't feel too strongly about it either and I want to work on other stuff now 😁

PR #791, opened 2016-09-22
Fix ajaxform interop w/ back-forward cache

toolness:
This fixes ajaxform's interoperability with the browser's [back-forward cache](https://developer.mozilla.org/en-US/Firefox/Releases/1.5/Using_Firefox_1.5_caching), whereby on some browsers a page containing an ajaxform is shown in its post-redirect state when the user clicks their browser's "back" button.

Solution gleaned from [stack overflow](http://stackoverflow.com/a/13123626).

To reproduce the bug this is fixing, do the following on Safari (probably in IE as well, according to yesterday's user research session):

1. Visit any ajaxform, e.g. `/styleguide/ajaxform`.
2. Fill the form out properly and click submit.
3. Click the back button.

You should see the ajaxform in its "loading" (i.e. faded out) state.

### DIFF
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -17,7 +17,17 @@ class Delegate {
   }
 
   redirect(url) {
-    // http://stackoverflow.com/a/13123626
+    // Some browsers will actually cache the state of our page, including
+    // its DOM and JS, when the user navigates away from it, so that if/when
+    // the user navigates back, the cached version is shown.
+    //
+    // In the case of ajaxform, this ultimately means that the user could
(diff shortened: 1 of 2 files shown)


frontend/source/js/data-capture/ajaxform.js
```diff
@@ -11,16 +11,28 @@ const $ = jQuery;
...
-let delegate = {
+class Delegate {
+  constructor(window) {
+    this.window = window;
+  }
+
   redirect(url) {
-    window.location = url;
-  },
+    // http://stackoverflow.com/a/13123626
```
jseppi: A little more of a description of why this is being done would be nice I think

jseppi:
Super weird bug o_O
👍

PR #792, opened 2016-09-22
add separate step for displaying rows with errors

### LINKED ISSUE #745: Move data upload errors to a separate sub-step instead of the current combined step
Whoever takes this on should make sure they're familiar w/ the new `Steps`/`StepRenderer` class, which is fully documented in [`data_capture/views/common.py`](https://github.com/18F/calc/blob/develop/data_capture/views/common.py#L17).

In particular, since the error presentation will constitute a "sub-step", we'll have to use one of the techniques discussed in https://github.com/18F/calc/pull/734#issuecomment-246826154.

jseppi:
This PR adds a "sub"-step after Step 3 (upload) to show the error rows found during price list validation. If there are no validation errors then the user is sent to Step 4 directly.

Here is what it looks like when a price list with no valid rows is uploaded:

screen shot 2016-09-22 at 3 59 07 pm

And here's the button row when there are valid rows in addition to error rows:

screen shot 2016-09-22 at 4 00 45 pm

I think some word-smithing and probably styling changes are needed, but I'd like those to happen in separate PRs.

Closes #745

To Do:
- [x] split out price list upload errors to new sub-step
- [x] tests
- ~~nail down language~~ do in separate issue/PR
- ~~improve styling~~ do in separate issue/PR


data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -5,6 +5,7 @@
 {% block step_body %}
 <h2>Validate data</h2>
 
+{# TODO: remove this block? #}
 {% if not is_preferred_schedule %}
```
jseppi: This message now shown in `step_3_errors`, so should I remove it here, or does it make sense to have it in both places?

data_capture/migrations/0004_auto_20160922_1539.py
```diff
@@ -0,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
jseppi: This is an unrelated migration, looks like just the `contractor_site` choice labels changed and a migration was never committed for it. I can remove it from this PR if it doesn't make sense to have here.

data_capture/templates/data_capture/bulk_upload/region_10_step_1_form.html
```diff
@@ -18,7 +18,7 @@
     <div class="submit-group">
       <button type="submit" class="button-primary">Next</button>
       <span class="submit-label">
-        Confirm you want to replace existing data
+        Confirm you want to replace existing data.
```
jseppi: added a period for consistency, not really related to this PR

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -49,25 +26,16 @@
...
   <div class="form-button-row clearfix">
     <a href="{% url 'data_capture:step_3' %}" class="button button-previous">Previous</a>
 
     {% if gleaned_data.valid_rows %}
       <div class="submit-group">
-        {% csrf_token %}
         <button type="submit" class="button-primary">Next</button>
         <span class="submit-label">
-          Submit to approver
+          Submit for approval.
```
jseppi: modified language slightly, added a period for consistency

jseppi:
ready for review @toolness

jseppi:
@hbillings: I'm pretty sure I saw some mock-ups you made with nice styling for this error sub-step, but I have not been able to locate them.

data_capture/templates/data_capture/price_list/step_4.html
hbillings: I've been thinking that in step 4 (sans errors) we should display all the data collected to that point. So I think it makes sense to remove this particular bit in favor of adding more info next sprint.

hbillings:
@jseppi derp, maybe I never uploaded them to drive. Regardless, I think your idea about styling being a separate PR is 👍

data_capture/templates/data_capture/price_list/step_4.html
jseppi: I just realized that actually if the user does upload a different schedule AND it has no error rows, then they won't see the step 3 error page, so they won't ever see this `not is_preferred_schedule` stop in that case.

jseppi: blehhhhh I guess we should keep it in for now. This message will actually never come up for anyone during first release because S70 is the only type of price list we accept.

data_capture/templates/data_capture/price_list/step_3_errors.html
```diff
@@ -0,0 +1,66 @@
...
+      <a href="{% url 'data_capture:step_3' %}">try uploading it again</a>.
+    </p>
+  {% endif %}
+
+  <form method="post">
+    {% csrf_token %}
+    <div class="form-button-row clearfix">
+      <a href="{% url 'data_capture:step_3' %}" class="button button-previous">
+        Fix errors and re-upload price list
+      </a>
```
toolness: We'll eventually want to make this page actually contain a form w/ an upload widget for re-uploading the file, right?  IIRC, one of our user testers had chosen "Fix errors and re-upload" before actually fixing the errors in Excel, which meant they were now looking at an upload form that had no information about what they needed to fix... Should we file a separate issue for that?

data_capture/migrations/0004_auto_20160922_1539.py
toolness: Cool cool, no probalo.

data_capture/tests/common.py
```diff
@@ -51,6 +51,17 @@ def uploaded_csv_file(content=None):
...
 
+def create_csv_content(rows=None):
+    if rows is None:
+        rows = [
+            ['132-40', 'Project Manager', 'Bachelors', '7', '15.00']
+        ]
+    headers = ['sin', 'service', 'education', 'years_experience', 'price']
+    all_rows = [headers] + rows
+    as_str = '\n'.join([','.join(x) for x in all_rows])
+    return as_str.encode('utf-8')
```
toolness: Noice.  If we ever need to get fancy I guess we can use a [`csvwriter`](https://docs.python.org/3/library/csv.html#writer-objects) but it doesn't seem like we need to (and anyways, `csvwriter`s are a bit annoying because they work on file objects rather than strings).

data_capture/tests/test_price_list_views.py
```diff
@@ -46,6 +48,17 @@ def test_permission_is_required(self):
         res = self.client.get(self.url)
         self.assertEqual(res.status_code, 403)
 
+    def test_cancel_clears_session_and_redirects(self):
+        if not self.should_handle_cancel:
+            raise unittest.SkipTest()
```
toolness: So one really annoying thing I realized since advising you to use `SkipTest` last time is that skipped tests show up in test runs as `s` instead of `.`.  This makes it seem like the test is *supposed* to be not-skipped, but is being skipped because of some temporary problem with the codebase that will eventually get resolved.  Blerg.  Looking at the [docs](https://docs.python.org/3.4/library/unittest.html#unittest-skipping) though, it does kinda seem like we're using it for the right reasons, I guess, so I can't think of anything else to use here.

toolness: Hmm, the one alternative I can think of is to have a `HandlesCancelMixin` that just defines `test_cancel_clears_session_and_redirects`.  Not sure if py.test will try testing `HandlesCancelMixin` by itself, though, which could get annoying.

data_capture/views/price_list_upload.py
```diff
@@ -167,9 +171,40 @@ def step_3(request, step):
...
+@handle_cancel
+def step_3_errors(request):
+    step = steps.get_step_renderer(3)
+    gleaned_data = get_nested_item(request.session, (
+        'data_capture:price_list',
+        'gleaned_data'
+    ))
+
+    if gleaned_data is None:
+        return redirect('data_capture:step_3')
```
toolness: Is there a test that verifies this?  I guess it'd basically be exactly like `Step4Tests.test_gleaned_data_is_required()`... I wonder if we could reuse that somehow, e.g. via a mixin class or in a similar way you're testing the canceling behavior here.

Also, this new step kinda makes me want to bring back that `requires_gleaned_data` decorator, but I guess since we're only using it twice, maybe we shouldn't do that unless we start needing it on more steps.

data_capture/templates/data_capture/price_list/step_3_errors.html
jseppi: Filed in https://github.com/18F/calc/issues/804

data_capture/templates/data_capture/price_list/step_4.html
jseppi: Filed in https://github.com/18F/calc/issues/805

data_capture/views/price_list_upload.py
jseppi: Right on, test added.

data_capture/tests/test_price_list_views.py
jseppi: Cool! I forgot about multiple-inheritance in Python.
I moved the `handle_cancel` and `gleaned_data` test methods to mixin classes.

jseppi:
Requested changes made and issues filed, @toolness

toolness:
Nice, this looks great! Feel free to merge.

Issue #570, opened 2016-08-18
ajaxform component doesn't work w/ radio inputs


toolness:
Because of the way radio buttons work, and the way our ajaxform component needs to manually populate a `FormData` instance (see #520), our ajaxform doesn't set the correct value for a radio input.  Instead, it will submit form data such that it always looks like the very last radio button in a group has been selected (I think).

This isn't a problem _for now_ because we don't have radios in our one ajaxform.  But we might want to at least detect the radio situation and throw an exception if it looks like there's a radio in the ajaxform or something (though obviously, even better would be to just support radios).

toolness:
Also, there may be other funky types of inputs that our ajaxform doesn't work with, I'm not sure. Blegh.

jseppi:
Oh geez :\ Are there any other ajax-form like components in the wild we can use or see how they deal with other form inputs?

Here's one I just found real quick: https://github.com/rnicholus/ajax-form
and here's how it deals with radios: https://github.com/rnicholus/ajax-form/blob/master/ajax-form.js#L239

PR #793, opened 2016-09-22
[WIP] Better page transitions from ajaxforms

toolness:
This adds some optional transitioning in the way ajaxforms redirect users to new pages by supporting the concept of a "visual anchor".  Basically, if an ajaxform specifies an `id` for a visual anchor element, then the ajaxform will smoothly scroll to the visual anchor before redirecting the user's browser.  Once redirected, the loaded page will check to see if it's coming out of a transition and, if so, automatically scrolls the page to the visual anchor.

ajaxform-transition

## Limitations

As can be seen from the above animated GIF, there is an unfortunate side effect of a "flicker" that occurs while the new page is loading. Unfortunately, I don't know of a way to get rid of this without progressively enhancing the entire workflow into a single-page app, which is a fairly big lift.

However, the implementation also adds a `html.ajax-form-transitioning` class at the  beginning of the redirected page load, which gives us room to add other transitional effects, such as the color transitioning of the newly-completed step in the steps widget.

Also, this transition effect only works with ajaxforms, which means that it currently could only work with the steps that involve upload widgets.  In order to make this transition happen across all steps of a workflow, we'd need to convert every step to use an ajaxform, which means tackling issues like #570.

## To do

- [ ] Add documentation.
- [ ] Add tests.
- [ ] Consider making the amount of time spent transitioning (currently 250ms) into a SASS variable.  It'd be nice if the JS knew the value of this but we might have to just duplicate it for now and add a comment reminding devs to keep it in-sync with the SASS variable.


toolness:
Just added a quick fade so that there's no flicker when the page transition happens:

> ajaxform-transition-with-fade

hbillings:
I'm conflicted about this. I really like the fade-in for the step, but if we feel like we need to automatically scroll the page, that might be an indication that the visual design isn't working. If there's too much fluff at the top of the page, we should rework that information hierarchy so that it's more streamlined. (I totally didn't catch yesterday that you were doing anything other than the color transition or I would have spoken up sooner -- my bad!!)

*takes off information designer hat*

@jseppi and @tram, whatcha think? @toolness, what else are you hoping to use this for?

toolness:
I think you're right @hbillings!  We should definitely think about refactoring the IA... But, I also think that the auto-scroll might still be appropriate for really small screens?  Hmm.

I actually decided to remove the "scroll up to the top of the form before fading out" behavior--before I added the fade-out, I think it was useful because the next page started at the same place. But then we had the flicker, which is why I added the fade-out... But then I think the scroll *with* the fade-out might've been a bit too much, and possibly even disorienting to the user.  So now the form just fades-out and fades-in with the top of the screen anchored to the visual anchor thingy.  And then the current step fades in.

toolness:
Ok, Imma close this one in favor of changing the IA of the site and some other stuff.  We can bring it back later if we want!

Issue #790, opened 2016-09-21
Remove Contract Year and adjust Contract Start/End labels


jseppi:
I think every single interviewee has had questions about what that field is and why it is needed.

hbillings:
I know we've had some discussion about this, but documenting here... is contract year necessary? Can we extrapolate from the start/end date + escalation rate what the current and out year pricing would be?

jseppi:
Per today's discussion:

* Remove the Contract Year radios
* Change "Contract start" to "Contract or current option period start"
* Change "Contract end" to "Contract or current option period end"

Does this sound correct @hbillings @tram @toolness?

jseppi:
Done in #794

PR #794, opened 2016-09-22
remove contract_year and update contract_start and _end labels

jseppi:
Removes `contract_year` and changes the labels for the `contract_start` and `contract_end` fields as discussed in #790.

I also changed the date widget's example to use "2016" as the year since that will be more similar to the year that users will generally enter.

To Do:
- [x] add migration (waiting until #792 is merged since it also has a migration)


jseppi:
ready for review, @toolness

frontend/templates/frontend/date.html
```diff
@@ -1,4 +1,4 @@
-<span class="form-hint" id="{{ hint_id }}">For example: 04 28 1986</span>
+<span class="form-hint" id="{{ hint_id }}">For example: 04 28 2016</span>
```
toolness: I liked 1986 😞

PR #795, opened 2016-09-23
Show highlight animation when section of styleguide is navigated to

toolness:
This uses the [`:target`](https://developer.mozilla.org/en-US/docs/Web/CSS/:target) CSS pseudo-selector to show the user where the section they just navigated to is.

target-highlight


jseppi:
👍

PR #798, opened 2016-09-23
surround Add Price Data link with perm check

### LINKED ISSUE #788: "Add price data" shows as an option even if user doesn't have permission
We gotta test for the right perms.

jseppi:
Closes #788

As @hbillings pointed out in #788, we don't really have a use case for users without the `add_submittedpricelist` permission. However, this check is slightly helpful to us because it is an easy visual queue during dev and test/staging deployments to show whether we've initialized the test user accounts with proper permissions.

### DIFF
--- a/data_explorer/templates/base.html
+++ b/data_explorer/templates/base.html
@@ -52,7 +52,9 @@
                     {% if user.is_staff %}
                     <li><a href="{% url 'admin:index' %}">Site admin</a></li>
                     {% endif %}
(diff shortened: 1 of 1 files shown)


toolness:
Oh neat, I did not know about [`perms`](https://docs.djangoproject.com/en/dev/topics/auth/default/#permissions).  COOL.

toolness:
Oh, should we have a similar check in the "Region 10 Upload" link in the admin UI?

PR #799, opened 2016-09-23
Rename _ajaxform.css to _ajaxform.scss.

toolness:
Aside from consistency, our SASS pipeline doesn't seem to detect changes to this file otherwise.


jseppi:
👍

PR #800, opened 2016-09-23
Fade-in current step

toolness:
This salvages the step fade-in from #793 as a CSS animation.

target-highlight

If the user's browser doesn't support CSS animations, this PR has no effect (in other words, the current step will still appear "filled-in" from the start, rather than appearing "empty" and never being "filled").

### DIFF
--- a/frontend/source/sass/components/_steps.scss
+++ b/frontend/source/sass/components/_steps.scss
@@ -1,4 +1,23 @@
 @import '../base/variables';
+
+@keyframes step-fade-in {
+  from {
+    background-color: $color-white;
+    color: $color-primary;
+  }
+
+  to {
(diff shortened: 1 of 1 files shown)


frontend/source/sass/components/_steps.scss
```diff
@@ -1,4 +1,23 @@
 @import '../base/variables';
+
+@keyframes step-fade-in {
+  to {
+    background-color: $color-primary;
+    color: $color-white;
+  }
+}
+
+@supports ((animation-name: test) and (animation-fill-mode: forwards)) {
```
jseppi: whoa, never seen `@supports` before

jseppi:
p cool 👍

PR #801, opened 2016-09-23
Capture and use Escalation Rate for out-year pricing

### LINKED ISSUE #770: Capture Escalation Rate in S70 Price List Upload flow

jseppi:
This PR implements capturing of escalation rate (in Step 2 of the Price List upload flow) and usage of that rate to escalate hourly prices for the `Contract` models that get created when a Price List is approved.

Closes #770 

To Do:
- [x] confirm escalation logic with Kelly
- [x] retarget to `develop` once #794 is merged
- [x] implement escalation logic
- [x] tests for new `Contract` model methods
- [x] tests for changes to `approve()` method
- [x] clear up TODO comments/questions (potentially by filing new issues for what they bring up)
- [x] file issue about validating prices again `FEDERAL_MIN_CONTRACT_RATE` during price list upload - filed https://github.com/18F/calc/issues/815
- [x] handle case when contract starts in future date
- [x] check `contract_end` when setting price fields
- [x] handle `contract_end` when escalating price fields


data_capture/tests/test_models.py
```diff
@@ -24,7 +26,9 @@ def create_price_list(self, **kwargs):
             is_small_business=False,
             contract_number='GS-123-4567',
             vendor_name='UltraCorp',
-            schedule=self.DEFAULT_SCHEDULE
+            schedule=self.DEFAULT_SCHEDULE,
+            contract_start=datetime.strptime('2016-09-01', '%Y-%m-%d').date(),
+            contract_end=datetime.strptime('2021-09-01', '%Y-%m-%d').date(),
```
jseppi: is there a better way to construct a `date` object? I'm a bit confused by `datetime` and `date` problems, especially with regards to Django models. It seems like the model itself accepts `str`, `datetime`, or `date` values.

data_capture/models.py
```diff
@@ -82,19 +92,62 @@ def approve(self):
...
-                current_price=row.hourly_rate_year1,
-                next_year_price=None,
-                second_year_price=None,
                 contractor_site=self.contractor_site,
                 schedule=self.get_schedule_title(),
                 business_size=self.get_business_size_string(),
                 sin=row.sin,
             )
+
+            # TODO: Move all this stuff to a method on Contract model ?
```
jseppi: Got several `TODO` comments throughout this section.
It seems a bit weird to me to have this logic in `approve()`, especially since it means if a contract gets unapproved and then approved again, the values will all change. I think, however, that is a weakness of the current Contract model's setup (with `current_year`, `next_year_price`, and `second_year_price` being fields in the model instead of values calculated on the fly).

jseppi:
@toolness - pre-completion review/thoughts are very welcome for this one

data_capture/tests/test_models.py
jseppi: When I go through the actual web flow, the `contract_start` and `contract_end`  fields are `date` objects by the time the code gets to the `SubmittedPriceList.approve()` method.

contracts/tests/test_contract.py
```diff
@@ -46,10 +67,8 @@ class ContractSearchTestCase(TestCase):
     ]
 
     def assertCategoriesEqual(self, results, categories):
-        self.assertEqual([
-            result.labor_category
-            for result in results
-        ], categories)
+        result_categories = [r.labor_category for r in results]
+        self.assertEqual(sorted(result_categories), sorted(categories))
```
jseppi: Uh, so, after I added `test_escalate_hourly_price_fields_raises_on_invalid_rate`, the original assertion here started failing. I didn't touch any of the code related to `ContractSearchTestCase` so I have no idea why that started happening.
The failures looked to be related to ordering, so that is why I modified `assertCategoriesEqual` to not fail on ordering differences.

<details>
```sh
contracts/tests/test_contract.py:73: in assertCategoriesEqual
    ], categories)
E   AssertionError: Lists differ: ['Interpretation Services Class 2: French, [135 chars]ish'] != ['Disposal Services', 'Interpretation Servi[135 chars]ian']
E
E   First differing element 0:
E   Interpretation Services Class 2: French, German, Italian
E   Disposal Services
E
E   - ['Interpretation Services Class 2: French, German, Italian',
E   -  'Disposal Services',
E   ? ^
E
E   + ['Disposal Services',
E   ? ^
E
E      'Interpretation Services Class 4: Afrikan,Akan,Albanian',
E   -  'Interpretation Services Class 1: Spanish']
E   ?                                            ^
E
E   +  'Interpretation Services Class 1: Spanish',
E   ?                                            ^
E
E   +  'Interpretation Services Class 2: French, German, Italian']
```
</details>

contracts/models.py
```diff
@@ -206,3 +210,75 @@ def get_education_code(text):
...
+        curr_contract_year = int(
+            (current_date - self.contract_start).days / 365.25) + 1
+
+        # TODO: Should we have lower and upper limits
+        # on the year fields in contract_start and contract_end
+        # in the upload form?
+
+        self.contract_year = curr_contract_year
+
+        if curr_contract_year > 5:
```
jseppi: I had original made this if-condition raise a `ValueError`, but on further consideration, I decided to just make it set the price fields to `None`

jseppi:
Hrm, I am not clear on what this line https://github.com/18F/calc/blob/develop/data_capture/schedules/s70.py#L254 or how this new escalation rate logic changes things.

data_capture/models.py
```diff
@@ -117,9 +123,7 @@ class SubmittedPriceListRow(models.Model):
         choices=EDUCATION_CHOICES, max_length=5, null=True,
         blank=True)
     min_years_experience = models.IntegerField()
-    hourly_rate_year1 = models.DecimalField(max_digits=10, decimal_places=2)
-    current_price = models.DecimalField(
```
jseppi: I removed this `current_price` field from `SubmittedPriceListRow` (and everywhere else in `data_capture`) because I am pretty certain it is not needed any longer.

jseppi:
Read for review, @toolness

jseppi:
Whoa CodeClimate randomly checked in on this PR o_O

contracts/tests/test_contract.py
toolness: Ah thanks for doing this! I actually have noticed this for a while, but it's only happened when I've specifically only run the tests for the `contracts` app; if I run *all* the tests, then the error didn't occur.  Which was super weird.  Anyways, thanks for perma-fixing it...

Actually, now I'm wondering, though, if those tests *are* trying to assert that the ordering of the categories are equal?  That could mean that we need to add an ordering parameter somewhere?  Should we file a separate issue for it and address that later?

toolness:
Lol, I like the idea of CodeClimate having random whims and mood swings.

contracts/tests/test_contract.py
jseppi: Hrm, I don't see why they would necessarily be in order, but, yeah, the intent is hard to glean here.

Also I just did a local experiment and removed the calls to `sort()`, and the tests still pass, so something non-deterministic appears to be going on.

jseppi:
Accidentally uncovered an untested and un-handled case while I was demoing to Kelly: when the `contract_start` is in the future.

I need to fix the current contract year calculation and `hourly_rate_yearX` to deal with this case, which is a valid situation. A contract option can be finalized but start in the future, after the price list is uploaded.

jseppi:
bah, I think I also need to take `contract_end` into account and not set price fields for years that are greater than the `contract_end`

jseppi:
gah, I guess the current year and end year should also be taken into account in `escalate_hourly_rate_fields`.

toolness:
@jseppi is this ready for final review+merge?  Or did kelly say there were changes that need to be made?

jseppi:
Kelly liked it, but I realized there are some issues in dealing with `contract_end`, so I'm working on that now. Blerg, brain fried.

jseppi:
Alright, @toolness, this should be ready for final review.
Beware, thar be dragons!

jseppi:
hrm, just did a manual run-through and noticed that if Escalation Rate is left blank, I get an `IntegrityError` once I get to the final submit. Will fix soon.

toolness:
I noticed that the escalation rate field is rather wide:

> screen shot 2016-09-29 at 9 18 04 am

@hbillings should we be using a narrower field for this? (Or @jseppi should we deal w/ styling in a separate PR?)

contracts/models.py
```diff
@@ -206,3 +210,91 @@ def get_education_code(text):
     @staticmethod
     def normalize_rate(rate):
         return float(rate.replace(',', '').replace('$', ''))
+
+    def update_price_fields(self):
+        '''
+        Set current, next, and second year price fields based on this
+        contract's contract_year.
+        '''
```
toolness: Should we call this over all contracts in a periodic `rq_scheduler` job or something?  Maybe we should file a separate issue about it?

contracts/models.py
```diff
@@ -206,3 +210,91 @@ def get_education_code(text):
...
+            self.current_price = self.get_hourly_rate(curr_contract_year)
+        else:
+            self.current_price = None
+
+        if -1 < curr_contract_year < contract_end_year:
+            self.next_year_price = self.get_hourly_rate(curr_contract_year + 1)
+        else:
+            self.next_year_price = None
+
+        if -2 < curr_contract_year < contract_end_year - 1:
```
toolness: I am confused by the above two conditionals... when would `curr_contract_year` ever be a negative number?

contracts/models.py
```diff
@@ -206,3 +210,91 @@ def get_education_code(text):
...
+
+            self.set_hourly_rate(i, next_rate)
+
+    def adjust_contract_year(self, current_date=None):
+        if current_date is None:
+            current_date = datetime.today().date()
+
+        if not self.contract_start:
+            # TODO: What should happen in this case? Could also early return
+            raise ValueError('contract_start must be defined')
```
toolness: When would this happen?  I know the contract start date is a required field in the price list upload, at least... Do bulk uploads sometimes lack contract start dates or something?

contracts/models.py
```diff
@@ -206,3 +210,91 @@ def get_education_code(text):
...
+
+        return contract_end_year
+
+    def get_hourly_rate(self, year):
+        if year < 1 or year > 5:
+            raise ValueError('year must be in range 1 to 5')
+        return getattr(self, 'hourly_rate_year{}'.format(year))
+
+    def set_hourly_rate(self, year, val):
+        if year < 1 or year > 5:
```
toolness: Hmm, I wonder if it'd be useful to define a `NUM_CONTRACT_YEARS = 5` constant or something.

data_capture/tests/test_models.py
jseppi: doh, I knew there must've been an easier way :P

contracts/models.py
jseppi: Well, I _think_ a better fix in the end will be to just calculate those fields on the fly and remove them from the database altogether. Whatever we chose should definitely be in another issue/PR though :)

contracts/models.py
jseppi: when the `contract_start` is in the FUTURE 😎

contracts/models.py
```diff
@@ -206,3 +210,91 @@ def get_education_code(text):
...
+
+        if -2 < curr_contract_year < contract_end_year - 1:
+            self.second_year_price = self.get_hourly_rate(
+                curr_contract_year + 2)
+        else:
+            self.second_year_price = None
+
+    def escalate_hourly_rate_fields(self, base_year_rate, escalation_rate):
+        if escalation_rate < 0 or escalation_rate > 99:
+            raise ValueError('escalation_rate must be between 0 and 99')
```
jseppi: 👍 done

contracts/models.py
jseppi: the `Contract` model itself allows null `contract_start` and `contract_end` fields, and yeah, in `Region10Loader` those fields can be not set when they are blank in the region 10 export.

contracts/models.py
toolness: WHOA.

contracts/models.py
jseppi: 👍  done

data_capture/templates/data_capture/price_list/step_2.html
```diff
@@ -35,6 +35,11 @@
     </fieldset>
   </div>
 
+  {{ form.escalation_rate.label_tag }}
+  {{ form.escalation_rate.errors }}
+  {{ form.escalation_rate }}
+  <p class="helptext">{{ form.escalation_rate.help_text }}</p>
```
jseppi: 👍 done

hbillings:
I can handle any presentation stuff in another PR.

jseppi:
> I noticed that the escalation rate field is rather wide

Filed as https://github.com/18F/calc/issues/new

jseppi:
Ok, @toolness, I think I addressed everything. Thanks for good suggestions!

toolness:
Wooot feel free to merge!

jseppi:
🚀

PR #802, opened 2016-09-23
Wrap Region 10 Upload link in perms check

jseppi:
Wraps the "Region 10 Upload" link in the Admin site header in a permissions check, per https://github.com/18F/calc/pull/798#issuecomment-249223920


toolness:
NOICE

PR #803, opened 2016-09-26
make <expandable-area> work in print, add CSS transition.

### LINKED ISSUE #797: <expandable-area> may not appear expanded when printed
Not that anyone would print the upload form that we currently use it on, but if we eventually want to use this widget in e.g. the *About* page, we might want to make it more printer-friendly.  Should just involve wrapping its CSS in a `@media screen` query, I think.

toolness:
This fixes #797 and adds a dinky "fade in" CSS transition when expanding so that users can distinguish what's just been added from the rest of the document.  (Ideally this would be a slide but I dunno how to do that without some fancy JS, and we've got bigger fish to fry anyhow.)

expandable-area

Browsers without support for CSS animations will just see the content appear instantly (i.e., the current behavior).

This PR also refactors the SASS to use nesting.

### DIFF
--- a/frontend/source/sass/components/_expandablearea.scss
+++ b/frontend/source/sass/components/_expandablearea.scss
@@ -1,23 +1,44 @@
 @import '../base/variables';
 
-expandable-area > *[role="button"][aria-expanded="false"]:before {
-  content: '\25B6';
-  padding-right: 4px;
-  display: inline-block;
+@keyframes expandable-area-fade-in {
+  from {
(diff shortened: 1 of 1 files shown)


jseppi:
I'm thinking 1 second might be a little too long, but I don't have anything to base that on really. Otherwise, 👍

toolness:
Ok, if folks seem to be perturbed by the delay we can shorten it :)

PR #807, opened 2016-09-26
upgrade django to 1.8.15

jseppi:
security upgrade: https://www.djangoproject.com/weblog/2016/sep/26/security-releases/


jseppi:
cc @toolness

Issue #808, opened 2016-09-26
Create a template tag/filter for rendering <fieldset> elements


toolness:
See https://github.com/18F/calc/pull/796#discussion_r80550026.

PR #816, opened 2016-09-28
Make date range error styling a bit more modular.

toolness:
So in trying to do #808 I noticed that the one place an `as_fieldset` helper wouldn't work with our current markup is in the `date-range` part of step 2.

But @hbillings mentioned she wasn't a fan of that CSS anyways, so I took a look at what it currently renders as:

current-date-range-error-css

There are a few quibbles I have with this styling:

* The word **to** is now at the end of the line rather than in between the fields on the same line, which looks a bit weird.

* If only *one* of the dates in the range has errors, the entire range is given a red background, rather than only the date with the error.

This **EXPERIMENTAL** patch changes things so it looks like this:

new-date-range-error-css

Now if only one of the dates is in error, only that date appears with the error styling; the other date appears normally.

Furthermore, the word **to** appears on the same line in the same manner as it does on the non-error version.

And finally, because the individual `<fieldset>`s in the `date-range` are rendered in the exact same way as other fieldsets in the form, it paves the way for encapsulating that markup in a macro/helper (see #808).

### DIFF
--- a/data_capture/templates/data_capture/price_list/step_2.html
+++ b/data_capture/templates/data_capture/price_list/step_2.html
@@ -20,14 +20,14 @@
     {{ form.contractor_site }}
   </fieldset>
 
-  <div class="date-range{% if form.contract_start.errors or form.contract_end.errors %} fieldset-error{% endif %}">
-    <fieldset>
+  <div class="date-range {% if form.contract_start.errors or form.contract_end.errors%}fieldset-error{% endif %}">
+    <fieldset {% if form.contract_start.errors %}class="fieldset-error"{% endif %}>
       {{ form.contract_start.errors }}
       {{ form.contract_start.label_tag }}
       {{ form.contract_start }}
     </fieldset>
     <p>to</p>
-    <fieldset>
+    <fieldset {% if form.contract_end.errors %}class="fieldset-error"{% endif %}>
       {{ form.contract_end.errors }}
       {{ form.contract_end.label_tag }}
       {{ form.contract_end }}
--- a/frontend/source/sass/components/_forms.scss
+++ b/frontend/source/sass/components/_forms.scss
@@ -39,17 +39,16 @@ form ul {
   list-style-type: none;
 }
 
-fieldset,
-.fieldset-error {
+fieldset {
   max-width: 53rem;
 }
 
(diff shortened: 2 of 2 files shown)


data_capture/templates/data_capture/price_list/step_2.html
```diff
@@ -20,14 +20,14 @@
     {{ form.contractor_site }}
   </fieldset>
 
-  <div class="date-range{% if form.contract_start.errors or form.contract_end.errors %} fieldset-error{% endif %}">
-    <fieldset>
+  <div class="date-range {% if form.contract_start.errors or form.contract_end.errors%}fieldset-error{% endif %}">
```
hbillings: Do we need the `if` clause on the `div` still, if we're rendering it directly on the `fieldset` now?

frontend/source/sass/components/_forms.scss
```diff
@@ -39,17 +39,16 @@ form ul {
...
-fieldset,
-.fieldset-error {
+fieldset {
   max-width: 53rem;
 }
 
 /* Django's default form rendering uses a <ul class="errorlist">
    for errors, and <span class="helptext"> for help text. */
-.fieldset-error {
+fieldset.fieldset-error {
```
hbillings: If we remove `.fieldset-error` from the `div`, we probably don't need this selector.

data_capture/templates/data_capture/price_list/step_2.html
toolness: Yeah, it has to have that to maintain the ol' `3rem` bottom margin as part of the `<div>` rather than the contained `<fieldset>`s.  If we keep that bottom margin as part of the `fieldset.fieldset-error` rule, then there's a massive margin underneath each errored date, which makes it so the **to** is no longer aligned properly with at least one of the dates.

If only the [`:has`](https://developer.mozilla.org/en-US/docs/Web/CSS/:has) pseudo-class was actually supported, we could just use that, but otherwise I don't see any obvious ways around still requiring the `if`. 😞

toolness:
Oof, so this PR is actually merged because I merged #820 into `develop`, which contained this stuff.  So even though GitHub says it's "closed", it is actually in the codebase!

PR #817, opened 2016-09-28
Add a {% fieldset %} template tag

### LINKED ISSUE #808: Create a template tag/filter for rendering <fieldset> elements

toolness:
**Note: This is a PR against #816, not `develop`.**

This fixes #808 by adding a `{% fieldset %}` tag, which basically functions like a macro.

### DIFF
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -9,6 +9,7 @@ from frontend.radio import UswdsRadioSelect
 
 class Step1Form(forms.ModelForm):
     schedule = forms.ChoiceField(
+        label="Which schedule is associated with this price list?",
         choices=registry.get_choices
     )
 
--- a/data_capture/templates/data_capture/price_list/step_1.html
+++ b/data_capture/templates/data_capture/price_list/step_1.html
@@ -1,4 +1,5 @@
 {% extends 'data_capture/step.html' %}
+{% load frontend %}
 
 {% block subtitle %}Upload price list{% endblock %}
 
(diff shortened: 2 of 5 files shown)


toolness:
Hmmm, according to my reading of Heydon Pickering's [Inclusive Design Patterns](https://www.smashingmagazine.com/2016/06/inclusive-design-patterns/), we probably want to make a bunch of our `<label>`s into `<legend>`s, particularly for radio buttons and probably the date widgets too.  I can file a separate issue for that, though, since this PR is just about refactoring things to use the `{% fieldset %}` helper.

hbillings:
Ah crap, MDN referred to `legend` as optional, so I thought I was okay, heh. `legend`s get moved to the top of the `fieldset`, which is normally fine but problematic with our error styling, which renders the error message above the `label`. I'll do some research to see if styling legends cross-browser has gotten any more reliable since the last time I tried to do it, but if not, we may want to do something like hide the `legend` from sighted users and render the `label` with `aria-hidden=true`?

toolness:
Oh, weird!  Bleh, yeah, I guess we can do that if all else fails.  The `legend` is really nice for screenreaders when coupled with form fields that have multiple inputs (radios, checkboxes, and our multi-input date widgets) because then the screenreader will contextualize a particular label (e.g. "month") with the legend (e.g. "start contract date").

toolness:
Err... should I re-file this as a new PR?  It looks like it was merged but then reverted in #824...

hbillings:
@toolness see: this morning's slack rant

toolness:
Ok, I ended up just merging #820, which built upon this, directly into `develop`--so these changes are in `develop` now!

PR #822, opened 2016-09-28
Adjust margin-left of styleguide section headers

toolness:
This just re-aligns the styleguide section headers. We're now defining them in terms of `em` so hopefully we don't need to adjust them as styles change anymore.


jseppi:
👍

Issue #810, opened 2016-09-27
Price Block inner borders don't extend all the way to top of blocks


jseppi:
<img width="258" alt="screen shot 2016-09-27 at 10 20 35 am" src="https://cloud.githubusercontent.com/assets/697848/18882414/465ef43a-849c-11e6-8b83-fe7f65a14ec4.png">

jseppi:
somewhat related: https://github.com/18F/calc/issues/560

jseppi:
Closed by #823

PR #823, opened 2016-09-29
Visual bugfixes

### LINKED ISSUE #560: proposed-price-block margin is too big
image

The block only shows up when you enter a Proposed Price in the box at the upper right of the bar chart. Probably needs to be added to the rules like in https://github.com/18F/calc/pull/558/commits/4809d65aae220e73273288422abdbbc6cef08954
Though maybe a single class for all three of the blocks would be easier to manage...

### LINKED ISSUE #735: Currency symbol floats above price when Price List table is narrow
This is likely due to the way I implemented the `.currency-symbol` styling, but I don't know how to make it better. Maybe remove the `width: 50%;` at smaller screen sizes?

screen shot 2016-09-13 at 9 39 33 am

screen shot 2016-09-13 at 9 39 16 am

### LINKED ISSUE #806: Bottom of text in Data Explorer dropdown filters is cut-off
screen shot 2016-09-26 at 10 41 48 am

hbillings:
For [this trello card](https://trello.com/c/eapicFYx/143-as-a-user-i-want-the-front-page-of-calc-to-have-a-consistent-display-and-style-that-is-free-of-visual-bugs-so-that-i-don-t-get-d)
- Fixed dropdown text being cut off when selected (closes #806). Unfortunately, the update to the filter language creates a new issue...
screen shot 2016-09-29 at 5 59 32 am

- Added flexbox for average/standard deviations (closes #560 & #810). This isn't perfect; the proposed price block is too large.
screen shot 2016-09-29 at 5 52 21 am

- Fixes #735


frontend/source/sass/data_explorer.scss
```diff
@@ -228,47 +228,47 @@ input[type="checkbox"]:checked ~ label {
...
-
 .avg-price-block,
 .standard-deviation-block {
   display: inline-block;
   margin: 0 auto;
-  padding: .4em 0;
+  padding: .4em;
   text-align: center;
-  width: 32%;
+  /*width: 32%;*/
```
toolness: Should we just delete this line, or is there a reason we should leave it commented-out?

toolness:
Nice, just tried it out on IE11 and it looks good there, too!

The filter cut-off actually looks a lot less janky on IE11, too:



Still not ideal but at least it doesn't look like a visual bug.

I wonder if we could use some fancy styling that makes the "Legacy Language" appear on a second line, slightly grayed-out and in a smaller font or something?

frontend/source/sass/components/_tables.scss
```diff
@@ -154,3 +153,9 @@ th.column-labor_category {
   overflow-wrap: break-word;
   max-width: 22rem;
 }
+
+@media all and (min-width: 650px) {
+  td > .currency-symbol {
+    width: 50%;
+  }
```
toolness: Hmm, so this seems to result in the `$` appearing closer to the "Minimum years of experience" than to the actual price when viewports are less than 650px wide:



Should the `$` be right-aligned or something?  Or brought closer to the actual price by some other means?

frontend/source/sass/data_explorer.scss
hbillings: ha! Clearly I am on my game today. I'll nix it.

frontend/source/sass/components/_tables.scss
hbillings: The `$` is in a `span`, which makes it kinda weird. I could try floating it, though then I need to float the price, too, and that promises to be all kinds of fragile. Or I could take another crack at implementing it as a `:before` pseudoelement, which seems less fragile. (And then we could even just drop the dolla on small screens if we wanted.)

toolness: Ok, it's up to you really--and if you want to merge this PR and address it in a separate issue that's fine too.

hbillings:
So I think in order to style the option, we'll either need to use some JS to render the name of the legacy schedule, or use `:after` to do something like:

```
.option[value="FABS"]:after {
  content: 'Legacy FABS';
  color: #d0d0d0;
  font-size: 1rem;
}
```
^^ I'm not totally sure this will even work, but I'm noting it here so I don't forget about it.

toolness:
That selector is bad ass.

Yeah, it would be nice to keep things semantic+accessible if possible... Hmmm.

hbillings:
Okay, I'm going to merge this with a fix for the dollar sign and open a new issue for the filter!

PR #828, opened 2016-09-29
validate contract_start is before contract_end

### LINKED ISSUE #825: Validate Start Date is not after End Date in Step 2 Form
Having an End Date that is before the Start Date of a SubmittedPriceList will cause issues with escalation rate calculation and setting of current/next/second year prices.

jseppi:
Closes #825

screen shot 2016-09-29 at 10 41 33 am

### DIFF
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -63,12 +63,12 @@ class Step2Form(forms.ModelForm):
 
     def clean(self):
         cleaned_data = super().clean()
-        if 'contract_start' in cleaned_data and 'contract_end' in cleaned_data:
-            start = cleaned_data['contract_start']
-            end = cleaned_data['contract_end']
-            if start > end:
-                self.add_error('contract_start',
-                               'Start date must be before end date.')
+        start = cleaned_data.get('contract_start')
+        end = cleaned_data.get('contract_end')
+        if start and end and start > end:
+            self.add_error('contract_start',
+                           'Start date must be before end date.')
+        return cleaned_data
 
     def clean_escalation_rate(self):
         value = self.cleaned_data['escalation_rate']
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -61,6 +61,15 @@ class Step2Form(forms.ModelForm):
         required=False,
     )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if 'contract_start' in cleaned_data and 'contract_end' in cleaned_data:
+            start = cleaned_data['contract_start']
+            end = cleaned_data['contract_end']
+            if start > end:
+                self.add_error('contract_start',
+                               'Start date must be before end date.')
+
     def clean_escalation_rate(self):
         value = self.cleaned_data['escalation_rate']
         if value is None:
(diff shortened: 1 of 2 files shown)


data_capture/tests/test_forms.py
```diff
@@ -16,6 +18,16 @@ def test_clean_escalation_rate_works(self):
         form.is_valid()
         self.assertEqual(form.cleaned_data['escalation_rate'], 2.5)
 
+    def test_start_and_end_date_is_validated(self):
+        form = Step2Form({
+            'contract_start': datetime.date(2020, 1, 1),
```
jseppi: halp! the `form.errors` always says `contract_start` and `contract_end` are missing

jseppi: cc @toolness

data_capture/tests/test_forms.py
```diff
@@ -16,6 +18,16 @@ def test_clean_escalation_rate_works(self):
...
+            'contract_end': datetime.date(2015, 1, 1),
+            'contractor_site': 'Both',
+            'is_small_business': True,
+        })
+        self.assertFalse(form.is_valid())
+        print(form.errors)
```
toolness: Remove `print`?!

data_capture/forms/price_list.py
```diff
@@ -61,6 +61,15 @@ class Step2Form(forms.ModelForm):
...
     )
 
+    def clean(self):
+        cleaned_data = super().clean()
+        if 'contract_start' in cleaned_data and 'contract_end' in cleaned_data:
+            start = cleaned_data['contract_start']
+            end = cleaned_data['contract_end']
+            if start > end:
+                self.add_error('contract_start',
+                               'Start date must be before end date.')
```
toolness: Doesn't `clean()` need to return `cleaned_data`?  I thought it did... that's what our other `clean()`s do, I think.

toolness: The other pattern I've seen in these that reduces DRY a bit is:

```py
start = cleaned_data.get('contract_start')
end = cleaned_data('contract_end')
if start and end and (start > end):
  self.add_error('blah')
```

But it's up to you really.

data_capture/tests/test_forms.py
jseppi: WIP!!! 😛

data_capture/forms/price_list.py
toolness: Oh, I recommend you return `cleaned_data` so that subclasses of your form can do what you did here (i.e. call superclass `clean()` and do something with its data).

toolness:
oh LOL sorry I didn't see the WIP!!

jseppi:
So the test doesn't actually test that the right error message (about start being after end) is added to the form. I can't seem to get the `Step2Form` to take in the `contract_start` and `contract_end` values in the test method. It works when I manually check through the web interface, but I can't figure out how to do it programmatically 😕

toolness:
Oh! That's because we're using the funky USWDS style date fields.  See our [price list upload tests](https://github.com/18F/calc/blob/develop/data_capture/tests/test_price_list_views.py#L118-L120):

```python
{
        'contract_start_0': '1985',
        'contract_start_1': '07',
        'contract_start_2': '08',
}
```

Basically form constructors expect the raw POST data being sent by the browser, and because we're using our fancy tri-input USWDS-style date widgets, that means we need to provide three separate values with weird names. 

I guess it might be useful to make some sort of `add_uswds_date(post_data, key, value)` test helper that adds a date to POST data so that this is easier for us.  I dunno.

jseppi:
ok, now ready for review, @toolness

toolness:
NOICE

PR #829, opened 2016-09-29
validate price_including_iff is above FEDERAL_MIN_CONTRACT_RATE

### LINKED ISSUE #815: Surface error when rates < FEDERAL_MIN_CONTRACT_RATE during Price List upload validation
When an uploaded S70 spreadsheet contains rate(s) less than `FEDERAL_MIN_CONTRACT_RATE`, we should probably show a validation error to the user (in Step 3 Errors).

It looks like https://github.com/18F/calc/blob/develop/data_capture/schedules/s70.py#L225 prevents `Schedule70PriceList` rows from having current_price < `FEDERAL_MIN_CONTRACT_RATE`, but that never gets seen by the user who uploads the price list.

jseppi:
Closes #815

screen shot 2016-09-29 at 11 14 06 am


jseppi:
ready for review, @toolness

toolness:
Nice, I like it!

Do we want to mention that the price is the "federal minimum contract rate"?  Would that be something an end-user would find useful?

Regardless, the actual code here is great so I say MERGE AWAY :shipit:

jseppi:
Yeah, I thought about adding something about "federal minimum contract rate", but then the error was too dang long.

PR #830, opened 2016-09-29
Link to S70 PPL template

### LINKED ISSUE #753: For schedule 70, offer a download link to the PPL template

toolness:
Fixes #753:

>


jseppi:
Sweet 👍

jseppi:
sorry, accidentally clicked the wrong button 😭

PR #831, opened 2016-09-29
Make <legend>s work on NVDA

toolness:
Arg, NVDA (and possibly other screen readers) *do* work with `<legend>` elements, but only if they're direct descendants of `<fieldset>`s, i.e. not wrapped in `<span>`s so we can style them.  So this adds a `<legend class="sr-only">` for the screenreaders and an `<label aria-hidden="true">` for the sighted folks.

😞


hbillings:
burn it all down.

PR #833, opened 2016-09-29
Use defusedxml and ensure xlrd is using it

toolness:
This is an alternate approach to fixing #749. It just calls `defusedxml.defuse_stdlib()` whenever `data_capture` is imported, and adds a test to verify that `xlrd` is using it, and that we aren't vulnerable to the [Billion Laughs](https://en.wikipedia.org/wiki/Billion_laughs) exploit.

This means that `xlrd.open_workbook()` can now throw a `defusedxml.common.EntitiesForbidden` exception, but that should be fine since our schedule 70 gleaner catches *any* exceptions thrown from `xlrd.open_workbook()`.

To do:

- [x] Actually call `open_workbook()` with a billion-laughs XLSX file and ensure it throws an `EntitiesForbidden` error.

### DIFF
--- a/data_capture/tests/test_xlsx_security.py
+++ b/data_capture/tests/test_xlsx_security.py
@@ -1,0 +1,33 @@
+import io
+import xlrd.xlsx
+from django.test import TestCase
+from defusedxml.common import EntitiesForbidden
+
+
+# https://en.wikipedia.org/wiki/Billion_laughs
+BILLION_LAUGHS_XML = """\
+<?xml version="1.0"?>
+<!DOCTYPE lolz [
+ <!ENTITY lol "lol">
+ <!ELEMENT lolz (#PCDATA)>
+ <!ENTITY lol1 "&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;&lol;">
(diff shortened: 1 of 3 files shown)


toolness:
@jseppi what do you think of this approach?

data_capture/tests/test_xlsx_security.py
```diff
@@ -0,0 +1,48 @@
...
+    zf.close()
+    return blob.getvalue()
+
+
+class BillionLaughsTests(TestCase):
+    def test_xlrd_open_workbook_raises_entities_forbidden(self):
+        with self.assertRaises(EntitiesForbidden):
+            xlrd.open_workbook(file_contents=build_zipfile({
+                'xl/_rels/workbook.xml.rels': BILLION_LAUGHS_XML,
+                'xl/workbook.xml': BILLION_LAUGHS_XML
```
toolness: This required digging in to `xlrd`'s source code to see what XML files it was looking for.  This test might stop working as we upgrade to new versions of `xlrd` but fixing it should just involve adding more files to the built ZIP file so that `xlrd` eventually gets around to opening one of the XML files, which will trigger the `EntitiesForbidden` exception.

jseppi:
Seems good to me!

Did you try running your test without the defusedxml patch to see if it blew up your computer?

toolness:
Yep--unfortunately it is a DoS attack so it locked everything up :(

PR #836, opened 2016-09-30
Update style for validate data page

hbillings:
**Note: This is against `upload-error-page`**
Goals: compress whitespace, add previously-entered details to the page.

Currently:
screen shot 2016-09-29 at 7 05 54 pm


toolness:
Looks like [flake8 doesn't like some things](https://travis-ci.org/18F/calc/builds/163882976#L1118-L1119):

```
./data_capture/views/price_list_upload.py:236:5: E303 too many blank lines (2)
./data_capture/views/price_list_upload.py:265:80: E501 line too long (92 > 79 characters)
```

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -3,29 +3,39 @@
...
+        <p>
+          <strong>Note:</strong> You specified that the price list you were
+          uploading was for <em>{{ preferred_schedule_title }}</em>, but it
+          appears to be for <em>{{ gleaned_data.title }}</em>. If this is
+          incorrect, please upload a different file.
+        </p>
+      </div>
+      {% endif %}
+    </li>
+  </ul>
```
toolness: I wonder if a [`<dl>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) would be more appropriate here? That way you could use `<dd>`s instead of `<span>`s for the values and it could be a better experience on screen readers, but it's up to you.

frontend/source/sass/components/_tables.scss
```diff
@@ -23,7 +23,8 @@ table {
     }
   }
 
-  &.sortable {
+  &.sortable,
+  &.price-list-table{
```
toolness: I'm not sure if we have a CSS style guide, but putting the opening `{` right after the selector (without any whitespace) makes me really twitchy for some reason, at least when it's inconsistent 😛  It's just a nit but a space before the `{` would be cool!

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -37,9 +47,9 @@
 
     {% if gleaned_data.valid_rows %}
       <div class="submit-group">
-        <button type="submit" class="button-primary">Next</button>
+        <button type="submit" class="button-primary">Submit</button>
```
toolness: Form style guides I've read say that we should prefer to use a more user-relevant verb than "Submit" for submit buttons, but I guess "Submit" really *is* the most user-relevant verb here, so 👍 !

data_capture/templates/data_capture/price_list/step_4.html
hbillings: Oh, that's an interesting thought. Will do!

data_capture/templates/data_capture/price_list/step_4.html
hbillings: Ha! Yeah, this is one of those fuzzy gut feeling things, but to me "Next" makes it feel like you're not at the last step. I could do something like "Submit for review" or "Send to admin" if that's clearer.

toolness: Oo I like "Submit for review"!  That really telegraphs that it's going to be reviewed before being added to CALC, but it's up to you.

PR #841, opened 2016-09-30
Improve selenium testing

toolness:
## What this does

- [x] Makes the tests run more consistently across a wider variety of browsers.
- [x] If `WD_TESTING_BROWSER` is undefined, or if it's defined to be something other than `phantomjs`, runs the QUnit tests via Selenium.
- [x] Integrates with Travis and Sauce Labs so that our ~~selenium~~ QUnit tests are run under IE11 on CI.
- [x] Adds a `selenium.md` that explains how to set up selenium testing using various configurations, and is linked-to by `README.md`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ python:
 - 3.4
 addons:
   postgresql: 9.3
+  sauce_connect: true
 matrix:
   fast_finish: true
 services:
@@ -31,6 +32,14 @@ env:
   - PHANTOMJS_TIMEOUT=15
   - DEBUG=yup
   - DATABASE_URL=postgres://postgres@localhost/hourglass
+  - DJANGO_LIVE_TEST_SERVER_ADDRESS=127.0.0.1:8000
+  - WD_TESTING_URL=http://127.0.0.1:8000
+  - WD_TESTING_BROWSER=internet explorer
+  - WD_TESTING_BROWSER_VERSION=11.103
+  - WD_JOB_VISIBILITY=public
+  - WD_SOCKET_TIMEOUT=60
+  - SAUCE_USERNAME=atulvarma
+  - secure: "RbgG35nc1SHgjdxVbGUDlumePeR/+dXq5fMbfTXOW8uFYw5kGh75px2KdFDkL1JIA1vbfxyNFm09x32uXMcHsBKu9fJS7xbpMrRDbJoV2wVy5wd6qtvZ7yFbR87Fqo3/tuGu13jNNDvUgKjpUlrzLZKaL7qvv5Gd7CVhkySUa1o="
 deploy:
   edge: true
   skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ install:
 - pip install -r requirements-dev.txt
 - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
 - npm install
-- gulp build
 script:
 - phantomjs --version
 - python manage.py ultratest flake8
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,11 @@ env:
   - PHANTOMJS_TIMEOUT=15
   - DEBUG=yup
   - DATABASE_URL=postgres://postgres@localhost/hourglass
+
+  # These environment variables affect the running of Webdriver/Selenium
+  # tests. Note that if we're not in a trusted environment (e.g. a PR
+  # is being issued from an untrusted fork), we will default to using
+  # PhantomJS locally, rather than a real-world browser via Sauce Labs.
   - DJANGO_LIVE_TEST_SERVER_ADDRESS=127.0.0.1:8000
   - WD_TESTING_BROWSER="internet explorer"
   - WD_TESTING_BROWSER_VERSION=11.103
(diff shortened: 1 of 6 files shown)


toolness:
Hmm, I just ran the selenium tests on sauce labs via an AWS-based docker machine.

The good news:

* All the tests passed!

The bad news:

* There were lots of socket timeouts, but our retry logic kicked in so the tests didn't fail. However, this resulted in a lot of extra time spent running the tests.

* Total test run time was 84.792s. Adding this on to every travis CI job would not be fun, especially if we add more selenium tests... Blerg.

toolness:
Alternatively, we could *only* run the qunit tests on saucelabs during travis CI. this would at least help us regularly ensure that they don't fail on ie11.

toolness:
Ok, I've modified `test_qunit` so that it runs the tests via Selenium if proper information is provided in the environment... That way we can still disable the other selenium tests (which I've now called `DataExplorerTests`) in Travis, but still have QUnit tests run on IE11, so at least that's neat.

toolness:
Ok, this PR is now [running QUnit tests against IE11 on Travis](https://saucelabs.com/beta/tests/531793b6800f40749f024afbab68ae7d/commands#8)!

toolness:
Blerg. Ok I think this is ready for review @jseppi.

Right now the creation of the Sauce Connect tunnel in Travis takes about 10-12 seconds, and running the QUnit tests via Sauce doesn't take too long (I can't time it precisely but it seems to only take a few seconds more than running the tests in Phantom).

Ideally in the future we'll change this so it runs the QUnit tests across a variety of browsers, rather than just IE11.

selenium.md
```diff
@@ -0,0 +1,111 @@
...
+Below is an example configuration for running tests in your
+host machine's Chrome browser; they assume that you're using a native
+version of Docker that exposes CALC through `localhost` at the
+default `DOCKER_EXPOSED_PORT` of `8000`.
+
+1. Download [Selenium Standalone Server][].  To run it, you will need to
+   install Java too. Start it up with `java -jar` followed by the
+   name of the `.jar` file you downloaded.
+
+2. Download [ChromeDriver][], decompress its executable and put it in the
```
jseppi: FYI, you can install both the Standalone Server and ChromeDriver with homebrew

```sh
brew install selenium-standalone-server
brew install chromedriver
```

(I'm trying it now to make sure it works)

selenium.md
```diff
@@ -0,0 +1,111 @@
...
+   ```
+   docker-compose run --service-ports app python manage.py test frontend
+   ```
+
+   If you need to run this multiple times but aren't changing any
+   front-end code between runs, set `SKIP_STATIC_ASSET_BUILDING=yup` in
+   your `.env` to make things run faster.
+
+
+### Running Selenium and a headless Chrome browser in Docker
```
jseppi: I think I would move this section to be at the top since it is easier, and I think I'd rather run the tests this way in most cases.

selenium.md
jseppi: The brew installations worked great. I also had to start the services:

```sh
brew services start selenium-server-standalone
brew services start chromedriver
```

jseppi:
Great documentation 💯! I was able to pretty easily get everything setup as it was written in `selenium.md`, except that I used `brew` for installing the server and chromedriver. 

It was fun to watch the Chrome windows launch and move around on their own, but I do.n't think I'd want to run it that way very often, hence the comment about moving the docker-based instructions up, but it's up to you really.

👍

.travis.yml
```diff
@@ -31,6 +31,18 @@ env:
...
+  # These environment variables affect the running of Webdriver/Selenium
+  # tests. Note that if we're not in a trusted environment (e.g. a PR
+  # is being issued from an untrusted fork), we will default to using
+  # PhantomJS locally, rather than a real-world browser via Sauce Labs.
+  - DJANGO_LIVE_TEST_SERVER_ADDRESS=127.0.0.1:8000
+  - WD_TESTING_BROWSER="internet explorer"
+  - WD_TESTING_BROWSER_VERSION=11.103
+  - WD_JOB_VISIBILITY=public
+  - WD_SOCKET_TIMEOUT=60
+  - SAUCE_USERNAME=atulvarma
```
jseppi: It's a little weird to me to use personal accounts for this type of thing, but we don't really have clear policy on using team or agency accounts :shrug:

toolness: Oh yeah, sorry I didn't comment on that--I originally asked on #infrastructure if it was possible to get a sauce labs account for a product rather than a person, and they said it wasn't possible unless the product has an email address, which CALC doesn't seem to, so for expediency I just got an account for myself (it is associated w/ my GSA email though, and paid for by 18F, so it's not like this is my *personal* personal acct).

Anyhow, I definitely agree we should use a product-specific acct here in the long run, will file a separate issue about it!

jseppi:
I think this also closes, #330, right?

toolness:
Hmm good question... So I think that we now have two different cases:

1. **Running Selenium tests using PhantomJS** - This is what #330 is really referring to, and it's still potentially an issue because of the fact that *untrusted* PRs will still run via Phantom.

2. **Running Selenium tests using Sauce Labs** - The only big issue here is the fact that while running *all* the Selenium tests does seem to work consistently, they're *very slow*,  taking 84s when running via Docker Machine + AWS at least (not sure why this would be faster on Travis without Docker, but it's possible).  So we might still want to leave the non-QUnit tests disabled on Travis for this reason (in fact, that's how this PR works--everything in `test_selenium.py` is still ignored).

So I'm not sure what to do. I guess maybe we could rename #330 to be "Selenium tests under PhantomJS randomly fail on Travis" and file a new issue called "Selenium tests run really slowly under Suace Labs"?  Blerggh...

selenium.md
toolness: Awesome, just mentioned this in the instructions.

selenium.md
jseppi: Looks good 👍 !

jseppi:
Ooof, thanks for clarifying #330 and the slowness when running under Sauce Labs. I think your suggestion of renaming #330 and filing a new issue is a good course of action.

toolness:
Woot!  Thanks, will do.

Issue #839, opened 2016-09-30
Create user stories for website analytics


toolness:
So right now we have Google Analytics and [DAP](https://www.digitalgov.gov/services/dap/) enabled, but we don't have a strategy for how we want to use them.

Over the past few months @tram has brought up some use cases, such as:
- As a researcher, I should be able to analyze information about people who somehow "fell off" the Price List Upload flow without completing it. I should be able to see what step they reached, and I should be able to follow-up with them individually if needed.
- As a developer or researcher, I should be able to analyze information about users who opted-in to safe mode. I should be able to see what percentage of users who were presented with the opt-in actually opted-in, and I should be able to see what page the opt-in was displayed on. If necessary, I should be able to follow-up with the user who was presented with the opt-in (if they were logged-in).

Depending on stories like these, we can figure out what kinds of [custom dimensions & metrics](https://support.google.com/analytics/answer/2709828?hl=en) and [events](https://support.google.com/analytics/answer/1033068?hl=en) we need to set up in GA.

I'm not very familiar with DAP and how its capabilities differ from GA, so we might want to do some research on that.

toolness:
I started watching the [DAP 201 Virtual Training](https://www.youtube.com/watch?v=rPo3zFbhXnk) and it looks like there definitely *are* some things that DAP provides which our 18F Google Analytics account doesn't. A specific one is the ability to create [custom funnel reports](https://support.google.com/analytics/answer/6180923?hl=en), which is part of DAP but *not* part of 18F GA.

I suspect custom funnel reports in particular will be useful for us, since the whole multi-step price list upload process constitutes a "funnel" of sorts.

toolness:
**Update:** We're currently using the [HEART framework](https://library.gv.com/how-to-choose-the-right-ux-metrics-for-your-product-5f46359ab5be#.gdiihef7n) to identify these user stories in the [HEART x GSM metrics exercise](https://docs.google.com/spreadsheets/d/1EnlEaRs6iKsgGsJtZqvnq1lEYKhTMnQ-zAdHUOXKX9o/edit#gid=0) google doc.

jseppi:
Good to close this? Or is there additional action needed here, @toolness?

toolness:
Yep good to close!

PR #842, opened 2016-09-30
Improve google analytics tracking

toolness:
This addresses some of the user stories outlined in #839:

* It defines [custom dimension](https://support.google.com/analytics/answer/2709828?hl=en#configuration) index 1 to be the user ID of the currently logged-in user.  This isn't PII by itself so we should be fine storing it in google analytics.

* It defines custom dimension index 2 to be `safe mode enabled` if safe mode is enabled.

* Whenever the user is shown the safe mode opt-in, an [event](https://support.google.com/analytics/answer/1033068?hl=en) of category `safe mode` and action `show` is triggered.

* Whenever the user expands or collapses an `<expandable-area>`, a custom event of category `expandable area` and action `expand` or `collapse` and the label of the expander (i.e. the text the user clicked on) is triggered.

* Moves the google analytics script to the `<head>`.  This is the recommended best practice since the GA script loads asynchronously and users might technically navigate to a different page before GA kicks in, if we load it at the bottom of the page.  Also, this way our code bundles can access `ga` for custom events at any point in their execution (the events will just be queued up if the analytics script hasn't been loaded yet).

* Adds a `common/ga` module that wraps the `ga()` global function, providing a no-op if GA is disabled (e.g. for unit tests, or if we eventually make `ga` optional for deploys).

**Note that all custom dimensions must be registered via the Google Analytics administrative UI too, or else they won't have any effect.**

I think this is enough for one PR, but I'll file separate issues for more things we ought to be tracking, such as ajaxform events and such.

### DIFF
--- a/frontend/source/js/common/ga.js
+++ b/frontend/source/js/common/ga.js
@@ -1,0 +1,11 @@
+/* global window */
+
+/**
+ * This is just a wrapper around Google Analytics' ga() function, which
+ * should be included inline at the top of the page. If it wasn't included
+ * at the top of the page, however, we'll just provide a no-op function
+ * so that code which calls it still works, allowing GA to be an optional
+ * dependency.
+ */
+
+module.exports = typeof window.ga === 'function' ? window.ga : () => {};
--- a/data_explorer/templates/base.html
+++ b/data_explorer/templates/base.html
@@ -11,6 +11,30 @@
     <script src="{% static 'frontend/js/vendor/history.min.js' %}"></script>
     <![endif]-->
 
+    <script>
+        (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+        (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+        m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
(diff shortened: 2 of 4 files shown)


frontend/source/js/common/ga.js
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * This is just a wrapper around Google Analytics' ga() function, which
+ * should be included inline at the top of the page. If it wasn't included
+ * at the top of the page, however, we'll just provide a no-op function
+ * so that code which calls it still works, allowing GA to be an optional
+ * dependency.
+ */
+
+module.exports = typeof window.ga === 'function' ? window.ga : () => {};
```
jseppi: nice!
I wonder if a simple `console.log` here would be a good idea here to remind devs to include the "real" GA?

jseppi:
👍 Looks great!

jseppi:
> Note that all custom dimensions must be registered via the Google Analytics administrative UI 

Has this been done?

frontend/source/js/common/ga.js
toolness: Hmm, I think though that it should be possible to disable GA entirely... And this will also be the case on QUnit tests, so it'd be nice to not have them spam the console. Hmmmmm.

toolness:
Nope, not yet! I will get on it.

frontend/source/js/common/ga.js
jseppi: Right on

Issue #845, opened 2016-10-03
Bulk create accounts from list of CO and Data Administrator email addresses


jseppi:
So that we have initial user accounts
- [x] Contract Officers
- [x] Data Administrators

jseppi:
Will run `bulkcreateusers` on pre-prod once #819 is merged.

jseppi:
Just made the Contract Officer users in the prod database

jseppi:
Shelly has been added as a Data Administrator to the pre-prod db. Anyone else, @tram?

jseppi:
COs and Data Admin (Kelly and Shelly) have been created in GovCloud prod instance.

PR #846, opened 2016-10-03
management command to bulk create users

jseppi:
Adds a management command called `bulkcreateusers` to create a bunch of users from a specified file, and optionally add them to a group with a specified name.

Called like:

```sh
python manage.py bulkcreateusers schedule_70_contract_officers.txt --group "Contract Officers" --dryrun
```

I modified the command a bit from my first commit. Now it just takes a line-delimited file containing email addresses, and a single group can be specified in the command-line call (via the `--group` option). As suggested by @toolness, there is also a `--dryrun` flag to indicate whether or not the command should actually commit anything to the database.

I tested the individual functions used internally to the command, but am unsure how to best test the command itself.

Will be useful for #845


toolness:
Noice!

I always like having a `--dry-run` option so I can make sure that the command is going to affect the database in a way I expect (since undoing the operation is hard).  In the past I've done this by wrapping the command's operations in a `with transaction.atomic()` and throwing a special `DryRunFinished` exception at the end if `--dry-run` is enabled. That's just what I like doing though, I'm not sure if there's an easier way to accomplish the same thing... You can see an example of it [here](https://github.com/toolness/hive-django/blob/master/directory/management/commands/importorgs.py#L337-L342) if you want.

meta/management/commands/bulkcreateusers.py
```diff
@@ -0,0 +1,57 @@
...
+@click.option('--group', 'groupname', type=click.STRING)
+@click.option('--dryrun', default=False, is_flag=True)
+def command(user_file, groupname, dryrun):
+    '''Expects a text file with one email address per line'''
+    if groupname:
+        try:
+            group = Group.objects.get(name=groupname)
+        except Group.DoesNotExist:
+            click.echo('"{}" group does not exist. Exiting.'.format(groupname))
+            return
```
toolness: Can we exit with a nonzero exit code here?  I think just raising a [`CommandError`](https://docs.djangoproject.com/en/1.10/howto/custom-management-commands/#django.core.management.CommandError) should be fine (doing things the `click` way seems to result in an unhelpful stack traceback).

toolness: Just saying this b/c we never know if/when this command might end up being used by some shell script, at which point the writer will expect something like `manage.py bulkcreateusers myusers.txt --group blarg && echo "SUCCESS"` to behave predictably...

jseppi: 👍  Totally agree

jseppi:
Ok, @toolness.
- Modified to raise CommandError on invalid group
- Filled out docstring and parameter help strings

toolness:
Noice, I like it!

Issue #811, opened 2016-09-27
Revise copy for uploading new labor price data


tram:
Recommendations from 18F Writing Lab in [18F/writing-lab/issues/199](https://github.com/18F/writing-lab/issues/199)

# Global changes
- Header: 18F style discourages capitalization of Government — in all instances of U.S. government, please use the lowercase.
- Headings and titles: Our style guide also recommends sentence case for headings and titles. Throughout, please update headings, titles, and form-field descriptions to sentence case.

# Calc home click login at bottom
- I’d update the short description to the following: “CALC lets you conduct market research on professional service labor categories quickly and easily, helping you make better informed decisions.”
- In the header, I’d add a period after “Help us improve.”
- Under Worldwide rates, I’d change the description to “Globally applicable rate” (to parallel the construction of the other descriptions).
- Instructions above the search bar: I’d update to “Enter your search terms below, separated by commas.” (Stronger call to action.)
- Under Optional filters, please lowercase the y in “Contract year.”
- Footer: I love the tone of the “Get in touch” section — very friendly and welcoming!

# Fake cloud gov login
- Remove the “Just” from “Just enter your email address below…” (“Just” is one of our “Words to avoid” in the style guide.)
- I was a bit confused by the content in the note — will users already have received information on how to have their email addresses registered in the CALC database? If not, it might be useful to update the note to say something like, “You’ll need to have registered your email address in the CALC database to log in here. Visit our README for registration details.”

# Click add price data
(No notes that aren’t already captured in the other categories)

# Add company and contract information
In the short description, I’d combine the first two sentences as follows: “Your price lists help make CALC a more robust tool, and sharing them is easy…”

# Add contract details
Question: Will users know how to differentiate small from regular businesses? If not, additional instructions here might be warranted.

# Upload price data
(No comments here — looks good!)

# This xlsx doesn’t have the proper sheet
(No comments here — looks good!)

# Sheet error
I might recommend a few small tweaks to the error message: “Oops! Please correct the following error.” 

# Upload correct xlsx
(No comments here — looks good!)

# One row has an error
Please update “double-check” to “double check.”

# Success!
So, our style guide favors active over passive voice (it’s easier for folks to read). In light of this, I might recommend that you update the success message to “You’ve submitted your data for review — it should be available soon!” (I also think it’s OK to remove the bit about the CALC admin, as you’ve already clarified who will be doing the reviewing.)

toolness:
Cool!

[...]

PR #848, opened 2016-10-03
modify Step 1 intro text

### LINKED ISSUE #784: Consider adding text that data capture is for FINAL price list upload
I've noticed in several of the interviews, subjects seem to want to upload non-finalized price lists.

jseppi:
Somewhat minor modifications to the introductory text in Step 1. My aim was to make sure that the user knows we are looking for a finalized price list (ref #784).

I also noticed that the sequence of events mentioned in the "Once you upload a file..." sentence was a bit incorrect/out-of-date, so I took a stab at amending it also.

screen shot 2016-10-03 at 11 52 56 am

I think the spacing looks a bit too big. @hbillings, thoughts?

Closes #784


hbillings:
yassss 👍 

I agree the spacing is large. I'm noticing that with forms, too, and I'll address it in my current PR.

toolness:
Nice!

The one nit I might make is the phrase "sharing them is easy".  I forget where I read it, but I recall reading something advising content writers avoid phrases like "it's easy", "just", and "simply".  This led to a quick [conversation on #g-content](https://gsa-tts.slack.com/archives/g-content/p1475528292000883) and the content team concurred.

I'm not really sure what should be put in place of "sharing them is easy", if anything--that said, it seems like a pretty tiny and self-contained task, so might be worth posting as a quick 10 minute [writing lab task](https://github.com/18F/writing-lab/issues/new)?  Alternatively, that might be overkill, but since this is the first copy the user will see when they try to upload a price list, maybe it's worth the polish...

jseppi:
I agree, @toolness. Perhaps "and sharing them is easy" could be replaced with something more in-line with the actual goal of CALC, like "and this data will help with future proposed price research."

I am 👍 to getting Writing Lab support.

hbillings:
Didn't the writing lab look at this already? Or am I losing it?

Emileigh Barnes (in my critique group) encouraged me to take out "just" in the original, but didn't mention anything about "easy." I have critique again on Thurs and could bring it up.

jseppi:
Yeah, I think you're right @hbillings. The writing lab had no comments during the review that @tram had them do (ref https://github.com/18F/calc/issues/811).

Merge this now with the "sharing them is easy" and open new issue/PR if necessary?

toolness:
Oh, I thought this PR contains new copy though? The writing lab did indeed look at this in https://github.com/18F/writing-lab/issues/199, but I thought this PR contains new copy, and as such might warrant running by the writing lab again.

Oh weird, now that I'm actually looking at the diff, I'm noticing that the "easy" part was in the original version too. My bad!

jseppi:
Yeah, the copy here is a little new, but mostly just reordering. The newest part is the first sentence, which is to close #784. That was an issue I created myself, so I don't want to be too unilateral in these decisions :)

toolness:
Sorry, I think this is good to merge in that case! :shipit:

Issue #849, opened 2016-10-03
R10 Upload - Page titles are incorrect


jseppi:
I'm pretty sure they R10 Upload pages used to say "Upload Region 10 Data" in the `{% block title %}`, but now they say "Add Price Data".

Issue #588, opened 2016-08-23
R10 Upload - Add XLS screenshot to Step 1


jseppi:
In the R10 bulk upload flow, add a screenshot and/or descriptive text about the file we need for the Region 10 Bulk Upload.

toolness:
I'm kinda wondering if it makes sense for these "excel screenshots" to "fake" a screenshot e.g. by just using a `<table>` that we style to look like an excel one.

Among other things, it'd make the "screenshots" accessible...

jseppi:
That does make sense, they'd be easier to make small changes to that way. Though those spreadsheets are pretty wide and the S70 price list one is in a fairly complicated sheet, so styling could be a somewhat big lift.

toolness:
Hmm, I'm curious if copying the cells in excel (on either Windows or OS X) sets an HTML type on the clipboard.  If it does, just retrieving its content might actually give us something that looks decent, but it's kind of a wild guess.

PR #850, opened 2016-10-03
R10 Upload changes/fixes

jseppi:
- Add R10 spreadsheet example to R10 Upload Step 1: Closes https://github.com/18F/calc/issues/588
- Fix R10 upload page heading: Closes https://github.com/18F/calc/issues/849
- Minor R10 step template rearrangements

screen shot 2016-10-03 at 3 28 01 pm


frontend/source/sass/components/_exceltables.scss
```diff
@@ -32,7 +32,8 @@ table.excel {
   }
 }
 
-table.excel.excel-schedule-70 {
+table.excel.excel-schedule-70,
+table.excel.excel-region-10 {
```
jseppi: Not too sure why the `.excel-schedule-70` rules are not just in the `.excel` class. @toolness?

jseppi:
ready for review, @toolness

frontend/source/sass/components/_exceltables.scss
toolness: Modularity, basically--I wanted the `.excel` class to be like a "vanilla" excel spreadsheet that we could then build separate ones off of, depending on the styling of different price lists or CSVs.  Explained in the [style guide](https://calc-dev.apps.cloud.gov/styleguide/#excel-tables)!

That said, the schedule 70 one was intentionally designed to look like the S70 PPL table.  Having a region 10 one might make it seem (to devs) like that is how "the official region 10 bulk upload" sheet looks... If there *is* no such official spreadsheet, it might make sense to call the new class `table.excel.excel-generic` or something, to indicate that it's not meant to reflect a particular "official" style.

data_capture/templates/data_capture/price_list/upload_examples/region_10.html
```diff
@@ -0,0 +1,30 @@
...
+
+{% if show_debug_ui %}
+  <p class="dev-note">
+    <strong>Developers:</strong>
+    Download <a href="{% static 'data_capture/r10_export_sample.xlsx' %}"><code>r10_export_sample.xlsx</code></a>
+    to use as an example Region 10 bulk export to load.
+  </p>
+  <p class="dev-note">
+    <strong>NOTE:</strong> If you have the Debug Toolbar enabled, you will not be able to import large datasets.
+    You can disable it (and other dev/debug affordances) by setting <code>HIDE_DEBUG_UI=yup</code> in your local <code>.env</code> file.
```
toolness: Is this still true, or was this made irrelevant once we moved to the async way of doing things?

toolness:
Nice, this looks good to me!

toolness:
As an aside, I noticed the first paragraph on step 1 reads:

> Upload an exported Region 10 pricing database spreadsheet into CALC. **This will delete and replace all of the existing Region 10 data in CALC!**

That first sentence reads odd to me... I'm not sure how to describe it, but It sounds like the name of a title or heading rather than the first sentence of a paragraph.  If it doesn't sound weird to you then I'm probably just imagining things, but otherwise maybe we should file a separate issue about it?

frontend/source/sass/components/_exceltables.scss
jseppi: cool, I like that idea 👍

data_capture/templates/data_capture/price_list/upload_examples/region_10.html
jseppi: Just checked, and it indeed works now that the processing is async

jseppi:
Ok, I
- changed to just using the base `table.excel` styling,
- removed the irrelevant note about disabling djdt
- changed the intro text to be a more good sentence, and simplified the part about replacing existing r10 data.

PR #851, opened 2016-10-03
Use steps as headings

### LINKED ISSUE #844: Remove headline and use progress bar to indicate current step
This reduces redundancy and will help compress the top of the page. The downside to this is that I don't think the steps widget is currently being read by screenreaders, so getting rid of the headline will also get rid of the accessible title of the page. I need to think about how to address that.

Idea:
screen shot 2016-09-30 at 3 23 44 pm

hbillings:
Closes #844 

This eliminates the need for the repeated "Add price data to CALC" heading, instead using a faux-tab system to indicate your current page. The steps widget moves to the top of the page, the language for the steps is clearer, the steps widget is slimmer (and matches the widths used in the WDS).

screen shot 2016-10-07 at 5 33 24 pm

This also affects the CALC homepage:
screen shot 2016-10-07 at 5 33 10 pm


hbillings:
Hrm. So I like the additonal color and weight, but I'm not sure it's actually enough to serve as the heading. For instance:
screen shot 2016-10-03 at 3 40 22 pm

I'm wondering if maybe something like this would work?
screen shot 2016-10-03 at 3 42 55 pm

hbillings:
Actually, I sort of like this combo best:
screen shot 2016-10-03 at 3 55 04 pm

frontend/source/sass/components/_header.scss
```diff
@@ -141,21 +139,34 @@ nav {
...
+        a {
+          color: $color-primary-darkest;
+          text-decoration: none;
+        }
       }
       a {
         padding: 0;
         font-weight: normal;
-        font-size: $small-font-size;
+        font-size: 11px; // $small-font-size - 2 to accommodate uppercase
```
jseppi: Could you use `font-size: $small-font-size - 2px` here? I am pretty sure SASS supports subtraction.

jseppi:
Just pulled and ran this branch. It looks like the new faux-tab is not working for the About page when the user is not logged-in:

screen shot 2016-10-11 at 12 32 06 pm

jseppi:
Minor nit: padding to the right of "ABOUT CALC" looks a bit larger than the padding on the left, and in the "ADD PRICE DATA" tab:

screen shot 2016-10-11 at 12 35 26 pm

toolness:
Hmm, on narrower viewports--not just mobile, but small windows or really zoomed-in screens on desktop--there appears to be a gap under the selected tab:

> gap-under-tab

hbillings:
Interesting! Wonder what that's about😑

I've got some modifications I'll push in a few; let's see if that helps st
all.


-- 
Heather Battaglia
Innovation Specialist, 18F
Chicago, IL

hbillings:
Okay, let's try this one more time! I rearranged a bunch of the HTML to make structuring this easier.
screen shot 2016-10-13 at 3 00 59 pm

screen shot 2016-10-13 at 2 50 28 pm

hbillings:
Ugh, so...the fix for anon tabs makes the button jump:
screen shot 2016-10-13 at 3 46 46 pm
screen shot 2016-10-13 at 3 46 37 pm

What I'd *really* like to do is just move the login, which would make the button placement consistent across both experiences, but I suspect that's a discussion to have outside this PR:
screen shot 2016-10-13 at 3 52 37 pm

frontend/source/sass/components/_header.scss
```diff
@@ -102,60 +104,77 @@ header {
...
+      background-color: rgba(255, 255, 255, 0.4);
+      text-align: center;
+      border-radius: $border-radius;
       &:first-child {
-        margin-left: 0;
+        margin-left: 0rem;
       }
       &:last-child {
-        margin-right: 0;
+        margin-right: 0rem;
```
jseppi: Is `0rem` different from `0`?

jseppi:
👍

frontend/source/sass/components/_header.scss
hbillings: derp! No, and actually the front-end guide specifically recommends against using units with `0`.

PR #852, opened 2016-10-04
Make ultratest's help more helpful

### LINKED ISSUE #649: "manage.py ultratest --help" isn't very helpful
```
$ python manage.py ultratest --help
Usage: manage.py ultratest [OPTIONS] [TESTTYPE]...

  Management command to test and lint everything
```

It would be nice if we documented exactly what `TESTTYPE` meant and its possible values!

jseppi:
Looks like this now:

```sh
$ docker-compose run app python manage.py ultratest -h
Usage: manage.py ultratest [OPTIONS] [flake8|eslint|bandit|py.test]

  Test and lint everything!

  Optionally specify one or more test names to run only those tests.
```

I used the `metavar` kwarg in `@click.argument` to display the dynamically generated testnames options. To do this I also had to move some code around slightly (mostly just moving variables outside of the function definition).

Closes #649


jseppi:
ready for review, @toolness

meta/management/commands/ultratest.py
```diff
@@ -69,13 +69,64 @@ def get_coverage():
...
+        'cmd': 'bandit -r .'
+    },
+    {
+        'name': 'py.test',
+        'cmd': 'py.test --cov-report xml --cov-report term --cov'
+    },
+]
+
+
+def get_testtype_names(testtypes=None, joiner=", "):
```
jseppi: changed the default value of `testtypes` here to `None` to prevent the mutability issue.

meta/management/commands/ultratest.py
```diff
@@ -69,13 +69,64 @@ def get_coverage():
...
+        'name': 'eslint',
+        'cmd': ESLINT_CMD
+    },
+    {
+        'name': 'bandit',
+        'cmd': 'bandit -r .'
+    },
+    {
+        'name': 'py.test',
+        'cmd': 'py.test --cov-report xml --cov-report term --cov'
```
jseppi: `--cov-report term` used to only be added when `is_verbose`, but that wouldn't work with the refactor since the verbosity level is not known until the command is called. It doesn't really hurt to add the terminal reporting all the time, and nothing additional is output at verbosity level 1 since at that level, the command's `stdout` is set to `None` anyways.

toolness:
Noice, I dig it!

The one thing that `--help` doesn't seem to make fully clear is the fact that it's actually possible to specify multiple test types, e.g. `manage.py ultratest flake8 eslint`, but I'm not sure how to document that properly or whether it's easy to do with `click`.  It's not much of a big deal anyways--if someone discovers it on their own, that's cool, and if they don't, well, whatevs.

I say :shipit: !

jseppi:
The docstring for the `command` function says that one or more test names can be specified, the dosctring is shown when run with `-h/--help`.

toolness:
Oh, I didn't realize that--noice!

PR #853, opened 2016-10-04
Mark functions that modify database as atomic

### LINKED ISSUE #580: Make sure requests that write/modify DB are atomic
https://docs.djangoproject.com/en/1.10/topics/db/transactions/#controlling-transactions-explicitly

jseppi:
I've gone through the codebase and marked views and management commands that modify the database with the `@transaction.atomic` decorator.

Alternatively, we could just set `ATOMIC_REQUESTS = True` in `settings.py` to wrap every view in a transaction, though this has a performance cost. Ref: https://docs.djangoproject.com/en/1.8/topics/db/transactions/#tying-transactions-to-http-requests While this is certainly easier to do than to remember to manually wrap methods in the decorator, _most_ of our views do not modify the database, so it does seem like quite a bit of unnecessary overhead (though quantitatively I'm not sure how much or if it would actually be an issue with the amount of traffic CALC gets).

Closes #580


toolness:
Noice!

I wish Django has an `ATOMIC_POST_REQUESTS` setting that only made `POST` requests atomic--we don't have many of those endpoints, they won't get hit often, and when they do, it's probably safe to assume they should happen in a transaction... It might be possible to write middleware that does that... But I dunno.

Anyhow, this looks good to me!  I like that the [`atomic` docs](https://docs.djangoproject.com/en/dev/topics/db/transactions/#django.db.transaction.atomic) mention that atomic transactions can be nested... That way e.g. if the admin view logic is already nesting stuff in an atomic transaction when it calling the approve/unapprove logic, our use of `atomic` can't hurt.

Issue #282, opened 2016-06-14
It should be possible to disable analytics scripts during development


toolness:
If I'm on a really slow internet connection or want to develop in an offline setting, it'd be nice if I could easily disable the analytics (DAP, google analytics, ethn.io, etc).

toolness:
Once we fix #887, I believe that all our tracking scripts will load themselves asynchronously, so this won't be a problem anymore.

toolness:
Ok, I'm closing this since it's no longer possible for any of our analytics scripts to block page loading.

PR #855, opened 2016-10-04
Set and get GA_TRACKING_ID from env

jseppi:
This PR removes the hard-coded GA tracking ID from `base.html`, instead allowing it to be set via env var (`GA_TRACKING_ID`). Instead of raising an exception on being unset, I opted to have it default to a blank string since some downstream forks of CALC might not be interested in using GA.

Ref https://github.com/18F/calc/issues/282

ready for review, @toolness


hourglass/context_processors.py
```diff
@@ -9,3 +9,8 @@ def api_host(request):
 def show_debug_ui(request):
     '''Include show_debug_ui in all request contexts'''
     return {'show_debug_ui': settings.DEBUG and not settings.HIDE_DEBUG_UI}
+
+
+def google_analytics_tracking_id(request):
```
toolness: Oh, the only context processor in `frontend` is `is_safe_mode_enabled`, which I put there because I wanted all safe-mode related functionality (or at least, as much as possible) to be encapsulated in the `frontend` app--it *seemed* appropriate because "safe mode" is part of our front-end strategy of progressive enhancement.  I suppose it might be even *more* encapsulated if there were just a separate `safe_mode` Django app, but meh.

I think that my personal "litmus test" for deciding whether something goes in `frontend` sort of depends on a combination of whether it's related to our JS/SASS asset pipeline *and* whether it's something that can potentially be reused across other 18F/gov projects in the future, while `hourglass` is more for site-wide CALC stuff that isn't relevant to the asset pipeline *and* doesn't have potential for reuse across other 18F/gov projects in the future... But that might be a blurry distinction?

Anyways, I think `hourglass`'s context processors is a fine place to add the google analytics one!

toolness:
Nice, this looks great to me! Feel free to merge.

I guess the main thing to remember is just to make sure that production has this new `GA_TRACKING_ID` defined in it.  Even though it's not directly related to this repository's source code, should we file a separate issue for that and associate it with the [Pre-Prod Launch](https://github.com/18F/calc/milestone/4) milestone?

I wonder if we should also have a separate tracking ID for staging, so that we can perform manual tests on the site to make sure it's tracking the stuff we want to track...

jseppi:
> make sure that production has this new GA_TRACKING_ID defined in it

Already added it yesterday :)

> separate tracking ID for staging

I'm not sure that is necessary since I think you can see everything that you'd want to verify by using the [Google Analytics Debugger Chrome Extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en)

Example from staging:

screen shot 2016-10-05 at 10 16 59 am

toolness:
Noice!

PR #856, opened 2016-10-04
Minor changes to the Authentication and Authorization section of the README

jseppi:
Biggest change is that only users in the **Data Administrators** group can bulk upload data


toolness:
This looks good to me!

One of the things that I find a bit funky about our current permissions and roles is the fact that a **Data Administrator** must also be **Staff**, i.e. a non-Staff Data Administrator simply doesn't make much sense.  I suspect that one day, a Data Administrator (or superuser) will make a new Data Administrator but forget to also mark them as staff and then become confused. 😞

Oh, also, do you think we should add a note to `initgroups.py` where `ROLES` is defined, and mention that any changes there should be reflected in the part of the `README.md` you just edited?  I mention this b/c I didn't actually realize that we had this section in the readme until this PR 😁

jseppi:
| Data Administrator must also be Staff

Yeah, this is weird. Just to be clear the `is_staff` is necessary to access the Django Admin, right? Is that easily changed? In `hourglass/urls.py` we are using our own `staff_login_required` function to wrap `admin.site.login`. I wonder if we could just change that decorator to also allow users with the `BULK_UPLOAD_PERMISSION`?

jseppi:
Moved discussion of `is_staff` to #864

PR #859, opened 2016-10-05
Document optional env vars in selenium.md

toolness:
Just documenting the rest of the optional environment variables that weren't covered in #841.


jseppi:
👍

PR #862, opened 2016-10-05
Embed upload form in /data-capture/step/3/errors

### LINKED ISSUE #804: Allow user to upload corrected spreadsheet on Step 3 Errors page
After #792 is merged there will be a separate step for showing errors found in the upload price list spreadsheet. Rather than redirecting the user back to the previous step to upload a corrected spreadsheet, where they will lose the messages about what to actually correct, the error page itself should have a form and upload widget.

toolness:
This fixes #804.

I basically implemented the mockup @hbillings described in https://github.com/18F/calc/issues/804#issuecomment-250782942:

screen shot 2016-09-30 at 10 57 44 am

Additional notes/caveats/adjustments:

* The error alerts no longer hyperlink the phrase "try uploading it again" to step 3.  An alternative might be to leave them hyperlinked, but change the behavior so that the page scrolls down to the upload form and gives it a temporary visual highlight to attract the user's attention to it.

* If no valid rows were found, we don't display the text "Otherwise, we'll discard the rows with problems when we upload your original spreadsheet", since continuing without fixing the errors isn't an option. An alternative might be to display entirely different text if the user *must* upload a fixed form to continue--for instance, rather than "Want to try that again?" we could have "You need to try that again."  Although that might not sound very friendly, which is why I altered the copy as little as possible. 😁 

* The form now `POST`s to `/data-capture/step/3`, *not* to `/data-capture/step/3/errors`.  It is also now an ajaxform, since it needs to be in order to support the upload widget. This was by far the simplest way to embed an upload form in the errors page.

  However, this does come with a few drawbacks. In particular, it means that in the unusual case where the form is invalid--e.g. if the user uploads an empty file--the re-injected ajax form will actually be *the form from step 3*, which has different buttons at the bottom. I tried this out and I don't find it disorienting, but if users do (or if y'all think users will) then we can fix it, it'll just take more code and testing.

### DIFF
--- a/data_capture/templates/data_capture/price_list/step_3_errors.html
+++ b/data_capture/templates/data_capture/price_list/step_3_errors.html
@@ -13,8 +13,7 @@
           <h3>No valid rows found!</h3>
           <p>
             Your uploaded price list contains no valid rows. Please fix the errors
-            shown above in the original file and
-            <a href="{% url 'data_capture:step_3' %}">try uploading it again</a>.
+            shown below in the original file and try uploading it again.
           </p>
         {% else %}
           <h3>{{ total }} row{{ total|pluralize:" has,s have" }} errors</h3>
@@ -23,7 +22,7 @@
             invalid and <strong>will be discarded</strong> when you upload your
             price list.
             If you'd like, you may correct {{ total|pluralize:"this row,these rows" }}
-            in your original spreadsheet and <a href="{% url 'data_capture:step_3' %}">try uploading it again</a>.
+            in your original spreadsheet and try uploading it again.
           </p>
         {% endif %}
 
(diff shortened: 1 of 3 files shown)


jseppi:
> The error alerts no longer hyperlink the phrase "try uploading it again" to step 3.

I think just an anchor to the upload widget section could pretty helpful, especially since as @hbillings [pointed out](https://github.com/18F/calc/issues/804#issuecomment-251137034), the list of errors could potentially be pretty long, so the user might not even realize there is an upload form down near the bottom. Happy to leave that up for research/discussion/another issue though.

> If no valid rows were found, we don't display the text ...

👍

jseppi:
👍  Nice, looks good!

PR #869, opened 2016-10-05
Additional improvements to Google Analytics

toolness:
* If the user is logged out or safe mode is disabled, we set the respective custom dimension to the empty string rather than omitting it entirely. This will allow us to e.g. use [user-level scope](https://support.google.com/analytics/answer/2709828?hl=en#example-user) in a sensible way.

* GA pageviews are now triggered on ajaxform HTML injection. The strategy here is to report analytics that are consistent with what would happen in the no-JS version of the site.

* Move GA dimension-setting after `ga('create')`, because this is what a bunch of Google's own [example code](https://support.google.com/analytics/answer/2709828?hl=en) seems to do.

* Disable the user ID dimension; we're going to pursue #897 instead, hopefully, and in retrospect I'm not sure how compliant sending the user ID is, so let's just leave it out in the meantime.

To do:

- [x] ~~Use Google Analytics' "User ID" feature instead of a custom dimension (#897)~~ Looks like it might take some time to figure out if we can actually use the User ID feature, so let's not deal with this now.
- [x] ~~Use [exception tracking](https://developers.google.com/analytics/devguides/collection/analyticsjs/exceptions) for ajaxform failures.~~ Spun this off into #942.
- [x] ~~Consider using exception tracking for `error` events in general (does GA already do this by default?)~~ Spun this off into #943.
- [x] ~~Consider adding the [`download` attribute](https://davidwalsh.name/download-attribute) to download links, and track them in GA. I believe that DAP tracks *off-site* downloads but I'm not sure if it tracks them when they're located on the same domain.~~ No need for this, looking at the [DAP source code](https://github.com/digital-analytics-program/gov-wide-code/blob/master/Universal-Federated-Analytics.js#L684-L685) it seems DAP will trigger an event with category `Download` on every click of an anchor that ends with a variety of [common extension types](https://github.com/digital-analytics-program/gov-wide-code/blob/master/Universal-Federated-Analytics.js#L50) (including XLS, XLSX, CSV, and PDF). That said, it's iterating through all the anchors on the page on load, so anchors injected via Ajax (like the ones in the table on the data explorer) probably won't get caught. (Filed this as #914.)


toolness:
Ok @jseppi I've reduced the scope of this PR and I think it's ready for review.

jseppi:
👍 cool, thanks!

PR #872, opened 2016-10-05
[WIP] Show validation error on duplicate contract number

### LINKED ISSUE #865: Error on duplicate Contract Number in Step 1 Data Capture?
Since modifications are not yet implemented, should we show an error message when a duplicate Contract Number is entered in Step 1?

The error message could even say that the modification feature is coming soon.

@tram: thoughts?

jseppi:
Adds new validation in Step1Form to check that a `SubmittedPriceList` with the same `contract_number` does not already exist.

screen shot 2016-10-05 at 2 11 25 pm

Closes #865


jseppi:
I wonder if it should only validate against `SubmittedPriceList`s who have `is_approved=True`?

toolness:
Wait... so one advantage of the way we currently have this set up is that if COs are doing things the way that (IIRC) they're *supposed* to be doing things, it's actually possible to support price modifications, so long as the CO uploads a complete replacement of the price list, rather than just the changed rows. 

If they do that, then the data administrator needs to un-approve the old price list and approve the new one. However, this workflow will become impossible with the validation logic in this PR.

Hmm, I can't think of any really straightforward solutions here... There are lots of different approaches swimming around in my head right now! I wonder if we should just put this on the back-burner for now until we discuss it more w/ Kelly and folks?

jseppi:
I am fine with tabling it for now. I think there might be two conflated situations here, _maybe_.

There are actual contract modifications, which I think could result in some weird baseline pricing logic (ie, the prices in the uploaded spreadsheets are actually not a base price, but some mid-term price). My understanding is that this is what the `Contract Year` field was initially supposed to solve for, but actually that whole situation is not clear to me. This is what my additional validation logic was supposed to prevent until we get that solved.

Then there is the case of modifying an already uploaded price list spreadsheet, with the use case of just maybe fixing something that is wrong. Leaving this PR's validation logic out will allow this use case (though the complicated part of the workflow will be on the Data Administrator to sort out).

toolness:
Yep, I agree!

Another potential alternative for the "modifying an already-uploaded price list" is to actually provide an edit view for not-yet-approved price lists... lots of possibilities!

hbillings:
Coming in late here...I think we potentially will have multiple price lists with the same contract number, just from different vendors. If I understand correctly, different vendors can be awarded different parts of a contract, but they all roll up to the same contract number. (This is definitely something we should check with Kelly on. Just wanted to raise that!)

jseppi:
Ah, thanks, @hbillings - I didn't realize that could be the case!

jseppi:
Looks like we'd be ok just checking the Contract Number. Kelly has clarified:

> One vendor per contract number. However, there could be vendors with multiple contract numbers.

hbillings:
:party_parrot:

jseppi:
Closing in favor of actually implementing a replacement flow!

Issue #880, opened 2016-10-06
Hyperlink "try uploading it again" to embedded upload form in step 3 errors


toolness:
In #862 we embedded the upload form in step 3 errors rather than redirecting the user to step 3.

However, the "try uploading it again" copy at the top of the page no longer hyperlinks to anything. As @hbillings [pointed out](https://github.com/18F/calc/issues/804#issuecomment-251137034), the list of errors could potentially be pretty long, so the user might not even realize there is an upload form down near the bottom, so we might want to re-hyperlink the text to the bottom of the page.

jseppi:
@toolness - want to take this one now that #884 is merged?

toolness:
Sure!

PR #884, opened 2016-10-06
Smoothly scroll between in-page links

toolness:
This adds smooth scrolling between in-page links.

Right now I think it only really does anything on our style guide, because that's the only "CALC 2.0" part of the site that has in-page links. But once we address #880, it will be useful for the price list upload page too.

## Limitations

* This uses `history.replaceState()`.  As such, it will clobber any existing use of `pushState()`, which we fortunately don't have--but it might befuddle folks in the future who may decide that they want to use `pushState()`, so we might want to document it.

* This seems to work fine on IE11, though the `popstate` stuff only seems to work intermittently. I think that's OK though, since it instantaneously scrolls the page to the right place.

* When navigating between history entries using the back/forward buttons, there is an almost-imperceptible flash on most non-Chrome browsers because the browser's default scrolling behavior has kicked in on `popstate`, and we immediately override it. Chrome makes it easy for us to override this default behavior via its fancy [scroll restoration API](https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration) but I don't think any other browsers have implemented it. Since it's almost imperceptible, I don't *think* it should be that big a deal, but let me know if you disagree.

* This doesn't currently work with links whose `href` attribute doesn't start with a hash. For instance, if the in-page links are all actually *absolute* links, this won't work. It's okay though, since the browser will just fall back to its default non-smooth-scrolling behavior.

## To do

- [x] Unit tests, I guess.
- [x] Currently if you have three history entries and press the back button twice consecutively, the page will first scroll to the previous one and then scroll to the next one. It'd be nice if it "aborted" the first scroll as soon as it noticed that the user wanted to see the second scroll. We could deal with this in this PR or just file a separate issue.


toolness:
@jseppi aiight i think this is good to go. code got a bit bloaty due to making it testable.

styleguide/templates/styleguide_section.html
```diff
@@ -1,4 +1,4 @@
-<h3 class="styleguide-section" id="{{ section.id }}">
+<h3 class="styleguide-section" id="{{ section.id }}" tabindex="-1">
```
toolness: Ehhh, on second thought I really don't like this effect; it doesn't make a lot of sense because the element isn't interactive anyways, and browser's don't normally provide focus outlines on named anchors, so I removed the focus outline in 89e6bdb8ba18ae1ad198bdc6e286966a1a0ad1a7.

frontend/source/js/tests/smooth-scroll_tests.js
```diff
@@ -0,0 +1,82 @@
...
+    '<h1>hello</h1>',
+    '<h2>there</h2>',
+    '<h2 id="foo">foo</h1>',
+  ].join('\n'));
+  iframe.contentDocument.close();
+
+  addHashChangePolyfill(iframe.contentWindow);
+
+  const done = assert.async();
+  const steps = (function* runSteps() {
```
jseppi: whoa, this is some fancy javascript up in here

jseppi:
Pretty sweet! The smooth scroll effect works well and seems to be at a nice speed to not feel annoying.

Browsers, JavaScript, and A11Y definitely seem to be at odds in a lot of ways :\ 

As you said, some documentation about `pushState` clobbering might be nice, but I don't think it is strictly necessary (I can't really foresee another place we'd use it). Perhaps in a Style Guide entry about Smooth Scroll?

👍 Feel free to merge when you are ready

jseppi:
Oh yeah, like I mentioned previously, I think it would be 💯 if you were able to publish this to npm!

frontend/source/sass/base/_typography.scss
```diff
@@ -41,6 +41,23 @@ h6 {
...
+    // The tabindex="-1" has only been provided to work around a
+    // hasLayout bug in IE that prevents sequential keyboard navigation
+    // on named anchors from working properly. We don't want non-IE
+    // browsers to actually provide a focus outline on these elements,
+    // because the elements aren't actually interactive and browsers
+    // don't provide focus styling on named anchors by default.
+    //
+    // For more information on the IE bug, see:
+    //
+    // http://juicystudio.com/article/ie-keyboard-navigation.php
```
toolness: Yup. I can't believe it's been a bug in IE since at least 2005 and they never fixed it. 😞

styleguide/templates/styleguide_section.html
jseppi: um, I just stopped and re-`docker-compose up`'d, and now it the focus outline is not showing up.

toolness: Oh lol! Ok awesome, I will go ahead and merge then!

toolness:
Sweet, thanks!  Yeah, I think I may decouple it from jQuery and publish it to npm. The only part I'm not looking forward to is setting up a whole gulp/eslint/babel pipeline for the one dinky script... Hmm. And I'd have to set up the test suite too and integrate it with travis... Hmmm.

styleguide/templates/styleguide_section.html
jseppi: I have no idea why it happened/stopped happening.

frontend/source/js/tests/smooth-scroll_tests.js
toolness: Haha, I was hoping it wasn't too off-the-wall 😁 

**Aside:** we were actually writing code like this for Firefox extensions back in the 2000s, since SpiderMonkey has supported generators--and a bunch of other ES2015 features--for a pretty long time. It was especially funny when coworkers who learned JS at Mozilla would interview incoming candidates who came from the web world: at interview debriefs in 2008 they'd say things like "wtf, this person claimed to know JS, but they don't even know what `let` is."

Issue #875, opened 2016-10-05
Get a deployment review by an expert cloud.gov/cloud foundry engineer


jseppi:
Specific things we need help with:

- General review of practices, manifests, resource sizing, etc.
- Zero-downtime deploys
  - manual blue-green deploys or cf-autopilot?
- Possible manifest inheritance
- Proper way to migrate production database during deployment
- What to do about rqworker and rqschedule instances
  - have both `a` and `b` versions of each (like we do currently, but that probably won't work as I intended)
- ~~Figure out this weird `panic: send on closed channel` error when attempting to push all 3 apps: https://github.com/18F/calc/issues/877~~ upgrading `cf-cli` fixes it

jseppi:
@toolness: feel free to clarify/modify/addify to the list of topics in the root comment

rogeruiz:
Here are some recommendations. I would also ping in #cloud-gov-support for any other opinions.

> General review of practices, manifests, resource sizing, etc.

This is something that I would say falls on the specific needs for your application. You can run the `cf org <ORG_NAME>` to get information about the memory size available for you across all the spaces in the organization. Keep in mind that things like zero-downtime deployments eat into double the amount of resources during the deployment because two exact instances of your application are built. So two instances at 1G each will consume 4G total when performing a zero-downtime deployment.

> Zero-downtime deploys
> manual blue-green deploys or cf-autopilot

I recommend using `cf-autopilot` based on anecdotal evidence that it works better. It also enforces having `manifest.yml` files as they're required to use it. When using this plugin, your CI server will need to install it before it deploys your code.

> Possible manifest inheritance

Since manifests are configuration, I would go against inheritance as there is no way to override properties inherited from an earlier manifest. This is especially important for routes. So I would be careful not to inherit any route properties across your manifests. Also, any property that can be declared within an `application` stanza can be declared outside of it in the manifest file. This allows for inheritance within a single manifest file.

> Proper way to migrate production database during deployment

Super tricky subject. Migrations that may adversely affect your current application should be tested on in staging to avoid issues like that. Because spaces share bound service between instances of the same application, you may want to look into binding a new service temporarily to the application and migrating data between them and shutting the other service and application down. This would be a complicated process though that I would avoid unless the migrations absolutely needed to be automated with zero-downtime. For DB migrations, nothing beats a clear documented guide and SLA for database migrations which could always inadvertently cause the application to crash.

> What to do about rqworker and rqschedule instances
[...]

PR #885, opened 2016-10-06
Switch to autopilot plugin for zero-downtime deploys

jseppi:
This PR switches us from manual blue-green production deploys to using the CF autopilot plugin for doing zero-downtime deploys.

An important change to note is that previously we were basically "ready to go" to switch over to the prod app as soon as soon as we got the word on the ATO. However, since now I have the prod app back to being named `calc-prod` (instead of `-a` and `-b`), I can't actually "pre-deploy" it because it would override the current `calc-prod` app. Not a huge deal -- I've verified that doing a manual deploy will work in the `calc-dev` space. And the prod services (calc-db, calc-redis, calc-env) are already setup and ready to go in `calc-prod`.

To Do:

- [x] setup `autopilot` on Travis
  - [x] validate that it works. This will probably involve temporarily setting the target branch to this PR's branch.
- [x] remove `calc-prod-a` and `calc-prod-b` in favor of only a single prod manifest (since the plugin will do the work of duplicating the web app on deploy).
- [x] update **Production Servers** section of `deploy.md`
- [x] add Staging checks back to New Relic 🎉  -- NR checks on staging were previously too annoying because the app would go down on every push into `develop`. However, now that we have zero-downtime deploys on staging, this shouldn't be an issue. So now we'll know if a push breaks staging!
- [x] delete `calc-prod` space `-a` and `-b` apps and routes

Ref: advice in https://github.com/18F/calc/issues/875


jseppi:
FYI, the new `deploy-travis.sh` is based on https://github.com/18F/epa-notice/blob/master/deploy.sh

jseppi:
ready for review, @toolness

.travis.yml
```diff
@@ -29,6 +29,7 @@ env:
   global:
   - TRAVIS_NODE_VERSION=6.5.0
   - PHANTOMJS_TIMEOUT=15
+  - CF_CLI_VERSION=6.22.1
```
rogeruiz: Nice! This is important because of your multi-application manifest! 🎉

deploy.md
```diff
@@ -3,8 +3,11 @@
 **Only of interest to 18F team members**
 
 If you don’t already have one, request a Cloud Foundry account in #devops
```
rogeruiz: I would change this from `#devops` to `#cloud-gov-support` to be consistent. There is no longer a `#devops` channel. Even though, everyone should automatically get a ☁️.gov account now.

deploy.md
```diff
@@ -30,18 +33,14 @@ so only production dependencies will be installed.
...
   - `calc-prod` (production) space:
-    - `calc-prod-a`
-    - `calc-rqworker-a`
-    - `calc-rqscheduler-a`
-    - `calc-prod-b`
-    - `calc-rqworker-b`
-    - `calc-rqscheduler-b`
+    - `calc-prod`
+    - `calc-rqworker`
+    - `calc-rqscheduler`
```
rogeruiz: Nice clean up. Not sure what the `-a` versus `-b` were.

jseppi: Sweet!
The `-a` and `-b` versions of everything was my attempt at (manual) blue-green deploys. Basically duplicated versions of everything so that while one version was being deployed, the other would be live. The autopilot plugin is much better!

.travis.yml
```diff
@@ -43,17 +44,15 @@ env:
   - WD_SOCKET_TIMEOUT=60
   - SAUCE_USERNAME=atulvarma
   - secure: "RbgG35nc1SHgjdxVbGUDlumePeR/+dXq5fMbfTXOW8uFYw5kGh75px2KdFDkL1JIA1vbfxyNFm09x32uXMcHsBKu9fJS7xbpMrRDbJoV2wVy5wd6qtvZ7yFbR87Fqo3/tuGu13jNNDvUgKjpUlrzLZKaL7qvv5Gd7CVhkySUa1o="
+before_deploy:
+- export PATH=$HOME:$PATH
+- travis_retry curl -L -o $HOME/cf.tgz "https://cli.run.pivotal.io/stable?release=linux64-binary&version=$CF_CLI_VERSION"
+- tar xzvf $HOME/cf.tgz -C $HOME
+- travis_retry cf install-plugin autopilot -f -r CF-Community
```
toolness: Oh neat, I didn't know about `travis_retry`. I assume it just retries the given command some number of times so that random network hiccups don't fail the build?

jseppi: Yup, it tries it 3 times before failing

PR #892, opened 2016-10-06
Validate Unit of Issue in S70 Price List uploads

### LINKED ISSUE #886: Validate "UNIT OF ISSUE" is "Hour" for every row in Price List upload
While the "UNIT OF ISSUE" is not strictly necessary when we construct `Contract` models from an uploaded price list, we could easily ensure that the Unit of Issue value in each price row is "hour". This would help avoid getting non-hourly rate data in CALC, which is already a problem with some rows of the Region 10 bulk data (see screenshot).

Think this is worth doing @tram?

related: #876 

screen shot 2016-10-06 at 12 53 32 pm

jseppi:
Adds simple validation of the "UNIT OF ISSUE" column of the S70 price list spreadsheet. The validation is a case-insensitive check for the values "hour" or "hourly" because we don't actually want any non-hourly rates.

This also standardizes the Unit of Issue to always be "Hour" (which is what the proposed price list template says it should be) IFF it matches the hour/hourly regex.

To Do:
- [x] tests
- [x] clean/coerce Unit of Issue to "Hour"

Here is how it looks, @hbillings:

Error table:
screen shot 2016-10-07 at 10 29 01 am

Good data table:
screen shot 2016-10-06 at 5 11 40 pm

Closes #886


jseppi:
Ready for review, @toolness

toolness:
This looks good to me. Do we think that there might be other variations here that are valid but that we don't yet know about, e.g. `per hour` or something like that?  I wonder how we'd surface that back to us, so that we could use the data to figure out that there are a preponderance of values that we should consider accepting or something.  Hrm.

jseppi:
> Do we think that there might be other variations

I just manually spot-checked about 20 of the sample price lists that @tram collected. They by and large had "hour" or "hourly" as their Unit of Issue value. Two of them had "HR" and "per hour" though. I think after this PR gets merged I'll resurrect my test harness and see if those variations are actually very common or not.

data_capture/schedules/s70.py
```diff
@@ -192,6 +193,10 @@ class Schedule70Row(forms.Form):
     min_years_experience = forms.IntegerField(
         label="Minimum years of experience"
     )
+    unit_of_issue = forms.CharField(
+        label="Unit of issue",
+        validators=[hourly_rates_only_validator]
```
jseppi: ok done. The value has to be coerced, not cleaned, because it is the coerced values that make it into `gleaned_data`, which is what eventually gets rendered in the tables. I think.

jseppi:
ready for review again, @toolness

toolness:
Nice, looks good to me!

PR #893, opened 2016-10-07
Remove parsing/validation of non-used columns of the S70 price list

### LINKED ISSUE #876: Failing S70 Price List Uploads when missing non-necessary columns
The S70 price list table has a whole bunch of columns, many of which are not used at all during the data-capture process (like `'COMMERCIAL LIST PRICE (CPL) OR MARKET PRICES'` and `'PRICE OFFERED TO GSA (excluding IFF)'`). However, the uploaded file will not be accepted if any of even these non-necessary columns are missing. 

Should we "soften" our validation logic to not care about those fields? Pretty sure it would be as simple as removing them from `DEFAULT_FIELD_TITLE_MAP` in `s70.py`.

screen shot 2016-10-05 at 5 22 49 pm

cc @toolness @tram

jseppi:
Opened against #892 

Closes #876


jseppi:
ready for review, @toolness

toolness:
This looks splendid!

PR #895, opened 2016-10-07
Link to Sauce Labs results when selenium tests fail

toolness:
This uses the instructions in [Building links to test results](https://wiki.saucelabs.com/display/DOCS/Building+Links+to+Test+Results) to display a link to sauce labs results that is visible in the `stdout` log when selenium tests fail.

It also hides successful test results in QUnit, so that hopefully information relevant to the failing test(s) is visible in the Sauce Labs screencast.

You can see what this looks like in #896, which adds a trivial failing test to the build:

travis-log

I guess it's not super easy to notice, but it's there if you know where to look... Definitely a step up from not having any idea and needing to scour the Sauce Labs logs for clues.

Here's what the QUnit test output looks like when we only show failing tests:

qunit-output

### DIFF
--- a/frontend/tests/test_qunit.py
+++ b/frontend/tests/test_qunit.py
@@ -10,6 +10,7 @@ from .test_selenium import WD_TESTING_BROWSER, SeleniumTestCase
 
 urlpatterns += [tests_url]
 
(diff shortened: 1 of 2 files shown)


jseppi:
Sweet! 👍

PR #899, opened 2016-10-07
In 'common/ga', always proxy to window.ga() instead of permanently caching it

toolness:
From [how analytics.js works](https://developers.google.com/analytics/devguides/collection/analyticsjs/how-analyticsjs-works):

> The JavaScript tracking snippet defines a global `ga` function known as the "command queue". It's called the command queue because rather than executing the commands it receives immediately, it adds them to a queue that delays execution until the analytics.js library is fully loaded.

The same page later states:

> Once the analytics.js library is loaded, it inspects the contents of the `ga.q` array and executes each command in order. After that, **the `ga()` function is redefined**, so all subsequent calls execute immediately.

(Emphasis mine.)

The fact that `ga` is *redefined* means that we shouldn't be caching it in our `common/ga` module, because if `ga` fails to load before the bundle is loaded, `common/ga`'s exported function will always be the "command queue" version, rather than the newly-redefined one.

This PR just makes `common/ga` export a "wrapper" around `window.ga` so that the current `window.ga` is always proxied-to.

### DIFF
--- a/frontend/source/js/common/ga.js
+++ b/frontend/source/js/common/ga.js
@@ -3,9 +3,16 @@
 /**
  * This is just a wrapper around Google Analytics' ga() function, which
  * should be included inline at the top of the page. If it wasn't included
- * at the top of the page, however, we'll just provide a no-op function
+ * at the top of the page, however, we'll just no-op
  * so that code which calls it still works, allowing GA to be an optional
  * dependency.
  */
 
-module.exports = typeof window.ga === 'function' ? window.ga : () => {};
+function ga(...args) {
+  if (typeof window.ga === 'function') {
+    return window.ga.apply(this, args);
+  }
+  return undefined;
+}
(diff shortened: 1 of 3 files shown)


jseppi:
👍 How did you come across this -- were you noticing missing events or just good documentation sleuthing?

toolness:
Just documentation sleuthing--I knew that GA did some funky trickery to allow itself to queue stuff up before it was actually loaded, and reading about the trickery led me to try setting `window._origGa = ga` right after our GA snippet... then I loaded the page and compared `_origGA` to `ga` and they weren't the same thing so I knew our current implementation would be janky.

That said, I wonder if the "command queue" version itself checks to see if analytics.js has already been loaded, and proxies to it if so... That'd be cool... In any case though, our solution should work so I'll go ahead and merge it!

PR #906, opened 2016-10-07
Init New Relic agent in wsgi.py

### LINKED ISSUE #902: New Relic not reporting on Staging :(
I misspoke at stand-up today. New Relic is _not_ working on staging due to my overlooking how autopilot works.

Basically, we can't use `cf set-env` to set environment variables because autopilot won't know about them. The only thing we are using `set-env` for is `NEW_RELIC_LICENSE_KEY`, as documented in `deploy.md`.

So I need to figure out another way to get our private NR key into cloud.gov.

jseppi:
This PR changes how we run the New Relic agent in staging/prod from using `newrelic-admin run-program` to initializing the NR agent from within our `wsgi.py`. This gives us the ability to store `NEW_RELIC_LICENSE_KEY` in our User Provided Service credentials instead of using `cf set-env`.  Not using `set-env` is important because of how the autopilot zero-downtime deploy plugin works (it creates a new app based on the manifest, so you can't ever have manually set environment variables).

Closes #902 (I hope)


hourglass/wsgi.py
```diff
@@ -4,14 +4,27 @@
 It exposes the WSGI callable as a module-level variable named ``application``.
 
 For more information on this file, see
-https://docs.djangoproject.com/en/1.7/howto/deployment/wsgi/
+https://docs.djangoproject.com/en/1.8/howto/deployment/wsgi/
```
jseppi: updated this link to proper version

hourglass/wsgi.py
```diff
@@ -4,14 +4,27 @@
...
 
 import os
 from django.core.wsgi import get_wsgi_application
-
+import newrelic.agent
+from hourglass.settings_utils import load_cups_from_vcap_services
 
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "hourglass.settings")
+
+load_cups_from_vcap_services('calc-env')
```
jseppi: So now `load_cups_from_vcap_services` is called both in this file and in `settings.py`. I can't really foresee this being a problem, but it is kinda ugly to me.

newrelic.ini
```diff
@@ -50,17 +50,7 @@ ssl = true
...
-# URL and send it as the request parameters for display in the
-# UI. When "true", it is still possible to exclude specific
-# values from being captured using the "ignored_params" setting.
-capture_params = false
-
-# Space separated list of variables that should be removed from
-# the query string captured for display as the request
-# parameters in the UI.
-ignored_params =
+attributes.exclude = request.parameters.*
```
jseppi: `capture_params` is a deprecated setting, so I replaced it with the new equivalent setting and value

requirements.txt
```diff
@@ -7,7 +7,7 @@ django-cors-headers==1.1.0
 djangorestframework==3.4.3
 djorm-ext-pgfulltext==0.9.3
 gunicorn==19.6.0
-newrelic==2.68.0.50
+newrelic==2.70.0.51
```
jseppi: uuuuuuupgrade

hourglass/wsgi.py
toolness: The only *potential* problem I see involves the duplication of the `calc-env` string... I can't imagine that string ever changing, but if it does, there's a possibility that it might only get changed in `settings.py`, which would then break the zero-time downtime deploys and nobody would find out about it until the deployment broke... I guess we could make the first argument of `load_cups_from_vcap_services` just default to `calc-env` in `settings_utils.py` so that it wouldn't have to be repeated in both files, but it's not a big deal either way IMO.

hourglass/wsgi.py
```diff
@@ -4,14 +4,27 @@
...
+
+
+nr_license = os.environ.get('NEW_RELIC_LICENSE_KEY')
+nr_app_name = os.environ.get('NEW_RELIC_APP_NAME')
+if nr_license and nr_app_name:
+    nr_settings = newrelic.agent.global_settings()
+    nr_settings.license_key = nr_license
+    nr_settings.app_name = nr_app_name
+    newrelic.agent.initialize()
```
toolness: I wish newrelic wasn't so "magical"... why doesn't it just provide an actual function that wraps a WSGI app we give it? Bleh.

toolness:
Super cool, this looks 👍 to me!

hourglass/wsgi.py
jseppi: Oh, it does have a WSGI wrap: https://docs.newrelic.com/docs/agents/python-agent/installation-configuration/python-agent-integration#wsgi-application

I cribbed this from though: https://github.com/18F/epa-notice/blob/c4ad0987c032ffacc7bf49233d92bda9c464fd5a/notice_and_comment/newrelic.py

jseppi: Ugh, so I just tried the wrapper, and then when I start the app it tells me `newrelic.core.agent WARNING - The Python Agent is not enabled.` I'm just going to leave it as I have it.

hourglass/wsgi.py
jseppi: Cool, I'll change `load_cups_from_vcap_services` to use the default value of `calc-env`.

PR #912, opened 2016-10-11
Increase esclation_rate input size

### LINKED ISSUE #911: Escalation rate more big
Whoops, looks like the escalation rate input needs to be slightly bigger because rates like "1.5" won't quite fit in there. Sorry, I should noticed this before merging #908.

screen shot 2016-10-11 at 2 30 49 pm

jseppi:
I bumped it up from `5rem` to `8rem`:

screen shot 2016-10-11 at 2 39 47 pm

Closes #911

Let me know if ok, @hbillings


hbillings:
Can we make it `7rem`, or is that still too small? (`5rem` is the size of the month and year date boxes above the escalation rate; the year box is `7rem`. I'm obsessive, sry >_<)

jseppi:
`7rem` won't fit 3 digits after the decimal:

screen shot 2016-10-11 at 3 11 01 pm

I don't know what actual escalation rates will look like however, so maybe we should find that out first :)

hbillings:
aaaaargh *shakes fist*

okay fine make it `9 rem` :-p

jseppi:
done!

PR #915, opened 2016-10-12
use IS_SUPPORTED in check to activate

jseppi:
Whoops, should have caught this in the #884 review. Just replaced repeated conditions with `IS_SUPPORTED`.


toolness:
Doh, good catch! Thanks.

PR #916, opened 2016-10-12
Make DAP auto tracker work w/ dynamic content

### LINKED ISSUE #914: PDF downloads aren't being tracked
As mentioned in #869:

> Looking at the [DAP source code](https://github.com/digital-analytics-program/gov-wide-code/blob/master/Universal-Federated-Analytics.js#L684-L685) it seems DAP will trigger an event with category `Download` on every click of an anchor that ends with a variety of [common extension types](https://github.com/digital-analytics-program/gov-wide-code/blob/master/Universal-Federated-Analytics.js#L50) (including XLS, XLSX, CSV, and PDF). That said, **it's iterating through all the anchors on the page on load**, so anchors injected via Ajax (like the ones in the table on the data explorer) probably won't get caught.

toolness:
This is an attempt to fix #914 by using a mutation observer to automatically ~~re-initialize~~ notify DAP's auto link tracker whenever new links are added to a page as per the instructions in https://github.com/digital-analytics-program/gov-wide-code/pull/47. ~~However, it also suffers from the same~~ It works around the issues described in https://github.com/digital-analytics-program/gov-wide-code/pull/47#issuecomment-253232382. This is done by temporarily overriding `document.getElementsByTagName('a')`, which DAP's initializer calls, to return _only_ the new links that have been added to the page.

To do:
- [x] Determine if we actually want to use a solution like this.
- [x] Add tests.
- [x] Make the new script code actually run on the data explorer (currently it only runs on the data explorer pages). This will involve addressing #917.


toolness:
Ok, I filed https://github.com/digital-analytics-program/gov-wide-code/pull/48 to work around the problem mentioned in https://github.com/digital-analytics-program/gov-wide-code/pull/47#issuecomment-253232382.

frontend/source/js/common/ga.js
```diff
@@ -16,3 +16,87 @@ function ga(...args) {
...
+  window.document.getElementsByTagName = () => links;
+
+  try {
+    initAutoTracker();
+  } finally {
+    window.document.getElementsByTagName = oldGet;
+  }
+}
+
+if ('MutationObserver' in window) {
```
toolness: `MutationObserver` is the same API that our Custom Elements polyfill is using--it works on IE11 and I verified it.

toolness:
@jseppi any thoughts on this weird approach?  It's working on the data explorer now, so you can try clicking on PDF links and it should report them to GA.

jseppi:
Ooof. The code looks good to me. It is a shame that the DAP API doesn't provide a better way to handle this.

toolness:
Ok this should be ready for review @jseppi !

jseppi:
👍 

Hopefully your https://github.com/digital-analytics-program/gov-wide-code/pull/48 gets merged somewhat soon and we can remove the hax!

PR #918, opened 2016-10-13
Add Modal Dialog for Canceling

### LINKED ISSUE #909: Make "Cancel" friendlier
Currently, hitting Cancel bounces you back to CALC's main page and deletes all your data, which is pretty destructive if you don't know what it's going to do. Login.gov has a similar issue, and what they do is throw up a modal that says, "Are you sure you want to cancel? X and Y will happen, and you'll lose Z." 

We could also call the cancel link something more descriptive, like "Cancel price list upload" or "Exit upload flow."

jseppi:
This PR adds a11y-friendly, modal dialog support (via the [a11y-dialog](https://github.com/edenspiekermann/a11y-dialog) library).

The "Cancel" buttons in both the Price List Upload and Region 10 Upload flows have been progressively-enhanced so they pop up a modal dialog instead of simply canceling without warning. In the case of no JS, then the Cancel buttons will work as they previously did.

Here's how it looks:
screen shot 2016-10-14 at 10 26 48 am

Of note is that the example in the Style Guide will not work properly with screen readers because the `{% example %}` block that holds the example markup is inside the page's `<main>` element, which has an `aria-hidden="true"` attribute added to it when a modal is shown.

To Do:
- [x] hook up cancel `POST` 
- [x] style review by @hbillings 

Closes #909


toolness:
O man, modals are hard accessibility-wise. One helpful resource might be [a11y.js's accessible dialog tutorial](https://allyjs.io/tutorials/accessible-dialog.html).

Also, is there a way this can still work in a progressively enhanced way?  even if, say, the current non-confirming behavior is retained for non-JS folks, and the modal is a progressive enhancement? (Alternatively we could have a separate interstitial page view but I'm guessing that would be harder.)

jseppi:
> even if, say, the current non-confirming behavior is retained for non-JS folks, and the modal is a progressive enhancement? 

Yup, that's the idea. Right now it the modal is hooked up [via a class, `.js-cancel`](https://github.com/18F/calc/pull/918/files#diff-8b94e3152fbf486d9742b81f1838a86bR18), though I'm not sure if that'll end up being the final way I hook it up.

frontend/source/js/data-capture/modal-dialogs.js
```diff
@@ -0,0 +1,38 @@
+/* global $, document, window */
+
+const A11yDialog = require('a11y-dialog');
+
+$(document).ready(() => {
+  const dialogs = {};
+
+  $('[data-modal-target]').each((i, el) => {
```
jseppi: hrm, I wonder if I should make this a Web Component (using the `is` attribute) instead?

jseppi:
Ready for review, @toolness and @hbillings

hbillings:
I'm about to invoke those famous last words, "How hard would it be to...?" 
- set a border radius of `$border-radius` on the modal itself
- remove the `:focus` style around the headline (I think this should just be something like `h1:focus { box-shadow: none; }`
- left-align the buttons (or center everything in the modal)
- make the `p` the visual equivalent of an `h3`, or make the `h1` equivalent to an `h3` and bold the `p` (the discrepancy between the sizes makes it easy to gloss over "you will lose your work, danger, danger")
- maybe tweak the language slightly to make it active voice, like, "If you choose to cancel, CALC will discard all of the information you've entered."

jseppi:
That all sounds doable!

jseppi:
@hbillings: I've made styling and language as you suggested. The one thing I didn't do is remove the focus outline from the `h1` because if it is removed, the element is still focusable, but the user can't see that it is. I can't just take off its `tabindex` and make it un-focusable because something in the modal needs to be focusable (other than the buttons) so that focus can jump to the top of it once it appears.

frontend/source/sass/components/_dialogs.scss
```diff
@@ -0,0 +1,54 @@
...
+    width: 500px;
+  }
+
+  form {
+    margin-bottom: 0;
+    display: inline-block;
+  }
+
+  h1 {
+    font-size: $h3-font-size;
```
jseppi: i copied all three of these rules from `_typography.scss`'s `h3` rule. Is that what you had in mind, @hbillings?
Perhaps a cleaner way to do this would be to add an additional selector of `.dialog-content h1` to the `h3` ruleset in `_typography.scss` instead of having it repeated here.

hbillings: I think adding it to `_typography.scss` makes sense! That way they stay in step if one changes. Maybe we even have some sort of more generic `h1.secondary` rule that we can apply to things that need to be `h1`s but are too big styled as such. (Though `h1.secondary` sort of makes my skin crawl...)

hbillings:
@jseppi w00t! Um, so, I hid the focus ring on error alerts ages ago because I thought the focus was only important for screen readers and that it wouldn't mean anything to non-visually-impaired folks. 😬  Should I go back and re-add the focus there?

jseppi:
> Should I go back and re-add the focus there?

Hrm, I _think_ that would be best for keyboard users, but I'm a neophyte with the a11y stuff. I bet @toolness knows!

toolness:
Oy, that's a really good question, and I'm not sure what the best answer is--I've just filed this as #925 and suggest you keep focus styling out for now, and depending on the result of that issue we can easily change things later if need be!

jseppi:
Ok, I removed the focus outline on the dialog's `h1` and I moved the rest of the `.dialog-content h1` rules over to `_typography.scss`. This should be good to go now!
Plz code review, @toolness.

data_capture/templates/data_capture/step.html
```diff
@@ -3,37 +3,65 @@
...
-      <label>Validate data</label>
-    </li>
-    <li {% if step.number == 5 %}class="current"{% endif %}>
-      <label>Submit data</label>
-    </li>
-  </ol>
-{% endblock %}
-
-{% block step_heading %}{% endblock %}
+  <div class="container">
```
jseppi: This block of changes is just indentation. I like indentation in template blocks so I can see where they start and end.

hbillings:
Okay, it looks like we need the focus outline. Does the focused element need to be the h1? Could we make it something smaller above the h1 or something? hrrrrrm...

jseppi:
> Could we make it something smaller above the h1 or something

Hrm, so in the case of the Cancel modal, it actually doesn't seem bad to have the focus jump to the "No" button at first, so I could just remove the `tabindex=0` from the `h1`. And it works really well with VoiceOver because it reads the `h1` first still.

We could then decide what to do for other modal dialogs as they come up I suppose.

jseppi:
Changes implemented and screenshot updated

toolness:
Hmm, so I noticed two problems when testing this with NVDA:

1. It doesn't read the `<h1>` in the modal, instead just saying "No, keep working" as soon as the user activates the Cancel button. Is it possible to put a `tabindex="-1"` on the entire modal and focus it instead? This would make screen readers announce the `<h1>` and it also means pressing <kbd>Tab</kbd> would immediately take the user to the "No, keep working" button.
2. Without NVDA, pressing <kbd>Enter</kbd> or <kbd>Space</kbd> on one of the dialog buttons works fine, but *with* NVDA, pressing either of those keys toggles the Django Debug Toolbar widget, which is completely bizarre. I think I'll have to do some further investigation to figure out why this is happening.

frontend/source/js/data-capture/modal-dialogs.js
```diff
@@ -0,0 +1,34 @@
...
+      // Hook up listeners to add/remove the 'no-scroll' class to the
+      // document body on hide/show of the modal dialog
+      dialogEl.addEventListener('dialog:show', () => {
+        const className = `${document.body.className} no-scroll`.trim();
+        document.body.className = className;
+      });
+
+      dialogEl.addEventListener('dialog:hide', () => {
+        const className = document.body.className.replace('no-scroll', '').trim();
+        document.body.className = className;
```
toolness: Is there a reason we're not just using [jQuery's class attribute functions](https://api.jquery.com/category/manipulation/class-attribute/) here? Without them or `classList` I'm thinking of weird edge cases like if we end up making up a new class called `foo-no-scroll`, which could make this code mess up and stuff...

data_capture/templates/data_capture/step.html
```diff
@@ -3,37 +3,66 @@
...
+        </h1>
+
+        <p id="cancel-dialog-content">
+          If you choose to cancel, CALC will discard all of the information you've entered.
+        </p>
+
+        <div class="dialog-buttons">
+          <button class="button-primary" data-a11y-dialog-hide>
+            No, keep working
+          </button>
```
toolness: So the `a11y-dialog` docs claim that this button "does need to have the `type-"button"` attribute".  I was hoping this might be the key to fixing the NVDA issue, but it's not... I'm wondering why the `type-"button"` attribute is supposed to be on here, though... Wish they explained that in the docs.

frontend/source/js/data-capture/modal-dialogs.js
jseppi: Good point! I was originally doing this without jquery, but then decided to just go with it. I'll make that adjustment.

data_capture/templates/data_capture/step.html
jseppi: Hrm, I don't see anything in the `a11y-dialog` source code that would necessitate having the `type="button"` attribute.

toolness:
So it looks like the NVDA issue with the button activation actually only happens on Chrome--the buttons work fine on Firefox+NVDA, and that's what most people use NVDA with anyways, so I'm going to chalk it up to being a NVDA bug.

jseppi:
Does the NVDA issue happen only in the Style Guide example, or also on the Cancel dialogs in the Price List Upload flow?

jseppi:
> It doesn't read the <h1> in the modal, instead just saying "No, keep working"

It sounds like from #925 we can put the `tabindex` back on the `h1`. Does this fix the issue with NVDA skipping to the button?

toolness:
Oh sorry, I actually haven't even tried the style guide example yet--all the issues I've had are done when using the cancel button on step 2 of price list upload.

Also, it looks like the issue with the dialog title not being announced on NVDA is *also* just a Chrome+NVDA bug. Basically, nobody should ever use NVDA with Chrome. The dialog title gets announced just fine on Firefox.

So in other words this looks great to me, just gotta fix up the class name manipulation stuff!!

jseppi:
Right on. Done!

toolness:
Yass 💯 !  Merge away.

PR #920, opened 2016-10-13
Hyperlink 'try it again' to upload form

### LINKED ISSUE #880: Hyperlink "try uploading it again" to embedded upload form in step 3 errors

toolness:
This fixes #880.


frontend/source/js/data-capture/smooth-scroll.js
```diff
@@ -81,7 +81,7 @@ export function activate(window, options = {}) {
   const scrollMs = options.scrollMs || DEFAULT_SCROLL_MS;
   const onScroll = options.onScroll || (() => {});
 
-  $('body', window.document).on('click', 'a[href^="#"]', e => {
+  $('html', window.document).on('click', 'a[href^="#"]', e => {
```
toolness: This took me a little while to debug.

Basically, in our styleguide, we include the data capture bundle at the bottom of the `<body>`. Binding to `<body>` works for smooth scrolling because the `<body>` element exists at the time of binding.

However, in the data capture pages, we include the bundle in the `<head>`--*before `<body>` has been defined*.  Because of this, `$('body')` evaluates to an empty query and this line is basically a no-op. Causing smooth scrolling to appear broken.

One way to fix this is to wait for `DOMContentReady` before activating ourselves, but an even easier way is to just bind to `<html>` instead of `<body>`.

frontend/source/sass/base/_typography.scss
```diff
@@ -58,6 +58,21 @@ h1, h2, h3, h4, h5, h6 {
   }
 }
 
+:target {
```
hbillings: Good call, @seppi. I think `_utils` makes sense!

frontend/source/js/data-capture/smooth-scroll.js
toolness: Good question, just filed #930 to address this.  Due to the potential side effects of doing that, though, I think we should address it separately from this PR.

toolness:
Ok, the changes should be good to go!

jseppi:
Sweet! 👍

PR #922, opened 2016-10-13
Improve button focus styling

toolness:
Currently, not all of our buttons have great focus styling. Here's me tabbing through all the [buttons in our styleguide](https://calc-dev.apps.cloud.gov/styleguide/#buttons):

> before

So it's very hard/impossible to tell when default, disabled, gray, and link buttons are in focus, which makes interaction particularly challenging for keyboard users.

This PR changes styling so it's easy to tell when any button is focused:

> after


jseppi:
Seems 👍  to me!

PR #929, opened 2016-10-14
Remember last known scroll position and go to it when reloading page

### LINKED ISSUE #921: Browsers w/ manual scroll restoration support don't auto-scroll to last known place on page
So I was developing and noticed that on Chrome and Firefox, when I'm in the middle of a page and reload it, I'm looking at the top of the page, instead of the middle, where the scroll position used to be.  (Normally browsers behave in the latter way.)

I am pretty sure this is because both browsers support the [scroll restoration API](https://developers.google.com/web/updates/2015/09/history-api-scroll-restoration), which we set to `manual` as of #884, and so the browser no longer scrolls to the last known place on the page when the user reloads it.  I guess this means we will have to implement that behavior ourselves now, probably via `window.sessionStorage`.

Fortunately, the browser *does* auto scroll to a named anchor on the page if it's in the URL hash, at least upon the first visit of a page.

toolness:
This fixes #921.

I'm opting to not use jQuery for this new code to make it easier to eventually decouple this module from jQuery and publish it on npm.

To do:

- [x] Add tests.
- [x] Make sure this works okay on browsers in private browsing mode, where accessing `sessionStorage` may throw an exception or something weird.

### DIFF
--- a/frontend/source/js/data-capture/smooth-scroll.js
+++ b/frontend/source/js/data-capture/smooth-scroll.js
@@ -77,6 +77,49 @@ function smoothScroll(window, scrollTop, scrollMs, cb) {
   });
 }
 
+/**
+ * Some modern browsers support the scroll restoration API, which lets
+ * us have full control over how the web page scrolls, eliminating
+ * flicker caused by our code and the browser trying to "fight over"
+ * the current scroll position.
+ *
+ * However, this also means we have to take care of some desirable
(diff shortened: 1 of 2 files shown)


toolness:
Ok this is ready for review @jseppi !

jseppi:
Looks good to me! 👍

toolness:
Hooray!

PR #931, opened 2016-10-17
Use browserify for data-explorer

### LINKED ISSUE #917: Treat common, data-explorer bundles just like data-capture
Currently the data-capture bundle supports ES6 because we're running it through Babel, and we're also doing all kinds of linting on it to ensure the style is awesome.

However, we're not  doing this with any of our legacy code--the common and data-explorer bundles--because, well, it's legacy.

But as we're starting to transition into actually working on those parts of the codebase, it'd be nice to take advantage of all the features of our data-capture toolchain.

So perhaps we should treat the legacy bundles the same as data-capture, but wrap all their files in `/* eslint-disable */` / `/* eslint-enable */` chunks to make the linter chill. Then we can add new modules and stuff that are compliant with our new standards (and possibly even migrate the legacy code to the new standards snippet by snippet).

toolness:
This fixes #917.

Instead of a common bundle and a data-explorer bundle, there are now:

* A **common vendor bundle**, which contains only third-party vendor dependencies for both data-explorer and data-capture (e.g. jquery)
* A **data-explorer vendor bundle**, which contains only third-party vendor dependencies for data-explorer (eventually this should be phased out and moved to `npm`).
* A **data-explorer bundle**, which contains all the first-party code for data-explorer, as well as any third-party dependencies brought in via npm (currently there are none).

Note that **there is no common bundle**; this is because our common first-party code is pretty small right now, so it's easy to just import it as needed in the data-explorer and data-capture bundles, and have browserify just build the common code into both bundles.

This won't be very smart once we actually have large common dependencies that are being brought in via `npm`, however, so we should eventually figure out how to tell browserify to keep common dependencies (whether first-party or third-party via `npm`) in separate bundles and be able to load them from completely different bundles (e.g., so that the data-explorer bundle can import a module from the common bundle). Currently our only workaround for this is to have such common third-party dependencies just be included in the common vendor bundle, and access them as globals in each bundle (this is how both bundles are able to access a common jQuery, for instance).

To do:

- [x] Figure out why `hourglass.js` isn't being linted.


jseppi:
> Figure out why hourglass.js isn't being linted.

@toolness: check `.eslintignore`

toolness:
Ok I think this is good to go @jseppi!

toolness:
Oh, before manually testing this out, you might want to `rm -rf frontend/static/frontend/built` just to make sure that all the legacy bundles are removed.

jseppi:
>  you might want to `rm -rf` ...

Cool. There is also a `gulp clean` task for this :)

jseppi:
👍 Looks great!

Somewhat related, but not really: when I run `docker-compose run app gulp` the process seems to hang and I can't <kbd>ctrl</kbd>+<kbd>c</kbd>. Any idea why that might be?

toolness:
Weird, was running `docker-compose run app gulp` hanging before this PR, or was the hanging introduced by this PR?

I've run into times when `docker-compose run` has hung and Ctrl+C hasn't helped, but I wasn't sure what was going on, alas. Just had to `kill` the docker process from a separate terminal session. 😞

jseppi:
I checked it out in `develop` before making that comment, and it was happening there as well. I just happened to run into it because I did a standalone `docker-compose run app gulp` while reviewing.

toolness:
Oh ok, weird. Hrm. Well I'm going to merge this now but that's a weird bug and worth investigating!

PR #933, opened 2016-10-17
Admin styling

hbillings:
This is a whole mess of styles/templates for the admin interface. I was trying to do something lightweight at first, but it became really evident that Django's lack of admin updates resulted in some cringeworthy code.

There's still optimization to do here, but I think this is a solid base PR. Things that are annoying:
- Lots of `!important` still because I haven't created overrides for all the default spreadsheets.
- The buttons are all Skeleton's default white and floated right, instead of our default blue and floated left.

Index page:
screen shot 2016-10-26 at 11 04 04 am

User page:
screencapture-0-0-0-0-8000-admin-auth-user-1477504481814

Price list page:
screencapture-0-0-0-0-8000-admin-data_capture-submittedpricelist-1477504581296

Price list edit page:
screencapture-0-0-0-0-8000-admin-data_capture-submittedpricelist-2-1477504691077

Price list rows:
screencapture-0-0-0-0-8000-admin-data_capture-submittedpricelistrow-1477504732805


jseppi:
oh my, looks so much nicer!

jseppi:
This certainly looks better!

Are the template and scss files in this PR pulled from Django? If so, I have two concerns:

1. I think each file that is copied (even if heavily modified) should specify that it came from the Django project, and contain a note and link to that project's license (since our code is public domain and we cannot re-license the Django code to be under that same license).
2. If (or when) we update Django there is a pretty good chance that those templates will have problems. I don't really know how to mitigate this other than it being something we need to watch out for when we upgrade.

hbillings:
@jseppi 

1) Great idea! I'll update with comments noting the source for the templates! (The CSS is just named the same thing, but is from-scratch.)

2) Yeah, I'm a little trepidatious about it, but the research I did into it suggested that we *might* be okay/it might not be super painful to switch over. And if it is...there's always Grappelli. 😆

hourglass/templates/admin/change_list.html
```diff
@@ -0,0 +1,90 @@
+{# adapted from https://github.com/django/django/blob/stable/1.8.x/django/contrib/admin/templates/admin/change_list.html #}
```
jseppi: On each of these, could you also add a comment like

```
{# Copyright (c) Django Software Foundation and individual contributors. License: https://github.com/django/django/blob/stable/1.8.x/LICENSE #}
```

I think we need to be explicit that this code is not covered by the CC0 license in our repo.

hbillings:
@jseppi oops, done!

jseppi:
Ha, well, the `{# adapted from ... #}` comments were also good so future us knows where they came from in the Django codebase :) Maybe add those back in also, then I think good to go!

jseppi:
screen shot 2016-10-28 at 11 41 35 am 😆

frontend/source/sass/admin/overrides.scss
```diff
@@ -0,0 +1,234 @@
...
+.mini.quiet {
+  color: $color-gray-medium;
+  font-size: $small-font-size;
+}
+
+/* forms.css */
+.inline-group {
+  border: none !important;
+  .tabular td.original p {
+    display: none; /* this appears to simply append "submitted price list row" to lists */
```
jseppi: Screenshot for posterity:

screen shot 2016-10-31 at 10 22 01 am

jseppi:
Looks 👍 to me

jseppi:
eep, actually just noticed this tiny mis-alignment or something (the tiny blue line under the Site Admin tab in the screenshot below) in the tabs when on the admin site.

screen shot 2016-10-31 at 11 21 34 am

jseppi:
uh oh, just noticed while demoing (or DEMONing 😈) that I couldn't actually click on the checkboxes  in a price list list view (such as at `/admin/data_capture/approvedpricelist/`):

screen shot 2016-10-31 at 11 57 05 am

jseppi:
Also there don't appear to be sort indicators in any of the table views, and there are some weird numbers that show up above the table headings once you click on more than one of them (I guess these are to show the order of columns that are sorted?)

screen shot 2016-10-31 at 11 58 42 am

jseppi:
But you could probably take care of those in separate issues/PRs so that this one can get merged before it gets bigger. The only really "breaking" issue is that the checkboxes can't be clicked.

hbillings:
I wonder if the checkbox bit has to do with the way USWDS styles checkboxes -- I've had to override that in a few places. 😞 

The underline in the nav is weird! Are you zoomed in at all, by any chance?

Which page are you seeing the number headings on?

I'll fix the checkboxes and file separate PRs for the others. Thanks for catching them! 😈

jseppi:
Nope, not zoomed in with the underline.

Those numbers appear on any page with sortable table views (like `/admin/data_capture/approvedpricelist/`) but you have to click at least two of the table headings to make them start showing up.

jseppi:
If I do zoom to 110% the underline goes away though

hbillings:
waaaaattt

jseppi:
Sweet, checkbox width fix works!

Somewhat related, but not breaking anything: the checkboxes on the "Submitted price list row" (`/admin/data_capture/submittedpricelistrow/`) view are a bit wide (note the focus outline):

screen shot 2016-11-01 at 7 27 19 am

But that's not really breaking anything, so feel free not to worry about it.

PR #934, opened 2016-10-17
Fix qunit tests on phantomjs

toolness:
This fixes our QUnit tests on PhantomJS by simply ignoring the parts of our smooth scroll tests that use `scrollTop`.

(They weren't triggering Travis fails because we're now running the QUnit tests against IE11.)

### DIFF
--- a/frontend/source/js/tests/smooth-scroll_tests.js
+++ b/frontend/source/js/tests/smooth-scroll_tests.js
@@ -1,10 +1,12 @@
-/* global $ QUnit document */
+/* global $ QUnit document window */
 
 import {
   IS_SUPPORTED,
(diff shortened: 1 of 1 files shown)


jseppi:
👍 
So based on some searching, it looks like PhantomJS just doesn't have `scrollTop` because it doesn't really have a viewport (?). It does have this [`scrollPosition`](http://phantomjs.org/api/webpage/property/scroll-position.html) property than can be set, but that looks like it would involve messing with `runner.js`, and I'm not even clear on if that would "enable" `scrollTop`. Blerg.

toolness:
Yeah blerrrrg. :(

PR #936, opened 2016-10-18
Fix problems w/ upload widget drag/drop indication

### LINKED ISSUE #919: Upload widget stuck in "drag over" mode if file is dragged off widget
I just noticed that if I drag a file over the upload widget but then change my mind and drag the file off the page, the widget gets stuck in the "Drop file here!" mode:

> 

This happens even if I just drag the file somewhere else on the page and drop it--it doesn't have to be a drag off the whole page.

toolness:
This fixes #919, as well as another bug that hasn't been documented yet: if the user drags and drops a file into the widget, and then drags a file over the widget and then out of the widget, the drag indication wouldn't go away, because we weren't resetting `dragCounter` on the `drop` event.

### DIFF
--- a/frontend/source/js/data-capture/upload.js
+++ b/frontend/source/js/data-capture/upload.js
@@ -155,6 +155,7 @@ class UploadWidget extends window.HTMLElement {
     this.isDegraded = false;
 
     let dragCounter = 0;
+    let lastDragEnterTarget = null;
 
     const finishInitialization = () => {
       if (this.uploadInput instanceof UploadInput) {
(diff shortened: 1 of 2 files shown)


frontend/source/js/data-capture/upload.js
```diff
@@ -215,10 +216,20 @@ class UploadWidget extends window.HTMLElement {
     $el.on('dragenter', e => {
       this._stopAndPrevent(e);
 
-      dragCounter++;
-      $el.addClass('dragged-over');
+      // Firefox spuriously fires multiple `dragenter` events, which
+      // we need to work around by keeping track of the most recent
+      // event target. For more information, see:
+      //
+      //   https://bugzilla.mozilla.org/show_bug.cgi?id=804036
```
jseppi: Just tell users to enable `e10s`! https://bugzilla.mozilla.org/show_bug.cgi?id=804036#c21

Whatever that is...

jseppi:
Sweet, 👍

frontend/source/js/data-capture/upload.js
toolness: Haha, e10s is the long-running project to make Firefox use a separate OS process per tab, like Chrome does, so that if one web page crashes it doesn't bring down the whole system. I think it also improves performance on multi-core systems but I'm not sure.

PR #937, opened 2016-10-18
Expand introduction to styleguide

### LINKED ISSUE #923: Add a "philosophy" section to the style guide
So it's not obvious at a glance, but we've actually been building our part of CALC (data capture) with Aaron Gustafson's [three maxims for progressive enhancement with JS](http://adaptivewebdesign.info/1st-edition/read/chapter-4.html#planning-and-restraint):

> 1. Make sure all content is accessible and all necessary tasks can be completed without JavaScript turned on.
> 2. Use JavaScript to generate any additional markup it needs.
> 3. Apply no style before its time.

If this seems like a good approach, maybe we should codify it in an introductory section of the styleguide?

toolness:
This fixes #923 by expanding the introduction of the style guide to include a link to the 18F Content Guide and information about our progressive enhancement strategy:

> screen shot 2016-10-18 at 10 02 36 am

### DIFF
--- a/styleguide/templates/styleguide.html
+++ b/styleguide/templates/styleguide.html
@@ -16,11 +16,38 @@
   <h1>CALC Style Guide</h1>
 
   {% guide %}
+
+  {% guide_section "Introduction" %}
+
+  <p>
+    Welcome to the CALC Style Guide!
+  </p>
+
+  <p>
+    CALC's content is structured using 18F's standard voice, which is
+    described in the <a href="https://pages.18f.gov/content-guide/">18F
+    Content Guide</a>.
+  </p>
+
   <p>
     CALC uses <a href="http://getskeleton.com/">Skeleton</a> as the foundation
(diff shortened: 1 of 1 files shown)


styleguide/templates/styleguide.html
```diff
@@ -16,12 +16,39 @@
...
+  {% guide_section "Introduction" %}
+
+  <p>
+    Welcome to the CALC Style Guide!
+  </p>
+
+  <p>
+    CALC's content is structured using 18F's standard voice, which is
+    described in the <a href="https://pages.18f.gov/content-guide/">18F
+    Content Guide</a>.
```
toolness: Note that we determined we were using 18F's standard voice after consultation on our Writing Lab issues, e.g. https://github.com/18F/writing-lab/issues/197.

jseppi:
Nice! 👍

PR #941, opened 2016-10-18
raise exception instead of returning it

jseppi:
Very minor correction: `raise NotImplementedError()` instead of `return NotImplementedError()`


toolness:
Nice catch, thanks!

Issue #940, opened 2016-10-18
Consider using parallel tracking for Google analytics instead of two script tags


toolness:
I noticed that DAP supports something called [Parallel Tracking (PUA)](https://www.digitalgov.gov/event/dap-learning-series-parallel-tracking-pua-benefits-implementation/), which might make it possible for us to send the same Google Analytics events to both DAP and 18F GA.

Currently, I believe these are completely separate accounts that receive different kinds of events: for example, DAP tracks download and offsite link clicks, while our custom 18F GA is set up to treat ajaxform form refreshes as page visits. This could make research difficult by siloing data in different places.

Just asked about this on [#g-analytics](https://gsa-tts.slack.com/archives/g-analytics/p1476804286000142).

toolness:
So I did a bit of digging and it looks like parallel tracking is implemented in the [DAP script](https://github.com/digital-analytics-program/gov-wide-code/blob/master/Universal-Federated-Analytics.js) by simply creating multiple trackers named <code>GSA_ENOR<em>n</em></code>, where <code><em>n</em></code> is an integer.  The DAP tracker is always `GSA_ENOR0`, while specifying additional GA tracking ids via the `pua` querystring argument in the DAP script tag registers additional ones that point to the respective ids. All GA operations are then multiplexed to all the trackers on the client-side by the DAP script.

It seems like it makes sense to use PUA for some things, but not others: for instance, changing the current page's querystring arguments when `pushState`-based transitions occur like #944 does would make sense to multiplex, but multiplexing the setting of custom dimensions wouldn't, since DAP and our 18F GA installation use completely different custom dimensions.

Ultimately it might just make sense to multiplex on a case-by-case basis, concentrating all such logic in `common/ga`. Making it so DAP and 18F GA have the same conception of what a page view is, for instance, seems like a useful goal to avoid confusion.

toolness:
Unifying page views, at the very least, would be really helpful. Right now things are quite confusing because the TTS GA account claims we had 111,018 pageviews with non-empty labor category search queries on the CALC homepage, while the DAP GA account thinks we had 1,854.  That's because the former is counting `pushState`-based transitions, while the latter isn't.

Aside from that, it's also confusing to have to look in separate places for custom events: right now events like clicking on the "I don't know what to upload" expander are reported to TTS GA, while external download clicks are reported to DAP GA.  If PUA helps unify these, it'd make things way less confusing.

toolness:
[...]

PR #944, opened 2016-10-18
Set GA location, send pageview on search

### LINKED ISSUE #924: Google analytics isn't tracking searches done in CALC
So after installing the handy [Google Analytics Debugger extension](https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en) that @jseppi recommended, I discovered that the single-page data explorer app isn't reporting searches to GA.  It's changing the page URL via `pushState`, which is great, but by default GA doesn't track those changes.

We can configure this using GA's guide on [Single Page Application Tracking](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications).

However, it should also be noted that since the front-end is using the CALC API, which is served via api.data.gov, api.data.gov *should* be recording metrics on its usage.  So it's also likely that we can access the information we need via its analytics. I'm [asking how to do that on #g-analytics](https://gsa-tts.slack.com/archives/g-analytics/p1476373395000137) now.

toolness:
This is an attempt to fix #924 by setting the [GA `location` field](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#location) and issuing a `pageview`.

## Notes

* The **GA Single Page Application Tracking guide** specifically advises us to [not update the document location](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#do_not_update_the_document_location), but this seems weird to me because that's all we're changing, and we really *want* those new querystring arguments to be tracked by GA. And since we're not using GA's campaign fields (e.g. `utm_source`, `utm_campaign`, etc) I don't *think* it should break its session tracking...

* As with the ajaxform improvements in #869, the way this is currently written, it will *only* report the pageview to 18F GA's account, *not* to DAP.  We could report it to both by fixing #940.

## To do

- [x] Import `../common/ga` instead of using the global.

### DIFF
--- a/frontend/source/js/data-explorer/explorer.js
+++ b/frontend/source/js/data-explorer/explorer.js
@@ -1,3 +1,5 @@
+import ga from '../common/ga';
+
 const hourglass = require('../common/hourglass');
 
 /* eslint-disable */
@@ -275,7 +277,7 @@ const hourglass = require('../common/hourglass');
 
       history.pushState(null, null, href);
 
-      if (didSearchChange && typeof(ga) === 'function') {
+      if (didSearchChange) {
         ga('set', 'location', window.location.href);
         ga('send', 'pageview');
       }
--- a/frontend/source/js/data-explorer/explorer.js
+++ b/frontend/source/js/data-explorer/explorer.js
@@ -278,7 +278,8 @@ const hourglass = require('../common/hourglass');
       history.pushState(null, null, href);
 
       if (didSearchChange) {
-        ga('set', 'location', window.location.href);
+        ga('set', 'page', window.location.pathname +
+                          window.location.search);
         ga('send', 'pageview');
       }
     }
--- a/frontend/source/js/data-explorer/explorer.js
+++ b/frontend/source/js/data-explorer/explorer.js
@@ -271,7 +271,14 @@ const hourglass = require('../common/hourglass');
 
     if (pushState) {
       var href = "?" + hourglass.qs.format(data);
+      var didSearchChange = window.location.search !== href;
+
       history.pushState(null, null, href);
+
+      if (didSearchChange && typeof(ga) === 'function') {
+        ga('set', 'location', window.location.href);
+        ga('send', 'pageview');
+      }
     }
 
     updateExcluded();


frontend/source/js/data-explorer/explorer.js
```diff
@@ -271,7 +271,14 @@ const hourglass = require('../common/hourglass');
...
     if (pushState) {
       var href = "?" + hourglass.qs.format(data);
+      var didSearchChange = window.location.search !== href;
+
       history.pushState(null, null, href);
+
+      if (didSearchChange && typeof(ga) === 'function') {
+        ga('set', 'location', window.location.href);
+        ga('send', 'pageview');
+      }
```
toolness: We're using this `didSearchChange` variable because this code appears to be executed multiple times per search, with identical arguments.  I didn't want to modify the legacy behavior of the app, but I also didn't want to report multiple hits on GA, so I decided to only report the hit when the page's search terms actually changed.

frontend/source/js/data-explorer/explorer.js
```diff
@@ -271,7 +271,14 @@ const hourglass = require('../common/hourglass');
 
     if (pushState) {
       var href = "?" + hourglass.qs.format(data);
+      var didSearchChange = window.location.search !== href;
+
       history.pushState(null, null, href);
+
+      if (didSearchChange && typeof(ga) === 'function') {
```
toolness: Oops, I should really be importing `ga` from `../common/ga`, in which case I don't need to check to see if `ga` exists. Forgot that #931 has been merged!

jseppi:
Hrm, the GA page you [linked](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#do_not_update_the_document_location)  also says updating `location` will break referral tracking, which might be something we don't want to break. I at least think it is pretty interesting knowing how much traffic comes from various channels.

I wonder if we could instead issue an `event` when a new search is done?

toolness:
Hmm, so I found this [autotrack](https://github.com/googleanalytics/autotrack) library that Google Analytics made, which includes a [`urlChangeTracker`](https://github.com/googleanalytics/autotrack/blob/master/docs/plugins/url-change-tracker.md) plugin for automatically informing GA about `pushState`-based location changes.

Based on its source code at [`url-change-tracker.js`](https://github.com/googleanalytics/autotrack/blob/master/lib/plugins/url-change-tracker.js), it looks like it's actually setting the GA `page` field to the concatenation of the current page's path *and* its query string. So even though all the examples on GA's documentation set `page` to just a path name, it looks like it's actually possible to provide query string arguments there too?

So confusing.

toolness:
Oh, using the analytics debugger chrome extension, it does look like the `page` property reported to GA by default *does* include the query string!

> ga_page

So that's cool. I guess I will just update this PR to send the `page` property instead of `location` and we should be set.

toolness:
Ok, this should be good to go now @jseppi.

jseppi:
nice! 👍

PR #945, opened 2016-10-19
Use browserify cache

toolness:
In #597 we started logging JS bundle writes because they were taking so long, but I poked around a bit and found out that we're not enabling browserify's cache when using watchify. The [watchify documentation](https://github.com/substack/watchify#watchifyb-opts) mentions that we "MUST set the `cache` and `packageCache` properties" but we're not doing that.

Adding these properties improves the bundling performance on my machine: making minor edits to `ajaxform.js` used to take 3-5 seconds; it now takes under 2 seconds.

I'm wondering if telling all the different browserify bundles to use the *same* cache would speed things up even more, but I'm not sure. Regardless, this seems like a pretty significant improvement for now!


toolness:
I really wish these `cache` and `packageCache` options to `browserify` were actually documented somewhere. 😞

jseppi:
Nice, good find! 👍

Issue #938, opened 2016-10-18
Should admins have a way to "reject" price lists?


jseppi:
Right now we have the boolean `is_approved` fields on our `SubmittedPriceList` model, but I think we actually need a field with more options instead so that we can capture when a price list has been reviewed, but is _rejected_ instead of merely just not approved.

Perhaps a better solution is to remove `is_approved` and replace it with an enum field (maybe something like `review_status`), which we can then use to store various statuses ("new", "approved", "rejected", "replaced", etc.). 

Thoughts, @toolness?

jseppi:
Oh also I think this new field, or even if we don't have a new field, should have a couple related fields: `status_changed_by` (FK to `User`), and `status_changed_at` (`DateTime`).

PR #927, opened 2016-10-14
User Account and Submitted Price List Views

jseppi:
This PR adds an account page (at `/account`) that lists the user's recent activity and a short summary of all their price list submission activity.

It also adds a price list listing page (`/data-capture/price-lists`) that lists all of a user's submitted price lists, separated into lists by status, and a price list details page (`/data-capture/price-lists/<ContractNumber>`) that shows all the details of a submitted price list.

To Do:
- [x] test new views
- [x] implement suggestions from @hbillings (https://github.com/18F/calc/pull/927#issuecomment-256960779)
- [x] use `<dl>` as in #961 (merge in `develop` first)
- [x] move inline styling to sass file
- [x] figure out what to do on Details page if there are multiple contracts with the same contract number -- update: currently just showing the price list with the most recent `created_at` for the given `contract_number`. I think we need a larger plan/discussion over handling duplicated contracts. 

Screenshots - I'll try to keep these somewhat up-to-date:

User Profile Page:
screen shot 2016-10-28 at 2 47 56 pm

<hr>

Price List Listing page:
screen shot 2016-10-28 at 2 47 37 pm

<hr>

Price List Details page:
screen shot 2016-10-28 at 2 47 28 pm

PR #951, opened 2016-10-19
Refactor SubmittedPriceList "is_approved" to new "status" field

jseppi:
This PR removes `is_approved` from `SubmittedPriceList`, replacing it with a `status` field so that we have finer granularity in describing the status of a `SubmittedPriceList`. Two additional fields, `status_changed_at` and `status_changed_by` are also added.

I tried to keep functionality pretty much the same in this PR, so there are only three available statuses: 
- `STATUS_NEW`: Only for new `SubmittedPriceLists`, as made during the Price List Upload flow
- `STATUS_APPROVED` - For approved `SubmittedPriceLists`
- `STATUS_UNAPPROVED` - For `SubmittedPriceLists` that were previously approved, but now are not

If this PR is merged, I think we should implement (relatively soon) another status, `STATUS_REJECTED` for `SubmittedPriceLists` that have been reviewed but that the reviewer wants to explicitly reject.

I could also see a future `STATUS_REPLACED` status for price lists that have been superceded by a new price list for the same contract (and probably an associated `replaced_by` FK field).

Here's how the admin page looks after this PR (with `status` instead of `is_approved`):
screen shot 2016-10-20 at 10 04 47 am

Closes https://github.com/18F/calc/issues/938


jseppi:
Before I get in too deep, what do you think about this (and #938), @toolness?

jseppi:
I think this will be beneficial for better reporting Submitted Price List status both to COs and to Data Administrators/the CALC Admin.

Ref #927 also.

data_capture/tests/test_price_list_views.py
```diff
@@ -385,6 +386,9 @@ def test_valid_post_creates_models(self):
         self.assertEqual(p.contractor_site, 'Customer')
         self.assertEqual(p.is_small_business, False)
         self.assertEqual(p.submitter, user)
+        self.assertEqual(p.status_changed_by, user)
+        self.assertEqual(p.status, SubmittedPriceList.STATUS_NEW)
+        self.assertEqual(p.status_changed_at.date(), datetime.now().date())
```
jseppi: This test is slightly inaccurate since it isn't testing the time, just the date, but it should be good enough I think.

The `freezegun` package looks like a pretty sweet way of testing time if we want to be better about it: https://github.com/spulec/freezegun

data_capture/views/price_list_upload.py
```diff
@@ -259,6 +262,10 @@ def step_4(request, step):
         # fell back to a schedule other than the one the user chose.
         price_list.schedule = registry.get_classname(gleaned_data)
 
+        price_list.status = SubmittedPriceList.STATUS_NEW
```
jseppi: `STATUS_NEW` is the default, but I figured it would be good to be explicit here

data_capture/admin.py
```diff
@@ -162,17 +163,17 @@ def has_add_permission(self, request):
...
-        if obj and obj.is_approved:
+        if obj and obj.status is SubmittedPriceList.STATUS_APPROVED:
             return self.fields
         return self.readonly_fields
 
 
 @transaction.atomic
 def approve(modeladmin, request, queryset):
-    unapproved = queryset.filter(is_approved=False)
+    unapproved = queryset.filter(~Q(status=SubmittedPriceList.STATUS_APPROVED))
```
jseppi: I _think_ this is the only way to do "not equal" filters, which seems weird to me since there are special ways of doing greater-than, less-than, etc (like with `.filter(myfield__gt=5)`).

data_capture/email.py
```diff
@@ -39,8 +41,9 @@ def price_list_unapproved(price_list):
     ctx = {
         'price_list': price_list
     }
-    if price_list.is_approved:
-        raise AssertionError('price_list.is_approved must be False')
+    # TODO: Change to STATUS_REJECTED ?
```
jseppi: I think this should be left to a future PR. I can remove the `TODO` comments if that is best.

data_capture/models.py
```diff
@@ -45,7 +57,11 @@ class SubmittedPriceList(models.Model):
     )
 
     submitter = models.ForeignKey(User)
-    is_approved = models.BooleanField(default=False)
+
+    status = models.IntegerField(choices=STATUS_CHOICES, default=STATUS_NEW)
+    status_changed_by = models.ForeignKey(User, related_name='+', null=True)
```
jseppi: I used `related_name='+'` to prevent back-linking from users to SubmittedPriceLists that they have changed the status of, but I'm not sure if this is best. Happy to hear thoughts on this decision.

data_capture/periodic_jobs.py
```diff
@@ -7,19 +7,15 @@
...
-    # we need to store an additional flag on that model to indicate
-    # if it has ever been manually unapproved so that we don't bug
-    # the admin with things that have been unapproved
-    logger.info('Processing periodic reminder about unapproved Price Lists')
+    logger.info('Processing periodic reminder about unreviewed Price Lists')
 
-    count_not_approved = SubmittedPriceList.objects.filter(
-        is_approved=False).count()
+    count_new = SubmittedPriceList.objects.filter(
+        status=SubmittedPriceList.STATUS_NEW).count()
```
jseppi: Now the reminder email will only be sent when there get to be a number of `STATUS_NEW` price lists

jseppi:
Ready for review, @toolness!

jseppi:
Hrm, is this a problem, @toolness?

```
RuntimeWarning: DateTimeField SubmittedPriceList.status_changed_at received a 
naive datetime (2016-10-20 20:45:49.821408) while time zone support is active.
```

data_capture/admin.py
toolness: Oh weird! Maybe it'd be useful to ask on `#django` in Slack?  It's ok if it's not possible I guess, ~Q doesn't look so bad either I suppose.

jseppi: Here's an issue about it, marked as `wontfix`: https://code.djangoproject.com/ticket/5763

toolness:
Oh man, I wonder if there's some way we could hook into [`django.contrib.admin.models.LogEntry`](http://stackoverflow.com/a/988202) to log the status changes to the model or something. Just brainstorming here, not part of the review!

data_capture/admin.py
```diff
@@ -245,28 +245,32 @@ class SubmittedPriceListAdmin(UndeletableModelAdmin):
...
+        'status',
     )
 
     inlines = [
         SubmittedPriceListRowInline
     ]
 
     def current_status(self, instance):
-        if instance.is_approved:
+        if instance.status is SubmittedPriceList.STATUS_APPROVED:
```
toolness: I wonder if it'd actually make sense to keep `is_approved` as a [`@property`](https://docs.python.org/3/library/functions.html#property) lookup to do the check you're doing here, to keep things nice and concise? Then again, I suppose from another viewpoint, it could be more obfuscating... Up to you really!

data_capture/models.py
```diff
@@ -112,8 +131,10 @@ def approve(self):
 
         self.save()
 
-    def unapprove(self):
-        self.is_approved = False
+    def unapprove(self, unapproval_user):
+        self.status = self.STATUS_UNAPPROVED
+        self.status_changed_at = datetime.now()
+        self.status_changed_by = unapproval_user
```
toolness: I get the impression that it might eventually be useful to have a `_change_status(new_status, user)` method but since the code is only repeated twice right now s'all gud.

data_capture/models.py
```diff
@@ -75,8 +91,11 @@ def get_business_size_string(self):
             return 'S'
         return 'O'
 
-    def approve(self):
-        self.is_approved = True
+    def approve(self, approval_user):
+        self.status = self.STATUS_APPROVED
+        self.status_changed_at = datetime.now()
```
jseppi: Should I use `timezone.now()` from `django.utils` here instead of `datetime.now()`, @toolness?

data_capture/models.py
jseppi: Good idea!

data_capture/models.py
toolness: That seems reasonable. We can always change it if we decide that for some reason we want the backwards relation, amirite?!

jseppi:
What's the verdict, pardner?

jseppi:
ping @toolness

toolness:
Ack sorry! The code looks good, gonna test things out on my local system now...

toolness:
Noice, this looks good to me. I noticed that the **Submitted price list rows** admin view is showing both `NEW` and `UNAPPROVED` price list rows, which is fine and preserves legacy behavior... and I think that you are addressing making it easier to only look at `NEW` rows in a separate PR.

Feel free to merge!

PR #953, opened 2016-10-20
Add ability to Reject price lists

jseppi:
~~Currently open against `price-list-status` (PR https://github.com/18F/calc/pull/951).~~

This PR adds the ability for an administrative user to Reject price lists, as a distinct action from Unapproving a price list.

This is basically done, but I think it would actually make sense to do some more work on the admin page (see PR https://github.com/18F/calc/pull/954).


jseppi:
ping @toolness

toolness:
Hey, I'm gonna take a look at the code here soon, but at the requirements level, could we talk this over w/ Kelly to make sure this is functionality he'd find useful?

jseppi:
Certainly! He has seemed please with how I've explained it a few of our stand-ups, but it would be good to show him.

data_capture/templates/data_capture/email/price_list_rejected.txt
```diff
@@ -0,0 +1,8 @@
+
+The price list you submitted to CALC on {{ price_list.created_at }} has
+unfortunately been rejected.
+
+The submitted price list was for contract {{ price_list.contract_number }}
+({{ price_list.get_schedule_title }}) with {{ price_list.vendor_name }}.
+
+If you have any questions, please contact the CALC administrator at calc@gsa.gov.
```
toolness: I wonder if we should actually insert the e-mail address of the data administrator who made the rejection here, so that the CO can easily communicate with them?

Also, should the data administrator be able to specify a *reason* for the rejection?  This could be useful both in the short-term for the CO, as well as in the long-term, if someone finds an old submitted price list and wants to know why it was rejected... I guess these are Kelly questions.

jseppi: Yeah, I think the actual person who rejected it should be mentioned. Though it is somewhat moot for the medium-term since email functionality won't be available for a while. It would probably be good to mention who initiated the status change in all the email notifications (approve, unapprove, and reject).

I also thought about adding a rejection reason but wanted to keep this a bit contained.

Open issues for both of these?

toolness:
Is there a way to undo a rejection?  Like if a data administrator accidentally rejected the wrong price list, or meant to approve it but accidentally rejected it instead?

jseppi:
In this PR, they could change from `STATUS_REJECTED` to any other status.

#954 changes that behavior so a price list with `STATUS_REJECTED` can only be changed to `STATUS_APPROVED`.

toolness:
O cool. This looks good to me!

jseppi:
Great, thanks! Will do a quick demo at stand today and merge if Kelly gives it the 👍

jseppi:
Kelly said 👍

PR #954, opened 2016-10-21
Price list admin status segmentation

jseppi:
Open against `price-list-reject` branch.

This PR implements multipl admin "views" of the `SubmittedPriceList` model based on `status`. This is done with [Proxy Models](https://docs.djangoproject.com/en/1.10/topics/db/models/#proxy-models) (see code in this PR's `admin.py`). Here's a screenshot of what it looks like on the first page of the admin site:

screen shot 2016-10-25 at 8 35 43 am

The available actions on a set of price lists with a given `status` is limited to only the subset that makes sense based on that `status`. Here is what I came up with.

| Current Status      | Available Actions |
|---------------------|-------------------|
| `STATUS_NEW`        | approve, reject   |
| `STATUS_APPROVED`   | unapprove         |
| `STATUS_UNAPPROVED` | approve           |
| `STATUS_REJECTED`   | approve           |

I'd love some thoughts/discussion around this, cc @tram @hbillings @toolness.


jseppi:
It would be cool if the Admin page showed a count of the Price Lists for each status. Like 

```
Approved price lists (100)     change
New price lists (5)            change
```

@toolness: any idea how that could be done?

jseppi:
@toolness: This is ready for review now

data_capture/templates/admin/data_capture/submittedpricelistrow/change_list.html
```diff
@@ -15,9 +14,9 @@
   changes.
 </p>
 <p>
-  Once you've muted any necessary rows, you can visit the
-  <a href="{% url 'admin:data_capture_submittedpricelist_changelist' %}">Submitted price list</a> admin page to approve the data
-  for inclusion into CALC.
+  Once you've muted any necessary rows, you can visit the corresponding
+  <a href="{% url 'admin:app_list' app_label='data_capture' %}">price list admin page</a>
```
jseppi: I couldn't link back just to the `submittedpricelist_changelist` page because that page no longer exists. This `SubmittedPriceListRow` `change_list` view currently lists all `SubmittedPriceListRow`s that are not approved (so it shows new, unapproved, and rejected rows).

Because the list of `SubmittedPriceListRows` could contain a mix of different `SubmittedPriceList` status, I just opted to link back to the generic `data_capture` admin page. Happy to hear/implement better suggestions for this though.

data_capture/admin.py
```diff
@@ -321,6 +318,62 @@ def get_readonly_fields(self, request, obj=None):
         return self.readonly_fields
 
 
+class ApprovedPriceList(SubmittedPriceList):
+    class Meta:
+        proxy = True
```
toolness: Cool! I did not know about [proxy models](https://docs.djangoproject.com/en/1.10/topics/db/models/#proxy-models), or at least about using them to provide multiple admin list views for the same Django model.

jseppi: Neither did I -- was lead to them by a fortunate google search 🍀

data_capture/admin.py
```diff
@@ -339,8 +392,11 @@ class SubmittedPriceListRowAdmin(UndeletableModelAdmin):
     )
 
     def contract_number(self, obj):
-        url = reverse('admin:data_capture_submittedpricelist_change',
-                      args=(obj.price_list.id,))
+        # get status label and derive url from it
+        status_display = obj.price_list.get_status_display()
```
toolness: Er where is this `get_status_display()` function being defined?  I feel like this should be really obvious to me, but for some reason it isn't. I might need more coffee.

toolness: ah nice, thanks!

toolness: Hmm, so it's good that we have a unit test that iterates through all the possible values of this, then. Otherwise I'd be a bit wary of linking human-readable names to view names!

PR #961, opened 2016-10-25
Use <dl> more properly on Step 4

### LINKED ISSUE #939: Incorrect usage of <dl> element in Step 4?
The current markup of the description in Step 4 of the Price List Upload flow looks like this:

```html
<dl>
  <li>Term: <dd>value</dd></li>
  ...
</dl>
```

I was just reading over the MDN page for description lists (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) and it looks like the items should instead be formed using sibling `<dt>` (definition term) and `<dd>` elements, like so:

```html
<dl>
  <dt>Term</dt>
  <dd>value<dd>
  ...
</dl>
```

and probably some accompanying css to add a separator after the `<dt>`s:

```css
dt:after {
  content: ": ";
}
```

Thoughts, @hbillings?

jseppi:
Attempt to use `<dl>`, `<dt>`, and `<dd>` elements more correctly in the `.contract-details` list on Step 4.

Here is what it looks like (the same as before!):
screen shot 2016-10-28 at 9 17 18 am

And here it is with the note when a different schedule was auto-detected:
screen shot 2016-10-28 at 9 01 34 am

Closes #939


frontend/source/sass/components/_forms.scss
```diff
@@ -113,12 +113,27 @@ dl.contract-details {
...
   }
+
   dd {
     font-weight: $font-weight-bold;
     margin-left: 5px;
-    display: inline;
+
+    &.contract-details__start-date:after {
+      font-weight: normal;
+      content: ' to ';
```
jseppi: @hbillings: Is this weird? I couldn't just put it as a child of the `<dl>`. It could _maybe_ go inside of a `<span>` with the start date `<dd>`, but I don't know if that is any better than this.

hbillings: Ooh, I like that a lot, actually. I might steal it for the form widget on Step 2 to avoid the extraneous markup there.

hbillings:
@jseppi Hmm... does it make sense, perhaps, to simply display the alert outside of the `dl`? The schedule will always be the last thing on the list, so visually there wouldn't be much of a difference from its current iteration.

jseppi:
@hbillings: works for me! (I also don't think that is going to come up often, or ever even)

jseppi:
Although, I kinda feel like the schedule should be closer to the top of the list, like after Contract Number, because it feels "bigger" than the rest of the details.

jseppi:
Ok done. I also set `margin-top: 0; margin-bottom: 0;` on `dl.contract-details` because it looked too spread out to me, especially when combined with the wrong schedule note being in a separate grid row. Screenshot in root comment has been updated.

jseppi:
Good to go, @hbillings?

PR #796, opened 2016-09-23
Add visual polish

hbillings:
This is two things:
a) some little things that have been bothering me, like some labels being bold and others not, border radius not being consistent, different styles being used for the same heading sizes on the same page.
b) a crack at better error styling (closes #779). 
screen shot 2016-09-23 at 7 26 23 am
~*~*~*~*~ and ~*~*~*~*~
screen shot 2016-09-23 at 7 39 07 am

UPDATE: New headers look like this now:
screen shot 2016-09-23 at 12 15 27 pm
screen shot 2016-09-23 at 12 15 15 pm

And the progress bar:
screen shot 2016-09-26 at 12 43 04 pm

To do:
- [x] Update styleguide headings
- [x] Remove unnecessary padding on CALC header
- [x] Make "help us improve" link the right color
- [x] Shrink progress bar
- [ ] ~~Style @jseppi's new separate error step once merged~~ Will do as a separate PR

PR #964, opened 2016-10-26
Fix minor typo ~~Remove focus outline from <alerts-widget>~~

### LINKED ISSUE #932: Remove focus outline from <alerts-widget>
As noted in #925, we don't need focus outlines on non-interactive elements that are outside the tab ordering, so we should remove it from the `<alerts-widget>` introduced in #635.

jseppi:
Well, I was going to do this real quick, but I am unable to get the `<alerts-widget>` example in the Style Guide (http://app.calc.docker/styleguide/ajaxform) to actually show the alerts widget. Any ideas, @toolness?

And then I noticed that it appears there is already a rule to remove the focus from `alert-widget`: https://github.com/18F/calc/blob/develop/frontend/source/sass/components/_alerts.scss#L40 (added by @hbillings in 82210e35ad247ef28cd52b7e308badee51744ff9).

So, I just corrected a minor typo I found.

Closes #932

### DIFF
--- a/styleguide/ajaxform_example.py
+++ b/styleguide/ajaxform_example.py
@@ -61,7 +61,7 @@ def view(request):
                     time.sleep(3)
                     return ajaxform.redirect(request, 'styleguide:index')
                 elif choice == CHOICE_500:
-                    raise Exception('Here is the 500 your ordered.')
+                    raise Exception('Here is the 500 you ordered.')
                 else:
                     return HttpResponse('Here is an unexpected response.')
 


toolness:
Oh, so I don't *think* `<alerts-widget>` is currently in the style guide--it's actually kind of hard to include, at least as static content that's visible on page load, because the very first thing it does it push browser focus to the widget, which we don't want to happen when someone visits the style guide.  I think it might be better to have a button on the style guide that dynamically adds an `<alerts-widget>` so it's easy to observe and manually test the functionality, though--what do you think?

jseppi:
Yeah, that would be annoying to focus on page load. There is text on the style guide page (https://github.com/18F/calc/blob/develop/styleguide/templates/styleguide.html#L130) that indicates an `alert-widget` will be shown when the sample ajax form is submitted with an error, but I couldn't get it to trigger.

toolness:
Oh I forgot about that part!  Hmm, I wonder if it's broken?  I tried submitting "5" as the answer to 2+2, and once I also uploaded a CSV file, it sent me back an error form with an `<alerts-widget>` in it, but it triggered a JS error (which activated the "safe mode" opt-in alert) of `Uncaught TypeError: Illegal invocation`.  Hmmm.

toolness:
So I noticed that the `document.activeElement` once I submit an ajaxform with errors in it *is* the
`<alerts-widget>` but there's no focus outline around it.  Was this bug already fixed or something?

So confused...​

jseppi:
Oh, it is showing the alerts? I must have been looking at the wrong thing. I was expecting an `.alert.alert-error` div.

And yes, as I mentioned in the original comment, it looks like @hbillings already had removed the focus outline a while back.

toolness:
Oh, yeah, it's not showing an `.alert.alert-error` div for some reason, but it is wrapping the alert content in an `<alerts-widget>` at least.  I'm not sure why it's not showing that div, I thought it used to...

jseppi:
That was an intentional change in https://github.com/18F/calc/pull/796. I guess we just forgot about it :P

jseppi:
I assume this extremely small typo fix is good to go? 😆

toolness:
Oh yeah! Except on my computer at least, the other parts of the visual polish in #796 aren't present, like the pink background around invalid fields. I guess that comes from putting them in `<fieldset>`s, which that ajaxform isn't doing. Imma file a separate issue/PR to address that.

This extremely small typo fix looks good :)

PR #969, opened 2016-10-26
Use {% fieldset %} in ajaxform example

### LINKED ISSUE #968: Ajaxform in style guide doesn't use fieldsets
The new error styling introduced in #796 makes the ajaxform look particularly weird when errors occur:

> 

I think this is because this style guide example isn't using `<fieldset>`s but I'm not sure. Either way, our style guide examples should conform to our style conventions.

toolness:
This fixes #968 so that the example ajaxform uses fieldsets and errors look better:

>

### DIFF
--- a/styleguide/templates/styleguide_ajaxform_example.html
+++ b/styleguide/templates/styleguide_ajaxform_example.html
@@ -7,6 +7,7 @@
 
       action="{% url 'styleguide:ajaxform' %}">
   {% csrf_token %}
+  {% load frontend %}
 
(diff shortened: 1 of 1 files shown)


jseppi:
👍 
The error on the file input still looks a bit weird, but maybe we never solved that styling.

toolness:
Yeah, I don't think we ever solved that one. I originally tried putting the file input in a `{% fieldset %}` too, but it made the upload widget as narrow as the text fields, which looked weird, so I decided not to pursue that.

PR #972, opened 2016-10-26
Make AlertsWidget extend HTMLElement

### LINKED ISSUE #967: Clicking on `<alerts-widget>` inside styleguide ajaxform example throws JS error
As mentioned in https://github.com/18F/calc/pull/964#issuecomment-256369123:

1. Visit [`/styleguide/ajaxform`](https://calc-dev.apps.cloud.gov/styleguide/ajaxform).
2. Put "5" in as the answer to "What is 2 + 2?" and drop any file with a `.csv` extension in to the file upload widget and submit the form.
3. Click the words "Uh oh" and open the JS console.

The JS console will contain the following:

No idea what's going on here and the traceback isn't very useful, particularly since our jQuery is minified. I wonder if we should keep the un-minified jQuery in dev builds to make debugging things like this easier.

Also, the above screenshot shows a `nrWrapper` in `ajaxform:8` but that doesn't display on my local dev environment.

toolness:
This fixes #967. For an explanation, see https://github.com/18F/calc/issues/967#issuecomment-256464907.

### DIFF
--- a/frontend/source/js/data-capture/alerts.js
+++ b/frontend/source/js/data-capture/alerts.js
@@ -11,7 +11,7 @@ import { dispatchBubbly } from './custom-event';
  * readers to announce the content of the widget immediately.
  */
 
-class AlertsWidget extends window.HTMLInputElement {
+class AlertsWidget extends window.HTMLElement {
   attachedCallback() {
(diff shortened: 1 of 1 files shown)


jseppi:
👍

PR #983, opened 2016-10-31
[WIP] Random vendor name generator

hbillings:
Output a variety of random vendor names. Currently not hooked up to anything.
TBH, I wasn't quite sure which direction to take this in.

I think I know how to do something like an autofill that displays a button on the first page of the flow that would generate fake data for the session (by using `localStorage`, maybe?) and apply it on a per-page basis before saving it to the model at the end.

I'm not sure how to implement something like the `manage.py seed` command [floated here](https://trello.com/c/OnU8RGHR/161-as-a-developer-i-want-to-autopopulate-fields-in-the-data-capture-flow-so-that-i-can-save-time-and-generate-fake-data-for-testing). I could probably figure out how to generate a JSON file with dummy data, which might be useful for a fixture? But I'm not entirely sure.

Also, I'm aware that this is very much low priority, and that there are UI tasks that require my attention more immediately. So I'm okay if the answer is "you should stop working on this now." ^_^

### DIFF
--- a/frontend/source/js/data-capture/random-data-generator.js
+++ b/frontend/source/js/data-capture/random-data-generator.js
@@ -1,0 +1,35 @@
+
+function randomize(item){
+  const max = item.length;
+  // Get random index of passed array
+  return item[Math.round(Math.random() * (max + 1))];
+}
+
+function generateVendorName(){
+  const vendorNames = ["Bailey", "Battaglia", "Brouilette", "Seppi", "Trammell", "Varma"];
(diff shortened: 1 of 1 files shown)


toolness:
YASSS 🎉 💥 💥

ericronne:
Oooh what's this, @hbillings?

toolness:
We gotta prioritize this for the next sprint!

[I made a card.](https://trello.com/c/hG2iLDCY/233-as-a-qa-tester-i-want-to-be-able-to-quickly-pre-fill-the-price-list-upload-form-in-with-hilariously-random-data-to-make-manual-t)

toolness:
Well, I guess since we're not actively developing the Data Capture code right now, it wouldn't be _super_ useful, but you're welcome to work on it just for fun... up to you!

PR #985, opened 2016-10-31
Validate duplicate contract number in Step 1

jseppi:
This PR adds duplicate contract number validation to Step 1 of the Price List Upload flow, which is in preparation for implementing a modification flow. When a `SubmittedPriceList` with the same (case-insensitive) contract number is found, an error message that directs the user to view that Price List's details page is shown (see screenshot in https://github.com/18F/calc/pull/985#issuecomment-263304748).

There is a bit of funkiness in deciding which `SubmittedPriceList` details page to link to, because `contract_number` is actually not a unique identifier, so there could be multiple `SubmittedPriceList`s with the same one.  As implemented here, the `SubmittedPriceList` that is linked to is the one with the most recent `status_changed_at` date. In the future, we might also want to take into account the `status` (like for instance to first show the most recent `STATUS_ACCEPTED` `SubmittedPriceList`).


jseppi:
wuh oh, I just realized that for the modification step to be a single page, `ajax-form` will need to finally get modified to work with radios (ref https://github.com/18F/calc/issues/570)

jseppi:
screen shot 2016-11-04 at 8 41 06 am
When the form is passed a `SubmittedPriceList` instance, the date fields aren't left-padded (like `01` instead of `1`). Minor, but looks funky when compared with the example text.

data_capture/forms/price_list.py
```diff
@@ -97,8 +100,10 @@ def clean(self):
         return cleaned_data
 
 
-class ModifyForm(forms.ModelForm):
-    file = fields.price_list_file
+class ModifyForm(PriceListFileForm, forms.ModelForm):
+    # The file field is not required when modifying a price_list
+    file = forms.FileField(widget=UploadWidget(required=False), required=False)
```
jseppi: Seems a little weird to have to specify `required=False` in two places

jseppi:
I've run into a bit of a snag with how to handle modifications of a price list by a user who was _not_ the original submitter.

I think these are the options for when a non-original user submits a modification:

1. Change the price list's `submitter` field to the new user, thus erasing the original user's association
1. Create a 'copy' of the price list, with the `submitter` field set to the new user. The old one will be unmodified. This will leave it up to the admin to sort out what should happen with the two (like approving one and unapproving the other).

We could also only allow price list modifications to come from the original user. This will not cover the use case of when a CO leaves or when maybe workload on a specific contract is moved to another CO. Maybe this is the "safest" option to start with while we think through how to deal with modifications from another user? Then again, option `2` above seems relatively "safe" as well.

@toolness, @tram, @hbillings: thoughts?

tram:
Makes sense to me to go with option `1` and just keep an eye out for the case where a CO needs to modify someone else's price list.

hbillings:
Would it be at all helpful to have mocks for this?

Also, could the "New page to update price list" be the same as the initial upload's Step 4 (but with editable fields)?

jseppi:
Yes and yes. I have it _mostly_ implemented already (in this PR's branch). Here's a screenshot (well, 2 of them since it is a long page) of the Modify page I'm working on:

screen shot 2016-11-08 at 1 03 41 pm
screen shot 2016-11-08 at 1 03 50 pm

jseppi:
Ok, so, as discussed in our virtual coworking session today, I am going to put this PR on hold for a little while, and instead work on allowing a user in the normal Price List Upload flow to go back to earlier steps, after the upload, but then not have to upload again.

Then, we could potentially use the normal Price List Upload flow with a pre-populated session to do a modification flow!

jseppi:
I think I am going to end up simplifying this PR to just have the Step 1 validation of `contract_number`. I'm not totally sure how it is expected to look. Here is what I'm currently working toward:

screen shot 2016-11-25 at 12 23 32 pm

`"View it here"` will be linked to the price list Details view for the associated price list.

However, finding the "best" associated price list is not so straightforward since there is no uniqueness constraint on `contract_number`, and we could potentially have multiple `SubmittedPriceList` models with the same `contract_number` but with varying `status` values and created/modified dates. We kinda discussed this, and I think we would want to default to showing the details for the most recent `approved` `SubmittedPriceList`.

jseppi:
cc @hbillings and @toolness for thoughts ^^

tram:
If there are multiples, showing the most recent makes sense to me, too.

jseppi:
There are kinda multiple versions of what "most recent" means because we have 3 date fields: `created_at`, `updated_at, and `status_changed_at`.

We probably don't want to use `created_at`. I think `status_changed_at` is probably the best option.

hbillings:
Maybe it makes sense to swap the order of the contract number and schedule. Then the error message would be near the field it applies to. 

I think maybe we should say something to the effect of "If you would like to submit a modification, you can do so here."

jseppi:
@hbillings like this?

screen shot 2016-11-28 at 8 36 16 am

Note that this "special" message replaces the normal "Oops! Please correct the following error." message.

toolness:
Plz don't hyperlink words like "here" as it's not very accessible -- explanation on the [18f content guide](https://pages.18f.gov/content-guide/urls-and-filenames/#link-text)!

jseppi:
screen shot 2016-11-28 at 9 06 33 am

jseppi:
Ok, since the actual modification logic will be done in a separate PR as I mentioned in https://github.com/18F/calc/pull/985#issuecomment-263008289, I think that error text needs to just say something about viewing the existing price list since there won't be a way to actually modify it yet.

screen shot 2016-11-28 at 9 40 38 am

hbillings:
@seppi makes total sense to add the modify language later. This looks good to me!

data_capture/forms/price_list.py
```diff
@@ -1,13 +1,18 @@
...
 from ..models import SubmittedPriceList
 from ..schedules import registry
 from frontend.upload import UploadWidget
-from frontend.date import SplitDateField
 from frontend.widgets import UswdsRadioSelect
+from frontend.date import SplitDateField
 from contracts.models import MIN_ESCALATION_RATE, MAX_ESCALATION_RATE
 
 
+class DuplicateContractValidationError(ValidationError):
```
jseppi: Is creating a custom class for a specific validation error an ok practice? I did it mostly so that it would be easy to create the `has_existing_contract_number_error` method (down below here).

data_capture/forms/price_list.py
```diff
@@ -88,6 +115,7 @@ def clean(self):
     class Meta:
         model = SubmittedPriceList
         fields = [
+            'vendor_name',
```
jseppi: moved `vendor_name` to Step 2

data_capture/templates/data_capture/messages.html
```diff
@@ -2,7 +2,7 @@
 <alerts-widget>
   {% for message in messages %}
     <div class="message message-{{ message.tags }}">
-      <h3>{{ message }}</h3>
+      <h3>{{ message|safe }}</h3>
```
jseppi: Added the `safe` filter so that we could pass in a link to the `SubmittedPriceList` details page

data_capture/models.py
```diff
@@ -152,6 +152,12 @@ def reject(self, user):
     def __str__(self):
         return "Price List for {}".format(self.contract_number)
 
+    @classmethod
+    def get_latest_by_contract_number(cls, contract_number):
```
jseppi: Right now this just returns the most recent (by `status_changed_at`) match for the given `contract_number`. We might want to change it to return the most recent in a particular status also, but I'm unsure right now.

data_capture/tests/test_integration.py
```diff
@@ -74,6 +73,7 @@ def test_data_capture(self):
         )
 
         form = self.get_form('form')
+        form.set_text('vendor_name', 'Battaglia Sausage Peddlers, Inc.')
```
jseppi: moved to Step 2

data_capture/views/price_list_upload.py
```diff
@@ -124,7 +125,21 @@ def step_1(request, step):
...
             }
             return redirect('data_capture:step_2')
-
+        elif form.has_existing_contract_number_error():
+            contract_number = request.POST['contract_number']
+            latest = SubmittedPriceList.get_latest_by_contract_number(
+                contract_number)
+            if not latest:
+                # User should never be able to get to this state
+                raise AssertionError('No price list found')
```
jseppi: What's the best thing to do in this situation? There should not be a pathway for the user to get here with an existing contract_number error but not have a SubmittedPriceList found with that contract_number.

jseppi:
Ok, @toolness, this is ready for review!

jseppi:
ping @toolness

data_capture/forms/price_list.py
toolness: I think it is fine, though another alternative is to use the [`code` parameter to `ValidationError`](https://docs.djangoproject.com/en/1.10/ref/forms/validation/#raising-validationerror).  I guess that actually seems to be the convention, though I didn't know about it until I saw [Loïc Bistuer's talk on Validation](https://www.youtube.com/watch?v=uzTaWKcMzos) a few days ago.

data_capture/forms/price_list.py
```diff
@@ -47,16 +52,38 @@ class Meta:
...
 
+    def clean_contract_number(self):
+        '''
+        Raises a DuplicateContractValidationError if a SubmittedPriceList
+        with the same contract number (case-insensitive) already exists.
+        '''
+        value = self.cleaned_data['contract_number']
+
+        num_existing = SubmittedPriceList.objects.filter(
+            contract_number__iexact=value).count()
```
toolness: It might technically be more efficient to use [`exists()`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.exists) here instead of `count()`, but it probably isn't that big a deal since this code likely won't be run very often. It could be a bit more readable though, I suppose.

data_capture/templates/data_capture/messages.html
toolness: Hmm, instead of this can you use [`mark_safe()`](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.safestring.mark_safe) when constructing the message?  That way we're explicitly blessing certain messages as safe, rather than treating *all* messages as safe, which could be dangerous.

data_capture/views/price_list_upload.py
```diff
@@ -124,7 +125,21 @@ def step_1(request, step):
...
+            details_url = reverse('data_capture:price_list_details',
+                                  kwargs={'id': latest.pk})
+            messages.add_message(
+                request, messages.ERROR,
+                "We found an existing price list for contract number {}.</br>"
+                "If you'd like, you may "
+                "<a href='{}'>see its details</a>.".format(
+                    contract_number, details_url))
```
toolness: As I mentioned earlier, this would be a great place to use `mark_safe()` instead of using the `safe` filter at the template level.

Also, I don't think we're currently validating any formatting for `contract_number`, are we?  If not, we should probably wrap `contract_number` in [`escape()`](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.html.escape), right?

data_capture/views/price_list_upload.py
```diff
@@ -124,7 +125,21 @@ def step_1(request, step):
...
-
+        elif form.has_existing_contract_number_error():
+            contract_number = request.POST['contract_number']
+            latest = SubmittedPriceList.get_latest_by_contract_number(
+                contract_number)
+            if not latest:
+                # User should never be able to get to this state
+                raise AssertionError('No price list found')
+            details_url = reverse('data_capture:price_list_details',
+                                  kwargs={'id': latest.pk})
```
toolness: Oof, I was about to suggest you use `latest.get_absolute_url()` instead, but I forgot that's still a PR (#1025).

data_capture/views/price_list_upload.py
toolness: Hmm, according to the documentation for [`latest()`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.latest), which your `get_latest_by_contract_number()` uses:

> Like `get()`, `earliest()` and `latest()` raise `DoesNotExist` if there is no object with the given parameters.

So we should be able to just count on that being propagated--I don't think your `AssertionError` can actually ever be raised!

data_capture/views/price_list_upload.py
```diff
@@ -131,7 +131,7 @@ def step_1(request, step):
                 contract_number)
             if not latest:
                 # User should never be able to get to this state
-                raise AssertionError('No price list found')
+                return HttpResponseBadRequest()
```
toolness: Ack, you made this change while I was doing my code review, so my [comment](https://github.com/18F/calc/pull/985#discussion_r90501959
) regarding this line just got marked as outdated, but it's still actually relevant:

> Hmm, according to the documentation for [`latest()`](https://docs.djangoproject.com/en/1.10/ref/models/querysets/#django.db.models.query.QuerySet.latest), which your `get_latest_by_contract_number()` uses:
>
>> Like `get()`, `earliest()` and `latest()` raise `DoesNotExist` if there is no object with the given parameters.
>
> So we should be able to just count on that being propagated--I don't think your `AssertionError` can actually ever be raised!

data_capture/forms/price_list.py
jseppi: Nice, didn't know that _exist_ ed

data_capture/forms/price_list.py
jseppi: I did read about the `code` parameter when I was researching how to do this, and actually implemented this functionality using it (though never committed), but it seemed weird to me for some reason.
I feel like the specialized class is more explicit or something. I don't have a good feeling for which is better.

data_capture/views/price_list_upload.py
jseppi: Hrm, should we maybe add some minor validation (alphanumeric and hyphens) to our `contract_number` field (in another PR)?

jseppi:
Ok, @toolness, all required changes have been made!

data_capture/forms/price_list.py
toolness: Yeah, me either--maybe ask on #django in Slack?

toolness:
NOICE looks great!!!!!!

data_capture/forms/price_list.py
jseppi: My question was so gnarly it killed all discussion in that channel

PR #989, opened 2016-11-01
make submitter read-only

### LINKED ISSUE #988: Admin - Submitter should be read-only field
On the Change Price List pages (like `http://app.calc.docker/admin/data_capture/newpricelist/1/`), the Submitter field should be read-only.

screen shot 2016-11-01 at 10 50 02 am

jseppi:
Simple change.

Actually I suppose this is up for debate. I can't think of any reason why the submitter should be a modifiable field. @toolness: thoughts, or 👍 ?

closes #988


toolness:
Looks good to me--yeah, I can't think of a reason why a data administrator would modify it, so I'm fine with making it read-only.  Is super easy to undo if there turns out to be a need for it!

PR #996, opened 2016-11-02
Use SubmittedPriceListRows for table; wrap error table in expandable-area

### LINKED ISSUE #993: Valid rows in price list detail don't reflect changes made via admin UI
It looks like the admin view for changing unapproved price lists (e.g. `/admin/data_capture/unapprovedpricelist/4/`) allows data admins to view and edit "submitted price list rows".  However, changing this information doesn't reflect any changes on the "Valid price list rows" section of the price list detail page (e.g. `/data-capture/price-lists/4`).

I *think* this is because the admin UI is presenting `SubmittedPriceListRow` models, while the price list detail page is presenting the deserialization of `SubmittedPriceList.serialized_gleaned_data`. While these start out identical, the former could be changed by data admins.

In any case, this could be confusing to data admins and COs.

jseppi:
This PR changes the Price List Details page to use related `SubmittedPriceListRows` instead of the original `gleaned_data` for constructing the table of valid rows for the given `SubmittedPriceList`. The benefit to this is that the table of valid rows will reflect any changes that have been made by an admin, and can also show if the row `is_muted` or not.

In this PR the table or error rows, which still comes from the original `gleaned_data` (since it can't be changed by an admin), has been wrapped in an `<expandable-area>` so it is not shown by default.

Screenshots:

screen shot 2016-11-02 at 9 53 20 am

<hr>

screen shot 2016-11-02 at 9 53 34 am

Closes #993 

cc @toolness


jseppi:
I could see wanting to use an icon for the "Is muted" status and maybe adding a little descriptive sub-title to explain what that column means, but I'll leave that styling stuff for @hbillings :)

hbillings:
This is an awesome idea!! I was just thinking yesterday that it seems weird to make those errors so prominent when the admin can't do anything about it. I'll make an issue for styling, which reminds me that I was going to look in to how accessible Django's default little true/false icons for cols like that are...

jseppi:
I _think_ adding an `aria-label` attribute should make them accessible, right?

jseppi:
@toolness and @hbillings: review plz!

PR #997, opened 2016-11-02
Add experimental price list analysis

toolness:
This adds some experimental price list analysis functionality so we can at least figure out what our outliers are. A lot of the early work was done in a [Jupyter notebook](https://github.com/18F/calc-analysis/blob/master/contract-analysis/contract-analysis.ipynb).

As we improve the algorithm, we can potentially use this to provide a report that COs can use to guide their negotiations (this would involve having them upload *proposed* price lists to CALC, not awarded ones).

> screen shot 2016-11-02 at 5 26 53 pm

The hard part is "broadening" the search from the labor category of a price list row to be more generic, while still being useful and not requiring administrators to constantly manage some kind of hand-crafted classification system. Here's how it currently works:

1. First, we ask the database for a list of words in the corpus of labor categories that occur at least a certain number of times (currently 100).
2. We filter against this list to find the most salient parts of a given row's labor category. For instance, it's how we toss out words like "Ridiculously", "Awesome", and "Doom" in the above screenshot.
3. Finally, we permute that filtered list of words until we find a combination that a decent number of existing contracts match against. If no contracts match, then we don't have any advice to give, which is the case with "Hamburger Chef" in the above screenshot.

Notes:

* The links in the screenshot point to CALC searches for the given labor category and education level. ~~The min/max years of experience is set to be the two years around the minimum experience level of the price list row: so if the price list row requires a minimum of 7 years experience, the linked-to CALC search has a minimum experience of 5 years and a maximum experience of 9.~~ First a search for labor categories matching the *exact* same education and experience level is done; if that doesn't result in a minimum number of comparables, then a search is done to find contracts that are *greater than or equal to* the education and experience level of the source row.
* ~~Currently a labor category like `junior administrative engineer` would result in a CALC search for `engineer` rather than `junior engineer` due to the way the algorithm works.~~ Fixed in c673afdbc8a90cd894fe5ad84eac6c6bd3f282a5.
* The "vocabulary" of frequently-occurring words described above is actually quite small and can fit in memory pretty easily. **Update:** Because the minimum number of comparables is now really low (4 versus the old value of 100), we need to have a much more granular vocabulary too. I'm not sure how big it is, though it does still seem to fit into memory.
* Trying out different permutations of words can take a little while, as each permutation requires a separate database query to determine if it yields a significant number of results. This could get slow under load, but I suppose one advantage is that it's somewhat scalable, if we have multiple database replicas. We can also use caching to speed things up, some of which is already being done. I've also added a word co-occurrence matrix to speed things up.
* Because we're not explicitly designing a classification system, some of the classifications the system comes up with can be a bit weird. ~~For instance, it might compare `Junior hamburger chef` to `Junior`, which isn't *really* a labor category (using part-of-speech tagging wouldn't be very useful here, since `Junior` is both an adjective and a noun). (That said, who knows, actually doing a search on `Junior` constrained by experience and education could actually be useful!)~~ Many "weird" classifications are eliminated if `nltk` is installed, in which case we use part-of-speech tagging to make sure that the broadened query we've come up with actually has nouns in it. 

To do:

- [x] If a labor category is particularly long and contains lots of frequently-used words, the permutation phase could take a *really* long time. We need to limit the maximum number of permutations. (I added something to cap this in 09b746b9af646fad98801836bc5f09cfb7ee44d6.)


data_capture/templatetags/analyze_contract.py
```diff
@@ -0,0 +1,182 @@
...
+
+@register.filter
+def analyze_contract_row(row):
+    # TODO: Currently this only works w/ fake schedule rows. Should
+    # figure out how to make it work independent of schedules; that
+    # might mean abandoning this weird template filter approach.
+
+    if (row['price'].errors or row['years_experience'].errors or
+            row['education'].errors or row['service'].errors):
+        return None
```
toolness: The indentation here is super weird to me, but the style that is natural to me currently violates [`E129 visually indented line with same indent as next logical line`](https://github.com/PyCQA/pycodestyle/issues/386).  From the conversation on that issue, though, it looks like it's advised we just globally disable that feature, since PEP 8 specifically gives no guidance on it.

jseppi: That looks totally normal to me o_O But you can turn off E129 if you want, I don't really mind either way.

Or maybe

```py
has_errors = (row['price'].errors or row['years_experience'].errors or
              row['education'].errors or row['service'].errors)
if has_errors:
    return None
```

toolness: Haha, if it looks totally normal to you, I'm fine with it then :)

toolness:
Just modified the analysis text so it's a bit less jargony:

screen shot 2016-11-16 at 1 37 00 pm

In particular, we're only saying whether it's above/below +1/-1 standard devation from the mean, rather than specifying how many standard deviations from the mean it is.  However, the number of *a*'s in *way* is equal to the number of standard deviations, so saying "waaay below" actually means it's three standard deviations below.

toolness:
Here's a screenshot of what the analysis looks like now:

> screen shot 2017-01-12 at 5 01 27 pm

toolness:
@jseppi -- just FYI, I moved the `xlsxwriter` dependency into the `Dockerfile`.  This makes it easier for us to switch between this PR and other `develop`-based PRs, since the `requirements.txt` stays the same and Docker just keeps the price analysis dependencies in a separate filesystem layer (we were already doing this for `nltk` actually).

toolness:
@jseppi any idea what's up with [these flake8 errors](https://travis-ci.org/18F/calc/builds/199006083#L505-L510)?

```
./data_capture/tests/test_s70.py:94:39: F821 undefined name 'deepcopy'
./data_capture/tests/test_s70.py:95:39: F821 undefined name 'deepcopy'
./data_capture/tests/test_s70.py:106:39: F821 undefined name 'deepcopy'
./data_capture/tests/test_s70.py:107:39: F821 undefined name 'deepcopy'
./data_capture/schedules/s70.py:128:25: F821 undefined name 're'
./data_capture/schedules/s70.py:128:54: F821 undefined name 're'
```

It seems like these might have come out of merging in #1284 but I'm not sure...

jseppi:
Weird, #1284 was passing everything: 

screen shot 2017-02-07 at 9 13 27 am

jseppi:
Imma fixing

toolness:
Hmm, I did recently merge `develop` back in, and it included #1324, which I think upgraded some of our linting packages--could that be responsible?

jseppi:
No, it looks like those imports actually were missing...

toolness:
Ummm that is super weird.  @jseppi can you issue a PR that fixes the breakages, or are you busy w/ ATO/other stuff right now?

jseppi:
Whoops, pushed directly to this branch before I saw your request that it be in a PR.

toolness:
Haha no worries, direct commit is good too!

jseppi:
omg what is going on

jseppi:
Hrm, looks like maybe something just wrong with the changes to `.travis.yml` now? I pulled and ultratest'd locally and everything is good.

jseppi:
Something to do with sauce? 

`The command "python frontend/tests/sauce_connect.py py.test --ignore=frontend/tests/test_selenium.py --cov" exited with 1.`

(ref https://travis-ci.org/18F/calc/builds/199255894#L5734)

toolness:
Oh, one of those errors is happening because some nltk data isn't downloaded. I'll add it to the travis.yml.

toolness:
Um. I am confused as to how these tests *ever* passed on travis.

jseppi:
Something ultrafunky is going on...

toolness:
I am reaallly confused right now.  I pushed edits to `.travis.yml` in 6ca266641418bc5f5008a803fb08242791378b67 to hopefully fix it, but the travis build log seems to be running the *old* `.travis.yml`... WUT.

toolness:
Dude what is up with travis using versions of `.travis.yml` from previous commits. I am so friggin confused.

toolness:
Oh @jseppi I think that the broken linting/actually broken module might have been an artifact of merging #1320 into this PR... I think git might have *thought* it was a conflict-free merge, when in fact it resulted in completely broken code.

.gitignore
```diff
@@ -32,7 +32,11 @@ calc-prod_manifest.yml
 tag-message-*.txt
 release-*-instructions.txt
 release-*-instructions.html
+<<<<<<< HEAD
```
jseppi: merge conflict, @toolness

toolness:
Ok, this big thing is now updated with the latest `develop` branch!

KellyBailey:
Some guidance for us to incorporate or consider for price reasonableness analysis-

https://dap.dau.mil/acquipedia/Pages/ArticleDetails.aspx?aid=1ae4716b-d8f3-4839-9785-adbd19d5a26f

https://www.acquisition.gov/gsam/current/html/Part538.html

These references are mostly for me to use to define the parameters of the analysis.

toolness:
So the main thing that would make this difficult to offer in production is the current algorithm for how searches are "broadened" to find a desirable number of comparables in `data_capture/analysis/vocabulary.py`.  By "broadened" we mean that words from a labor category are removed so that doing a CALC search for them yields more results (e.g. "Ridiculously Awesome Engineer of Doom II" is broadened to "Engineer II").

Currently I think that algorithm involves a lot of memory _and_ a lot of computation.  It basically figures out all possible permutations of words in the labor category and ranks them by the maximum number of matches they might have, given a word co-occurrence matrix that is actually pretty big (the matrix didn't used to be as big--originally it only contained words that occurred at least 100 times in the database but it was reduced to 4, making the matrix much larger).  So ideally this algorithm should be different, or else it runs the risk of crashing or DOS'ing CALC as CALC's database grows.

Aside from that, though, I know there is also some statistical things that @EricSchles had concerns about: if I recall correctly, he thinks that terms like "median" and "standard deviation" only really make sense in a normal distribution, and when we're making up our own search criteria from scratch, we're quite likely to get something that isn't a normal distribution. But since the current algorithm is a reflection of how people are already using CALC manually, it might be a bigger problem than just this algorithm....

hbillings:
@KellyBailey would like to be able to at least demo this on staging in a couple of weeks.

ericronne:
@tadhg-ohiggins do you happen to know what's going on with the conflict?

tadhg-ohiggins:
@ericronne Yes, and it should go away once I push my last change to this branch (which should occur in the near future).

ericronne:
Is near almost here? If so, i'll buy you a 🍺!*

_*virtual_

tadhg-ohiggins:
@hbillings Since I've made a bunch of changes here I thought I'd ask you to give it a quick look over; I did take the quickest route to getting the tests to pass and so `multi_phrase_search` and `get_queryset` are duplicated in two classes, which probably isn't ideal, but I don't feel familiar enough with the code yet to work out a better approach.

tadhg-ohiggins:
Actually we're not going to merge this until we fix some of the content issues (and possibly tweak some of the user flow).

ericronne:
Todo
- [ ] error checking #2077
- [ ] tabular-format analysis #2093
- [ ] just added: implement new intro content #2098

abrouilette:
@hbillings to discuss putting this on production and behind a login w/ @tadhg-ohiggins. Price analysis testers will be given price list upload privileges. (This way we can use existing group.)

tram:
Test with R2, R6, IT70.

toolness:
O hai! Let me know if y'all have any questions about this crazy branch, I am happy to hop on a non-FOIAble channel.

ericronne:
A bunch of the data* in the download is carried to about seven or nine decimal points, which is visually overwhelming and probably not meaningful. Should we round to, say, two decimal points?

image

*Avg experience, avg price, % dif from avg, and standard deviation data

tadhg-ohiggins:
@toolness We're trying to decide on what number of occurrences to go with in production, and I wanted to verify that my basic understanding of what's going on is reasonable, so here's how I think this works:

If the DB has the following occurrences of titles in it:

+   50 × Engineer of Doom
+   30 × Assistant Engineer of Doom
+   4 × Grandmaster Engineer of Doom
+   2 × Deathless Grandmaster Engineer of Doom

Then these are the number of matches we'd see for the following values of the minimum occurrence number, assuming a search for `Engineer`:

| Min. Occurrences | Num. Results |
| --- | --- |
| 100 | 0 |
| 50 | 50 |
| 40 | 50 |
| 30 | 80 |
| 10 | 80 |
| 4  | 84 |
| 3  | 84 |
| 2  | 86 |

Thanks in advance for any help!

contracts/models.py
```diff
@@ -205,6 +205,52 @@ def order_by(self, *args, **kwargs):
 
         return queryset
 
+    def multi_phrase_search(self, query, *args, **kwargs):
```
Jkrzy: Duplicate of `ContractsQuerySet.multi_phrase_search` and looks like it's sole invocation can be replaced by the `CurrentContractManger` implementation. Recommend that change -- `.objects.multi_phrase_search` -- be made and this, along with`get_queryset` below, be removed.

data_capture/analysis/core.py
```diff
@@ -0,0 +1,234 @@
...
+        result['dist_from_avg_percent'] = (result['dist_from_avg'] / avg) * 100
+        result['preposition'] = ('below' if price < avg else 'above')
+
+        result['labor_category'] = fc.phrase
+        result['url'] = get_data_explorer_url(fc)
+        result['comparable_search_criteria'] = {
+            'exp': fc.finder.get_exp_comparable_search_criteria(),
+            'edu': fc.finder.get_edu_comparable_search_criteria(),
+        }
+        result['description'] = render_to_string(
```
Jkrzy: As of 7f2aa57ac7cc7bf0decc9c7760d0a50f661d2d6b, `description` is no longer used in the final rendered template. It, along with the template, can be removed.

Jkrzy:
@tadhg-ohiggins ,

My take on  (https://github.com/18F/calc/pull/997#issuecomment-420014557)...

`DEFAULT_MIN_NDOC` -- minimum occurrence for inclusion in the comparison vocabulary.

Your specific example would return all 86 contracts for every value of `DEFAULT_MIN_NDOC`. We're `multi_phrase_search`ing the original value before going down the rabbit hole in `broaden_query` and a search for "Engineer" will return all of those contracts.

`DEFAULT_MIN_NDOC` comes into play when the provided labor category returns *no* search results.

Let's search for **"Awesome Machinist"** within the following data.

Category | Contract count
--------|------
Machinist | 10
----

`broaden_query` will return 0 results, and no analysis, unless `DEFAULT_MIN_NDOC` is 10 or less.

The lower the value of `DEFAULT_MIN_NDOC`, the greater the size of the vocabulary and subsequently the possible permutations, w/ matched vocab values, of an incoming labor category. If a wacky labor category comes in with 100 words in the vocab, `powerset` is going to explode when it tries to generate all permutations. 

The memory requirements of the cooccurrence matrix also increases along with the vocabulary size. A quick check against a calc dataset of ~52k records gives us a few benchmarks below. Also attached here is a csv of the potential vocab words w/ their counts. [Vocab-Table.txt](https://github.com/18F/calc/files/2390088/Vocab-Table.txt)


Cooccurrence matrix size in Bytes | DEFAULT_MIN_NDOC | len(vocab)
-------|---------|-------
3603934 | 4 | 1609
2535254 | 10 | 816
1649533 | 25 | 407
1140834 | 50 | 237
941743 | 75 | 189
846095 | 100 | 167

With the current size of the calc dataset, 65350 records, and the limited number of users with access to this experimental feature, a `DEFAULT_MIN_NDOC` of 4 should be supportable. Though as @toolness noted earlier, when the underlying data grows and use of this tool evolves/expands the algorithm should be refined and optimized.

As a first step for that optimization, I'd expand caching around vocabulary, the cooccurence matrix, permutation generation, and permutation count lookups.

tram:
Describe analysis features as in alpha and potentially wonky.

data_capture/analysis/vocabulary.py
```diff
@@ -0,0 +1,244 @@
...
+        return upper_bound
+
+    def are_cooccurrences_valid(lexemes):
+        for a, b in combinations(lexemes, 2):
+            if vocab.get_cooccurrences(a, b) < min_count:
+                return False
+        return True
+
+    # Remove the first element, as it's the empty set.
+    permutations = list(powerset(lexemes))[1:]
```
Jkrzy: This invocation of `powerset` has the potential to 💥 when passed a long list of lexemes. We should cap that length, or otherwise limit the number of permutations we're calculating

data_capture/templates/data_capture/analyze_price_list/step_3.html
```diff
@@ -0,0 +1,89 @@
...
+            <th id="avg-price" scope="col">Average<br />price</th>
+            <th id="plus-minus-dollars" scope="col">+/- avg.</th>
+            <th id="plus-minus-percent" scope="col">% +/- avg.</th>
+            <th id="plus-minus-std-dev" scope="col">+/- std.<br />deviation</th>
+          </tr>
+        </thead>
+        <tbody>
+        {% for row in analyzed_rows %}
+          <tr>
+            <td headers="your-data proposed-service"><a href={{ row.analysis.url }} target="_blank">{{ row.labor_category }}<span> ({{ row.analysis.count }})</span></a></td>
```
Jkrzy: As this is an experimental feature, It may be helpful to display the matched labor category on the results page. Right now it is accessible only by following this link, or examining the URL, to the search results. I think this would help avoid confusion when the matching and/or analysis seems strange.

i.e. `Clerical I` matched to ~260 results for `Clerical` -- including Clerical II, Clerical III, Clerical 4, etc, etc.

data_capture/analysis/core.py
```diff
@@ -0,0 +1,234 @@
...
+                submitter_id=0,
+                escalation_rate=0,
+                status_changed_at=timezone.now(),
+                status_changed_by_id=0,
+            )
+            price_list.save()
+            gleaned_data.add_to_price_list(price_list)
+            for row in price_list.rows.all():
+                valid_rows.append(analyze_price_list_row(cursor, vocab, row))
+            transaction.savepoint_rollback(sid)
```
Jkrzy: We should generate the list of `SubmittedPriceListRow`s to iterate through for analysis without writing to the database. Neither the price list or individual rows need to be in the database for the analysis to complete.

Jkrzy:
@tadhg-ohiggins, @hbillings 

Here's a quick summary review....

There are a handful of issues that should be tackled before release, specifically towards preventing the analysis from failing w/ out of memory  and making clear the results of the matching for the price analysis. 

Beyond that there remains the potential of a lengthy or otherwise complex incoming price list taking too long to analyze. This is mitigated a bit by the limited set of users w/ access to the tool so long as they're made aware of the experimental nature of this feature and its limitations. We can work towards bringing the runtime down w/ additional caching of intermediary work and not writing to the database.

Additional logging throughout the analysis/matching, specifically the intermediate broadening steps, would also be helpful going forward to assist in the evolution of the algorithm by future developers. 

Additionally, there's a good amount of house keeping that can be done in the form of refactoring, removal of dead code, and outstanding TODOs throughout the PR. I'll submit new issues for the later.

Definitely address before release:

- Prevent `powerset` from 💥 (https://github.com/18F/calc/pull/997#discussion_r218212890)
- Update dependency minor versions (https://github.com/18F/calc/pull/997#discussion_r217448858)
- Make "broadening" results readily available to user (https://github.com/18F/calc/pull/997#discussion_r218215480)

Nice to have:
- Reduce cpu/db load of analysis (https://github.com/18F/calc/pull/997#issuecomment-422153369, https://github.com/18F/calc/pull/997#discussion_r218474641)

Jkrzy:
These earlier concerns from the world of stats are worth resurfacing and addressing as a separate issue once this is merged. 
https://github.com/18F/calc/pull/997#issuecomment-398168152:
> Aside from that, though, I know there is also some statistical things that @EricSchles had concerns about: if I recall correctly, he thinks that terms like "median" and "standard deviation" only really make sense in a normal distribution, and when we're making up our own search criteria from scratch, we're quite likely to get something that isn't a normal distribution. But since the current algorithm is a reflection of how people are already using CALC manually, it might be a bigger problem than just this algorithm....

tadhg-ohiggins:
I addressed a number of @Jkrzy 's comments in my last several commits, and those that we ran out of time for I created new issues around (https://github.com/18F/calc/issues/2179 and https://github.com/18F/calc/issues/2180). This should be ready for final review and (hopefully!) approval for merging into `develop`.

data_capture/analysis/vocabulary.py
```diff
@@ -23,6 +23,12 @@
 ANALYSIS_STOP_WORDS = []
 
 
+# Cap on number of lexemes that get_best_permutations will process.
+# This is a crude attempt to stop the CPU/RAM consumption of the function from
+# getting too large.
+MAX_LEXEMES_FOR_PERMUTATIONS = 4
```
Jkrzy: Recommend bumping this up to 8 so as to avoid handicapping the broadening algorithm's attempts to identify the "best" permutation while still preventing the process from exploding on weird input.

At 8, we'd avoid trimming lexemes for ~99.4% of the contracts we've seen thus far. At 4, that drops to ~85%. Explored by binning the number of lexemes matched to the vocab on a test dataset of ~55k labor categories.

PR #1001, opened 2016-11-02
Add support for radio and checkbox inputs to ajax-form

### LINKED ISSUE #570: ajaxform component doesn't work w/ radio inputs

jseppi:
This PR adds `input type="radio"` and `input type="checkbox"` support to our `ajax-form` web component.

I've also created a `UswdsCheckbox` widget for WDS-style checkboxes, in the same vein as our `UswdsRadio` widget. Styleguide examples have been updated to include some checkbox inputs.

Closes #570 

Needed for #985 

TO DO:
- [x] verify that the formData for checkboxes can be read by the Django backing form
- [x] figure out what to do about `<select>` and `<textarea>` form elements


frontend/source/js/data-capture/ajaxform.js
```diff
@@ -94,9 +94,6 @@ class AjaxForm extends window.HTMLFormElement {
 
       if (el.isUpgraded) {
         formData.append(el.name, el.upgradedValue);
-      } else if (el.type === 'radio' || el.type === 'checked') {
```
jseppi: I may have missed this before, but should that `el.type === 'checked'` clause instead have read `el.type === 'checkbox'`?

jseppi:
So, it looks like `ajax-form` will not work with `<select>` or `<textarea>` elements. Should we file an issue and/or put some error-throwing code in `ajaxform.js` to help out future devs?

jseppi:
Ready for review, @toolness!

frontend/source/js/data-capture/ajaxform.js
toolness: Oh heh, [yup](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox)! My bad.

toolness:
> So, it looks like `ajax-form` will not work with `<select>` or `<textarea>` elements. Should we file an issue and/or put some error-throwing code in `ajaxform.js` to help out future devs?

Yup that sounds good to me!

toolness:
This looks rad! 👍

PR #1005, opened 2016-11-03
[WIP] Add CALC log analysis (jupyter notebook)

toolness:
This adds the CALC log analysis I did last sprint.

It requires the CALC API logs, which are constantly changing, too large to include in the repo, and include potentially sensitive data. So the jupyter notebook has instructions on how to retrieve them instead.

[A rendered version of the HTML is also available.](https://s3.amazonaws.com/atuls-sandbox/CALC%2BLog%2BAnalysis.html)


jseppi:
Sauce!!!! :shake-fist:

toolness:
One thing I'm unsure about here is what our recommended way for getting jupyter notebook up and running is. Some options:

1. Install it in the `app` container.
2. Make a separate `jupyter` container for it.
3. Expect folks to just start a virtualenv for it on their local system, or use a standard data science community alternative like anaconda.

One advantage of putting it in a container is that it might make it easier to integrate with CALC's database, if we find a need to do that.

jseppi:
Making a separate container seems like a good idea to me. Maybe even just a Dockerfile in the `analysis/` directory, and can worry about linking to the db container later if there ends up being a need for it.

toolness:
Ok, based on a conversation in Slack yesterday, I've opted instead to put this in a separate [18F/calc-analysis](https://github.com/18F/calc-analysis) repository. Closing this PR now.

PR #1007, opened 2016-11-03
make status_changed_by and status_changed_at non-null

### LINKED ISSUE #1000: In admin interface, status_changed_by_email can cause 500 error
Discovered by @hbillings

If the `status_changed_by` relation is not set on a `SubmittedPriceList`, then the `status_changed_by_email` method in `admin.py`'s `BaseSubmittedPriceListAdmin` will throw an exception.

Note that this is currently causing problems in staging, so we should fix it ASAP.

jseppi:
This PR makes `status_changed_by` and `status_changed_at` non-null fields and provides a migration to set default values.

Closes #1000


jseppi:
Let me know if this looks ok, @toolness

toolness:
Looks good to me! 👍

PR #1022, opened 2016-11-08
check price_list_rows instead of gleaned_data.valid_rows

jseppi:
Relatively minor fix. The `if` statement in the details template was checking the wrong condition for showing the table of existing price list rows.


toolness:
Noice!

PR #1023, opened 2016-11-08
Skip GA_TRACKING_ID test if defined in env

toolness:
One of our tests is to ensure that the default value of `settings.GA_TRACKING_ID` is an empty string when it's not defined already in the OS environment.  However, if it *is* already defined in the OS environment (which is the case on my dev system right now), there's actually no way for us to test this, so we'll skip it.

### DIFF
--- a/hourglass/tests/tests.py
+++ b/hourglass/tests/tests.py
@@ -1,5 +1,7 @@
+import os
 import unittest
 import json
+from unittest import SkipTest
 from unittest.mock import patch
 from django.test import TestCase as DjangoTestCase
 from django.test import override_settings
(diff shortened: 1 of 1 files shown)


jseppi:
I restarted the failing pr check -- looks like stalled SauceLabs.

This is probably fine. Could we also potentially use [EnvironmentVarGuard](https://docs.python.org/3.6/library/test.html#test.support.EnvironmentVarGuard) for temporary setting/unsetting of env vars during testing?

toolness:
Hmm, that is nifty and I didn't know about it, but the main problem is that the code that sets the setting to its default value is executed upon the very first import of `settings.py`--once it's been imported, we can't re-import it without doing some weird things that aren't generally recommended. And it'll almost certainly have been imported already by the time the test (or the code within the `EnvironmentVarGuard`) first runs.

I mean, an alternative is to factor out the `os.environ.get('GA_TRACKING_ID', '')` in `settings.py` to a separate function which we then test, but that feels like overkill...

jseppi:
Ah yeah, true. Yeah, I don't think we need to go nuts with this, so 👍

PR #1024, opened 2016-11-08
Make docker_django_management.py output to stderr

### LINKED ISSUE #1003: docker_django_management.py should only write to stderr
Because it currently writes things like `Connection to database established` to stdout (via `print()`), it's difficult to use it with management commands that output structured data to stdout, like [`dumpdata`](https://docs.djangoproject.com/en/1.10/ref/django-admin/#dumpdata). Often such commands do allow sending output directly to a file, but this may not always be the case, and it's a hassle in any case.

If we write to `stderr` instead that should fix things.

toolness:
Fixes #1003.

Note that this doesn't actually seem to make the output go to `stderr` by default: in order to really send `stderr` in the Docker container to `stderr` in the host, you'll need to pass the [`-T` option to `docker-compose run`](https://docs.docker.com/compose/reference/run/), e.g.:

```
docker-compose run -T app python manage.py dumpdata > boop.json
```

Um, except that the output is very inconsistent, which seems to be due to https://github.com/docker/compose/issues/3267. Weird.

### DIFF
--- a/docker_django_management.py
+++ b/docker_django_management.py
@@ -78,17 +78,25 @@ CONTAINER_NAME = os.environ.get('DDM_CONTAINER_NAME')
 IS_RUNNING_IN_DOCKER = 'DDM_IS_RUNNING_IN_DOCKER' in os.environ
 
 
+def info(msg):
+    '''
+    Prints a message.
+    '''
+
+    sys.stderr.write('{}\n'.format(msg))
+    sys.stderr.flush()
+
+
 def warn(msg):
     '''
     Prints a warning message in red.
     '''
 
-    print(
+    info(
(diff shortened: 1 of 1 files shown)


jseppi:
With these changes, is stuff still logged to the console? I don't really know much about TTY stuff (like I don't even know what TTY is).

I had gotten around this by starting up `bash` within the `app` container and then just running commands like `dumpdata` there.

toolness:
Oh good idea about just running within `bash`.

Stuff is still logged to the console, yup!

toolness:
@jseppi is this ok to merge?

jseppi:
:+1:

PR #1025, opened 2016-11-08
[NEEDS DESIGN WORK] Add SubmittedPriceList.get_absolute_url()

### LINKED ISSUE #995: Define get_absolute_url() for submitted price lists
If we define a [`get_absolute_url()`](https://docs.djangoproject.com/en/1.10/ref/models/instances/#get-absolute-url) method for `SubmittedPriceList`, that points to e.g. `/data-capture/price-lists/4`, then the admin UI will automatically link to it via a "View on site" link.

This could be convenient for devs and data administrators, especially if there is information in the price list detail view that we don't surface in the admin (like the invalid row information).

It might also be nice to link the other way, i.e. if the user is a data administrator then have a "View in admin UI" link on the price list detail page, but we should probably file a separate bug for that if it's useful.

toolness:
This fixes #995.

However, it might actually be a bit confusing for users, the way it's currently displayed in the admin:

> screen shot 2016-11-08 at 3 48 45 pm

Because the admin UI now looks like the rest of the site, the phrase "View on site" might actually be confusing to admins.  I'm not sure what a better wording might be, though; nor do I know how hard it might be to alter the admin UI to display that wording.

We might also want to change the styling to make "History" and "View on site" appear on the same line, with some kind of delimiter between them.

Thoughts on any of the above, @hbillings?

To do:

- [x] Add a test for `get_absolute_url()`.

### DIFF
--- a/data_capture/models.py
+++ b/data_capture/models.py
@@ -1,4 +1,4 @@
-
+from django.core.urlresolvers import reverse
 from django.db import models
 from django.contrib.auth.models import User
 from django.core.validators import MinValueValidator, MaxValueValidator
@@ -148,6 +148,11 @@ class SubmittedPriceList(models.Model):
     def reject(self, user):
         self._change_status(self.STATUS_REJECTED, user)
         self.save()
+
+    def get_absolute_url(self):
+        return reverse('data_capture:price_list_details', kwargs={
+            'id': str(self.id)
+        })
 
     def __str__(self):
         return "Price List for {}".format(self.contract_number)
--- a/data_capture/tests/test_models.py
+++ b/data_capture/tests/test_models.py
@@ -52,6 +52,14 @@ class ModelTestCase(BaseLoginTestCase):
 
 @override_settings(DATA_CAPTURE_SCHEDULES=[FAKE_SCHEDULE])
 class ModelsTests(ModelTestCase):
+    def test_get_absolute_url_works(self):
+        p = self.create_price_list()
+        p.save()
+        self.assertEqual(
+            p.get_absolute_url(),
+            '/data-capture/price-lists/{}'.format(p.id)
+        )
+
     def test_add_row_works(self):
         p = self.create_price_list()
         p.save()


jseppi:
Looks like the "View on site" text is hardcoded into the template: https://github.com/django/django/blob/master/django/contrib/admin/templates/admin/change_form.html#L35

jseppi:
So I guess we could extend `admin/change_form.html` and override the `object-tools` block.

hbillings:
Yeah, altering the template shouldn't be hard at all! "History" should go away, at least on new price lists, because it doesn't have any. 

But now I'm wondering...what's the advantage of having the admin-specific views? The "new" lists are obviously editable, but for the other statuses, there aren't any actions you can take on the price list page itself. They share a lot in common with the `price-lists` page. Should we make the lists that aren't editable just render the corresponding `price-lists` page with admin-specific information?

(Sorry if this doesn't make any sense; I've been trying to put my thoughts into words off and on for about an hour.)

hbillings:
Example flow, from the data admin side: 
Approved price lists -> click contract number -> get `price-lists` view

toolness:
Good point about questioning the existence of the admin-specific views other than new/unapproved... I wonder if we could just add the appropriate admin-only buttons to the normal price list detail page for some of those. Hrm.

That said, I would caution against removing the **History** link, even on the "New Pricelists" view--while the vast majority of those will have no history, it's possible that some might not, e.g. a price list that an admin accidentally approved and then undid (this is kind of the price list analogue of viewing an e-mail but then wanting it to be re-marked as unread).  Or if an admin doesn't make an approval action on a new pricelist, but does edit some of its information, and so on.  Being able to access that history would be useful, so unless it really confuses folks, I think we should keep it!

hbillings:
I'mma mock something up real quick, but I'm thinking maybe that's a separate PR?

Oooh, I thought that any action on a new pricelist would change its status to something else. Good to know! I wonder if surfacing the most recent action on the page would be helpful...hmm...

toolness:
Hmm, I'm going to close this for now, we can revisit it later via #995 if we want!

PR #1027, opened 2016-11-08
Allow skip ahead on Step 3 if price list already uploaded

jseppi:
This PR adds the functionality so that if a user goes back to Step 1 or 2 after having already uploaded a price list file, they will not have to do the upload again.

To Do:
- [x] update tests for new interstitial step_3_data view


jseppi:
@hbillings: Any ideas for how Step 3 should look if they arrive there having already uploaded a file? Right now I just show a simple placeholder message:

screen shot 2016-11-08 at 4 44 02 pm

What should happen if the uploaded file had some error rows? Should the new "Review Data" button go to Step 3 Errors?

data_capture/templates/data_capture/price_list/step_3_errors.html
```diff
@@ -61,6 +61,8 @@ <h2 id="upload" tabindex="-1">Want to try that again?</h2>
     {{ form.file }}
 
     <div class="form-button-row clearfix">
+      <a href="{% url 'data_capture:step_3' %}" class="button button-previous">Previous</a>
```
jseppi: @hbillings: I added a previous button here since it was missing.

screen shot 2016-11-08 at 4 45 08 pm

jseppi:
Should the step indicators be linked to their respective step pages? That would give "power users" a way to navigate back to step 1 without first going through step 2.

Maybe that should be done in a separate PR.

hbillings:
Yeah, I think we'll want to link the previous steps (should be a separate PR because I'll need to wrangle the HTML a little for that).

What if, instead of the message + review data button, we rendered the data above the upload box? Similar to step 3A, except with good data + errors?

jseppi:
Cool, yeah, I can totally render the tables right there on Step 3. Then whatever the new button is will route to Step 4.

jseppi:
This is ready for some interaction review. Meeting scheduled with @hbillings for this afternoon.

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -60,6 +60,7 @@
 <form method="post">
   {% csrf_token %}
   <div class="form-button-row clearfix">
+    {# TODO: What to link to here if got here via the skip ahead button on Step 3? #}
```
jseppi: Right now this will go back to Step 3 Errors if there are any invalid rows. But what if the user got to Step 4 via the new button on Step 3? How can we even tell how the user got here?

jseppi:
Ok, I think this is ready for review, @toolness and @hbillings.

I would appreciate if you could actually run the code and try a few different uploads (like one with no errors, one with some errors and some valid, and one with only errors), and go back and forth through the steps to make sure I didn't miss anything.

toolness:
This is looking great!

One thing that does feel a little bit confusing is that if I upload a price list with errors, then go back to `/step/3`, and click "next" without uploading anything, I'm sent to `/step/3/errors`, which looks a whole lot like `/step/3` (because it's showing me my previously-gleaned data w/ invalid rows) but with a slightly different UI/copy.

I wonder if it might be better to skip `/step/3/errors` if the user simply clicked "next" without uploading anything in step 3, since we know the user has already seen the information about the invalid rows in the previous screen? We don't seem to be giving them any additional information or options in `/step/3/errors` in this particular case. We could also file this as a separate issue to address it later if it's too much work for this PR.

toolness:
Oh wait!  Another alternative would be to simply push them on to `/step/3/errors` if they go to `/step/3` but have any invalid data. This would also nicely handle the case where they have existing gleaned data that is completely invalid.

So in other words, `/step/3` is only shown when the user has *no* gleaned data or when the user has *100% valid* gleaned data. But if the user has *any* invalid gleaned data, they're sent to `/step/3/errors` instead. Does that make sense? Or am I missing some kind of edge case?

jseppi:
@toolness: If the user has an uploaded price list with invalid errors, and we direct them directly to `step/3/errors` when they try to go back to `step/3` (maybe coming from `step/2` most likely), then I think there would potentially be confusion because they wouldn't see any of the _valid_ data they've uploaded. 

I also agree that seeing the errors in `step/3` and then again `step/3/errors` is not great. I'm more on board with your initial suggestion of skipping `step/3/errors` if coming from `step/3` and having already submitted a price list.

Either way would not be hard to implement in this PR.

@hbillings?

jseppi:
Hrm, maybe on `step/4` there could be a small message that says something like `"{} rows were invalid and will not be included. <a href="step/3">Go back</a> if you'd like to upload a corrected price list."`

jseppi:
I went ahead and implemented @toolness's original suggestion, and updated some of the copy/styling on `step/3` when a price list is already present: 1) added a link to skip down to the upload, 2) added the "Want to try that again?" header as in `step/3/errors`

toolness:
Ohh got it, yeah, I forgot about `/step/3/errors` not showing the valid rows.  Skipping `/step/3/errors` if "next" was clicked w/o uploading anything new does sound fine though--as long as we remember to keep that file upload *required* if all previously-gleaned rows were invalid!

jseppi:
👍 I'll make sure that case is covered and tested

hbillings:
TBH, I think this has all gotten pretty confusing. :-/ Skipping steps that a user has already seen is bound to make people wonder if they missed something. 

One thing we might do is decouple the previously submitted data from the upload box completely. (A pre-step 3, if you will.) That page could function as an interstital that says, "Hey! You've got this previously uploaded stuff! You want to go upload new stuff, or you want to go right to submitting it?"

I feel like we're hacking the current flow to make this work, but I think what we really _should_ do is take a step back and see if there are ways to refactor the flow to make it work better for all of the cases involved. Which is probably bigger than the scope of this PR. 😭

hbillings:
One thing I was going to mention and forgot, 'cause brains: If we're going to skip steps at any point, the user should be the one to initiate that. Doing it for them is the part that gets confusing.

jseppi:
Hrm, so as it is now (before implementing the last suggestion @toolness made), if the price list is only invalid rows, then the user will go to `step/3/errors` where they will be presented with the `Your uploaded price list contains no valid rows...` message, and have to upload again to proceed. Would that be better since it better mirrors the original flow, @hbillings?

jseppi:
Alright based on discussion during coworking today, new idea is to add a new page, `step/3/data`, that will be kind of an interstitial when valid price list rows already exist in session and the user goes either back/forward to step 3. 

`step/3/data` will show both valid and invalid rows, and will have 4 buttons: `Previous` (back to step 2), `Upload new price list` (to regular step 3), `Next` (to step 4), and `Cancel`. 

Also, the `Previous` button `step/3/errors` will be changed to go to step 2.

data_capture/views/price_list_upload.py
```diff
@@ -137,63 +145,91 @@ def step_2(request, step):
...
-            template_name=step.template_name,
-            ajax_template_name='data_capture/price_list/upload_form.html',
+    session_pl = request.session['data_capture:price_list']
+    step_1_data = get_step_form_from_session(1, request).cleaned_data
+    schedule_class = step_1_data['schedule_class']
+
+    if request.method == 'GET':
+
+        gleaned_data = get_deserialized_gleaned_data(request)
+        force_show = request.GET.get('force')
```
jseppi: I couldn't think of a better way to make `step_3` not redirect when `gleaned_data` is present and valid other than to add and check for a query parameter.

jseppi:
Alrighty, new interstitial view, `step/3/data`, is done. Ready for another review, @hbillings and @toolness.

hbillings:
@jseppi I think this is pretty solid for now, but I think I may have just come up with something better. Screenshotting now. Please don't hate me. >_<

jseppi:
image

jseppi:
jk jk

hbillings:
@jseppi :joy: You know what, just merge this and we can iterate.

jseppi:
Haha, I'll take that as a 👍 from @hbillings.

Still need a code review from @toolness, especially given the changes from Friday.

toolness:
This is looking good! The one UX-related confusion I have is that I expect clicking between "previous" and "next" on steps should always take me through the same "path" of pages. But I found a path that breaks this expectation:

1. Go all the way to step 4, uploading a price list at least one valid *and* invalid row.
2. Go back to step 2 (vendor details).
3. Click **next**.  You should see a page that looks something like this:
  screen shot 2016-11-15 at 9 34 36 am
4. Click **next**. You should be at step 4 now.
5. Click **previous**. You should be at step 3 now, but it doesn't look the same as it did in the earlier screenshot. Now it looks like this:
  screen shot 2016-11-15 at 9 34 52 am

Is this intentional, or a bug?

jseppi:
I think that should be considered a bug. So I need to somehow keep track of if the user goes `step/3/data` -> `step/4`, and set the value of the "previous" link on `step/4` to `step/3/data`.

toolness:
Oh SNAP. It took me a while to realize why you would have to do that, but blugh, I guess you are right. Errgg.

I guess one potential solution might be to set something on the querystring, e.g. `/step/4?back_goes_to_errors=yup`.

jseppi:
This will also likely be pretty whacky if/when we have random step access, like through clicking on the Steps Widget. Or actually we kinda already do if the user manually changes the URL.

toolness:
The code here looks 👍 , just lemme know when you've fixed that bug--or we could also fix it separately from this PR if that's easier.

jseppi:
ok, fixed, @toolness. I implemented by changing Step 4 to take an optional `prev` query param that contains the url for what the Previous button should link to.

data_capture/views/price_list_upload.py
```diff
@@ -273,7 +311,10 @@ def step_4(request, step):
 
         return redirect('data_capture:step_5')
 
+    prev_url = request.GET.get('prev')
```
jseppi: Mitigated this by using `is_safe_url`

toolness:
Noice, this looks great!

I just noticed that technically we might want to do the same kind of thing for the URL of the **next** link on step 2, since it's possible to be at `/step/3/errors`, click **previous**, and then click **next** to arrive at `/step/3/data`, but seriously dude... just file a bug for that, if anything.

This PR is 👍 ! :shipit: !

Issue #1021, opened 2016-11-08
Bulk upload data / price lists can contain duplicate rows with different SIN numbers


toolness:
@jseppi just noticed [on Slack](https://gsa-tts.slack.com/archives/calc/p1478616897000083) that the CSV data for region 10 contains rows that are duplicates except for the SIN numbers in them.

[Here's an example on the current production CALC instance.](https://calc.gsa.gov/?sort=current_price&q=Budget%2FPolicy+Analyst&query_type=match_exact&education=BA&min_experience=5&max_experience=5&contract-year=current&experience_range=5%2C5)

We might want to add some logic to bulk / price list upload that notices such duplicate rows and merges their SIN numbers (since the SIN column is messy and some entries already contain e.g. SIN numbers separated by commas, we can just "merge" them through string concatenation).

toolness:
Note that duplicates are now part of our data quality report: https://calc.gsa.gov/data-quality-report/dupes/

One approach to dealing with these is simply to de-duplicate them in the data explorer, which would be particularly helpful in reporting accurate statistics. Alex Petralia's Pycon 2018 talk [What pandas and SQL Taught Me About Taking an Average ](https://www.youtube.com/watch?v=DlgG0QdrqAU) could help in figuring out the proper techniques to use.

PR #1029, opened 2016-11-09
[WIP] [Experiment] Added 'finddupes' management command (take 1)

toolness:
This attempts to help with #1021 by adding  a `manage.py finddupes` command that, as it's currently implemented, presents counts of `Contract` model rows that have identical vendor name, contract number, current price, and labor category, but which differ along other axes. Here's its output for the region 10 data that currently ships with CALC:

```
sin                       2818
contractor_site           770
education_level           244
min_years_experience      120
next_year_price           8
hourly_rate_year5         8
hourly_rate_year2         8
second_year_price         8
hourly_rate_year3         6
hourly_rate_year4         4
```

Note that the counts do overlap, e.g. rows that differ on `hourly_rate_year4` may also differ on `hourly_rate_year5`.

I implemented this using Django APIs so we could eventually reuise it in the site's UI somewhere, allowing us to pre-check incoming pricing data and make it possible to perform "sanity checks" on existing databases.

However, I didn't know about the existence of [Dedupe](https://dedupe.readthedocs.io/en/latest/), which will probably be a better tool to use here, so we should probably either close this PR or evolve it to use that tool!

### DIFF
--- a/data_capture/management/commands/finddupes.py
+++ b/data_capture/management/commands/finddupes.py
@@ -1,0 +1,49 @@
+import djclick as click
+
+from contracts.models import Contract
+
+
+@click.command()
+def command():
+    identical_fields = [
+        'vendor_name',
+        'labor_category',
+        'idv_piid',
+        'current_price',
+    ]
+    contracts = Contract.objects.raw(
+        'SELECT c1.*, '
+        'c2.id AS duplicate_id '
+        'FROM %(table)s as c1, %(table)s as c2 '
+        'WHERE c1.id != c2.id '
+        'AND %(identical_fields)s' % {
+            'table': Contract._meta.db_table,
+            'identical_fields': ' AND '.join([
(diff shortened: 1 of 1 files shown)


toolness:
Regarding how to automatically resolve the conflicts (in the R10 data at least):

* In the case of identical `sin` fields, we can "merge" them, since the SIN field is already a messy string to begin with (e.g. for some it's a comma-separated list of of numbers).
* For `min_years_experience`, I could see us deleting the row containing the larger of the two values.
* For `contractor_site`, we could merge the two by using the "both" value.
* For `education_level` we could interpret it to mean "minimum education" and delete the "larger" of the two values, e.g. deleting a "Ph. D." variant while keeping the "Masters" one.

The ones that involve pricing are super confusing though and might require individual, manual attention. Fortunately there seem to be very few of those.

toolness:
Ok, since we've got a duplicates report available as of #1875 and #1938, I'm going to close this issue.

PR #1043, opened 2016-11-17
Add edit button on step 4 of price list upload

toolness:
This attempts to add an **Edit** button below the vendor and contract details section of step 4 of price list upload.

If the user is in safe mode (or doesn't have JS) they are sent to a version of the page with an edit form on it.  Otherwise, the vendor/contract details slide up and are replaced by a form:

> expandable-area2

If the user submits anything invalid, they are re-shown step 4 with the edit form already displayed and the appropriate error(s) highlighted.

Otherwise, If their changes are valid, the price list is submitted and the user moves to step 5. (We could potentially change this to re-show step 4 without the edit form, so that the user is forced to re-verify that everything looks OK.)

To do:

- [x] Verify with @hbillings that this is actually what she wanted.
- [x] Add tests.
- [x] Fix any broken tests.
- [x] ~~Possibly move the JS out of an inline jQuery-powered `<script>` tag. I could turn it into an actual web component but I'm not sure what the best plan for actually making it reusable is...~~ Mehh, unless someone else feels really strongly about this I am gonna just leave it as-is. See my [rationale](https://github.com/18F/calc/pull/1043#discussion_r88679802) for why I think it's OK to not treat this the way we treat some of our other production JS.
- [x] Make sure that the edit form lists fields in the same order as the summary it's replacing (it doesn't currently).
- [x] Make sure the "previous" button still goes where the user expects it to, even during POST requests with invalid data. Urgh. Or just file it as an issue to be dealt with later.

### DIFF
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -8,6 +8,14 @@ from frontend.widgets import UswdsRadioSelect
 from contracts.models import MIN_ESCALATION_RATE, MAX_ESCALATION_RATE
 
 
+def ensure_contract_start_is_before_end(form, cleaned_data):
+    start = cleaned_data.get('contract_start')
+    end = cleaned_data.get('contract_end')
+    if start and end and start > end:
+        form.add_error('contract_start',
+                       'Start date must be before end date.')
+
+
 class EscalationRateField(forms.FloatField):
     def __init__(self):
         return super().__init__(
@@ -74,11 +82,7 @@ class Step2Form(forms.ModelForm):
 
     def clean(self):
         cleaned_data = super().clean()
-        start = cleaned_data.get('contract_start')
-        end = cleaned_data.get('contract_end')
-        if start and end and start > end:
-            self.add_error('contract_start',
-                           'Start date must be before end date.')
+        ensure_contract_start_is_before_end(self, cleaned_data)
         return cleaned_data
 
     class Meta:
@@ -170,11 +174,7 @@ class Step4Form(forms.ModelForm):
 
     def clean(self):
         cleaned_data = super().clean()
-        start = cleaned_data.get('contract_start')
-        end = cleaned_data.get('contract_end')
-        if start and end and start > end:
-            self.add_error('contract_start',
-                           'Start date must be before end date.')
+        ensure_contract_start_is_before_end(self, cleaned_data)
         return cleaned_data
 
     class Meta:
--- a/data_capture/forms/price_list.py
+++ b/data_capture/forms/price_list.py
@@ -8,6 +8,26 @@ from frontend.widgets import UswdsRadioSelect
 from contracts.models import MIN_ESCALATION_RATE, MAX_ESCALATION_RATE
 
 
+class EscalationRateField(forms.FloatField):
+    def __init__(self):
+        return super().__init__(
+            label="Escalation rate (%)",
+            help_text='CALC uses the escalation rate (as a percentage) '
+                      'to calculate out-year pricing. '
+                      'Leave this field blank or enter 0 if this contract '
+                      'does not have a fixed escalation rate.',
+            min_value=MIN_ESCALATION_RATE,
+            max_value=MAX_ESCALATION_RATE,
+            required=False
+        )
+
+    def clean(self, value):
+        value = super().clean(value)
+        if value is None:
+            value = 0
+        return value
+
+
 class Step1Form(forms.ModelForm):
     schedule = forms.ChoiceField(
         label="Which schedule is associated with this price list?",
@@ -50,16 +70,7 @@ class Step2Form(forms.ModelForm):
     contract_end = SplitDateField(
         label='Contract or current option period end'
     )
-    escalation_rate = forms.FloatField(
-        label="Escalation rate (%)",
-        help_text='CALC uses the escalation rate (as a percentage) '
-                  'to calculate out-year pricing. '
-                  'Leave this field blank or enter 0 if this contract does '
-                  'not have a fixed escalation rate.',
-        min_value=MIN_ESCALATION_RATE,
-        max_value=MAX_ESCALATION_RATE,
-        required=False,
-    )
+    escalation_rate = EscalationRateField()
 
     def clean(self):
         cleaned_data = super().clean()
@@ -69,12 +80,6 @@ class Step2Form(forms.ModelForm):
             self.add_error('contract_start',
                            'Start date must be before end date.')
         return cleaned_data
-
-    def clean_escalation_rate(self):
-        value = self.cleaned_data['escalation_rate']
-        if value is None:
-            value = 0
-        return value
 
     class Meta:
         model = SubmittedPriceList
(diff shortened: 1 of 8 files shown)


hbillings:
@toolness I think this is great for right now! (I do have some thoughts on design, but given that I'm going to be working on bringing the new designs over anyway, it probably doesn't make sense to go too deep in the weeds right now.)

Two thoughts:
Can we make the information follow the same order in both? (I don't really have an opinion as to order, so I say modify whichever is easier.)
Could we have a "save" button replace the "edit" button when the form fields are revealed? That might help with the back/forward issue too...

toolness:
Thanks for the feedback!

> Can we make the information follow the same order in both? (I don't really have an opinion as to order, so I say modify whichever is easier.)

Yep--I guess I recommend making the summary match the order of the form, because the summary is only in one place while the form is in multiple places (on steps 1/2 and in step 4).

> Could we have a "save" button replace the "edit" button when the form fields are revealed? That might help with the back/forward issue too...

Hmm what should happen to the **Submit for review** button at the bottom of the page, though?  If there's *both* a **Save** button *and* a **Submit for review** button, that could be confusing... if we hide the **Submit for review** one, should **Save** send the user back to step 4 with just the summary, so they can review their information one last time, or should it submit the whole shebang for review and move on to step 5?

hbillings:
I was thinking the "Save" button would essentially just write the new info and change the form-field view back to the original non-editable display of information, and then Submit maybe gets changed to "Save & Submit." Users tend to like to be able to trigger save events via buttons and don't always trust the system to save it for them.

data_capture/forms/price_list.py
```diff
@@ -135,3 +144,46 @@ def clean(self):
             cleaned_data['gleaned_data'] = gleaned_data
 
         return cleaned_data
+
+
+class Step4Form(forms.ModelForm):
```
jseppi: In my other modify PR, I extracted these fields to another file for reuse. Is that a good pattern?

ref: https://github.com/18F/calc/pull/985/files#diff-390edec5313f55a84bdd274f687c5817

toolness: I think that's ok, but in this case we're not actually going to reuse these exact fields long-term, as we actually want at least some of them to be displayed in a tabular/inline format rather than stacked... So for now I'm just copy-pasting.

Once I do the next PR I will probably either do what you did, or factor them out into some kind of base field classes or something weird.

data_capture/tests/test_price_list_upload_views.py
```diff
@@ -467,13 +472,46 @@ def test_gleaned_data_with_valid_rows_is_required_on_POST(self):
...
+        self.assertEqual(SubmittedPriceList.objects.all().count(), 0)
+        self.assertRedirects(res, Step4Tests.url)
+
+    def test_invalid_post_shows_errors(self):
+        self.login()
+        session = self.client.session
+        session['data_capture:price_list'] = self.session_data
+        session.save()
+        self.set_fake_gleaned_data(self.rows)
+        data = self.valid_post_data.copy()
```
toolness: We really need to refactor these tests so they don't all share the same first few lines of code, but that feels like something to address in a separate PR, so for now I'm just copy-pasting.

toolness:
Ok, I guess this is ready for review.

data_capture/tests/test_price_list_upload_views.py
jseppi: Yeah, agree

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -70,15 +113,49 @@
...
 
+{% if not is_safe_mode_enabled and not show_edit_form %}
+<script>
+$(function() {
+  var editBtn = $('[data-edit-btn]');
+  var contractDetails = $("[data-contract-details]");
+  var editContractDetails = $("[data-edit-contract-details]");
+
+  if (editBtn.length && contractDetails.length &&
+      editContractDetails.length) {
```
jseppi: Hrm, I don't know if I love that this is inlined in the template. For instance, it won't get linted and, less importantly, it's not ES2016, won't be minified, and will be less discoverable by a future dev.

I think I can guess why you put it in the template though -- it's clearly not necessary on any other pages since it is so specific to this one page. I suppose we don't have a good strategy for adding JS like this. Any thoughts on this situation?

data_capture/templates/data_capture/price_list/step_4.html
```diff
@@ -5,23 +5,57 @@
...
+  </div>
+
+  {% fieldset form.escalation_rate %}
+
+  <div class="form-button-row clearfix">
+    <div class="submit-group">
+      <button type="submit" class="button" name="save-changes">Save changes</button>
+    </div>
+
+    {# TODO: This should include any existing querystring args. #}
```
jseppi: Forgotten TODO?

data_capture/forms/price_list.py
```diff
@@ -135,3 +144,46 @@ def clean(self):
...
+    escalation_rate = EscalationRateField()
+
+    def clean(self):
+        cleaned_data = super().clean()
+        ensure_contract_start_is_before_end(self, cleaned_data)
+        return cleaned_data
+
+    class Meta:
+        model = SubmittedPriceList
+        fields = Step1Form.Meta.fields + Step2Form.Meta.fields
```
jseppi: heh, that's cool

data_capture/forms/price_list.py
```diff
@@ -135,3 +144,46 @@ def clean(self):
...
+
+    @classmethod
+    def from_post_data_subsets(cls, *post_datas):
```
jseppi: Small dosctring for this would be nice. I could easily see forgetting what this method is for.

data_capture/templates/data_capture/price_list/step_4.html
toolness: Yep, that's exactly why!  Bloop.  One alternative might be to just refactor the gulpfile so that it's easier to add more JS bundles to it--right now it's a bit cumbersome, as one has to add a bunch of entries to a bunch of different data structures, but I think it should be possible to refactor it so that it follows more of a "convention over configuration" philosophy. Then we can just include the `data-capture-step-4.min.js` bundle or whatever on the page.

That said, I think it might be better to add that in a separate PR, since the refactoring of the gulpfile would likely add a lot to this one?

data_capture/views/price_list_upload.py
```diff
@@ -282,41 +308,76 @@ def step_4(request, step):
         'escalation_rate': step_2_form.cleaned_data['escalation_rate'],
     }
 
+    show_edit_form = (request.GET.get('show_edit_form') == 'on')
```
toolness: Ah yeah, that makes sense. I decided to use `on` here just to follow the HTTP convention for checkboxes, since submitting a form with e.g. `<input type="checkbox" name="show_edit_form">` checked will result in `show_edit_form=on` in the querystring.  Aside from the conceptual consistency, I like it because it means we can easily make a form that results in the view we want, if we ever want to do that sort of thing.

So anyhow, that's why I'd argue for breaking our environment variable boolean convention and instead adhere to HTTP checkbox conventions.  I guess my preference in this case would be to rename the variable to `edit_form=on` and then change step 3 to explicitly check for `force=on`--but ultimately I'm fine either way!

jseppi: Ah cool, I actually didn't realize checked checkboxes submitted with `"on"`. In that case, I agree we should change the step 3 `force` to check for `"on"`. You want to do that here, or file an issue? Either way fine with me.

toolness: Cool, how about we file a separate issue?

jseppi: Filed https://github.com/18F/calc/issues/1061

toolness:
Ok, I've split out the script into its own file, so this should be ready to go @jseppi !

jseppi:
Looks 👍 , just got a minor eslint issue that Code Climate found.

PR #1046, opened 2016-11-17
Properly pluralize generic form error message

toolness:
The `Oops! Please correct the following error.` message wasn't properly pluralizing the word "error" before, even though it had some code that was trying to do so.  This should fix everything.


jseppi:
whoops!
👍

PR #1047, opened 2016-11-18
make ethnio id configurable by env var

### LINKED ISSUE #1030: Consider making it possible to disable ethnio during development
Somewhat related to #282, I realized now that we actually have a real ethnio screener, it shows up during development:

It's not that big a deal, but it can be a bit distracting during development. I think that Trammell might even be making it more eye-catching so we get more folks to opt-in, so it might be useful to have a way to disable it during development. (I thought that perhaps it'd remember that I closed it, but it doesn't seem to during development.)

jseppi:
Adds another env var setting, `ETHNIO_SCREENER_ID`, that when set will cause the ethnio javascript for the given ID to be included in CALC pages.

Closes #1030


toolness:
Looks good to me--just need to remember to set the cloud foundry UPS thingy as needed whenever we next deploy so that we don't lose our current ethnio script!

jseppi:
Sweet, updated `calc-prod`'s `calc-env` with the ethnio id so that it will be good to go when we switch over.

PR #1049, opened 2016-11-18
Add a selenium test for data capture.

toolness:
This adds a single selenium test for data capture, which goes through the whole data capture workflow by uploading the example CSV for the fake schedule. It shouldn't run into any weird race conditions because it's doing everything in compatibility mode, so all the progressive enhancement is disabled.

To use it:

1. Run `python manage.py test data_capture.tests.test_selenium`.
2. Once you've run that once, you can add `SKIP_STATIC_ASSET_BUILDING=yup` to your `.env`, which will speed things up a lot if you're just tinkering with stuff on the python side.

### DIFF
--- a/data_capture/tests/test_selenium.py
+++ b/data_capture/tests/test_selenium.py
@@ -48,7 +48,7 @@ urlpatterns += [
 class DataCaptureTests(SeleniumTestCase):
     screenshot_filename = 'data_capture.png'
 
-    def test_awesome(self):
+    def test_data_capture(self):
         registry._init()
         call_command('initgroups', stdout=io.StringIO())
         t = BaseLoginTestCase()
@@ -65,4 +65,22 @@ class DataCaptureTests(SeleniumTestCase):
             .send_keys('GS-123-4567')
         e = self.driver.find_element_by_name('vendor_name')
         e.send_keys('Battaglia Sausage Peddlers, Inc.')
+        e.submit()
+
+        self.driver.find_element_by_css_selector(
+            '[for="id_is_small_business_0"]'
+        ).click()
+
+        self.driver.find_element_by_css_selector(
+            '[for="id_contractor_site_2"]'
+        ).click()
+
+        self.driver.find_element_by_name('contract_start_1').send_keys('04')
+        self.driver.find_element_by_name('contract_start_2').send_keys('28')
+        self.driver.find_element_by_name('contract_start_0').send_keys('2016')
+
+        self.driver.find_element_by_name('contract_end_1').send_keys('03')
+        self.driver.find_element_by_name('contract_end_2').send_keys('27')
+        e = self.driver.find_element_by_name('contract_end_0')
+        e.send_keys('2017')
         e.submit()
--- a/data_capture/tests/test_selenium.py
+++ b/data_capture/tests/test_selenium.py
@@ -1,0 +1,68 @@
+import io
+import django.contrib.auth
+from django.conf.urls import url
+from django.test import override_settings
+from django.http import HttpResponse
+from django.core.management import call_command
+
+from hourglass.urls import urlpatterns
+from hourglass.tests.common import BaseLoginTestCase
+from data_capture.tests.common import FAKE_SCHEDULE
+from data_capture.schedules import registry
+from frontend.tests.test_selenium import SeleniumTestCase
+from frontend import safe_mode
+
+
+USERNAME = 'boop'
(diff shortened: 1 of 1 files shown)


toolness:
Huh, so this actually runs pretty fast locally via phantom--about 4 seconds--but it takes 40 seconds on sauce, which is a bummer. But it does output a nice video:

https://saucelabs.com/beta/tests/d11b632eb19f4865b5a17043ca0f7a70/watch

@jseppi what do you think? Should we commit it but not run it on travis, or commit it and run it on travis, or not commit it at all?

data_capture/tests/test_selenium.py
```diff
@@ -0,0 +1,111 @@
...
+        self.driver.find_element_by_name('contract_end_2').send_keys('27')
+        e = self.driver.find_element_by_name('contract_end_0')
+        e.send_keys('2017')
+        e.submit()
+
+        e = self.driver.find_element_by_name('file')
+        e.send_keys(FAKE_SCHEDULE_EXAMPLE_PATH)
+        e.submit()
+
+        self.driver.find_element_by_css_selector('.button-primary').submit()
```
toolness: A lot of the element-finding code here isn't very readable, and is potentially quite brittle... We could probably make it more readable in the future by adding some helper JS on the browser side, maybe. I don't want to do too much back-and-forth between the test suite and the WebDriver server, though, as I'm concerned it could make sauce take a *really* long time to run these.

data_capture/tests/test_selenium.py
```diff
@@ -0,0 +1,111 @@
...
+        raise Exception('Unable to authenticate')
+    django.contrib.auth.login(request, user)
+    request.session[safe_mode.SESSION_KEY] = True
+    return HttpResponse('{} is now logged in'.format(
+        user.email
+    ))
+
+
+urlpatterns += [
+    url(r'^autologin/$', autologin),
```
toolness: Making an auto-login endpoint available only on the test suite seems to be the only really easy way to manipulate the session in a `LiveServerTestCase`, at least that I can find.

toolness:
So I'm working on an alternative to this in #1052.  That other PR actually builds upon this one though, so it's OK if you still review this one @jseppi.

data_capture/tests/test_selenium.py
jseppi: Seems relatively understandable to me. I agree that it is brittle, but maybe that's good?

jseppi:
I certainly like and see the value in having a full data capture flow test, but the amount of time you say it takes on selenium is 😦 

I think the code looks fine, though I am not 100% clear on why the `autologin` endpoint is needed. Is it just that `LiveServerTestCase` does not provide access to `self.client.session`?

toolness:
Yep, that's exactly it--there's no `self.client` in `LiveServerTestCase`s because their whole point is to use an external HTTP client (like selenium) rather than Django's built-in one.  Which means that we'd have to either create a custom session and somehow hand its cookie over to Selenium, or get the cookie from Selenium and modify it on our end, or something weird like that.  In the end it just seemed easier to add a custom view that sets up the session and have Selenium ping that.

Regarding the amount of time it takes on sauce labs, I agree--I think #1052 can help with this, as it will allow us to just use `RoboBrowser` to vastly speed things up if we want.  I wish we could do something where we do things the "fast" way on commits but still have the "slow" version run on selenium at midnight every day, or somesuch.

toolness:
Oh, it looks like Travis supports [cron jobs](https://docs.travis-ci.com/user/cron-jobs/)!  We could have a daily cron job that runs all the selenium tests--even the ones that consistently work in the data explorer suite--on IE11, and then either not run them or run them in a deprecated mode (e.g. RoboBrowser) during commits.

jseppi:
whoa, yesssss!

jseppi:
Interesting about `LiveServerTestCase`. I've not used it before. I guess another option would be to override our auth mechanism completely for that test, right? Or maybe replace it with a `MagicMock`? Not necessarily advocating for that, I don't have an informed opinion either way.

toolness:
Hmm, I'm not sure how we'd do that... We'd also have to figure out how to forcibly enable safe mode so that no JS runs in the client browser to cause weird race conditions and potentially mess up test results (it appears there's no easy way to tell webdriver to disable JS on the target browser, so it's easiest to just enable safe mode, which we're doing in that `/autologin` view too).

toolness:
I guess I am gonna close this in favour of #1052.

PR #1051, opened 2016-11-18
[WIP] Reactify Data Explorer

jseppi:


jseppi:
So far this is a mess of getting `eslint` and the `airbnb` config working properly. `ultratest`'s `eslint` is ✅, but apparently codeclimate doesn't like some of the new stuff (the dangling commas specifically, which I also don't like). Going to try to figure out wtf those are for.

jseppi:
Airbnb's reasoning for the dangling commas are for cleaner diffs: https://github.com/airbnb/javascript#commas--dangling. I don't like them and they are causing problems in `gulpfile.js` and in codeclimate. Bascially you can only use them in an environment that supports ES2017 or that has gone through a pre-piler.

jseppi:
woot, codeclimate is happy again.

toolness:
@jseppi should this be closed since you decided that it'd be super hard to reactify?  or at least, even if we do reactify it, the data explorer code has changed a lot since #1053 and #1067 ...

jseppi:
Yeah, closing!

PR #1052, opened 2016-11-21
Add selenium-independent data capture integration test

toolness:
This is an alternative to #1049.

So I've had it with selenium / sauce labs.  Either it randomly fails or it's *really* slow or some other weird problem. But I'm not ready to abandon it *completely*.

One good thing is that because the data capture app is progressively enhanced, we don't really *need* to run JS in our integration tests.  In fact, all we really need is something that pokes at our app and acts like a browser at the HTTP level--specifically, we want to be able to "fill out" forms that we're presented with, which Django's test client can't do out of the box. Python's [mechanize](https://pypi.python.org/pypi/mechanize) does this, but it hasn't been updated in ages and might not even work in Python 3; through Stack Overflow I found a newer alternative by our own Josh Carp called [RoboBrowser](https://github.com/jmcarp/robobrowser).

But like I said, I'm queasy about abandoning Selenium *completely*; it still has lots of promise, if only it'd actually work consistently, which might still happen down the road, if we cross our fingers really hard.

So I thought it might be neat to have a layer of abstraction that lets us use *either* Selenium or RoboBrowser to drive our client. By default, we use RoboBrowser, because it's really fast, works consistently, doesn't require us to build our static assets, and allows us to test the very basics of integration without having to rely on a real browser that might explode in our face. But if we explicitly specify a browser via the `WD_TESTING_BROWSER` environment variable, the suite will instead use selenium as the client.

To do:

- [x] Decide if this is a dumb idea. Right now our abstraction layer has only a handful of methods that need to be implemented by the selenium/RoboBrowser-specific backends; but if our integration testing needs expand, we might have to add more methods, which could lead to an annoying amount of work if we need to always support both backends.

- [x] Move the `pip install` of RoboBrowser from the `Dockerfile` to `requirements-dev.txt`.

- [x] Add more documentation for the whole `data_capture.tests.browsers` package.

- [x] Consider using the [`abc`](https://docs.python.org/3/library/abc.html) module for the abstract base class(es) in this PR. I have no idea what the benefit of that is so I'll have to read up on it.

- [x] Document how `data_capture.tests.browsers` interprets the ~~`WD_TESTING_BROWSER`~~ `TEST_WITH_ROBOBROWSER` environment variable.


toolness:
@jseppi I think this is ready for you to take a look at. If the general approach looks OK to you, I can work on making Travis run the RoboBrowser version on normal travis builds, while running the sauce labs / selenium version on nightly cron-based travis builds.

data_capture/tests/browsers/base.py
```diff
@@ -0,0 +1,98 @@
...
+        '''
+        Given a text input's name, set its value.
+        '''
+
+        raise NotImplementedError()
+
+    @abc.abstractmethod
+    def set_radio(self, name, number):
+        '''
+        Set a radio button's value, given a its name and its order in the
```
jseppi: minor typo: `a its`

jseppi:
I definitely like the `RoboBrowser` approach and think that having this integration could catch some potential bugs. The code you have here so far is nice and understandable 👍 .

I just read over the Travis [cron job support](https://docs.travis-ci.com/user/cron-jobs/). Looks like you'll have to email them to add it.

toolness:
Cool!  For now how about I just make this use RoboBrowser on Travis so that we can merge it sooner rather than later, and I file a separate issue to make it cron-jobbable?

jseppi:
Yeah, sounds like a plan!

toolness:
@jseppi ok how does 9b9292be26d46a191bf00a9fe83873db3c5c2515 look?

jseppi:
👍

PR #1053, opened 2016-11-22
Lint (and minor refactoring) explorer.js

jseppi:
To Do:
- [x] fix problem with restoring filters, excluded rows, sort order, etc.
- [x] verify it all works


jseppi:
Well, I got it linted as best as possible. I had to do a few `eslint-disable-line` comments, mostly for `no-param-reassign` because of d3 stuff.

The diff is nearly impossible to read due to reordering. I was able to remove some dead code, but largely avoided actual logic refactoring so far.

I've done a lot of manual checking to make sure everything works, and it seems like all functionality is the same as before.

toolness:
Nice! Just tested it out on IE11 and Edge and it seems to work fine on them too.

I haven't read the source code since the diff is indeed nearly impossible to read.  So if you're ok with that, feel free to merge!

toolness:
(I figure if anything really weird happens, it will be reported via New Relic and/or users and/or eventually noticed by us.)

PR #1054, opened 2016-11-22
Refactor gulpfile to make it easier to add new bundles

toolness:
This reduces repetition in our gulpfile to make it easier to add new JS bundles for individual pages as we see fit, as per https://github.com/18F/calc/pull/1043#discussion_r89008703.

I'm not super happy with the way I refactored this but I think maybe we can refactor it more in the future to make it better. Right now it has been refactored to be sort of grunt-ish, in that `bundles` is now a configuration object that we iterate through to define actual tasks and stuff.  Hrm.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -161,7 +161,7 @@ gulp.task('sass', () => gulp.src(path.join(dirs.src.style, paths.sass))
 );
 
 // Compile and lint JavaScript sources
-gulp.task('js', ['lint'].concat(browserifiedBundles).concat(['js:legacy']));
+gulp.task('js', ['lint', 'js:legacy'].concat(browserifiedBundles));
 
 gulp.task('js:legacy', vendoredBundles);
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,6 +25,8 @@ const watchify = require('watchify');
 const babelify = require('babelify');
 const del = require('del');
 
+const BUILT_FRONTEND_DIR = 'frontend/static/frontend/built';
+
 const dirs = {
   src: {
     style: 'frontend/source/sass/',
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -83,8 +70,55 @@ const bundles = {
...
+        dirs.dest.scripts[name],
+        paths[outfileName]
+      )
+    );
+    browserifiedBundles.push(browserifiedBundleName);
+  }
+
+  if (vendor.length) {
+    const vendoredBundleName = `js:${dirName}:vendor`;
+    gulp.task(vendoredBundleName, () => concatAndMapSources(  // eslint-disable-line
```
toolness: I've added these `// eslint-disable-line`s because I didn't want to move the code for the functions they're calling and thereby make the diff harder to review.  I figure I can move the functions after this code has been 👍 'd, either as a commit before merging the PR, or as a separate PR.

toolness:
Oh, another thing we should probably file an issue for: I wonder if we should just move the babel ES2015 polyfills into the common vendor bundle that's included on every page?  That way for per-page and future bundles we don't need to worry about whether we've included the polyfills on each individual bundle's `index.js`, and we don't have to worry about minimizing the number of repetitions of the polyfill across bundles or other weird stuff.

gulpfile.js
```diff
@@ -127,11 +161,9 @@ gulp.task('sass', () => gulp.src(path.join(dirs.src.style, paths.sass))
 );
 
 // Compile and lint JavaScript sources
-gulp.task('js', ['lint', 'js:data-explorer',
-  'js:data-capture', 'js:styleguide',
-  'js:tests', 'js:legacy']);
+gulp.task('js', ['lint'].concat(browserifiedBundles).concat(['js:legacy']));
```
toolness: The two `concat`s here look weird but I wanted to preserve the ordering of the original dependency list just in case it mattered. If it doesn't actually matter, we can smoosh this down to just one `concat`.

jseppi: I think just that `'lint'` should come first, the order of the others doesn't appear to matter.

toolness:
@jseppi aside from the `// eslint-disable-line`s does this look OK?

gulpfile.js
```diff
@@ -83,8 +70,55 @@ const bundles = {
...
+  tests: {},
 };
 
+const browserifiedBundles = [];
+const vendoredBundles = [];
+
+Object.keys(bundles).forEach(name => {
+  const options = bundles[name];
+  const dirName = options.dirName || name;
+  const noBrowserify = options.noBrowserify;
```
jseppi: maybe use `!!options.noBrowserify` to coerce to an actual boolean. not a big deal really.

jseppi:
yeah, looks good! 👍

PR #1055, opened 2016-11-22
[WIP] Data analysis

EricSchles:
operationalization of data analysis tool.  Two commands were created during this PR:

load_lookup_data which is run by - `python manage.py load_lookup_data`
generate_arima_model which is run by - `python manage.py generate_arima_model`


jseppi:
oooweeee looks like you got some fun `flake8` issues to fix 😺

jseppi:
Added "[WIP]" (work-in-progress) to the PR name, which has been our practice so it's easy to scan the list to see PRs that need to be reviewed.

EricSchles:
👍 I'll work on the flake8 issues

toolness:
Nice! Looking through this now.

Hmm, when I run `load_lookup_data` the following warning is logged:

```
/usr/local/lib/python3.4/site-packages/django/core/management/__init__.py:346: DtypeWarning: Columns (2,3,4,5,17) have mixed types. Specify dtype option on import or set low_memory=False.
  self.fetch_command(subcommand).run_from_argv(self.argv)
```

The command is also taking a while to run--is there any easy progress reporting you could add?  As well as a message that everything completed successfully?

Oh, I just noticed you're using the logger to log info... hrm, Django logging is hard to configure on a per-module basis, and since there's not really a danger of spamming any logs in the commands, could you use `self.stdout.write()` instead of the logger?  This is what most built-in Django commands (e.g. [`loaddata`](https://github.com/django/django/blob/master/django/core/management/commands/loaddata.py)) use anyhow so it seems to be a best practice. If there are any logging messages that do seem spammy, you can always guard those with an `if options['verbosity'] >= 2`.

toolness:
Oops, I just realized this is still WIP lol. Sorry if my earlier comments were out of line!

toolness:
Hmm, I noticed there are some files in `data_analysis/` that don't seem to be required for production use, like emacs auto-backup files ending with `~`. That's not a very big deal since those files don't bloat the repo much and are easy to remove.

However, I'm also noticing a few larger files that bloat the repo a bit since they're larger than a megabyte:

```
-rw-r--r--  1 atulvarma  staff  1055752 Nov 25 11:26 categories.json
-rw-r--r--  1 atulvarma  staff  1943160 Nov 25 11:26 labor_category_to_high_level_category.json
-rw-r--r--  1 atulvarma  staff  5129829 Nov 25 11:26 list_of_labor_categories.json
```

Are these files required for the data analysis to work, or were they intermediate products of research that aren't needed anymore?

.env~
```diff
@@ -0,0 +1,4 @@
+DEBUG=yup
+DATABASE_URL=postgres://localhost/hourglass
+DOCKER_EXPOSED_PORT=8000
+EMAIL_URL=dummy
```
toolness: You might want to add a [global `.gitignore`](https://help.github.com/articles/ignoring-files/) to ignore files ending with `~`.

toolness:
@EricSchles do you think a more accurate title for this PR would be "Time series analysis"?  I just wanted to make sure it's easy to disambiguate between this one and #997, which adds experimental price list analysis...

PR #1057, opened 2016-11-23
Support fake filenames in upload widget

toolness:
As per yesterday's coworking discussion, this PR provides functionality that allows a progressively-enhanced upload widget to *appear* to be "pre-populated" on page load, with a fallback for the baseline experience:

> screen shot 2016-11-23 at 7 47 07 am

Note that this PR does *not* actually use this functionality anywhere yet (other than the styleguide). That comes in a later PR!

To do:

- [x] Ensure w/ the rest of the team that this is acceptable behavior.
- [x] Get rid of the `eslint-disable-line` in the code.
- [x] Maybe add a JS-side unit test.
- [x] Maybe add a python-side unit test.
- [x] Since there's more copy in the baseline version of the widget, if the user's screen is narrow or the filename is really long, the content can overflow the widget, which currently has a fixed height of 6em. We might want ameliorate the CSS.

### DIFF
--- a/frontend/source/js/data-capture/upload.js
+++ b/frontend/source/js/data-capture/upload.js
@@ -157,23 +157,6 @@ class UploadWidget extends window.HTMLElement {
     let dragCounter = 0;
     let lastDragEnterTarget = null;
 
-    const finishInitialization = () => {
-      if (this.uploadInput instanceof UploadInput) {
-        if (!this.isDegraded) {
-          this.uploadInput.upgrade();
-
-          const fakeInitialFilename = $el.attr('data-fake-initial-filename');
-
-          if (fakeInitialFilename) {
-            setCurrentFilename(fakeInitialFilename);  // eslint-disable-line
-          }
-        }
-        dispatchBubbly($el[0], 'uploadwidgetready');
-      } else {
-        $el.one('uploadinputready', finishInitialization);
-      }
-    };
-
     function setCurrentFilename(filename) {
       $('input', $el).nextAll().remove();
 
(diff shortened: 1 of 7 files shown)


frontend/source/js/data-capture/upload.js
```diff
@@ -161,6 +161,12 @@ class UploadWidget extends window.HTMLElement {
       if (this.uploadInput instanceof UploadInput) {
         if (!this.isDegraded) {
           this.uploadInput.upgrade();
+
+          const fakeInitialFilename = $el.attr('data-fake-initial-filename');
+
+          if (fakeInitialFilename) {
+            setCurrentFilename(fakeInitialFilename);  // eslint-disable-line
```
toolness: Blergh I don't want to increase the difficulty of reading this PR by moving `setCurrentFilename` above this code so I'm just disabling it for now and will do the move later.

@jseppi do you have an opinion on [`no-use-before-define`](http://eslint.org/docs/rules/no-use-before-define)?  I'm kind of inclined to just disable the rule globally but if you like it we can keep it... Blegh.

jseppi: I would say I have very little opinion on a lot of the eslint rules, but the point of standardizing on a specific config is to not have to think about them (especially across the org) :)

I also don't find moving code around to be necessarily that distracting when reviewing a PR.

hbillings:
yaaaaaaaaay!!

I'm going to echo what James said in Slack for the record (and because I think he's right, as usual) and say that the filename on the non-enhanced version should be more prominent. I'd go with something above the input that says, "You've already uploaded `boop.csv`. You can upload a new file below if needed." (Or something along those lines.)

toolness:
@jseppi ok I think this is ready for review!

frontend/upload.py
```diff
@@ -34,9 +37,30 @@ def render(self, name, value, attrs=None):
...
         id_for_label = final_attrs.get('id', '')
+        widget_attrs = {}
+
+        if self.degraded:
+            widget_attrs['data-force-degradation'] = ''
+
+        nojs_preamble = ''
+        if self.existing_filename:
+            if 'required' in final_attrs:
+                raise AssertionError(
```
jseppi: I guess I could maybe see a use case when we have both `required` and want to show the previously selected filename, but that also wouldn't work well with the hardcoded message in `nojs_preamble`.  We can cross that bridge if/when we come to it though.

frontend/upload.py
```diff
@@ -34,9 +37,30 @@ def render(self, name, value, attrs=None):
...
+                'You\'ve already uploaded '
+                '<code>{}</code>. '.format(
+                    escape(self.existing_filename)
+                ) +
+                'You can keep using it or select a new file to replace it.'
+            )
 
         return "\n".join([
-            '<upload-widget%s>' % degraded_str,
+            '<upload-widget%s>' % flatatt(widget_attrs),
```
jseppi: Didn't know about `flatatt` -- looks handy!

jseppi:
👍 looks great!

toolness:
woooooooooooo

Issue #1059, opened 2016-11-23
Add CSP Headers


jseppi:
As an additional layer of security against XSS attacks, we may want to use content security policy headers.

This looks like a promising library for adding CSP headers in Django: https://github.com/mozilla/django-csp

toolness:
Yup I've used that library before, it is nice!

toolness:
Note that we did add CSP for the data explorer *only* in #1063, but I guess we should leave this issue open until we've actually added CSP to the rest of the site.

PR #1060, opened 2016-11-23
Sanitize data explorer query string when parsed

### LINKED ISSUE #1058: Search box doesn't sanitize user input properly
See below: I "searched" for a script tag, and the JS was executed. This is a potential XSS vulnerability.

screen shot 2016-11-23 at 10 37 56 am

jseppi:
This PR enhances `hourglass.qs.parse` to strip `<` and `>` characters from query string values so that HTML entities (like script tags) won't get to the rest of the front end code, where they could potentially be injected into the DOM.

fixes #1058


jseppi:
@toolness: I think this is good now.

toolness:
Hmm, so because we're using this to interpret the value of the search query at a very early stage, does it mean that the user can't actually search for `<` and `>` characters as part of their query?  That is, won't the `&gt;`'s be ultimately passed to APIs and such?  I think this is OK, since it doesn't actually make sense to search for labor categories that include those characters, it still feels a bit wonky.

I feel like a cleaner way to fix this would be to use proper HTML escaping at the point that we present it to the user, just like we do with all our code in the data capture app...  But now that I'm reading the source of `explorer.js`, it seems there are lots of places we're passing potentially unsanitized values to jQuery's `.html()`, so that could take a lot of work. 😣 

Is it urgent to fix this ASAP for pentesting?  If so, maybe we should merge this PR for now but file a separate issue about using a more rigorous escaping strategy later on?

jseppi:
Yeah, I was originally going to try where it was injected via `.html()`, but as you also found, that happens in a whole lot of places. And there are other unsafe jQuery methods that would need to be checked (like `.append()`).

Even with this PR, the user _can_ still search with `<` and `>` in the query. They would, however, be protected if they click on a link that has those characters (and say, an embedded script tag) in the `q` parameter.

This didn't come up in pentesting (yet), but yeah, it seems like getting some kind of a fix in the near-term would be good.

toolness:
Urg, yeah, maybe fixing #1059 sooner rather than later really would help here... Aarggh.

frontend/source/js/data-explorer/explorer.js
```diff
@@ -52,7 +52,8 @@ const hourglass = require('../common/hourglass');
     name = name.replace(/[\[]/, "\\[").replace(/[\]]/, "\\]");
     var regex = new RegExp("[\\?&]" + name + "=([^&#]*)"),
       results = regex.exec(location.search);
-    return results === null ? "" : decodeURIComponent(results[1].replace(/\+/g, " "));
+    return results === null ? ""
+      : decodeURIComponent(results[1].replace(/\+/g, " ")).replace(/[<>]/g, '');
```
toolness: Maybe add a comment about why you're doing the replacement here, like you did in `hourglass.js`?

toolness: Also, I wonder if this function should be pulled out into `hourglass.js` so it can be unit tested? bleh might not even be worth it if we're just gonna rewrite this in react I guess.

jseppi: yeah, I mean, like very little is documented in `explorer.js` and very little (none?) of it is unittested :(

toolness:
OK FINE SHIPIT

PR #1063, opened 2016-11-25
Add nonce-based CSP to index page only

toolness:
This is a quick fix that adds some basic defense in-depth to mitigate any latent XSS vulnerabilities in the data explorer index page *only* (e.g. #1060).  It's a very broad CSP rule that basically allows anything except scripts that we "bless"--specifically, ones that originate from us/DAP/ethnio/GA and inline scripts that we've added a `nonce` attribute to.

The CSP doesn't actually do anything particularly helpful on IE or Edge or any [other](http://caniuse.com/#feat=contentsecuritypolicy2) browsers that only support CSP version 1, because it takes advantage of the [nonce-source](https://blog.mozilla.org/security/2014/10/04/csp-for-the-web-we-have/) feature of CSP 2 to make it easy for us to quickly add CSP without having to un-inline all our scripts.

Note that Firefox applies CSP directives to script elements injected by add-ons, so it might prevent some add-ons from working, and this could result in spurious error messages in the console. You can see these script elements by running the following in the web console:

```js
document.querySelectorAll('script:not([src]):not([nonce])');
```

To do:

- [x] ~~I think that this might mess up the Django Debug Toolbar (on the index page only) because it might have some inline scripts in it. Need to check.~~ Nope, DjDT doesn't include any inline scripts.

### DIFF
--- a/data_explorer/templates/base.html
+++ b/data_explorer/templates/base.html
@@ -11,7 +11,7 @@
     <script src="{% static 'frontend/js/vendor/history.min.js' %}"></script>
     <![endif]-->
 
-    <script>
+    <script {{ csp_nonce }}>
         (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
         (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
         m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
@@ -32,7 +32,7 @@
     <link rel="stylesheet" href="{% static 'frontend/built/style/main.min.css' %}"/>
 
     <![if gt IE 8]>
-    <script>
+    <script {{ csp_nonce }}>
       var API_HOST = "{{ API_HOST }}";
     </script>
 
@@ -117,7 +117,7 @@
       {% endif %}
 
       <![if gt IE 8]>
-      <script>
+      <script {{ csp_nonce }}>
         if (typeof aight === 'object') {
           d3.select('body')
             .classed('ie', true)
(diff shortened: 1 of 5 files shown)


data_explorer/templates/base.html
```diff
@@ -11,7 +11,7 @@
     <script src="{% static 'frontend/js/vendor/history.min.js' %}"></script>
     <![endif]-->
 
-    <script>
+    <script {{ csp_nonce }}>
```
toolness: Note that this will just show up as `<script >` in all pages except the index, since `csp_nonce` is only defined on our index page.  (We can add CSP support for the whole site later.)

data_explorer/views.py
```diff
@@ -1,7 +1,35 @@
...
 
 
 def about(request):
     return render(request, 'about.html', {
         'current_selected_tab': 'about'
     })
+
+
+def index(request):
+    csp_nonce = get_random_string(length=10)
```
toolness: Not sure how processor-intensive it'll be to generate one of these on each request. Hopefully not very!

toolness:
Ok, assuming this strategy doesn't seem horrible as a quick fix, it's ready for review @jseppi!

data_explorer/views.py
```diff
@@ -1,7 +1,35 @@
...
+    script_src = ' '.join([
+        "'self'",
+        "*.googleapis.com",
+        "dap.digitalgov.gov",
+        "www.google-analytics.com",
+        "ethn.io",
+        # Browsers that don't support CSP v2 need this to work.
+        "'unsafe-inline'",
+        # For browsers that *do* support CSP v2, the following will
+        # override our earlier 'unsafe-inline' directive.
```
jseppi: 👍

jseppi:
👍 Sweet, I think this looks good!

toolness:
woooot

PR #1066, opened 2016-11-28
Replace Skeleton with Neat

hbillings:
This removes the Skeleton dependency and substitutes [Neat](http://neat.bourbon.io/). The grid default changed to 50px cols with 30px gutters (relatively defined; this is at max width). Max width of containers is now 930px. Typography and some other elements have been redone in line with the new styleguide as needed to prevent removal of Skeleton from causing visual oddities.


jseppi:
Non-logged-in About page "tab" doesn't meet the page body:

screen shot 2016-11-30 at 4 24 54 pm

Admin page "tabs" are still rounded:

screen shot 2016-11-30 at 4 25 14 pm

margin-top on `.alert` `h3` seems a bit big:

screen shot 2016-11-30 at 4 31 27 pm

The main site header is quite a lot bigger than it used to be, I'm sure there are lots of factors making this the case and you probably already know about it:

screen shot 2016-11-30 at 4 41 01 pm

Region 10 Upload "No, Keep the existing data" button looks funky:

screen shot 2016-11-30 at 5 09 49 pm

The two columns of text in the left of the footer seem like they align less well with the logos than they previously did:

screen shot 2016-11-30 at 5 10 59 pm

hbillings:
Oh man, thanks for the detailed review!

The header should be more manageable once we switch away from buttons/tabs
altogether, like the new designs call for. That's part of the PR I'm
working on now. (Similar issue with the footer.)

My computer just died, so I'll get fixed for the rest in as soon as I can
charge up!


-- 
Heather Battaglia
Innovation Specialist, 18F
Chicago, IL

jseppi:
Couple more smallish things -

On admin list view (eg `/admin/data_capture/newpricelist/`), button looks embossed:

screen shot 2016-12-01 at 9 18 16 am

Same view, checkboxes look slightly misaligned:

screen shot 2016-12-01 at 9 19 08 am

hbillings:
@jseppi @toolness This look gtg?

jseppi:
I think this looks good! I noticed one fairly minor thing, which is that the site banner in the admin views is _slightly_ larger than in other views. I think it might be the bottom margin or padding on the logged-in user's name:

screen shot 2016-12-05 at 4 44 16 pm

screen shot 2016-12-05 at 4 44 22 pm

Feel free to either not worry about this or handle it in a separate PR.

jseppi:
Oh also, we might want to consider installing `neat` with `npm` as described at https://github.com/thoughtbot/neat#installing-with-npm-and-using-a-node-based-asset-pipeline

It should hopefully make it a bit easier to upgrade if necessary, and it keeps unnecessary stuff out of our repo.

jseppi:
Went ahead and opened a PR for using `npm`'d `neat` against this branch: #1081

hbillings:
@seppi I'm implementing some changes to the header in #1078, so I'll fix it over there! Thanks for the `npm` PR!

PR #1067, opened 2016-11-28
Modularize explorer.js into separate files

toolness:
**Note:** I'm not sure how useful this is, but it helped me get my head around the code a little bit. It's OK if we decide that this shouldn't be merged.

This builds upon #1053 by factoring out a few new files from `explorer.js`:

* `autocomplete.js` encapsulates the **Type a labor category** autocomplete field functionality.
* `histogram-to-img.js` encapsulates the **Download graph** functionality.
* `histogram.js` encapsulates the d3-based histogram rendering.
* `table.js` encapsulates the results table with its sortable columns.
* `util.js` contains a bunch of miscellaneous utility functions.

After this refactoring, `explorer.js` weighs in at just under 500 lines of code, which feels less bloaty.

The main confidence I have that this refactoring actually works comes from the linting made possible by #1053--in fact, the linting was invaluable in the process, as I basically refactored by cutting and pasting code snippets into other files and seeing what undeclared globals the linter complained about. But that said, if we decide to go with this refactoring, we should definitely run the code through its paces on a variety of browsers (especially IE11, since some of the JS only seems to run on it) before merging.

Also, the new modules aren't *necessarily* all that modular because many of them still use jQuery and pass in global selectors and stuff.  I'm hoping that they don't store stuff in the DOM that other modules retrieve, but who knows.  At least this refactoring shrinks the lexical scope of a bunch of the code, if nothing else...


frontend/source/js/data-explorer/explorer.js
```diff
@@ -708,12 +155,12 @@ function update(error, res) {
...
   if (res && res.results && res.results.length) {
     updatePriceHistogram(res);
-    updateResults(res);
+    updateResults(form, res, submitFn);
   } else {
     res = EMPTY_DATA; // eslint-disable-line no-param-reassign
     // updatePriceRange(EMPTY_DATA);
     updatePriceHistogram(res);
-    updateResults(res);
+    updateResults(form, res, submitFn);
```
toolness: We could probably refactor these branches to not share so much code in common, but I held back on doing too much clever/error-prone stuff for now. Diff is hard enough to read as-is!

frontend/source/js/data-explorer/table.js
```diff
@@ -0,0 +1,257 @@
...
+
+function excludeRow(form, id, submit) {
+  const idString = String(id);
+
+  const excluded = getExcludedIds(form);
+  if (excluded.indexOf(idString) === -1) {
+    excluded.push(idString);
+  }
+  form.set('exclude', excluded.join(','));
+  submit(true);
```
toolness: I was able to get around the `eslint-disable-line no-use-before-define` directive here by passing `submit` in as an argument.  Man, that function is getting passed around all over the place now.

toolness:
Things seem to work fine on IE11.

I thought I found a regression, but the bug seems to exist on legacy CALC too, so I filed it separately as #1068.

jseppi:
👍 Looks good, thanks! I manually tested in Chrome and everything seems to work as expected.

toolness:
Yay! I will merge now.

Issue #569, opened 2016-08-18
Improve docker node_modules


jseppi:
Per article at http://jdlm.info/articles/2016/03/06/lessons-building-node-app-docker.html
as discussed/used in Checklistomania: https://github.com/18F/checklistomania/pull/100#issuecomment-240492439

toolness:
Hmm, so I just realized that the lines of our `Dockerfile` that make it possible to deploy our container to the cloud contain the following:

```
COPY . /calc/
```

that means that if the user has `node_modules` at the time that they build the container, it could overwrite `/calc/node_modules` in the container, which could lead to unpredictable results... Blergh.

toolness:
Ack, ignore my last comment--I completely forgot about `.dockerignore` and the fact that ours has `node_modules` in it. Woot!

PR #1071, opened 2016-11-29
make eslint ensure import statements are valid.

toolness:
Currently our eslint configuration doesn't ensure that `import` statements in our code are actually valid, so if we have e.g.:

```js
import nonExistentMethod from './util';
```

then eslint won't complain, and the code will actually run fine, until `nonExistentMethod` gets used somewhere, at which point it will be `undefined`.

This just enables an eslint rule to ensure that the `import` statements make sense, so our test suite will fail if they don't.  It also fixes some warnings raised by the addition of the new rule.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,6 +1,7 @@
 {
   "extends": "airbnb-base",
   "rules": {
-    "no-underscore-dangle": ["error", { "allowAfterThis": true }]
+    "no-underscore-dangle": ["error", { "allowAfterThis": true }],
+    "import/named": 2
   }
 }
--- a/frontend/source/js/data-capture/ajaxform.js
+++ b/frontend/source/js/data-capture/ajaxform.js
@@ -6,7 +6,7 @@ import * as supports from './feature-detection';
 
 import ga from '../common/ga';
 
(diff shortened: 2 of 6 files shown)


jseppi:
Sweet, does this actually work in our docker container?

jseppi:
👍 btw

toolness:
Yup! Er, is there a reason you think it wouldn't work in our docker container? I've only tried it in the docker container actually, so I'm not sure how well it works outside our docker container. Heh.

Merging now!

jseppi:
I figured that's where you were running it. I just remembered some weird `NODE_PATH` issues from earlier due to how it is setup in our docker container, and was wondering if eslint was properly taking it into account. Looks like it is!

toolness:
Oh geez, good point, I'd forgotten about that.

It's actually possible that we don't use `import` with any npm-based modules... If we use `require()` for them, then I don't think this rule kicks in. And all the `import` statements I can think of are bringing symbols in from our own modules, so the paths are relative.

jseppi:
Related: https://github.com/18F/calc/issues/569

PR #1073, opened 2016-11-29
A few more data explorer refactorings

toolness:
These are a few more refactorings that try to make the data explorer a little bit easier to understand:

* In `table.js` and `histogram.js`, all global CSS selectors are now constants defined at the top of the file. I thought this might make it easier to figure out what global state the modules were modifying. I wanted to do this with `explorer.js` too but there were just way too many global selectors. (Outside of these three files, none of the modules reference global CSS selectors.)

* Made  `table.js` better encapsulated by moving `parseSortOrder` from `util.js` into it, and also moving some table-specific code out of `explorer.js` into it.

### DIFF
--- a/frontend/source/js/data-explorer/explorer.js
+++ b/frontend/source/js/data-explorer/explorer.js
@@ -233,6 +233,8 @@ function popstate() {
 }
 
 function initialize() {
+  initializeTable(form, submit);
+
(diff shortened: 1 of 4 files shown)


jseppi:
woot, nice!
👍

PR #1074, opened 2016-11-30
Factor out a <RestoreExcluded> JSX component in data-explorer

toolness:
This is an alternative attempt at migrating the Data Explorer to React.

Rather than the top-down approach motivating #1051, where we were going to encapsulate the whole explorer in React and gradually break it down into smaller components, this attempts to Reactify just one small part of the explorer UI--in this case the **★ Restore N rows** widget.

To do:

- [ ] Determine whether this is a good idea or not; we were thinking of doing a rewrite instead of a gradual migration, but after the linting (#1053) and modularization (#1067), I'm really starting to wonder if it's possible to migrate it instead. Not sure which would be faster, but some advantages of doing it incrementally like this could include:

  * It could make it easier for @hbillings to learn how React works, since it's so piecemeal.

  * It might also be a somewhat more agile way to do things, since by Reactifying piece-by-piece we'll always have a shippable product that we can send to production. As opposed to a monolithic replacement that we push out, which feels more waterfall-y.

  * Because we convert everything piece-by-piece, it's a bit more likely that we reproduce the exact behavior of the original explorer. (If we did a complete rewrite, it's possible we could forget to implement some non-obvious functionality, since we wouldn't be consulting the original source code very often, but I could be wrong about this.)

  * I don't think this approach actually rules out doing a rewrite later on: for example, the `RestoreExcluded` component I've added here is encapsulated well enough that it could be reused in a rewrite, should we eventually decide that the remaining un-Reactified code is just too tangled to migrate.

- [ ] Add unit tests.

@jseppi thoughts on this approach?


frontend/source/js/data-explorer/table.js
```diff
@@ -19,19 +24,18 @@ function getExcludedIds(form) {
...
+    React.createElement(RestoreExcluded, {
+      excluded: getExcludedIds(form),
+      onClick: e => {
+        e.preventDefault();
+        form.set('exclude', '');
+        submit(true);
+      },
+    }),
+    $(RESTORE_EXCLUDED)[0]
+  );
```
toolness: I could've renamed this file to `.jsx` and used JSX syntax here instead of `React.createElement()` directly, but that felt weird for some reason. I can change it to that if preferred.

jseppi:
So, I do agree in theory with doing it piecemeal vs a monolothic rebuild. However, something else to consider is how much we are going to want to modify the entire data explorer from a UX perspective. It's not something we've really dug into (research-wise) very much thus far. My concern is that we might end up wanting to go with very large changes to the UX that might make any piecemeal refactoring somewhat wasted effort.

.eslintrc
```diff
@@ -1,5 +1,13 @@
 {
-  "extends": "airbnb-base",
+  "extends": ["airbnb-base", "plugin:react/recommended"],
```
jseppi: I think we should just use https://www.npmjs.com/package/eslint-config-airbnb instead of building up our own eslint with react stuff added. Could be done as part of a separate PR though.

toolness: Ah ok. I noticed you seemed to be doing something like this in #1051, but then I noticed you had to make a bunch of minor changes to unrelated files, so I decided not to go down that path. That PR was still really helpful in figuring out what I needed to do to get JSX to work with our build setup, though.

frontend/source/js/data-explorer/table.js
jseppi: I think `createElement` is 👍  here

toolness:
Ah yeah, good point about the design changes! HMM. Perhaps we could talk it over during coworking or somesuch.

jseppi:
I'm also wondering, like, _how far_ a piecemeal replacement will be able to get us, given how intertwined everything still feels. I just don't feel like I have enough experience with doing something like this (and really I only have like 1.5 React projects under my belt), and I was having a lot of difficulty the first time I tried it here. However, made the combination of changes from linting and from your recent PRs will make it easier to think about.

toolness:
Yeah, as I mentioned in the description (though I added it very recently so you might not have seen it), I think that's OK--as long as we're migrating the piecemeal stuff in a reasonable, nicely encapsulated way, I think that we could eventually probably just *do* a rewrite but reuse all our piecemeal components in it. In particular I think the stuff that's left in `explorer.js` is *really* thorny and might be worth just rewriting entirely, especially since the way data flows in redux/react is so different from the way it's flowing in the existing code.

toolness:
I ended up doing the rest of the migration work in #1075 and targeting that against `develop`, so I'm closing this PR.

Issue #547, opened 2016-08-11
"Proposed price" field has no label


toolness:
The HTML looks like this:

``` html
<div class="proposed-price">
  <input type="number" step="0.01" name="proposed-price" placeholder="Proposed price" class="two-decimal-places" />
  <button class="button-primary go" />Go</button>
</div>
```

It needs a `<label>` associated with it to be accessible.

Related: #546

toolness:
Fixed by #558!

toolness:
Ack, wait... I just noticed that #558 sets the `for` attribute of the label to the `name` attribute of the input... but it's supposed to be the `id` of the input, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label).  Which is kind of annoying because it means we have to give the input an `id`. 😞 

Also, a weird thing I just discovered is that VoiceOver apparently uses the placeholder text to identify the form input, which means the `<label>` probably isn't necessary on VoiceOver... but it's still a best practice--I guess not all screen readers may consult the placeholder text?--so we should probably Do The Right Thing, I guess.

Reopening this issue!

hbillings:
Reopening because derp.

Issue #1068, opened 2016-11-29
Browser history in data explorer doesn't seem to work


toolness:
Try doing the following in the data explorer (i.e., front page of CALC):

1. On a brand new visit in a brand new browser tab/window, the search field should be empty and all advanced options should be set to their default. There should be upwards of 50k results.

2. Type "engineer" into the search field and press <kbd>Enter</kbd>. There should now be about 7k results.

3. Press the browser's **back** button once. Note that while there is a brief loading animation on the front-end, nothing actually changes once the loading is finished: the search is still for "engineer" and the same number of results are shown. Even the URL hasn't changed. (One would expect the search to be for the empty string, and for 50k results to be shown.)

4. Press the **back** button again. Note that again, while a loading animation is shown, nothing changes once the loading is finished.

5. Press the **back** button a third time. Note that while the URL has changed--there is no longer a querystring in it--nothing else has. Our search is still for "engineer" and there are about 7k results.

This bug manifests itself on calc.gsa.gov, which I guess means it's been there since before CALC 2 started.

toolness:
I think this line in `explorer.js`'s `popstate()` function is part of the problem:

```js
  // read the query string and set values accordingly
  const data = hourglass.extend(
    form.getData(),
    hourglass.qs.parse(location.search)
  );
```

Because we're not assigning defaults to the return value of `hourglass.qs.parse()`, it means that items left out of the querystring won't be in the object that we extend atop the form data, so the existing form data values will persist.  This explains why going from `/?exclude=1` to `/` via popstate doesn't do anything, though it doesn't explain why we have so many extra history entries.

toolness:
This was just fixed by #1075! Woot!

Issue #1056, opened 2016-11-22
Education Filter can have duplicate, yet unapplied, values selected


jseppi:
To reproduce: 

1. Select one or more education filters from the custom checkbox dropdown. 
2. Reload the page
3. Click on the education filter again. Nothing will be highlighted, but the previously selected values will still be shown.
4. Select more education filters. They will be appended to the existing list.

Screenshots:
Before reload:
<img width="246" alt="screen shot 2016-11-22 at 5 03 37 pm" src="https://cloud.githubusercontent.com/assets/697848/20545647/e90ae88c-b0d5-11e6-9cc4-0db87a2badc8.png">

After reloading and again selecting filters:
<img width="241" alt="screen shot 2016-11-22 at 5 03 52 pm" src="https://cloud.githubusercontent.com/assets/697848/20545652/edd3c758-b0d5-11e6-8db5-f3791399aee3.png">

jseppi:
Note I thought this was due to my eslint refactoring in #1053, but it is also present in the current prod code.

toolness:
Just fixed by #1075!

PR #1075, opened 2016-11-30
Convert data explorer to use Redux and React

toolness:
This converts the entire Data Explorer to use React/Redux for most of the UI logic instead of jQuery/d3. It also fixes a bunch of bugs and security vulnerabilities.

jQuery is still used in a few places, which can eventually be converted to React if we want. They all rely on third-party jQuery libraries:

* Tooltips
* Labor Category Autocomplete
* Min/Max Experience Slider

Also, d3 is still used for the histogram.

To do (many/all of these can be spun off into separate issues):
- [ ] File separate issues for the CSS transition/animation related regressions.
- [x] Fix the regression where [prices of $0 show up as rows](https://github.com/18F/calc/pull/1075#issuecomment-270399280).
- [x] Fix formatting of cents on histogram
- [x] Get rid of the `#results-actions` id and css selector
- [x] Remove the [TODO about filters on initial page load](https://github.com/18F/calc/pull/1075#discussion_r94601655)
- [x] Blank education level cells should display "N/A" as done in current CALC
- [x] Remove logging middleware in production
- [x] Put a `<noscript>` message in the data explorer
- [x] Add docs for `StoreRatesAutoRequester`
- [x] Add docs for `serialization.js`
- [x] Add docs for `history.js`
- [x] Ensure sort key is valid in `serialization.js`.
- [x] Figure out if we want to keep the new "Proposed Price" field behavior, which updates on a per-keystroke basis. If we do, then we should get rid of the "Go" button; otherwise we should revert the behavior.
- [x] Make the loading indicator accessible to screen readers.
- [x] Merge `develop` back into this. There's one big merge conflict, which is that b14e560f1d82961bb69e9ed99c3b0ee211110eb6 (and to a much lesser extent bd61edd03340d18ec3b909fd730c46a9eea43f9e and 336af1caf7ca32bcd64ffb2220d1aa37ac551d3e) made significant changes to the data explorer's `index.html` and I have no idea how to reconcile them. 😞 
- [x] Factor out a `<LoadingIndicator>` JSX component to encapsulate the `<div class="loading-indicator">` in the page.
- [x] Factor out a `<ContractYear>` JSX component to encapsulate the **Contract year** filter.
- [x] Factor out a `<Schedule>` JSX component to encapsulate the **Schedule** filter.
- [x] Factor out a `<BusinessSize>` JSX component to encapsulate the **Business size** filter.
- [x] Factor out a `<Site>` JSX component to encapsulate the **Worksite** filter.
- [x] Factor out a `<QueryType>` JSX component to encapsulate the **Contains words / Contains phrase / Exact match** radio buttons.
- [x] Factor out a `<LaborCategory>` JSX component to encapsulate the labor category/text search field and its associated autocomplete logic (we can keep the jQuery-based autocomplete for now, but this new component should manage its initialization via React lifecycle methods).
- [x] Once all of the form-related components are completed, we can get rid of `formdb.js`.
- [x] Once all of the above are completed, we should be able to create a new `<DataExplorer>` component that encapsulates all of these components and any additional HTML, at which point we can get rid of the `data-embed-jsx` attribute, `temp-redux-middleware.js`, and a lot of the non-React/Redux code in `explorer.js`.
- [x] Make sure all form inputs "light up" when they have non-default values.
- [x] Make sure there are no stored XSS vulnerabilities if bad data is added to the database (e.g. if there is a labor category with the name `<script>alert('hi')</script>`). (Looks like there were, which I reported as #1118, and then fixed in this PR.)
- [ ] There are a number of constants in `constants.js` that are duplicated from server-side Python code. We should consolidate it.
- [x] ~~Remove code from `hourglass.js` that is no longer needed.~~ Er, I thought there were some candidates for this, but it turns out they're used by `hourglass.API`. We should probably file a separate bug to streamline this file a bit though, since it's duplicating some features already provided by ES6 polyfills, among other things.
- [x] Get rid of the `location` export in `util`.


toolness:
Ok, here's a quick status update. The areas outlined in red below have been converted to React components:

screen shot 2016-12-05 at 9 54 46 am

The whole front-end is now using a Redux store as the sole "source of truth" for generating views; URL changes and `/rates/` requests are made as a result of state changes.

Bugs fixed by this PR so far include:

* #1068 - Browser history now works properly
* #1056 - The **Education level** filters are now properly populated on page load / state changes

A number of other smaller, undocumented bugs have also been fixed.

There are also a few UX improvements:

* The **Proposed price** indicator is now dynamically updated on a per-keypress basis; e.g., typing `150` into the field will automatically show the label in the histogram, without the user needing to press enter or click the **Go** button.

The only major downsides of this PR right now are:

* Because the migration to redux/react isn't complete, we haven't dropped any dependencies, yet we've added redux and react as dependencies, so the page weight will increase. Even once the migration is complete, I still see us needing the largest dependencies, like jQuery and d3; we'll be able to drop the smaller ones like formdb but that won't save us much weight. (Fortunately, at least Redux isn't very big, but that's not the case for React.)

* When React-ifying some of the components, I've dropped some of the jQuery transitions, which results in more of an abrupt UX for some parts. I've added TODOs in the code, though, and we can bring the transitions back, it'll just take a bit more work than they did in jQuery (perhaps there are some third-party libraries that will make this easier).

toolness:
At this point, this PR plugs all the XSS holes in the data explorer, so it offers a more robust fix for #1058 than that offered by #1060.

toolness:
Ok, I've just made a number of accessibility fixes that are broken in the old data explorer:

* It previously wasn't possible to toggle sorting via the keyboard; now it is.
* Sort headers previously had `aria-role` attributes; as React helpfully informed me, this isn't actually a valid attribute--it's just `role`, not `aria-role`.
* The education level dropdown didn't previously have any kind of ARIA role or `aria-expanded` attribute associated with it; now it does.
* Focusing on other page elements outside the education level dropdown didn't collapse it.
* It wasn't possible to see tooltips for keyboard-focused UI elements. Now, if there's a tooltip associated with a focusable item on the page, it's now automatically shown whenever that item receives keyboard focus. (This isn't yet happening with the "What's this?" next to the "Schedule" dropdown, though, as that particular UI hasn't yet been converted to React.)
* I fixed #547, where the "Proposed Price" field had no label associated with it.

toolness:
Ok, I've added a list of TODOs that need to be finished for a complete migration of the explorer to React/Redux.

By "complete" I mean:

* Not using any jQuery for first-party code, though we'd still use third-party jQuery libraries for tooltips, the education slider, and autocomplete. We can migrate away from these third-party libraries in the future if we want.
* Only using d3 for the histogram. We can migrate away from d3 in the future if we really want.

Anyways, aside from that I think we can actually merge this code as-is, and just file separate issues for all the TODOs. In fact it might even be good for us to merge it sooner rather than later, as it fixes security issues in a more robust way, which is now more important since we lost CSP due to #1092.

toolness:
So I tried merging `develop` back into this but it looks like the DOM changes introduced by the reactification are messing things up.  I think it will probably just be easiest to finish up the reactification and then migrate the new app component over to the new styles.

toolness:
Hooray for coworking!! @hbillings helped me fix up the `develop` merge so I guess I'm just going to merge this PR into the other PR it's building on.

toolness:
Er actually I will just re-target this PR to merge against `develop`, and close #1074.

toolness:
@jseppi um I think this is ready for review.

toolness:
Just merged in the changes from `develop`, including mirroring the changes from #1156, #1157, and 39588e9ef9d3e4d9d4c6306e00dc359fdcfee46b.

frontend/source/js/data-explorer/components/loading-indicator.jsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { connect } from 'react-redux';
+
+function LoadingIndicator({ error }) {
+  return (
+    <div className="loading-indicator">
+      <p className="message">Loading results...</p>
+      <div className="error-message">
```
jseppi: maybe this entire `<div>` should be hidden when `!error || error === 'abort'`

toolness: I agree--I only had it do what it currently does to maintain the exact same behavior as the original data explorer, but I think there are better solutions we can pursue--especially if we want this loading indicator to be accessible to screen readers!

frontend/source/js/data-explorer/components/proposed-price.jsx
```diff
@@ -0,0 +1,90 @@
...
+    const id = `${this.props.idPrefix}proposed-price`;
+
+    return (
+      <div className="proposed-price">
+        <label htmlFor={id} className="sr-only">Proposed price</label>
+        <input id={id} type="text" name="proposed-price"
+               className="form__inline"
+               placeholder="Proposed price" value={this.state.typed}
+               onChange={this.handleChange} />
+        <button className="button-primary go">Go</button>
```
jseppi: Does this button do anything? I can't see how it is "hooked up" to any handler.

toolness: Uh, yeah, since I made the proposed price thingy automatically take effect on a per-keystroke basis, that "Go" button doesn't actually do anything anymore, lol.  If my new behavior is preferred, then I can remove the "Go" button, or alternatively I can revert the field to work exactly like it did before.

jseppi: Well, when the code is running, when I click the Go button it does seem to trigger a reload of results (I see the "Loading results" message appear briefly). I can't figure out what code path it is triggering though o_O

toolness: Oh heh, I think it is actually triggering the form's `submit` behavior purely by virtue of being a `<button>` in a `<form>` (and we do have an `onSubmit` handler attached to that form via React, IIRC).  Forms are really weird that way.

jseppi: Ah, ok, that was my suspicion

frontend/source/js/data-explorer/components/loading-indicator.jsx
toolness: Ok, just fixed that!

toolness:
Any thoughts on what's left to do before merging this PR?  Is it just figuring out what to do about the "Go" button next to the "Proposed Price" field, or is there more?

jseppi:
Hrm, I was kinda waiting until we finished up with our "soft launch" but I guess this could be part of that since it should be transparent to users! Do you think it is READY FOR PRIMETIME?

toolness:
YEAH! Especially with the fancy unit tests. And the a11y fixes. And the security fixes. And the other miscellaneous bug fixes!!!!!!!

frontend/source/js/data-explorer/serialization.js
```diff
@@ -0,0 +1,125 @@
...
+import {
+  parsePrice,
+} from './util';
+
+const parseSort = val => {
+  if (!val) {
+    return DEFAULT_SORT;
+  }
+
+  // TODO: Ensure the key is valid.
```
jseppi: Able to do this one?

frontend/source/js/data-explorer/components/results-table/index.jsx
```diff
@@ -0,0 +1,94 @@
...
+import * as ExcludedColumn from './excluded-column';
+import * as LaborCategoryColumn from './labor-category-column';
+import * as ExperienceColumn from './experience-column';
+import * as PriceColumn from './price-column';
+import * as ContractNumberColumn from './contract-number-column';
+
+const COLUMNS = [
+  ExcludedColumn,
+  LaborCategoryColumn,
+  createSortableColumn({
```
jseppi: Blank education level cells should display "N/A" as done in current CALC

frontend/source/js/data-explorer/rates-request.js
```diff
@@ -0,0 +1,77 @@
...
+export function getRatesParameters(state) {
+  const data = getSerializedFields(state, ratesFields, {
+    omitEmpty: true,
+  });
+  return Object.assign(data, {
+    experience_range: `${data.min_experience},${data.max_experience}`,
+  });
+}
+
+export default class StoreRatesAutoRequester {
```
jseppi: An explanatory comment for this middleware would be nice to have.

data_explorer/templates/index.html
```diff
@@ -39,477 +38,7 @@ <h2 class="description-secondary">CALC lets you conduct market research on profe
...
-          </tfoot>
-        </table>
-      </div>
-
-    </div><!-- /.container -->
-  </section><!-- /.results -->
-
-</form>
-</div><!-- .card -->
+<div data-embed-jsx-app-here class="card--half-padding"></div>
```
jseppi: Stick a `<noscript>` message in here for any users with JS disabled.

toolness:
@jseppi, I fixed what I *think* was an unintentional mutation-related bug in d4db6ff0210f899689df8efcc52cccfc2d5fd0e0. Can you check that to make sure my assumptions were correct?  I also had to tweak an existing snapshot test because it was using assuming props that depended on an earlier test case.

frontend/source/js/data-explorer/tests/business-size.test.jsx
```diff
@@ -57,7 +57,7 @@ describe('<BusinessSize>', () => {
   });
 
   it('matches snapshot', () => {
-    const { wrapper } = setup();
+    const { wrapper } = setup({ size: 'contractor' });
```
jseppi: You could just update the snapshot instead of passing props into this `setup()`, since an empty setup is the pattern used in the other tests' `'matches snapshot'`s.

toolness:
@jseppi do you think there's a way to tell eslint that everything under our jest tests directory should have `expect, describe, it, jest` as default globals?  Would cut down on some of the boilerplate...

jseppi:
Yeah, I do think there is a way. I meant to look into it on my original PR but didn't get around to it.

frontend/source/js/data-explorer/components/app.jsx
```diff
@@ -0,0 +1,224 @@
...
+                      ⬇ Download graph
+                    </a>
+                  </div>
+
+                  {/* TODO: The following id actually reproduces a bug from
+                    * the legacy implementation that this is based on.
+                    * It's actually required to maintain the same
+                    * layout as the legacy version, so we're keeping
+                    * it for now, but we should eventually fix it! */}
+                  <div id={prefixId('results-actions four columns')}>
```
jseppi: What bug are you referring to here? The bug that `id="results-actions four columns"` ? Maybe just remove that whole id and the corresponding `#results-actions` styling in `data_explorer.scss` since they aren't actually doing anything.

frontend/source/js/data-explorer/components/description.jsx
```diff
@@ -0,0 +1,132 @@
...
+    results += 'with ';
+  } else {
+    filtersClasses.push('hidden');
+  }
+
+  // TODO: The original version of this faded-in (but never out)
+  // whenever it changed. We might want to do that too, or choose
+  // a different animation.
+
+  // TODO: The original version of this didn't show the filters on
```
jseppi: Seems like the original version not showing the filters on first request was probably a bug/oversight, not an intentional behavior. Probably good to remove this TODO.

frontend/source/js/data-explorer/components/education-level.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+import {
+  EDU_LABELS,
+} from '../constants';
+
+import {
+  toggleEducationLevel,
+} from '../actions';
+
+// TODO: We could just use jQuery for this, but I wanted to decouple
```
jseppi: Can probably remove the "TODO" and just leave the rest of this comment as a note

frontend/source/js/data-explorer/components/education-level.jsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    // Whenever users click outside of our dropdown, hide it.
+    if (!elementContains(this.dropdownEl, e.target)) {
+      this.setState({ expanded: false });
+    }
+  }
+
+  handleToggleMenu(e) {
+    // TODO: The original jQuery logic used .slideDown() here, it'd
```
jseppi: Any idea how to do that? Just a CSS transition? Maybe file an issue for this.

frontend/source/js/data-explorer/components/highlights.jsx
```diff
@@ -0,0 +1,68 @@
...
+export function Highlights({
+  stdDeviation,
+  avgPrice,
+  proposedPrice,
+}) {
+  const stdDevMinus = avgPrice - stdDeviation;
+  const stdDevPlus = avgPrice + stdDeviation;
+  const proposedPriceStyle = proposedPrice ? { display: 'block' } : null;
+
+  // TODO: The original implementation faded the proposed price in and
```
jseppi: File an issue for this?

frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -156,8 +200,11 @@ export default function updatePriceHistogram(data) {
     pp.append('line');
   }
 
+  // TODO: Format the proposed price so that if it has cents, it
```
jseppi: Does this price formatting still need to be fixed?

jseppi:
Noticed a bug in the table that results in prices of $0.00 being shown. It happens when the contract year button is changed from "Current" to "+1" or "+2", but it is possible it might manifest in other situations. I think if the price for the currently selected contract year is 0, then the row probably just shouldn't be shown.

With "Current" selected:
screen shot 2017-01-04 at 9 26 53 am

After selecting "+1":
screen shot 2017-01-04 at 9 27 00 am

toolness:
Oh, interesting--I noticed the [corresponding link on calc.gsa.gov](https://calc.gsa.gov/?q=Producer%2C%20&contract-year=1&query_type=match_exact) doesn't list any Prices with $0 in them.  So this seems to be a regression specific to the React rewrite, not present in the original jquery/d3-ish code?

frontend/source/js/data-explorer/components/histogram.jsx
toolness: ummm Yeah :/

frontend/source/js/data-explorer/components/education-level.jsx
toolness: unfortunately `slideDown()` is a big pain to do purely with CSS transitions because of the way height transitions work, but yeah, will file a bug!

jseppi:
Correct - I think this is a regression in this code.

frontend/source/js/data-explorer/components/app.jsx
toolness: Yeah, that bug. My original code split it out into `id="results-actions" class="four columns"` but that actually *broke* the styling, even though it seemed like the original intent, and even removing the `class` left it still broken, so yeah, I think that just removing the whole id and the corresponding `#results-actions` rule is the way to go.

frontend/source/js/data-explorer/components/histogram.jsx
jseppi: Able to just use `import { formatPriceWithCents } from '../../util'` ?

jseppi:
Re: Proposed Price - I personally like the auto-updating on keypress. I'm a little concerned that removing the Go button might make that input/functionality even more confusing. Maybe you could just attach a noop handler to the Go button to prevent page refreshes? That way it'll be mostly in-line with the legacy front end, and we don't have to rethink that whole widget right now.

toolness:
Ok, I think everything is taken care of, except for the filing of new issues, which I'll do as soon as this PR is merged.

jseppi:
I think this is good to go! I'd kinda like to wait to roll it into one of our new more official releases (so at the end of this sprint) if that's cool with you. That will allow us to continue focusing on current "soft launch" issues if any more arise.

toolness:
That sounds good to me, though if we need to make any low-priority changes to the front-end in the meantime, can we prefer to make them here instead of `develop`, just to save potential merge conflict headaches?

hbillings:
@toolness Do you mean changes to the data explorer particularly or the front-end in general? (Just want to make sure I put things in the right place!)

toolness:
Oh sorry, I just meant data explorer particularly, my bad!

jseppi:
So, I think we are not going to do another prod release until end-of-sprint at this point, which means, this should be GOOD TO MERGE!

toolness:
WOOOOOT 👍 💯 💃

PR #1077, opened 2016-12-02
Modify price list details

jseppi:
This PR adds functionality for modifying a `SubmittedPriceList`'s contract details fields (vendor name, is small business, dates, etc.) on the Price List Details page.

Update: For now I've implemented it so that a modification calls `unapprove()` on the target price list. I don't think this is exactly what we should do because modified price lists might get "lost" in the admin interface since they'll be mixed in with other "retired" price lists. We should definitely sort out what the desired STATUS should be and how to present it to the admins, but possibly in another PR/issue (I've opened https://github.com/18F/calc/issues/1082 to discuss this).

To Do:
- [x] Figure out and implement how modification changes should be reviewed. Like, what should the `status` of the modified price list be set to (`NEW`, `UNAPPROVED`, or a new status altogether)? What should happen to the associated `Contract`/`SubmittedPriceListRows` models?
  - Or maybe change `STATUS_NEW` to `STATUS_UNREVIEWED` and set both newly submitted price lists and modified price lists to that status?
- [x] Add additional test(s) for what is decided wrt ^^
- [x] Update: add some text to describe this when `status==STATUS_APPROVED`. Should we have a modal or other warning text that modifying a price list's details will remove it from CALC until it is reviewed? Maybe this warning could just be displayed if the price list is `STATUS_APPROVED` at the time of modification.


data_capture/forms/price_list.py
```diff
@@ -205,6 +201,25 @@ def clean(self):
         ensure_contract_start_is_before_end(self, cleaned_data)
         return cleaned_data
 
+    def is_different_from(self, price_list):
+        my_model = self.save(commit=False)
```
jseppi: @toolness: is this dangerous to do?

frontend/source/sass/components/_alerts.scss
```diff
@@ -40,11 +40,17 @@ alerts-widget:focus {
   outline: none;
 }
 
+
+// TODO: style these the same as alerts?
```
jseppi: @hbillings: maybe this should be handled separately, but do we want to think more about `alerts` vs `messages`? If there is a semantic and styling difference, perhaps we should mention it in the Style Guide.

frontend/source/js/data-capture/edit-details-form-enhance.js
```diff
@@ -0,0 +1,30 @@
...
+  if (hasAll) {
+    editBtn.click(e => {
+      contractDetails.slideUp();
+      editContractDetails.slideDown();
+      $('[data-edit-text]').each((_, el) => {
+        $(el).text($(el).attr('data-edit-text'));
+      });
+      e.preventDefault();
+    });
+  } else if (hasAny) {
```
jseppi: Modified this condition to instead check if any, but not all, of the required elements are present.

data_capture/views/price_lists.py
```diff
@@ -59,8 +63,69 @@ def price_list_details(request, id):
...
+                    "were found in the submitted form."
+                )
+                return redirect('data_capture:price_list_details',
+                                id=price_list.pk)
+
+            # otherwise, update the price list
+            price_list.submitter = request.user
+
+            # TODO: Should instead call unapprove()?
+            # or maybe make a another status and method (STATUS_MODIFIED) ?
```
jseppi: Really unclear what to do upon modification...will elaborate in slack or PR description.

data_capture/tests/test_price_list_views.py
```diff
@@ -87,3 +90,44 @@ def test_has_context_vars(self):
...
+            res, "Your changes have been submitted. An administrator will "
+                 "review them before they are live in CALC.")
+        pl = SubmittedPriceList.objects.get(id=self.price_list.pk)
+        self.assertEqual(pl.status_changed_by, user)
+        self.assertEqual(pl.status, SubmittedPriceList.STATUS_NEW)
+        self.assertEqual(pl.status_changed_at.date(),
+                         datetime.datetime.now().date())
+        self.assertEqual(pl.submitter, user)
+        self.assertEqual(pl.vendor_name, post_data['vendor_name'])
+        self.assertEqual(pl.is_small_business, False)
```
jseppi: This assertion fails. I can't figure out why. It might be something super obvious, but I can't figure it out. The `post_data` dict above has `is_small_business': 'False'`, and that field is set in my [view method](https://github.com/18F/calc/pull/1077/files#diff-a176b6a07f901ba3cc7b7a0f28cf74dfR97), so I have no idea why it is not `False` here. cc @EricSchles @toolness

data_capture/views/price_lists.py
```diff
@@ -59,8 +63,67 @@ def price_list_details(request, id):
...
+                # otherwise, update the price list
+                price_list.submitter = request.user
+
+                # TODO: Should instead call unapprove()?
+                # or maybe make a another status and method (STATUS_MODIFIED) ?
+                price_list.status = SubmittedPriceList.STATUS_NEW
+                price_list.status_changed_by = request.user
+                price_list.status_changed_at = timezone.now()
+
+                form_model = form.save(commit=False)
```
jseppi: @EricSchles @toolness: using the returned model from `form.save` instead of the `form.cleaned_data` dict to get the modified field values solved the problem I was having with `is_small_business` being a string instead of a boolean.

jseppi:
@toolness This is largely ready for review, though there is a somewhat big business logic TODO described in the PR description.

data_capture/tests/common.py
```diff
@@ -79,13 +79,6 @@ def r10_file(content=None, name='r10.xlsx'):
     DATA_CAPTURE_SCHEDULES=[FAKE_SCHEDULE],
 )
 class StepTestCase(ProtectedViewTestCase):
-    def assertHasMessage(self, res, tag, content):
```
jseppi: moved this out of `StepTestCase` and into new `BaseTestCase` in `hourglass/tests/common.py` so it could be used in more tests.

jseppi:
Ready for code review @toolness.

jseppi:
Hrm, since modifying details will have a side effect of unapproving the price list (thus removing its data from the Data Explorer), should we maybe have some text or even a modal to warn the user before saving?

data_capture/forms/price_list.py
toolness: Hmm, I don't think so!

data_capture/views/common.py
```diff
@@ -14,6 +17,36 @@ def add_generic_form_error(request, form):
...
+
+    Any keyword arguments that are `None` will be excluded, though:
+
+        >>> build_url('data_capture:step_4', a=None)
+        '/data-capture/step/4'
+
+    Any value to the  `reverse_kwargs` argument will be passed through as the
+    `kwargs` argument to `reverse`. This is useful for setting url parameters.
+
+        >>> build_url('data_capture:price_list_details', reverse_kwargs={'id': 5})  # noqa
```
toolness: Oh, doctests are meant to look like python REPL sessions, so you can continue a line by prefixing the next line with `...` instead of `>>>`, e.g.:

```
>>> bloop(1,
...       2)
```

Assuming that is the kind of thing you meant?

jseppi: yup, that's exactly what I meant!

toolness:
Yeah, I definitely think a warning of some sort would be really helpful here--something that sets the user's expectations so they have some idea of what's going to happen when they submit their changes.

Implementing a baseline first (something that doesn't require JS, i.e. not a modal, just a warning near the submit button) seems like a good initial implementation to me!

I was originally assuming that modifying a PL would actually leave its unmodified version in CALC (if it had been approved before) but that would introduce some weird complexity into things, wouldn't it.

jseppi:
Ok, I'll add some warning text (maybe only conditionally displayed if the current status is 'approved').

Yeah, I thought about trying to create a duplicate price list, but that gets pretty weird for dealing with the associated submitted price list rows and for how the user would differentiate the 'current' one and the modified 'pending approval' one. Blurpf.

hbillings:
Could we maybe put the warning near the "edit" button instead of the "save" button? That way folks know up front that what they're about to do will have consequences. *cracks knuckles*

data_capture/views/price_lists.py
```diff
@@ -59,8 +62,62 @@ def price_list_details(request, id):
...
+                price_list.submitter = request.user
+
+                form_model = form.save(commit=False)
+
+                price_list.vendor_name = form_model.vendor_name
+                price_list.is_small_business = form_model.is_small_business
+                price_list.contractor_site = form_model.contractor_site
+                price_list.contract_start = form_model.contract_start
+                price_list.contract_end = form_model.contract_end
+                price_list.escalation_rate = form_model.escalation_rate
```
toolness: Instead of this manual assignment, could you just bust `forms.PriceListDetailsForm(request.POST, instance=price_list).save()`?

toolness:
Super cool!  Assuming @hbillings is ok with this design, I am 👍 for mergification!

jseppi:
Help text says "Making changes to these details will remove this price list's data from CALC until an administrator reviews it again." and looks like this:

screen shot 2016-12-07 at 12 08 57 pm

data_capture/views/price_lists.py
jseppi: Sweet, yeah, that does work. I modified your line slightly to `.save(commit=False)` since `unapprove()` also calls `.save()`.

jseppi:
@hbillings let me know if 👍

Issue #1082, opened 2016-12-06
Determine STATUS for when a Price List is modified


jseppi:
Related: https://github.com/18F/calc/pull/1077

We need to determine how to handle the re-approval of a modified `SubmittedPriceList` model. Some outstanding question:

- How should modified price lists be presented to Data Admins in the admin interface?
- Should we create a new STATUS for modified price lists (like `STATUS_MODIFIED`)?
  - Alternatively, we could maybe rename `STATUS_NEW` to `STATUS_NEEDS_REVIEW` (or something like that).
  - If we do that should we have some other fields to indicate when a price list with `STATUS_NEEDS_REVIEW` is new or a modified existing price list? Does that matter?
- Should the `submitter` of the modified price list be changed to the user who does the modification (as it is currently implemented in #0177, the answer is "yes")?

cc @hbillings @tram @toolness

PR #1083, opened 2016-12-06
Change "NEW" to "UNREVIEWED" and "UNAPPROVED" to "RETIRED"

jseppi:
I've gone ahead and changed the `STATUS_NEW` status to `STATUS_UNREVIEWED` and updated all references to "new" to be something along the lines of "awaiting review" or "not yet reviewed."

To Do:
- [x] Decide if this is the right way to go for #1082 
- [x] If this is ok, then add a new method `unreview()` to `SubmittedPriceList` in the vein of `approve()` and `unapprove()`. Use this method in the modification logic of the `price_list_details` view.
- [x] Change `STATUS_UNAPPROVED` to `STATUS_RETIRED` and `unapprove()` to `retire()`


jseppi:
Ready for discussion, @toolness, @hbillings, @tram.

tram:
"Awaiting review" makes sense to me.

toolness:
Super cool!

For future reference, I'm fine with only changing these kinds of things in content that faces end-users, rather than internal variables and constants and stuff, unless it really impacts devs' understanding of the code--which I don't think it does in this case--but feel free to disagree!

jseppi:
Ready for review, @toolness

data_capture/models.py
```diff
@@ -140,19 +141,26 @@ def approve(self, user):
...
-
+        self._delete_associated_contracts()
         self.save()
 
     def reject(self, user):
         self._change_status(self.STATUS_REJECTED, user)
         self.save()
 
+    def _delete_associated_contracts(self):
+        for row in self.rows.filter(is_muted=False):
```
jseppi: @toolness: I extracted this logic from `unapprove()`. Is there a reason we only look at rows with `is_muted=False`? We should be able to just iterate over all `self.rows` and delete `row.contract_model` if it exists, right?

jseppi: Right on, will remove the `.filter` in that case!

toolness:
Hmm, the `unreview()` method looks technically sound, but I was at first confused about what the difference was between it and `unapprove()`. In particular, because `unapprove()` is the reverse of `approve()`, it made me think that there must be a `review()` method, but there's not...

Oy! At this point I think the terminology we're using within our own code is getting confusing. 😞 

While it might be a good idea to eventually do some renaming to make this less confusing (I'm thinking maybe renaming "unapproved" to "retired", which I think we were considering), for the time being could you add some documentation that makes things less confusing?

Another alternative I can think of that might make things less confusing is to add an optional keyword argument `status` to `unapprove` that defaults to `STATUS_UNAPPROVED` but which can be changed by the caller... That way instead of calling `unreview`, clients can call `pricelist.unapprove(user, status=pricelist.STATUS_UNREVIEWED)`, but I dunno. That might be confusing because "unapproved" is conflating an action with a status... blargghhh

jseppi:
There is no `review()` method because that is what `approve()` is. I agree that `unapprove()` should probably be `retire()` (and  `STATUS_UNAPPROVED` -> `STATUS_RETIRED`). Uh, I guess that wouldn't be too difficult to do in this PR also if that will make things clearer (in addition to docstrings for the methods).

toolness:
Cool let's do it!!!!!!!!!

jseppi:
Alrighty, @toolness, I made the switch from `STATUS_UNAPPROVED` to `STATUS_RETIRED`. Ready for final review!

jseppi:
oh wait, I haven't made the `is_muted` change, or added docstrings to the status change methods. Few more minutes!

jseppi:
ok, now good to go!

data_capture/models.py
```diff
@@ -142,16 +147,31 @@ def approve(self, user):
...
+        '''
         self._change_status(self.STATUS_RETIRED, user)
         self._delete_associated_contracts()
         self.save()
 
     def reject(self, user):
+        '''
+        Reject this SubmittedPriceList. This method should be called on
+        "unreviewd" price lists. It is the opposite of approving a price list.
+        '''
```
toolness: Do we want to put an assertion in the code here that verifies that the status is currently `UNREVIEWED`?

Also, looks like you made a typo above, `unreviewd` is missing an `e` near the end.

jseppi: yeah, that's not a bad idea. I'll add the assertion and a test to cover it, and I'll fix the typo. Then :shipit:!

PR #1085, opened 2016-12-06
Change step_1 to not overwrite other step session data

### LINKED ISSUE #1084: Step 2 isn't restored when user goes back to step 1
I've been doing a lot of back-and-forth navigation in implementing this card stuff, and I've found that this works and data is saved as expected:

Step 1 (add data) -> Step 2 (add data) -> Step 3 -> Step 2 (still has data)

but if I go all the way back to Step 1 from Step 3...

Step 1 (add data) -> Step 2 (add data) -> Step 3 -> Step 2 (still has data) -> Step 1 (still has data) -> Step 2 (no longer has data!)

jseppi:
Closes #1084


jseppi:
GitHub was having issues -- had to push multiple branches before the PR functionality would work for some reason.

toolness:
Noice, looks good!

PR #1086, opened 2016-12-07
Validate contract_number, spruce up some tests

jseppi:
This PR adds some minor regex-based validation to the contract_number field so that it can only contain alphanumeric and dash characters.

screen shot 2016-12-07 at 9 11 34 am


jseppi:
Ready for review @toolness.

Any reason _not_ to do this, @tram?

tram:
Looks good to me! Will we be able to capture when people are shown formatting errors? It would be helpful to see where folks are running into roadblocks.

jseppi:
We potentially could. I think it would involve some javascript that runs on page load and looks for validation error elements or text and fires a GA event if found.

toolness:
As long as actual real-world contract numbers actually obey the constraints we're adding here, I think this is good to merge!

jseppi:
Confirmed by Kelly!

PR #1078, opened 2016-12-02
Implement card-based visual layout and other visual changes

hbillings:
Relies on #1066.

Implementing @ericronne's designs one piece at a time. This constrains the width of the forms to 6 of the available 12 columns (with an extra column of padding on either side). 

Screenshots will update here as applicable.
Logged-out homepage. I'd really like to add the log in link to this new, less prominent nav:
screen shot 2016-12-15 at 9 25 09 am

Logged-in homepage header:
screen shot 2016-12-15 at 9 07 56 am

Narrow card view, with header:
screencapture-0-0-0-0-8000-data-capture-step-1-1481819148057

Wide card view, no header:
screencapture-0-0-0-0-8000-data-capture-step-3-errors-1481819805913

Narrow card view, no header:
screen shot 2016-12-15 at 9 23 21 am

To do:
- [x] Get Step 2 to reflect the schedule name
- [x] Update site admin page styles

PR #1089, opened 2016-12-07
Upload new price list for existing SubmittedPriceList

jseppi:
This PR adds the ability for a user to upload a new price list file for an already submitted price list. Once uploaded, the `SubmittedPriceList` is set to unreviewed so an admin must re-approve it for it to be live in the Data Explorer.

To Do:
- [x] Implement flow to upload a replacement price list
- [x] Show previously uploaded file name (ref https://github.com/18F/calc/pull/1110#issuecomment-266783693)
- [x] Save newly uploaded file name
- [x] Cardify the new steps once #1078 is merged
- [x] Tests for new flow
- [x] remove step widget
- handled by #1163 ~figure out how to "jump" back to top of Details page after finishing the replace upload flow~


jseppi:
Ready for code review, @toolness.
Ready for style/content review, @hbillings.

data_capture/templates/data_capture/replace_price_list/replace_step_1.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% block step_body %}
+  {% if upload_example %}
+    <expandable-area>
+      <p><strong>I don't know what to upload.</strong></p>
+      <div>
+        {{ upload_example }}
+      </div>
+    </expandable-area>
+  {% endif %}
```
toolness: I wonder if we should bust out this upload example bit into a separate partial to keep things DRY?  I dunno.

toolness:
So does this basically duplicate a bunch of the price list upload logic and modify it a bit, or am I misreading things?  While that's OK, especially since there are tests for the duplicate logic, it is a bit of a bummer that we couldn't reuse more logic as-is.  I trust your judgement on the approach though!

jseppi:
Yeah, there is definitely some overlap between the normal price list upload and this replacement flow. However, the upload flow is already somewhat confusing and I don't know if abstracting the shared pieces would help make anything easier to understand.

jseppi:
@hbillings here are some screenshots:

Price List Details page (the note text is only displayed when `status=='approved'`):
screen shot 2016-12-20 at 10 43 37 am

Replace Step 1:
screen shot 2016-12-20 at 10 44 52 am

Replace confirmation page:
screen shot 2016-12-20 at 10 47 07 am

After submitting the replacement price list, user is redirected back to Details page and shown a `.message`:
screen shot 2016-12-20 at 10 49 03 am

data_capture/templates/data_capture/replace_price_list/replace_step_1.html
```diff
@@ -0,0 +1,21 @@
+{% extends 'data_capture/replace_price_list/replace_step.html' %}
+
```
hbillings: Could we add `{% block body_class %}card--narrow card--no-header{% endblock %}` to this in order to remove the two-card visual? (See current step 3 for an example.)

data_capture/templates/data_capture/replace_price_list/replace_step_1_errors.html
```diff
@@ -0,0 +1,74 @@
...
+    {% with total=rows|length %}
+      <div class="message message-error" role="alert">
+        {% if not gleaned_data.valid_rows %}
+          <h3>No valid rows found!</h3>
+          <p>
+            Your uploaded price list contains no valid rows. Please fix the errors
+            shown below in the original file and <a href="#upload">try uploading it again</a>.
+          </p>
+        {% else %}
+          <h3>{{ total }} row{{ total|pluralize:" has,s have" }} errors</h3>
```
hbillings: Change to `We found errors in N rows of your data`

data_capture/templates/price_lists/details.html
```diff
@@ -6,102 +6,114 @@
...
             </div>
           </div>
-        </div>
+        {% endif %}
+
         <div class="row">
           <div class="columns twelve">
-            <h3>Valid price list rows</h3>
-            {% include 'price_lists/_price_list_rows_table.html' %}
+            <a href="{% url 'data_capture:replace_step_1' id=price_list.pk %}" class="button button-primary">Upload replacement price list</a>
```
hbillings: For approved lists: Is there any way to add a message here when the user has been returned to the page that tells them the data has been removed? I get scrolled down to this button automatically and miss the changed status alert at the top of the page. Maybe even do something like show the approved status briefly, and then change it to "needs approval" or something like that?

jseppi: Hrm... Should I perhaps change the success message from 

> Your changes have been submitted. An administrator will review them before they are live in CALC.

to 

> Your changes have been submitted and the previous data has been removed from CALC. An administrator will your changes before they are live in CALC.

when the user is modifying an approved price list?

I _think_ the weird scroll back down to the bottom of the page is caused by our smooth-scroll javascript, since it is remembering that that's where the scroll position was at the time of pressing the button. I'll see if there is a straightforward way to forget that position. cc @toolness

jseppi: What about: we modify smooth-scroll not restore scroll position, but instead scroll to the messages container (probably by targeting `#messages`) if it is there?

jseppi: Third option: we have the final step in the replace flow redirect to the messages anchor (ie `/data-capture/price-lists/1#messages`).

jseppi: This scroll stuff might be better served by its own issue/PR I think.

jseppi:
@hbillings: I made the first two changes you requested. Let me know if you think the language changes I suggested in https://github.com/18F/calc/pull/1089#discussion_r93450836 are reasonable.

Regarding the scroll position when being redirected back to the Details page after submission, I totally agree it is confusing that the scroll is not reset to the top of the page. I think, however, fixing that probably warrants a separate issue and some more thought, as indicated by my string of comments.

Oh, another potential way to solve that messaging issue is to add a "Success" page as a third step to this replacement flow, which would actually be more in-line with our regular upload flow. It might also alleviate the "two steps look weird" issue (#1150).

data_capture/templates/price_lists/details.html
jseppi: Another option: The original intent of the scroll-position-saving was for the original Price List Upload flow. IIRC, we wanted to have smooth scrolling when the user clicked the "Upload again" link on Step 3 Errors, but that required setting the browser scroll restoration to `'manual'`, so then we implemented our own scroll position restoration in JS (ref https://github.com/18F/calc/issues/921). 

So, maybe we should only have the scroll restoration/smooth-scroll javascript on only pages that need it instead of on all `/data_capture/` pages?

jseppi:
Per @tram and @KellyBailey, we are going to NOT make a "Step 3 Success" and are going to remove the Step Widget (since it only has two steps for this flow).

jseppi:
@toolness: any ideas about preventing the auto-scroll "jump" back down to the bottom of the page when going back to the details page after successfully uploading a replacement price list? See https://github.com/18F/calc/pull/1089#discussion_r93454626 and subsequent replies for some ideas of what we could do.

data_capture/templates/price_lists/details.html
toolness: Oh shoot, this is definitely a bug in the autoscroll code!  I can fix.

toolness: Just filed #1162 separately, will fix it tomorrow!

toolness: Just fixed #1162. I do think that smoothly scrolling from the top of the page to `#messages` is a good idea too, but I think we should tackle it as a separate feature for `smooth-scroll.js`.

jseppi:
Since #1163 is handled separately, I think I've addressed all of @hbillings review comments. She's not in this week so I will probably just merge without her review approval.

@toolness: could you take one last look at this to give me final 👍?

data_capture/templates/data_capture/replace_price_list/replace_step.html
```diff
@@ -2,18 +2,9 @@
...
-      {% elif step.number == 2 %}
-        Data verification
-      {% endif %}
-    </label>
-    <ol class="steps">
-      <li {% if step.number == 1 %}class="current"{% endif %}></li>
-      <li {% if step.number == 2 %}class="current"{% endif %}></li>
-    </ol>
-  </div>
+{# intentionally left empty #}
```
toolness: It might be helpful to mention why that spot is intentionally empty, e.g. b/c a two-step widget isn't very useful. (Actually, it might also be useful to mention this in the "Steps Widget" section of the style guide, too!)

Hmm. Makes me think that for #1161, we could even have the `StepsWidget` render method simply return an HTML comment if there are no more than two steps--that way the step flow can transparently "upgrade" to a widget if we ever add another step to the flow, without us having to do anything extra. Anyways, just brainstorming I guess.

PR #1100, opened 2016-12-09
build redirect_uri and pass to token exchange

### LINKED ISSUE #1099: Auth callback: set redirect_uri in request
In order to have multiple callback urls in our UAA registration (which is what we have/want with `calc-prod-fr` so that we can test with the non-prod URL), we need to pass `redirect_uri` with our token request.

jseppi:
Closes #1099 

I tested on the GovCloud calc-dev instance and it appears to work.


toolness:
I like it!

Issue #1092, opened 2016-12-08
CSP is blocking New Relic browser monitoring


jseppi:
<img width="284" alt="screen shot 2016-12-08 at 11 55 31 am" src="https://cloud.githubusercontent.com/assets/697848/21021288/6d7891b4-bd3d-11e6-8782-d5233e793541.png">

Can see it live in your dev console at https://calc-dev.apps.cloud.gov/

jseppi:
barf, looks like New Relic doesn't "support" CSP:

https://docs.newrelic.com/docs/browser/new-relic-browser/getting-started/compatibility-requirements-new-relic-browser#csp

jseppi:
Though it does look like we could copy/paste their script instead of having it injected (hopefully!!!):

https://docs.newrelic.com/docs/browser/new-relic-browser/installation-configuration/add-apps-new-relic-browser#copy-paste-app

toolness:
Hmm, I don't think I have a new relic login currently. Will have to sign up and then get access to CALC so I can get the snippet, I think.

jseppi:
The CALC New Relic account shows that you are an admin

jseppi:
Closed by #1106

PR #1106, opened 2016-12-12
Disable CSP because New Relic sucks.

toolness:
This works around #1092 by disabling CSP 2.0 on the data explorer index page.


jseppi:
Dang New Relic :(
This code looks fine to me.

PR #1108, opened 2016-12-12
Minor changes to install order in Travis

jseppi:
- move node installation to `before_install` section
- move gulp installation to `install` section
- move `after_success` section to appear after `script` section (doesn't change behavior, just file formatting)


jseppi:
ready for review, @toolness

toolness:
Looks good to me--what's the rationale for these changes, though?

jseppi:
What caught my eye was that we were outputting the node and npm versions and installing gulp before installing our updated version of node/npm (the line that begins with `rm -rf ~/.nvm`), so I moved that line up into `before_install`. 

Then it seemed that the installation of gulp should be down in the `install` section near the other `npm install`.

I moved the `after_success` block just because it seemed like it should lexicographically be later in the list of stuff.

Issue #1105, opened 2016-12-12
Admin - User list selection links display incorrectly and don't seem to work


jseppi:
When first loading the list of Users (at `/admin/auth/user/`), the text "All X selected" is visible, even though no users have been selected:

<img width="383" alt="screen shot 2016-12-12 at 9 27 16 am" src="https://cloud.githubusercontent.com/assets/697848/21104755/5a783cdc-c04d-11e6-99ba-a4fa3bd1bb96.png">

After clicking "Select all x users", the "All X selected" text is oddly moved to a different spot on the page, and no users in the list actually appear to be selected:

<img width="732" alt="screen shot 2016-12-12 at 9 32 25 am" src="https://cloud.githubusercontent.com/assets/697848/21105290/4f11341e-c04f-11e6-9114-5b24022f60e1.png">

hbillings:
@jseppi I can't reproduce this locally, and `calc-dev` looks like this to me:
<img width="550" alt="screen shot 2016-12-12 at 4 40 05 pm" src="https://cloud.githubusercontent.com/assets/509309/21121458/b2eea248-c089-11e6-81fb-f8efc2972056.png">

I don't see a "Select All Users" option at all, which makes me wonder if I've hidden part of the Django user management UI inadvertently. 😳

jseppi:
Huh, so I am also not seeing it on https://calc-dev.app.cloud.gov/admin/auth/user/ but I AM seeing it on https://calc-prod.app.cloud.gov/admin/auth/user/. `calc-prod` actually has DOM elements that are not being shown in `calc-dev`:

<img width="1251" alt="screen shot 2016-12-13 at 10 00 49 am" src="https://cloud.githubusercontent.com/assets/697848/21147562/102c7958-c11b-11e6-9f00-22cf0dcbc92d.png">

I _think_ the difference might be that `calc-prod` has enough users to go over a single page of users, so maybe that is why those extract DOM elements and functionality exist.

jseppi:
Yup, that seems to be it. Check out the `actions.html` template here: https://github.com/django/django/blob/3c447b108ac70757001171f7a4791f493880bf5b/django/contrib/admin/templates/admin/actions.html#L7

hbillings:
@jseppi ahhh! Thanks; I'll look at that now!

hbillings:
Closed by #1109; will be handled more gracefully in #1134.

Issue #1014, opened 2016-11-04
Left-align all table headers


hbillings:
Some table headers are right-aligned, probably because they're picking up styles from number cells.
![screencapture-0-0-0-0-8000-data-capture-step-4-1478296678853](https://cloud.githubusercontent.com/assets/509309/20024078/ba376f4c-a2a9-11e6-8fe2-1e7f9aab1a43.png)

jseppi:
Yeah, the `<th>` elements with columns of numbers also have the `.number` class applied to them.

PR #1109, opened 2016-12-12
UI bugfixes

### LINKED ISSUE #1093: Data Explorer - Reset button is invisible
Got some kind of a ghost button thing going on with the "Clear Search" button on the Data Explorer:

screen shot 2016-12-08 at 1 08 12 pm

Here it is on current prod CALC:

screen shot 2016-12-08 at 1 10 08 pm

The markup for it is a bit weird because it is an `<input>` and not ` <button>` o_O

```html
<input id="reset-button" class="reset button" type="reset" value="Clear search">
```

hbillings:
Note: This can be merged whenever; I'm just continuing to push bugfixes to it as necessary.

Closes #1093, #1014, #1105. 

Note that for #1105, I'm just hiding the links mentioned and exposing a hidden checkbox that allows selecting all users _on a page_. This is different from being able to select all users, period, obvi. We probably want to allow the latter to happen, but it's going to require tracking down some JS that's missing from the django admin (or possibly just rewriting it, because there's a lot of JS we _don't_ want in the admin).


toolness:
Yass let's merge these in!  Will also allow me to reflect the explorer changes in #1051 so it stays up to date.

PR #1110, opened 2016-12-13
Pre-fill step 3 when a file was successfully uploaded earlier

### LINKED ISSUE #1102: When returning to Step 3 after already uploading, show filled-in file input
If the user has already uploaded in Step 3, but then navigates back to Step 2 and then goes forward to Step 3, the file input should show the name of the previously uploaded file. The user should be able to progress to the next step (either Step 3 Errors or Step 4) based on the already uploaded data.

toolness:
This fixes #1102.

Currently this PR reverses a bunch of #1027, specifically the `/step/3/data` view.

Notes:

* Clicking **Previous** in `/step/3/errors` always takes the user back to `/step/3` (before, it took the user to `/step/2`). It felt weird to me to not take the user back to the previous page they were on, but we can change this back if folks feel differently.

* Do we want to pre-fill the field if a price list consisting *entirely* of invalid rows was uploaded earlier?  Currently, we don't do this, but we can if needed, I think.

To do:
- [x] ~~Figure out how we want to remember the filename of the file that the user previously uploaded; it's currently hard-coded to `foo.csv`.~~
- [x] Add database migration.


toolness:
At first I was thinking that this might be a good opportunity to just save the whole file, since Django's own [`FileField`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#filefield-and-fieldfile) and third-party libraries like [`django-storages`](https://django-storages.readthedocs.io/en/latest/index.html) keep track of filenames and Kelly mentioned that it might be useful in the future to allow users to re-download price lists they uploaded months ago.

However, these mechanisms only appear to store file names in a flat namespace, so that e.g. if one user uploaded `boop.csv` and another uploaded a file with the same name, the second one would be renamed to something unique like `boop2.csv`. Obviously this would be confusing for users so we should probably just keep track of the uploaded file name ourselves.

jseppi:
> Do we want to pre-fill the field if a price list consisting entirely of invalid rows was uploaded earlier?

I think we should not pre-fill in that case.

toolness:
@jseppi oh the reason for that is for the modify price list flow (#1089 I think) -- my understanding was that when the user comes back a billion years later to modify their price list, we want to show them the file field pre-filled just like it was when they first uploaded it. And the only way to do that is to store it in the DB.

toolness:
I guess this is ready for review.

jseppi:
Actually, back to the `existing_filename`, the way I am implementing the Upload Replacement File stuff in #1089 is going to be with a few new step-based views. I think it makes better sense to start with an empty file upload widget in the first step of that flow (since the user has already clicked something to get into the replacement flow), so I won't be pre-filling the widget with the file name.

@hbillings: do we actually need/want to show what the previous file name was somewhere in that step?

hbillings:
@jseppi I think if you're implementing the flow such that the first thing a user sees in that flow is the upload form, it makes sense to leave it blank. It might be a little confusing if they can go forward and see the old data, but if someone's making a concerted effort to click an "upload new file" button, they aren't very likely to do that, I'd think?

So, yeah, blank is 💯

jseppi:
@hbillings okie dokie, they won't be able to go forward without uploading. It's not actually the same flow as our full Price List Upload, but a very similar 3-step flow: upload (-> errors) -> review -> submit

toolness:
O ok.  Should we remember the filename anyways, since the code is in this PR anyways, just in case we eventually decide that we do want to change the UX so we show the user the name of their previously-uploaded file?

jseppi:
Feel free to merge this, @toolness. I'm waiting on it so I can handle updating the `uploaded_filename` in #1089.

toolness:
Ok! I noticed that one of your recently-merged PRs added a migration `0013` to `data_capture` as well, so even though the test suite and github didn't flag it (because the filenames are actually different), I've merged in `develop` and made the new migration depend on those other recently-added ones.

PR #1124, opened 2016-12-16
Fix #1006 (<select> and <textarea> in ajaxform).

### LINKED ISSUE #1006: ajaxform component doesn't work with select or textarea elements
And potentially other form elements I'm not thinking of right now.

Related: #570 (fixed via #1001)

toolness:
Er, I'm just making this change via github's web interface so hopefully this change will actually do what I think it should. 😁 

By which I mean, it should fix #1006.


jseppi:
nice, thx 👍

Issue #1129, opened 2016-12-16
Fix calc-dev deployment - fails on Step 4


jseppi:
Fails on `step/4` of price list upload. Not sure what the issue is.

<details>
  <summary>error log from calc-dev</summary>
 
```
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR [16/Dec/2016 17:36:25] ERROR [django.request:256] Internal Server Error: /data-capture/step/4
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR Traceback (most recent call last):
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR   File "/app/.heroku/python/lib/python3.4/site-packages/django/core/handlers/base.py", line 132, in get_response
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR     response = wrapped_callback(request, *callback_args, **callback_kwargs)
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR   File "/app/.heroku/python/lib/python3.4/site-packages/newrelic-2.70.0.51/newrelic/hooks/framework_django.py", line 503, in wrapper
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR     return wrapped(*args, **kwargs)
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR   File "/home/vcap/app/data_capture/views/common.py", line 156, in wrapper
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR     return func(*args, **kwargs)
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR   File "/app/.heroku/python/lib/python3.4/site-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR     return view_func(request, *args, **kwargs)
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR   File "/app/.heroku/python/lib/python3.4/site-packages/django/contrib/auth/decorators.py", line 22, in _wrapped_view
2016-12-16T11:36:25.69-0600 [APP/PROC/WEB/0]ERR     return view_func(request, *args, **kwargs)
[...]

PR #1135, opened 2016-12-16
rm nonexistent js reference in step_4.html

jseppi:
Fixes https://github.com/18F/calc/issues/1129

I verified that it fixed the broken Step 4 in our `calc-dev` deployment by doing a manual deploy and checking that everything worked as expected.

It stinks that a missing script file breaks an entire page. I wonder if there is some setting in Whitenoise that won't cause it to break? This was the pertinent line from the `cf logs`:

```
2016-12-16T11:36:25.70-0600 [APP/PROC/WEB/0]ERR ValueError: The file 'frontend/built/js/data-capture-price-list-upload-step-4/index.min.js' could not be found with <whitenoise.storage.CompressedManifestStaticFilesStorage object at 0x7f050339d0b8>.
```


jseppi:
Oh so a nice thing that this did let me manually verify, was that even after my manual deployment and having to login again, when I went back to the Upload flow, all my previously entered data was still there and I was able to easily get back to where it had broken earlier!

jseppi:
I am going to merge this since it just a removal of something that was already broken, and that is now causing things to break. Also I did a manual verification, so 👍

PR #1137, opened 2016-12-16
Fix to show total_unreviewed instead of total_new

### LINKED ISSUE #1131: On /account page, awaiting approval has no number
screen shot 2016-12-16 at 11 25 03 am

jseppi:
Fixes #1131


jseppi:
Merging since this was a previously accepted change.

Issue #1127, opened 2016-12-16
Homepage feature bullets not aligned


hbillings:
<img width="335" alt="screen shot 2016-12-16 at 10 33 57 am" src="https://cloud.githubusercontent.com/assets/509309/21272145/2ec7eb2e-c37b-11e6-99b6-e9eadab58fc7.png">

ericronne:
whoops mentioned in #1140

PR #1138, opened 2016-12-16
More UI bugfixes

### LINKED ISSUE #1125: Cardify 404 page
Doesn't look too gud.

### LINKED ISSUE #1130: Upload widget can overflow on long file names
Here's an example from the R10 upload, but I bet it happens on all the other instances of the widget as well:

screen shot 2016-12-16 at 11 26 39 am

### LINKED ISSUE #1133: Can't select text in main header
Not sure if bug or not, but seems a bit weird: Basically I can't highlight/select text in the header on the main page. It looks like, when using the DOM explorer, that the `<nav>` element is on top of all the text:

screen shot 2016-12-16 at 11 53 42 am

the `.row` that actually contains the main heading and intro text looks like it has no height:

screen shot 2016-12-16 at 11 54 06 am

@hbillings any ideas if this is bad or not or what is going on?

### LINKED ISSUE #1140: Improve the home-page "bullet" area
## Home-page header (for context, and not yet using these colors btw)
image

## Area in need of improvement
image

## Needs header text

 - [x] Rewrite text
"With …" is a placeholder; any better ideas?
"CALC offers …" seems to communicate the basic idea, but doesn't fit with all four: "CALC offers fully burdened" eg.

image

"CALC data includes …" works if we add "rates" after "fully burdened." (thx @hbillings!)
*Shall we implement this?*
image

"About CALC data" seems more fitting, but then it would be confused with the about section. We could link it to that page, but that feels weird to me.

## Visuals need adjustment

 - [x] Shrink the icons
 - [ ] Shrink the type
 - [x] Center stack the icons (#1127)
 - [x] Left align the text

We can pair on this if you like :)

thx!

hbillings:
Closes #1133 
~#1131 already closed~
Closes #1130 
~#1127 already closed~ 
Closes #1140 
Closes #1125 

(Edited by @jseppi so that it auto-closes all the issues once merged)


user_account/templates/account.html
```diff
@@ -12,7 +12,7 @@
           <h2>{{ total_submitted }} price list{{ total_submitted|pluralize }} submitted to date!</h2>
           <ul>
             <li><a href="{% url 'data_capture:price_lists' %}#approved">{{ total_approved }} approved and available in CALC</a></li>
-            <li><a href="{% url 'data_capture:price_lists' %}#new">{{ total_new }} awaiting approval</a></li>
+            <li><a href="{% url 'data_capture:price_lists' %}#new">{{ total_unreviewed }} awaiting approval</a></li>
```
hbillings: dueling commits!

jseppi: If this once gets kept, make sure the anchor is also changed from `#new` to `#unreviewed` :)

Issue #1136, opened 2016-12-16
Figure out how to stop missing static files from breaking everything


jseppi:
As documented in #1129, a missing JS file (one that had been deleted) referenced in a template (`step_4.html`) caused Whitenoise to bomb and thus a `500` error was issued. This is terrible. We either need to figure out a test to make sure that won't happen, or figure how to disable or capture that kind of Whitenoise exception.

jseppi:
I can't figure out how to not have this happen. @toolness?

jseppi:
Whitenoise docs say to check Django staticfiles: http://whitenoise.evans.io/en/stable/django.html#troubleshooting-the-whitenoise-storage-backend

jseppi:
Would running tests not in DEBUG mode have caught this?

toolness:
Hmm, so the really weird thing is that we added `frontend.tests.test_collectstatic` in #776 *specifically* to address this exact situation, because it had bitten us in the past in #772.  So I'm not sure why it didn't catch this problem. Looking into it now!

toolness:
Oh geez. Yeah, so I guess that `test_collectstatic` wasn't exploding because `manage.py collectstatic` was working fine, it was a template *requesting* a static file that didn't exist that made things explode.

Blergh.

I guess we should be running the tests without `DEBUG` enabled, though it's a bit of a tradeoff because I think there may be some sanity assertions in the code that are only run if `DEBUG` is actually present? I'm not sure... will look into it.

toolness:
But yeah, I don't think it's actually whitenoise that's breaking everything, as much as I want to blame it. 😛  It's delegating most of its behavior to Django's [`ManifestStaticFilesStorage `](https://docs.djangoproject.com/en/1.10/ref/contrib/staticfiles/#manifeststaticfilesstorage), I think, and that's the one that's raising the exceptions.

Imma fix this up somehow.

PR #1141, opened 2016-12-19
[WIP] Add a crotchety_staticfiles app that mimicks production behavior of the {% static %} template tag.

toolness:
This adds a `crotchety_staticfiles` app that mimicks the production behavior that caused #1129. It's a potential fix for #1136 but I am not really a big fan of it; I think a cleaner fix would be to make a custom static file storage backend, so I'll probably submit that as a different PR later.


toolness:
I despise computers.

toolness:
Closing this in favor of #1146, which simulates the compressed manifest static file storage backend a bit more faithfully.

PR #1144, opened 2016-12-19
Added canvas to 3 icons so all four match

ericronne:
Added blank space to the briefcase, globe, and mule icons. All four
icons are now 80px wide, and should be center-aligned when stacked.

@hbillings


hbillings:
If you could change them to the color in #1140 and push it to this branch, that would be spec-frickin'-tacular!

ericronne:
@jseppi i recolored the icons, could you review pls? thx!

jseppi:
Looks good to me. I don't think we can merge yet though since some additional (minor though) work will need to be done to adjust the CSS to account for the new sizes. Perhaps @hbillings (or you or me) could make those adjustments within this same branch to keep it all atomic.

Does https://github.com/18F/calc/pull/1145 (reverting colors) impact this PR or vice versa?

Also, not that it needs to be done here/now, but I think it would be better to use SVG for these (and probably any other) icons. Then we could style them with CSS instead of having to change binaries! Also we could inline them either in the SCSS or HTML to reduce the number of server requests.

jseppi:
Merging into the `ui-bugfixes` branch.

PR #1145, opened 2016-12-20
Revert colors to pre-WDS ones

hbillings:
This changes the colors back to the way prod calc currently is. After a conversation with @ericronne, we decided to save a color change for when we revamp the front page to make more of a splash.

It also removes all of the unused color variables from the WDS collection.

The only real oddity is that I don't know what to sub in for `$color-primary-alt`, currently used to underline the selected nav item. @ericronne, thoughts?

screen shot 2016-12-19 at 4 52 46 pm


ericronne:
Two of the colors don't meet 508 contrast levels. Here are new ones...

image

ericronne:
Also is it easy to boldface the active tab? I can do issue if you like.
E.g. "Explore the(sic!) data" here...
image

hbillings:
@ericronne Thanks! Here's what it looks like now. I think I need some icons in the right color.
screen shot 2016-12-20 at 3 30 45 pm

jseppi:
Hey @ericronne: can you give this one the 👍 if you think it is good to go?

jseppi:
@ericronne should this have the old background header image? I can add it in if so. Right now it has no image in the header, just a solid color.

ericronne:
Let's keep the bg image for the headers for now, but i'm killing it from the footer. 👍

jseppi:
Header image is back in the game!

PR #1146, opened 2016-12-20
Add frontend/crotchety.py to simulate production static file storage.

### LINKED ISSUE #1136: Figure out how to stop missing static files from breaking everything

toolness:
This is an alternative to #1141 that tries to fix #1136 by implementing a development-only static file storage backend whose `url()` method throws an exception when passed a non-existent file name, just like the @$💥%#💩!ing `CompressedManifestStaticFilesStorage` does in production.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,8 @@ install:
 script:
 - phantomjs --version
 - python manage.py ultratest flake8
+- npm run gulp -- build
+- python manage.py collectstatic --noinput --link
 - echo "For details on why we're ignoring selenium tests, see https://github.com/18F/calc/issues/330"
 - bandit -r .
 - py.test --ignore=frontend/tests/test_selenium.py --cov
@@ -34,6 +36,7 @@ env:
   - TRAVIS_NODE_VERSION=6.5.0
   - PHANTOMJS_TIMEOUT=15
   - CF_CLI_VERSION=6.22.2
+  - SKIP_STATIC_ASSET_BUILDING=yup
   - DEBUG=yup
   - DATABASE_URL=postgres://postgres@localhost/hourglass
 
--- a/frontend/crotchety.py
+++ b/frontend/crotchety.py
@@ -1,0 +1,30 @@
+# This middleware + static file storage backend makes development behave more
+# like production by disallowing references to nonexistent files from being
+# generated in templates.
+
+from django.contrib.staticfiles.storage import StaticFilesStorage
+from django.contrib.staticfiles import finders
+from whitenoise.middleware import WhiteNoiseMiddleware
+
+
+class CrotchetyWhiteNoiseMiddleware(WhiteNoiseMiddleware):
+    def get_name_without_hash(self, filename):
+        # The default implementation of this sometimes returns
+        # the names of nonexistent files (e.g. converting files with
+        # .min.js extensions to just .js) which then makes the static
+        # file storage backend throw an exception.
+        #
+        # Since this middleware should only be used in development,
+        # where hashes are never included as part of filenames, we'll
+        # just return the filename here.
+        return filename
+
+
+class CrotchetyStaticFilesStorage(StaticFilesStorage):
+    def url(self, name):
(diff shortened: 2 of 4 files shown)


toolness:

.travis.yml
```diff
@@ -24,6 +24,8 @@ install:
 script:
 - phantomjs --version
 - python manage.py ultratest flake8
+- npm run gulp -- build
+- python manage.py collectstatic --noinput --link
```
jseppi: Hrm, how was the frontend being built prior to this change? Just as a side effect in `test_collectstatic`? If so, this is much better!

toolness: We've got a helper function in [`frontend.tests.utils`](https://github.com/18F/calc/blob/develop/frontend/tests/utils.py) called `build_static_assets()` that we put in all the tests that need them, like the selenium and qunit tests.

I felt a bit weird doing things this way--I really wish there were a `Makefile`-esque way that our static build system could detect whether static assets *needed* to be rebuilt, so that it could just skip building the assets if needed. Without that, we have to resort to this weird `SKIP_STATIC_ASSET_BUILDING=yup` thing that I find myself frequently (but not always) setting in my local config when I am modifying just the python-side of integration tests... Bleh.

Anyways, the `test_collectstatic` is actually a bit different in that it is explicitly testing out static file collection using `CompressedManifestStaticFilesStorage`, which is the one that does the scraping of CSS files. It also collects them in a separate `static_compressed` directory to ensure that those files don't conflict with the normal, uncompressed static files that other tests look at. Blerg.

Honestly I'm not even sure why our selenium/phantom/robobrowser tests don't just use [`StaticLiveServerTestCase`](https://docs.djangoproject.com/en/1.10/ref/contrib/staticfiles/#django.contrib.staticfiles.testing.StaticLiveServerTestCase). That would at least obviate the need to constantly run `collectstatic`.

Part of me wants to consolidate `npm run gulp -- build` into a new `manage.py buildstatic` command, just so everything is discoverable and runnable from `manage.py`, but I'm not sure if it's going overboard or not. What do you think?

toolness:
Oh, so another side effect of this PR is that if someone checks out the repository and runs the tests *without* first running `gulp build`, most of the tests will fail because the crotchety static file backend will be complaining about the built assets not existing yet.  I decided it'd be better to just leave it like this rather than *always* building static assets at the beginning of every single test run, forcing devs to either deal with the slowness or using `SKIP_STATIC_ASSET_BUILDING` all the time... But we might want to document it.... blarlgalrglh

jseppi:
bah, yeah, that does strike me as a little odd that `gulp build` would have to be run first/separately. 

I don't think I have a good mental picture of all that is going on with python building/collecting/shelling out to other test runners. Maybe you can go over it all with me in a screenhero or something? I'm worried that adding another `manage.py` command won't necessarily clarify things. I actually never use `SKIP_STATIC_ASSET_BUILDING`, but I guess that's because I haven't been doing any front end stuff. Blerffffff.

toolness:
Yeah, honestly you don't even need to use `SKIP_STATIC_ASSET_BUILDING` unless you are trying to fix broken (or write brand-new) selenium or robobrowser tests from the back-end, which is uncommon.  Even when working on front-end you likely won't have to use it because then you'll just run the QUnit tests from your browser, rather than from the command-line.

Would be happy to talk about how it all works on a screenhero call--in fact I actually got confused myself about how things worked, so talking it over might help us figure out how to simplify it to make more sense.

toolness:
Should I go ahead and merge this for now, with the assumption that we might refactor things later to be simpler? Or should I hold off on merging?

jseppi:
I'm good for merging given that this will help us prevent the nasty bug we saw during the demo!

PR #1148, opened 2016-12-20
Use StaticLiveServerTestCase

### LINKED ISSUE #1147: Make tests use StaticLiveServerTestCase instead of LiveServerTestCase
As mentioned in https://github.com/18F/calc/pull/1146#discussion_r93267341, our live server tests don't use [`StaticLiveServerTestCase`](https://docs.djangoproject.com/en/1.10/ref/contrib/staticfiles/#django.contrib.staticfiles.testing.StaticLiveServerTestCase), which means we have to ensure that `collectstatic` is run before they execute, which complicates test setup, among other things.  I don't think there's a reason we shouldn't be using `StaticLiveServerTestCase` here, which would simplify things a bit.

toolness:
Fixes #1147.


toolness:
Ok I think this is ready for review @jseppi !

jseppi:
Django has a class for everything!

PR #1153, opened 2016-12-21
React testing with Jest

jseppi:
Open against `reduxify-restore-excluded`

This PR adds Jest+Enzyme tests for the new React components. It only involves component testing, and no redux-related testing (ie, there is no integration testing with redux actions/stores/etc).

To Do:
- [x] `<Experience>` component. Need to figure out how to mock or otherwise deal with global `$` in `components/experience.jsx`.
- [x] `<ExportData>` component
- [x] `<LoadingIndicator>` component
- [x] `<ProposedPrice>` component
- [x] `<Schedule>` component
- [x] `<Tooltip>` component
- Later: ~`<LaborCategory>` component~
- Later ~`<Histogram>` component~
- Later ~All the `<ResultsTable>` components~


toolness:
AW YISS 💥 🥇 💃 

This comment does not constitute an approval of the code; merely enthusiasm of its existence.

toolness:
This is looking great to me! I'm fine with you merging this into `reduxify-restore-excluded` if you want. Whatever's easiest!

jseppi:
Sweet, glad it looks good! It's been a good way to familiarize myself with the new react data-explorer and to learn jest/enzyme.

jseppi:
@toolness: I think I am going to punt on including the tests for those last 3 components in this PR and hope to handle them in future PRs. They are the more complicated ones since they rely on (or rely on things that rely on) global `$` and `d3` in ways that I haven't figured out how to mock out...yet.

Even better than mocking would be to switch to real imports for those libs, but that might be a pretty big lift that isn't worth the effort at this point.

Let me know if good to merge (into your PR) again, since several changes/additions have been made.

toolness:
Noice, I like it!

The only part folks might find confusing is the fact that some of our front-end tests are now in `/js/tests/` while others are now in `js/data-explorer/tests`.  I wish there were a way to make that less confusing but I can't think of one so IMMA MERGE

PR #1154, opened 2016-12-21
Bullet sizing etc

ericronne:
- Made all four icons the same width (30px)
- Recolored icons to better fit old colors
- Recolored the rule underneath “CALC offers …”
- Increased line height for intro paragraph to 1.5em

Ready for a looks @jseppi !


jseppi:
Could you target this PR to `ui-bugfixes` instead of `develop`?

ericronne:
Arg github app reverts back to `develop` every time, sry

jseppi:
Ew, that's weird o_O. Time to learn command-line git :) :) :)

frontend/source/sass/components/_header.scss
```diff
@@ -62,6 +62,7 @@ header {
   color: rgba(255,255,255,.6);
   font-size: $h4-font-size;
   font-weight: $font-weight-normal;
+  line-height: 1.5em;
```
jseppi: I am pretty sure @hbillings prefers to use `rem` for font-related stuff. To get the equivalent "leading" as you have here, it would be `2.7rem`.

Here's where she once told me that `rem` is generally for font sizes: https://github.com/18F/calc/pull/472/files/0e6a4400aab09a12fe8cdea24f6a16ef3d5da79f#r72784668

frontend/source/sass/components/_header.scss
```diff
@@ -74,7 +75,7 @@ header {
   }
   h3 {
     font-size: $base-font-size;
-    border-bottom: 1px solid $color-gray-light;
+    border-bottom: 2px solid $color-cool-blue-lighter;
```
jseppi: `$color-😎-blue-lighter`

frontend/source/sass/components/_header.scss
ericronne: I always use rem, but saw ems in use here (margin-bottom) and figured when in rome n all...

image

frontend/source/sass/components/_header.scss
ericronne: $color- 😎 - 🔵  - 🔥

jseppi: The EmojiMaster teaches the EmojInitiate

frontend/source/sass/components/_header.scss
```diff
@@ -56,14 +56,15 @@ header {
...
   font-size: $h1-font-size;
   margin-top: 0;
-  margin-bottom: .3em;
+  margin-bottom: .3rem;
 }
 .description-secondary {
   color: rgba(255,255,255,.6);
   font-size: $h4-font-size;
   font-weight: $font-weight-normal;
+  line-height: 1.5rem;
```
jseppi: Heh, this is probably not what you wanted:

screen shot 2016-12-21 at 4 35 18 pm

I think it needs to be `2.7rem` to be equivalent to the `1.5em` you previously had.

frontend/source/sass/components/_header.scss
```diff
@@ -56,14 +56,15 @@ header {
...
   letter-spacing: normal;
-  margin-bottom: 2em;
+  margin-bottom: 2rem;
```
jseppi: For margins/padding, I think you should keep sizes in `em`, so you probably should revert this line and line 65 back to `em`.

ericronne: I'm seeing a mix, actually...

image

image

For now, i'll just change everything i see in this doc to rem for type and em for everything else. We may want to do a rem/em bughunt.

jseppi:
@ericronne: all the changes from `rem` to `em` on the various `padding` and `margin` rules essentially cause the header to be bigger than it previously was. Is that what you want?

Looks like your changes result in 400px total header height:

screen shot 2016-12-23 at 1 30 03 pm

Compare with previous 365px height:

screen shot 2016-12-23 at 1 34 28 pm

ericronne:
Yeah the line spacing for the intro paragraph was too tight before. I'm loathe to fatten up that header, but we will soon be redoing it. Thxman

PR #1161, opened 2016-12-27
[WIP] Encapsulate steps widget HTML rendering in a StepsWidget class

toolness:
This attempts to encapsulate the steps widget rendering into a `StepsWidget` class, which uses its own template to render itself. This helps us stay DRY and easily add/change steps widgets where needed.

To do:

- [ ] Modify existing "manual" steps widgets to use this.

### DIFF
--- a/frontend/steps.py
+++ b/frontend/steps.py
@@ -1,0 +1,34 @@
+from collections import namedtuple
+
+from django.template.loader import render_to_string
+
+
+Step = namedtuple('Step', ['label', 'number', 'current'])
+
+
+class StepsWidget:
+    '''
+    The steps widget can be used to visualize the user's progress through a
+    multi-step process.
+
+    While it is described as a "widget", it is not actually a widget
+    in the Django sense--i.e., it is not used to present users with
+    a form UI.
+    '''
+
+    def __init__(self, labels, current):
+        self.labels = labels
+        self.current = current
+
+    def render(self):
+        steps = [
+            Step(label=label, number=i, current=(i == self.current))
+            for label, i in zip(self.labels, range(1, len(self.labels) + 1))
+        ]
+        return render_to_string('frontend/steps.html', {
+            'current_step': steps[self.current - 1],
+            'steps': steps,
+        })
+
+    def __str__(self):
+        return self.render()
--- a/frontend/tests/test_steps.py
+++ b/frontend/tests/test_steps.py
@@ -1,0 +1,14 @@
+from django.test import SimpleTestCase
+
+from frontend.steps import StepsWidget
+
+
+class StepsWidgetTests(SimpleTestCase):
+    def test_it_works(self):
+        steps = StepsWidget(
+            labels=('foo', 'bar', 'baz'),
+            current=2
+        )
+        html = str(steps)
+        assert 'bar' in html
+        assert 'Step 2 of 3' in html
--- a/styleguide/templates/styleguide.html
+++ b/styleguide/templates/styleguide.html
@@ -212,31 +212,22 @@
   </p>
 
   <p>
-    Use <code>li.current</code> to identify a step as being the current
-    one.
+    Vision-impaired users will hear an audio description of the
+    visualization, e.g. "Step 2 of 3", followed by the name of the
+    current step.
   </p>
 
   <p>
-    We recommend adding <code>aria-hidden="true"</code> to the widget, as
-    there isn't an easy way to communicate the information to vision-impaired
-    users without being overly verbose (thus annoying the user) or confusing.
-    However, <em>do</em> be sure to repeat the name of the current step
-    elsewhere in a header, so that the user knows what step they're currently
-    on.
+    Rather than writing out the raw HTML of the steps widget, we
+    recommend using {% pyobjname 'frontend.steps.StepsWidget' %}.
   </p>
 
-  {% example %}
-  <div class="step-bar" aria-hidden="true">
-    <label>
-      Validate data
-    </label>
-    <ol class="steps">
-      <li></li>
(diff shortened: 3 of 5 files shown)


toolness:
Hmm, so one funky thing about this widget is that it expects there to actually be as many steps in the widget as there are in the workflow. This is already the case for the bulk upload flow, where we have step 3 being called "complete", but it is not the case in the standard price list upload flow, where there is a 5th step that effectively represents completion but is not represented in the widget.

I also just noticed in https://github.com/18F/calc/pull/1089#issuecomment-269385077 that we aren't having a step 3 "complete" in the price list replacement flow, which makes me wonder if we'll also remove it from the bulk upload flow, leaving only the initial price list upload flow with a steps widget. 😢

jseppi:
> it is not the case in the standard price list upload flow

Does this mean this new widget doesn't work with that flow?

toolness:
Oh heh, sorry I didn't follow up on that--I basically added an actual step 5 to the widget called "Complete", so that both flows have a final "Complete" step. However, that was before seeing your conversation on #1089 mentioning that there should be no "Complete" step.  So if we decide we want to go that route, I can make the widget work with flows that have fewer defined step labels than actual workflow steps.

data_capture/views/common.py
```diff
@@ -190,6 +192,13 @@ def context(self, context=None):
         return final_ctx
 
     @property
+    def widget(self):
+        return StepsWidget(
+            labels=self.steps.labels or (['??'] * self.steps.num_steps),
```
jseppi: wait, does this mean if it is missing `self.steps.labels` it will print out `??` for all the missing labels? Maybe just make the labels required and raise an exception if they aren't specified? I don't think we'd ever want `??` getting through to the user so seems ok to raise an exception during dev.

toolness: Blerg, I just didn't want to have to change the docs for `Steps` I guess, lol. At first I was thinking of just giving each step a generic name like `Step 1`, `Step 2`... blah.

toolness:
Ok, I made a few changes:

* Exceptions are now raised if a widget is created for flows that don't have an explicit label for a step.

* A `title` attribute is now added to each `<li>` in the steps widget, which at least allows inquisitive desktop users to mouse-over the items to reveal a bit more information about them, e.g.:

  > tooltip

  (My mouse cursor isn't visible, but was hovering over step 2 in the above screenshot.)

* The labels are now defined via the `@step` decorator in the form of e.g.:

  ```py
  @steps.step(label='Validate data')
  def step_2(request, step):
    # ...
  ```

  This is actually what I originally wanted when writing up the `Steps` class earlier this year--I thought it'd both be a nice way to keep the step label closer to the step definition, and nice for documentation too, since it effectively tells readers the semantic purpose of the code they're about to read.

Thoughts?

data_capture/views/common.py
```diff
@@ -200,13 +215,25 @@ def wrapper(*args, **kwargs):
...
             return func(*args, **kwargs)
 
+        wrapper.label = label
         self._views.append(wrapper)
         return wrapper
 
+    def step(self, func=None, **kwargs):
+        if func is None:
+            # We were called in the form `@step()` w/ possible kwargs.
+            return lambda f: self._build_step_view(f, **kwargs)
```
jseppi: nice

frontend/steps.py
```diff
@@ -0,0 +1,48 @@
...
+                    step.number
+                ))
+
+    def render(self):
+        return render_to_string('frontend/steps.html', self.__dict__)
+
+    def __call__(self):
+        '''
+        A convenience for rendering the widget directly from a Django
+        template.
```
jseppi: Cool, didn't know about `__call__`

frontend/tests/test_steps.py
```diff
@@ -0,0 +1,30 @@
...
+        self.assertEqual(steps.current_step, steps.steps[1])
+
+    def test_rendering_works(self):
+        steps = StepsWidget(
+            labels=('foo', 'bar', 'baz'),
+            current=2
+        )
+        html = steps()
+        assert 'bar' in html
+        assert 'Step 2 of 3' in html
```
jseppi: What's the deal with using plain ole `assert` vs (in this case) something like `self.assertContains(...)`?

styleguide/templates/styleguide.html
```diff
@@ -212,31 +212,22 @@ <h3 class="example-heading">Example</h3>
   </p>
 
   <p>
-    Use <code>li.current</code> to identify a step as being the current
-    one.
+    Vision-impaired users will hear an audio description of the
```
jseppi: Better terminology might be "Users assisted by screen readers..."

frontend/tests/test_steps.py
toolness: I believe `assertContains` requires a Django `Response` object as its first argument, which is really annoying.  I guess I could use [`assertRegexpMatches`](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRegexpMatches) instead, at least that would (hopefully) raise an error that includes the actual output if the test fails.

jseppi: My mistake, I meant `self.assertIn`, but really my question was the difference between using `assert ...` vs `self.assertXYZ(...)`

toolness: Oh lol, well that's because I didn't even know about `assertIn`!  I was always annoyed at how I had to use `assert X in Y` instead of an `assertXYZ` method, but now I know better.

jseppi: haha, well then. I nearly always confuse `assertContains` and `assertIn` when I'm writing tests >.<

PR #1163, opened 2016-12-28
Store scrollTop in history.state instead of sessionStorage

### LINKED ISSUE #1162: Smooth scroller's manual scroll restoration can autoscroll on new page visits
As mentioned in https://github.com/18F/calc/pull/1089#discussion_r93361726, this is happening because we're using `sessionStorage` and keying pages by their URL.

This means, though, that if the user:

1. Starts out on page with URL A, clicks on a link to a page with URL B,
2. and then clicks on a link to page with URL A (distinct from pressing their browser's back button!)

... their browser will be auto-scrolled (thanks to `smooth-scroll.js`) to the last position they were at when viewing URL A before. This is not actually normal browser behavior--normal browsers would refrain from scroll restoration because the visit to URL A is a new one.

I think we can fix this pretty easily if we just use `window.history.state` to store the info rather than `sessionStorage`.

toolness:
This fixes #1162 by storing the last scrolled-to position on the page before navigation in `window.history.state` instead of using `window.sessionStorage` and keying off the URL.

### DIFF
--- a/frontend/source/js/data-capture/smooth-scroll.js
+++ b/frontend/source/js/data-capture/smooth-scroll.js
@@ -39,10 +39,14 @@ const DEFAULT_SCROLL_MS = 500;
 // support smooth scrolling.
 export const IS_SUPPORTED = window.history && window.history.replaceState;
 
+function buildPageState(object) {
+  return Object.assign({}, window.history.state || {}, object);
+}
+
 function rememberCurrentScrollPosition(window) {
-  window.history.replaceState({
+  window.history.replaceState(buildPageState({
     pageYOffset: window.pageYOffset,
-  }, '');
+  }), '');
 }
 
 function changeHash(window, hash, cb) {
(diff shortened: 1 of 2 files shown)


jseppi:
This code looks good, though I'm not very familiar with possible side effects of messing with history state like this.

I did notice something kinda weird: In the Price List Upload flow, if I manually changed the address bar URL to a differnt step, the URL "jumps." Llike say I was on `/step/3` and manually changed it to `/step/1`), the URL actually changes itself back to `/step/3` and then changes to `/step/1`. This seems like it could indicate something is wrong or that there might be an unexpected side effect.

toolness:
Oh blerg. I think I can fix that, hold onna sec.

toolness:
Ok, so now we only use `replaceState()` on page load, to assign the current visit a unique ID.  Then on `beforeunload` we store the `scrollTop` in `sessionStorage` like before, except now the key we use is based on the unique visit ID rather than the page URL. This should still fix #1162 without resulting in the weird address bar flickering behavior you described earlier.

jseppi:
Sweet, thanks for the fix! I think this is good to go 👍 

Something to ponder: would it make sense to only include a smooth scroll script on pages where we actually need/want it? Seems like we're hacking over a lot of stuff here, even though we really only have one use case for this smooth scroll, I think (the in-page link to upload a replacement price list on `step/3/errors`).

toolness:
Woo!

Hmm good question.

It seems like we actually frequently have internal in-page links that would be nice to smooth scroll between--for instance, even the data explorer has the "back to top" link at the bottom of it, the about page has a bunch of internal links (one of which is linked-to from the data explorer), and then there's your suggestion about having pages automatically scroll down to `#messages` if it exists, which is a good one.

Then there's also the fact that New Relic will be reporting any exceptions, and our safe mode opt-in... so unless the code is causing inconveniences that aren't triggering exceptions, I feel bullish about putting it in lots of places!

jseppi:
right on right on

PR #1167, opened 2016-12-28
Header anti aliasing

### LINKED ISSUE #1165: Implement anti-aliasing for all white text
Light text over dark tends to look bad. Anti-aliasing via css only works on macs, but judging by sessions which occurred over the last two months, mac users account for approximately 9% of our user base …
screen_shot_2016-12-28_at_10 12 22_am_480

So we should go ahead and set up anti-aliasing for reversed-out type: header, footer, buttons. (Buttons already have it set up, but the code needs to be amended for firefox.)

Here's the code.

`   -webkit-font-smoothing: antialiased;
  -moz-osx-font-smoothing: grayscale; `

ericronne:
Fixes #1165


jseppi:
Want to give this a looksie-do, @toolness?

frontend/source/sass/base/_utils.scss
```diff
@@ -146,3 +146,8 @@ th:hover.tooltips span {
 .no-scroll {
   overflow: hidden;
 }
+
+@mixin antialias {
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+}
```
toolness: Should we add any comments here explaining the rationale for this mixin, e.g. what @ericronne wrote in #1165, specifically the part about "Light text over dark tends to look bad, so we should set up anti-aliasing for reversed-out type"? Or will that be pretty well-understood by anyone reading this already?

jseppi: Not a bad idea. I'll add a comment!

PR #1168, opened 2016-12-28
Simplify footer button markup

jseppi:
@ericronne: This PR is against your footer PR. I have simplified the markup related to the login button.

- removed the wrapping `div` elements
- inherited from some already defined `.button` styling
- moved the button's styling rules to `_buttons.scss` to be consistent with where the other button-related styling is.
- changed the class to `.button-footer` to be a bit more consistent with class naming


jseppi:
Screenshot, though it looks the same as before:

screen shot 2016-12-28 at 2 21 57 pm

PR #1169, opened 2016-12-28
Minor text change in R10 upload widget

### LINKED ISSUE #1166: Region 10 Upload Widget text looks squished
Fairly minor since only one user, but it does look kinda yucky:

screen shot 2016-12-28 at 1 39 30 pm

Simple fix might be to just reduce the text inside to "Choose file or drag and drop here. Region 10 Export File (XLS or XLSX), please."

jseppi:
Fixes #1166 

Ready for review, @toolness 

screen shot 2016-12-28 at 3 15 50 pm


toolness:
Looks good, thanks!

PR #1170, opened 2016-12-28
Add jest to ultratest

jseppi:
Really it is adding a call to `npm test`, which runs our `jest` runner.


jseppi:
Ready for review @toolness

toolness:
NOICE 👍

PR #1171, opened 2016-12-28
Smoothly scroll to location hash on new page load

toolness:
This builds on the smooth scroll functionality to smoothly scroll the page on initial page loads, if there's a valid id in the location hash.

If this seems OK, we could additionally improve it by making it auto-scroll to `#messages` (or other configurable default id) if it exists *and* there's no other id defined in the location hash, as suggested in https://github.com/18F/calc/pull/1089#discussion_r93454626.

### DIFF
--- a/frontend/source/js/data-capture/smooth-scroll.js
+++ b/frontend/source/js/data-capture/smooth-scroll.js
@@ -110,6 +110,29 @@ export function getOrCreateVisitId(window) {
   return latestVisitId;
 }
 
+function onPageReady(window, cb) {
+  const doc = window.document;
+
+  if (doc.readyState === 'interactive' || doc.readyState === 'complete') {
+    cb();
+  } else {
+    window.addEventListener('DOMContentLoaded', cb, false);
+  }
+}
+
+function smoothlyScrollToLocationHash(window, scrollMs) {
+  onPageReady(window, () => {
+    const id = window.location.hash.slice(1);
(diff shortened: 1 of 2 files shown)


jseppi:
Is there somewhere I could see this in action? I tried going to `http://app.calc.docker/data-capture/price-lists#retired` on this branch but it didn't smoothly scroll to that id.

toolness:
Oh hrm, are we using smooth scrolling on that page?  I was testing it out using permalinks in the styleguide, e.g. `/styleguide/#buttons`.

jseppi:
Yeah, it uses the data-capture bundle, so it has smooth-scroll.

Smooth scrolling on page load also does not seem to be working in the styleguide for me o_O

toolness:
Hmm which browser are you using? Right now the smooth scroll logic is actually triggered as part of the manual scroll restoration logic, so it will only work on Chrome and Firefox, methinks.

jseppi:
Chrome 😭

toolness:
It will also smooth scroll only on the very *first* page load, btw--subsequent page loads won't be smooth scrolled, as that would get annoying. So you need to actually open the URL with the hash in a new tab, rather than open the un-hashed URL in a tab and click on a permalink and then reload the page...

jseppi:
Ooooooohhhhh, that must have been it. Sorry!

PR #1176, opened 2016-12-29
Ensure all config files use same python version

toolness:
This just loads a bunch of different configuration files (`.travis.yml`, `runtime.txt`, and `Dockerfile`) and ensures that they all specify the same version of Python to use, to maximize dev-prod parity.

I wanted to also add a test for postgres versions but noticed that we actually use 9.3 for travis and 9.4 for docker (!!!).


hourglass/tests/test_environment.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Version:
+    '''
+    Just a helper that makes it easy to do things with version numbers.
+
+        >>> v = Version('1.2.3')
+        >>> str(v)
+        '1.2.3'
+        >>> float(v)
+        1.2
```
toolness: the `float()` cast seems kind of pointless until we realize that it is actually how version numbers are represented in `.travis.yml` 😲

toolness:
Hmm, perhaps `test_configuration.py` is a better name than `test_environment.py`.

PR #1179, opened 2016-12-29
Adjust chart price callouts to accomodate new typeface

### LINKED ISSUE #1178: Font size too large in Proposed Price doohicky in chart
screen shot 2016-12-29 at 3 11 01 pm

Probably because we switched our typeface to Franklin without adjusting this element's font size.

jseppi:
Fixes #1178

It turned out that for the price callout thingies in the Data Explorer chart, the sizes and offsets are manually specified, which is not ideal in my opinion. However, I don't really want to mess with that code too much right now, so this PR just adjusts the sizes/offsets to accommodate our new typeface.

screen shot 2016-12-29 at 4 28 52 pm

I think the way to do it "properly" would be to use `textEl.node().getComputedTextLength()`, but I believe that requires the text node to actually be rendered first.


frontend/source/js/data-explorer/histogram.js
```diff
@@ -142,26 +142,27 @@ export default function updatePriceHistogram(data) {
       .attr('y', ppOffset - 25)
       .attr('x', -55)
       .attr('class', 'pp-label-box')
-      .attr('width', 110)
       .attr('height', 26)
+      .attr('width', 130)
```
jseppi: This `'width'` actually isn't used due to the `if` statement just below here which readjusts the size, but probably good to have the value specified be the same (`130`).

jseppi:
plz approve, @toolness

PR #1180, opened 2016-12-30
Adjust Proposed Price input width for new font

jseppi:
Before:

screen shot 2016-12-30 at 9 22 52 am

After:

screen shot 2016-12-30 at 9 22 45 am


jseppi:
plz review, @toolness

jseppi:
I think the whole thing needs to be reworked, honestly. Kinda weird to have such a large placeholder in that input.

Browser zoom seems to work fine with the px-specified width.

PR #1182, opened 2016-12-30
add jest env to tests/.eslintrc

jseppi:
`.eslintrc` files [cascade](http://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy), so I made a local one for our jest tests folder and added the jest `env` to it, which is a built-in config for setting up the jest globals.

Open against `reduxify-restore-excluded` branch


jseppi:
here you go, @toolness

toolness:
NOIIIICE 👌 thanks!

PR #1187, opened 2017-01-03
Use the X-Forwarded-Host header

toolness:
This hard-codes the use of the `X-Forwarded-Host` for now, which seems to be required on calc.gsa.gov because Django thinks all incoming requests to that are being made to calc-prod.app.cloud.gov and redirects users to it when building absolute URIs.

Ideally we should be exposing this as a configuration option to the environment IMO, but I discovered that we're also hard-coding `SECURE_PROXY_SSL_HEADER` (a change made in 3386469ba13487d9b1461398aacdf144b5adaa20, before the current team started working on the project), so for the sake of expediency it might be best to just hard-code this for now and add config options later.


jseppi:
Yeah, I _think_ that is all that is needed on our end.

PR #1193, opened 2017-01-04
Add hourglass.__version__ and tests for changelog.

toolness:
The `CHANGELOG.md` is nice because it's very accessible via github and a local repo checkout, it's human-readable, and it's easy to edit by hand.  However, because it repeats information a lot, it's also prone to become inconsistent. So this PR adds some tests that verify its basic integrity.

This PR also exposes the current version number at `hourglass.__version__` (which is imported from `hourglass.version`--I thought it'd be nice to make it a visible module in the filesystem for the sake of discoverability). We can use this to report the version number in `/healthcheck` and other places. A test has also been added to ensure that this version is in-sync with the latest version listed in the changelog.

Notes:

* This PR adds `beautifulsoup4` to `requirements.txt`, but it's not actually a new dependency--it was previously an implicit dependency via `robobrowser` but now that we're using it in our own first-party code, I've added it as an explicit dependency.

* [`packaging`](https://packaging.pypa.io/en/latest/version/) is a package that includes a parser/validator for [PEP 440](https://www.python.org/dev/peps/pep-0440/)-compliant version numbers, which are a superset of semver.  It just occurred to me that since we explicitly use semver (uh, I think we do at least), we might actually want to use [`semantic_version`](https://pypi.python.org/pypi/semantic_version/) instead of this package... Thoughts @jseppi ?


jseppi:
I have no idea what that weird PEP-440 versioning is, so unless you like it for some reason, then I'm more into semver versioning :)

toolness:
Haha ok I'll swap the PEP thing out for semver then :)

toolness:
Ok, just merged #1194 into this and replaced `packaging` with `semantic_version`. How does this look?

hourglass/version.py
```diff
@@ -0,0 +1 @@
+__version__ = '1.2.0'
```
jseppi: Oh, I guess a comment here about why this file exists in addition to `__version__.py` might be nice. I could see someone (me) wondering about that in the future.

jseppi:
Looks 👍

hourglass/version.py
toolness: Wait, there's a `__version__.py`? wat?

jseppi: lol, no, whoops, disregard

PR #1194, opened 2017-01-04
Report version number in /healthcheck.

toolness:
**Note: This is a PR against #1193, not `develop`.**

This adds the version number to the `/healthcheck` endpoint.

I'm actually a bit conflicted about whether this is a good place to put the version (it doesn't have to be the *only* place we put it, of course).  It's not *directly* related to the health of a deployment, but at the same time, it does give some significant *context* with which one could judge the deployment's health, so... I dunno. Feel free to just close this PR if you don't like it.

### DIFF
--- a/hourglass/healthcheck.py
+++ b/hourglass/healthcheck.py
@@ -3,6 +3,8 @@ from django.db.migrations.executor import MigrationExecutor
 from django.db import connections, DEFAULT_DB_ALIAS
 import django_rq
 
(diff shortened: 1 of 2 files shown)


jseppi:
I don't mind it being there, but I also can't think of how it would be used. Maybe a simple check that a deployment of a new version succeeded?

toolness:
Yeah, that seems useful! Noice.

PR #1201, opened 2017-01-04
Specify prod url in manifest

### LINKED ISSUE #1200: manifest-prod.yml needs to specify production url
Otherwise it will get dropped during a zero-downtime-deploy, as I just experienced.

jseppi:
Closes #1200 

I think


jseppi:
Ugh, no, I think this is wrong...

PR #1204, opened 2017-01-05
Specify prod domain via routes property

### LINKED ISSUE #1200: manifest-prod.yml needs to specify production url

jseppi:
Ok, I think this should work now. I'm using the [`routes` property](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest.html#routes) as recommended by Adam K.

For consistency, I've also changed the dev and staging manifests to use the same method of setting up their URLs even though they each only have a single URL (as opposed to prod, which has 2).

Closes #1200


jseppi:
I manually verified that this works on our staging deployment by specifying two separate `.app.cloud.gov` urls via `routes:`, and it worked, so I am pretty confident it will work for prod as well.

toolness:
Nice! I'm not fully clear on the details here but I trust yo' judgement and Adam K's advice.

PR #1211, opened 2017-01-05
use django's built-in settings for email from addresses

jseppi:
This PR removes our bespoke `SYSTEM_EMAIL_ADDRESS` setting in favor of Django's "built-in" `DEFAULT_FROM_EMAIL` and `SERVER_EMAIL`.

Note that all our cloud.gov environments will need their user-provided-services updated to include `DEFAULT_FROM_EMAIL` and `SERVER_EMAIL` after this PR is merged.


jseppi:
Ok, will merge tomorrow so I can make the UPS changes

PR #1215, opened 2017-01-06
Document release process and add associated tooling

toolness:
This adds a `RELEASE.md` file that documents how we can do releases.

To facilitate the release process, it also adds:

* A `manage.py releasehelp` command which generates a custom set of release instructions based on the target version number.

* A `manage.py unreleased_changelog` command which outputs the "Unreleased" section of `CHANGELOG.md` to stdout (useful for piping to `pbcopy` and pasting into a new PR).

* A `manage.py bump_changelog` command which makes it easier to "bump" `CHANGELOG.md` to reflect the latest version stored in `hourglass/version.py`.

Note that none of these management commands are *required* to perform a release; they're just tools that make the release process a bit more convenient.


hourglass/changelog.py
```diff
@@ -0,0 +1,102 @@
+import re
+import datetime
+import pathlib
+
+
+PATH = pathlib.Path(__file__).resolve().parent.parent / 'CHANGELOG.md'
```
toolness: I first learned about the `pathlib` module a few months ago, but I was looking at the [Python 3.4 `pathlib` docs](https://docs.python.org/3.4/library/pathlib.html), which claimed that it was a so-called "provisional" package:

> This module has been included in the standard library on a provisional basis. Backwards incompatible changes (up to and including removal of the package) may occur if deemed necessary by the core developers.

However, the [Python 3.6 docs](https://docs.python.org/3.6/library/pathlib.html) make no such claim, implying that the package is no longer provisional. Since the API seems largely unchanged, I figured we might as well use it, as it's way more convenient than `os.path`.

hourglass/changelog.py
```diff
@@ -0,0 +1,102 @@
...
+UNRELEASED_HEADER = '## [Unreleased][unreleased]'
+
+
+def get_contents():
+    '''
+    Return the contents of CHANGELOG.md.
+    '''
+
+    with PATH.open('r', encoding=ENCODING) as f:
+        return f.read()
```
toolness: Python 3.5 added [`pathlib.read_text()`](https://docs.python.org/3.6/library/pathlib.html#pathlib.Path.read_text), but we're still on 3.4...

toolness:
Doh, the `Add a cmd to make life easier` commit log message wasn't what I typed into my terminal--silly bash shell. Ah well, at least the message still makes some amount of sense on its own.

toolness:
Hmm, there's enough in this whole "changelog" thing to make it a separate Django app, really.  Especially if we eventually want to add an `/updates/` endpoint [à la cloud.gov](https://cloud.gov/updates/) that shows a nice view of the changelog embedded on the site itself... Hrm.

meta/management/commands/releasehelp.py
```diff
@@ -0,0 +1,64 @@
...
+
+    with (ROOT_DIR / html_filename).open('w', encoding='utf-8') as f:
+        f.write(HTML_TEMPLATE % dict(
+            version=v,
+            html=markdown.markdown(md)
+        ))
+
+    print('Wrote plaintext instructions to %s.' % markdown_filename)
+    print('Wrote HTML instructions to %s.' % html_filename)
+    print('Please open one of these to learn how release v%s.' % v)
```
toolness: It might also be nice to just open the HTML instructions in the user's web browser via [`webbrowser.open`](https://docs.python.org/3.6/library/webbrowser.html#webbrowser.open), but since the developer might be running it in a docker container, that complicates things.

toolness: I also considered making this a Django view instead of a management command, since the ideal way to read the customized instructions *is* in HTML, but I wasn't really sure if devs would want to be running CALC while making a release.

Also, I thought it might be nice to eventually package all these changelog-related utilities into a reusable Python package, and ideally that would be something decoupled from any particular web framework (or even the assumption that the product is web-based).

meta/tests/test_bump_changelog.py
```diff
@@ -0,0 +1,56 @@
+from unittest import TestCase
+from unittest.mock import patch
+from pathlib import Path
+from click.testing import CliRunner
```
toolness: Nice that Click comes with its own [testing utilities](http://click.pocoo.org/6/testing/)!

toolness: ... and that our commands appear to be subclasses of *both* Click commands and Django commands, apparently. 😅

.gitignore
```diff
@@ -29,3 +29,6 @@ docker-compose.override.yml
 cf-ssh.yml
 calc-deploy_manifest.yml
 calc-prod_manifest.yml
+tag-message-*.txt
```
jseppi: not a huge deal, but might also want to add these files to `.cfignore` also

meta/tests/test_bump_changelog.py
jseppi: Whoa, cool, I did not realize click had testing utils!

meta/tests/test_bump_changelog.py
```diff
@@ -0,0 +1,56 @@
...
+                with fakelog.open('r', encoding=changelog.ENCODING) as f:
+                    self.assertIn('9.0.0', f.read())
+
+                tagmsg = Path('tag-message-v9.0.0.txt')
+
+                with tagmsg.open('r', encoding=changelog.ENCODING) as f:
+                    self.assertIn('Fixed some stuff', f.read())
+
+
+del UtilTests  # So our test runner doesn't find and run them.
```
jseppi: o_O

toolness:
Just added the files to `.cfignore` too--if it looks OK feel free to merge!

Issue #1132, opened 2016-12-16
In Admin tables, datetimes are not localized


jseppi:
Looks like maybe they are UTC?

<img width="430" alt="screen shot 2016-12-16 at 11 25 38 am" src="https://cloud.githubusercontent.com/assets/697848/21272985/1c0f0d8a-c387-11e6-9d8e-2fab746a74a5.png">

@toolness: ideas on this one?

jseppi:
@toolness: just noticed that in our `settings.py`, we have `TIME_ZONE = 'UTC'`. That's probably why it is all in UTC.

toolness:
Er what should they be localized to, though?  EST?

jseppi:
It would be cool if they were shown localized to the current user's location, but probably not necessary. Some kind of indication that they are UTC might be an acceptable fix.

PR #1217, opened 2017-01-09
Display datetimes in emails localized to US/Eastern time

### LINKED ISSUE #1189: Emails - specify that datetimes are UTC or adjust to be in Eastern
Otherwise the content is a bit confusing since a user would likely think the time shown is localized.

screen shot 2017-01-03 at 3 27 03 pm

jseppi:
This PR adds a new template filter `readable_timestamp` to display datetimes in a localized, readable format, including the timezone code. By default, they are displayed in "US/Eastern", e.g.: `'Jan. 8, 2017, 3:51 p.m. (EST)'`. The new filter is used in all the emails to display `created_at` datetimes.

Closes #1189


data_capture/templatetags/data_capture.py
```diff
@@ -0,0 +1,20 @@
+from django import template
+from django.utils.formats import localize
+from pytz import timezone
+
+register = template.Library()
+
+
+@register.filter
+def readable_timestamp(value, tz_name="US/Eastern"):
```
jseppi: I couldn't think of a better name. Happy to entertain other options!

data_capture/tests/test_email.py
```diff
@@ -15,14 +18,16 @@ class EmailTests(ModelTestCase):
 
     def test_price_list_approved(self):
         price_list = self.create_price_list(
-            status=SubmittedPriceList.STATUS_APPROVED)
+            status=SubmittedPriceList.STATUS_APPROVED,
+            created_at=timezone.make_aware(datetime(2017, 1, 8, 20, 51, 0)))
```
jseppi: "timezone-aware" dates must be used to do timezone conversions/localizations

requirements.txt
```diff
@@ -21,3 +21,4 @@ django-debug-toolbar==1.5
 rq==0.6.0
 django-rq==0.9.2
 rq-scheduler==0.7.0
+pytz==2016.10
```
jseppi: necessary to use nice timezone names, like "US/Eastern"

data_capture/templatetags/data_capture.py
toolness: I think that's fine.  Django's own [`humanize`](https://docs.djangoproject.com/en/1.10/ref/contrib/humanize/) package prefixes their names with `natural` but I guess we don't want to confuse our version with theirs.

What if it was `tz_timestamp`?  That would convey that it's a timestamp that has timezone info in it, maybe.  I dunno.

data_capture/templatetags/data_capture.py
```diff
@@ -0,0 +1,20 @@
...
+    '''
+    This template filter formats the given timezone-aware datetime into
+    a more readable format (ex: "Jan 1, 2015, 1:55 p.m. (EST)"),
+    localized to the timezone specified by tz_name.
+    '''
+    try:
+        tz = timezone(tz_name)
+        timestamp = value.astimezone(tz)
+        return "{} ({})".format(localize(timestamp), timestamp.tzname())
+    except Exception as e:
```
toolness: Is there any way we could be a bit more precise in the exception type here?  Like catching a `ValueError` or something?  I have a knee-jerk aversion to blanket `Exception` catching but if there's nothing more specific we can use, I'm OK with it, especially if we've got good unit tests.

jseppi: I actually copied this pattern from Django's source https://github.com/django/django/blob/master/django/templatetags/tz.py#L56, but I see in [humanize.py](https://github.com/django/django/blob/master/django/contrib/humanize/templatetags/humanize.py#L117) they are more specific with which exceptions are caught. I'll amend to follow that pattern instead.

toolness: whoaaa funky!!!!! um cool

jseppi:
> Wait, is there a reason we're not using settings.TIME_ZONE to define the preferred time zone of the app, and using that as the default zone in the readable_timestamp filter instead of hard-coding US/Eastern there and having TIME_ZONE be UTC

I don't know what the original reason for `TIME_ZONE` be UTC was. Probably just that it appears to be the [default value specified in the Django project template](https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-TIME_ZONE). I feel like mucking with `TIME_ZONE` now _might_ not be a great idea, especially since we already have data in prod (though since it is only R10 and S70 bulk data I guess not that big a deal). I would say I kinda like the simplicity of having it set to UTC. Here's another note from Django's docs [arguing for UTC](https://docs.djangoproject.com/en/1.10/topics/i18n/timezones/#selecting-the-current-time-zone):

> Even if your website is available in only one time zone, it’s still good practice to store data in UTC in your database. The main reason is Daylight Saving Time (DST). Many countries have a system of DST, where clocks are moved forward in spring and backward in autumn. If you’re working in local time, you’re likely to encounter errors twice a year, when the transitions happen.

I suppose another benefit of setting `TIME_ZONE` to something other than UTC is that datetimes in the Admin views would be rendered in something more understandable (as reported in #1132).

I will also admit I'm having some difficulty time all of Django's timezone settings and behavior. For instance, from the [docs about `TIME_ZONE`](https://docs.djangoproject.com/en/1.8/ref/settings/#time-zone):

> When USE_TZ is False, this is the time zone in which Django will store all datetimes. When USE_TZ is True, this is the default time zone that Django will use to display datetimes in templates and to interpret datetimes entered in forms.

So how are they *stored* when `USE_TZ=True` AND `TIME_ZONE='US/Eastern'` (or whatever)? The docs don't seem to say. I could be overcomplicating things though...

jseppi:
Ok, I just did a test (locally) where I set `USE_TZ=True` and `TIME_ZONE='US/Eastern'`. I created a new R10 `BulkUploadContractSource`, and checked in the DB, where the `created_at` timestamp was in UTC - so that's good!

So, in short, using TIME_ZONE='US/Eastern' probably _is_ a good way to go 👍

jseppi:
Oh geez, actually changing `TIME_ZONE='US/Eastern'` breaks every test that dates/datetimes are examined...

```
15 failed, 250 passed, 17 skipped in 25.70 seconds
```

jseppi:
heh, good thing I tried to this at the current time, which actually is tomorrow in UTC, otherwise most of these tests probably would have passed until someone else test lated in the day.

This is probably a good opportunity to modify all those tests to use https://github.com/spulec/freezegun

toolness:
Cool--I agree that Django's time zone configuration is super confusing--maybe we should bring up our solution in #django on slack just to make sure it's OK?

toolness:
Ack sorry, didn't mean to close.

jseppi:
Ok, update per slack discussion: Going to keep `TIME_ZONE='UTC'` and just modify the views/templates to localize to 'US/Eastern' for now. That will leave open (an easier route) to possible future changes to localize per user.

I still want to add `freezegun` to our tests, but will do that in a separate PR.

jseppi:
I ended up deciding I liked your `tz_timestamp` name suggestion so I went ahead and made that change as well :)

Issue #766, opened 2016-09-18
Upgrade to Python 3.5


toolness:
I recently found a Python 3.5 feature that I wanted to use in CALC, but we're on 3.4.  Specifically, it's [PEP 448 - Additional Unpacking Generalizations](https://www.python.org/dev/peps/pep-0448/), which allows us to write things like this:

```
>>> dict(**{'x': 1}, y=2, **{'z': 3})
{'x': 1, 'y': 2, 'z': 3}
```

I think this could be particularly nice for easily combining a bunch of different Django template context-like dictionaries into a single context.

jseppi:
Sweet. I think this is the current Python buildpack that cloud.gov is using: https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.5.9

Just need to change `runtime.txt` for cloud.gov (and `.travis.yml` for Travis).

toolness:
Well, 3.6 came out recently, and it's one of the biggest releases since 3.0 (includes 16 PEPs). It'd be nice if we could just leapfrog to that.

Also, 3.5 supports type annotations, which would let us start using [mypy](http://mypy-lang.org/) to lint our code, which could be nice.

jseppi:
Yeah, 3.6 would be cool!

Just checked on GovCloud-cloud.gov. Current python buildpack is 1.5.13, which unfortunately doesn't have Python 3.6: https://github.com/cloudfoundry/python-buildpack/releases/tag/v1.5.13

toolness:
Oh, this was fixed by #1219. Closing!

PR #1219, opened 2017-01-10
Upgrade to Python 3.5

toolness:
This upgrades CALC to work on Python 3.5, fixing #766.

The upgrade actually surfaced a bug in two of our tests: we were calling `assert` methods on `Mock` objects that were  introduced in Python 3.6, but Python 3.4 wasn't throwing an `AttributeError` on them because the mock assumed any nonexistent method calls were mock calls!  Fortunately they fixed this in [Python bug 21238](https://bugs.python.org/issue21238) so that a `Mock` *would* throw an `AttributeError` if `assert` (or a common misspelling, `assret`) was in the attribute name, which thus made our tests break.

### DIFF
--- a/data_capture/tests/test_price_list_views.py
+++ b/data_capture/tests/test_price_list_views.py
@@ -135,13 +135,13 @@ class DetailsViewTests(ProtectedViewTestCase):
         self.assertEqual(pl.contract_end, datetime.date(1989, 4, 14))
         self.assertEqual(pl.escalation_rate, 0)
 
-    @mock.patch.object(SubmittedPriceList, 'retire')
-    def test_valid_post_calls_retire(self, mock):
+    @mock.patch.object(SubmittedPriceList, 'unreview')
+    def test_valid_post_calls_unreview(self, mock):
         self.login()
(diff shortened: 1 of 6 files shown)


data_capture/tests/test_price_list_views.py
```diff
@@ -135,13 +135,13 @@ def test_post_updates_price_list(self):
         self.assertEqual(pl.contract_end, datetime.date(1989, 4, 14))
         self.assertEqual(pl.escalation_rate, 0)
 
-    @mock.patch.object(SubmittedPriceList, 'retire')
-    def test_valid_post_calls_retire(self, mock):
+    @mock.patch.object(SubmittedPriceList, 'unreview')
+    def test_valid_post_calls_unreview(self, mock):
```
toolness: I realized that `retire` wasn't actually being called, but `unreview` was, so I renamed the test... hope that was the right way to fix this, aside from the `assert` method fix.

jseppi: Whoa, great!

PR #1220, opened 2017-01-10
Add experimental mypy integration.

toolness:
~~**Note: this PR is against #1219, not `develop`.**~~

This adds experimental [mypy](http://mypy-lang.org/) integration to CALC by adding it as one of `ultratest`'s subcommands, running it on travis, and telling it to only check a whitelist of files for now (specified by `mypy-files.txt`).

Notes:

* In Python 3.5 and above, type annotations are "real code" that's just thrown out of the AST during execution.  The downside of this is that some of our annotations need to come from mypy's `typing` package, which means that we need to include mypy in our *production* requirements, even though we're not actually using it.

* Because we need to import stuff from `typings`, and because in some cases we can only annotate via comments (as true variable annotations were only just introduced in Python 3.6), this means that `flake8` can complain about unused imports from `typings`, so we have to `# NOQA` those for now, which is a bit annoying. More details in [pycqa/flake8#118](https://gitlab.com/pycqa/flake8/issues/118).

* `docker_django_management.py` can be executed via `manage.py`, which can be run from the host system, which may only have Python 2 installed.  Because of this, I wanted to keep the file Python 2-compatible and used [mypy's Python 2 comment style](http://mypy.readthedocs.io/en/latest/python2.html)  for it.


hourglass/settings.py
```diff
@@ -130,7 +131,7 @@
     'styleguide',
     'meta',
     'frontend',
-)
+)  # type: Tuple[str, ...]
```
toolness: Agh, by default mypy seems to treat tuples like the way they are in strongly-typed languages, rather than as immutable lists, which is how we're using them here.  So we have to add a funky type annotation to convince mypy that this is a tuple of an arbitrary number of strings, or else it will complain when we later add stuff to it. (An alternative would be to simply make these lists instead of tuples, I guess.)

toolness:
I guess this seems ok for a first attempt at integration...

@jseppi what do you think of this?  Too weird?

I guess mypy won't be terribly useful because there aren't any official type annotations for Django.  There is [mypy-django](https://github.com/machinalis/mypy-django) but I'm not sure how maintained it is.

api/utils.py
```diff
@@ -10,15 +11,15 @@ def get_histogram(values, num_bins=10):
         raise ValueError('num_bins must be greater than 0')
 
     # convert values to floats
-    values = [float(v) for v in values]
+    fvalues = [float(v) for v in values]
```
toolness: Had to rename the variable because mypy doesn't like it when the type of a variable changes, I guess.

mypy-stubs/dotenv.pyi
```diff
@@ -0,0 +1,3 @@
+from typing import Optional
+
+def load_dotenv(dotenv_path: str) -> Optional[bool]: ...
```
toolness: Looking at the source for dotenv, it seems this function returns `None` if no `.env` file was found, but `True` if it was found or loaded.  We don't actually use the return value regardless, but I figured we might as well try to be accurate...

mypy.ini
```diff
@@ -0,0 +1,13 @@
+[mypy]
```
toolness: Docs for all these options can be found at http://mypy.readthedocs.io/en/latest/config_file.html.

jseppi:
Hrm, given that `mypy` is experimental and doesn't have bindings for Django, how useful do _you_ think it will be? I would say I like the idea of its promise of type checking, but don't know if the added overhead will be worth it for us, given the caveats mentioned.

toolness:
A FINE QUESTION INDEED.

So while mypy itself is indeed experimental, the type annotation syntax *and* its semantics are not: they're actually standardized in [PEP 484](https://www.python.org/dev/peps/pep-0484/).  So even if mypy decides to suddenly take a completely different direction or something, odds are that it won't suddenly render PEP 484 invalid.  (And it's primarily PEP 484 that guides the annotation of our own code, rather than anything mypy-specific. Um, I think.)

Another strong indicator about mypy is the fact that it's currently being worked on by Python's creator, Guido Van Rossum, and it's being actively used by his employer (Dropbox).  So it's not a side project that random folks are doing on their nights and weekends or anything.

It's also being used at Google and Facebook with success, FWIW: https://github.com/python/typeshed/issues/739

The other thing I'd argue is that the primary usefulness of Python type annotations really comes from its role as "executable documentation"--e.g. being able to look at a function's signature and immediately know what its inputs and output are.  In that context, mypy could really be regarded as more of an "documentation verifier", I guess.

The last thing is that I like the fact that mypy has been built to be something that projects like ours can incrementally adopt.  I like the fact that we can just have it run on these 3 or 4 files right now, and not use it anymore if we don't feel like it.  Unlike TypeScript, which is its own separate language that actually transforms our code, mypy doesn't really carry much of a maintenance burden with it--we can remove it from our build system and even leave the existing type annotations in our code and nothing changes.  So it's not very risky to adopt.  And if it *is* in there, I really like that I have the option to annotate any new code that I write.

So I guess overall, I think it's worth keeping it around in the codebase for now.  If it gets annoying to modify the handful of annotated files because of mypy errors, we can always either make the mypy rules more lax (I've actually ratcheted them up quite a bit in this PR) or we can do away with mypy entirely.

What you say?

docker_django_management.py
```diff
@@ -69,6 +69,9 @@
 import signal
 import subprocess
 
+if False:
+    # This is just needed so mypy will work; it's never executed.
```
jseppi: Why is this in an `if False:` block here but not in `utils.py`?

toolness: Oh it's annoying. Basically we want `docker_django_management.py` to work even if mypy isn't installed (as it will be run on the host system, which might just have vanilla python 2 on it).  The only way to do that without angering mypy or flake8 is to put the import inside an `if False` statement so it never gets executed, but so mypy sees it, and so flake8 doesn't complain.

It's weird, but this hack was actually [suggested by Guido himself](https://gitlab.com/pycqa/flake8/issues/118#note_3848035), so... 😁

toolness:
wooooo

PR #1224, opened 2017-01-10
Refactor WhiteListPermissions to not take array position into account

jseppi:
This PR is a semi-rewrite of the `WhiteListPermissions` class. That class used to check that _ONLY_ the penultimate IP in X-Forwarded-Hosts was in the list of whitelisted IPs. I've modified it to instead check that _ANY_ IP in that header is in the list of whitelisted IPs.

I also added tests since there were none.


api/tests/test_permissions.py
```diff
@@ -0,0 +1,64 @@
...
+        w = WhiteListPermission()
+        req = mock.MagicMock(
+            META={
+                'HTTP_X_FORWARDED_FOR': '5.5.5.5, 2.2.3.3',
+                'REMOTE_ADDR': '5.6.7.8'
+            }
+        )
+        self.assertFalse(w.has_permission(req, None))
+
+    def test_it_prefers_forwarded_for_to_remote_addr(self):
```
jseppi: Not sure if the behavior this tests for is absolutely necessary, but it was how the original implementation of `WhiteListPermission` worked so I kept the same behavior.

api/permissions.py
```diff
@@ -8,20 +8,23 @@ class WhiteListPermission(permissions.BasePermission):
...
+        ip_addresses = []
+
         forwarded = request.META.get('HTTP_X_FORWARDED_FOR')
         if forwarded:
             try:
-                ip_addr = forwarded.split(',')[-2].strip()
+                ip_addresses = [f.strip() for f in forwarded.split(',')]
             except:
-                ip_addr = forwarded
+                ip_addresses = [forwarded]
```
jseppi: I don't quite understand what the original logic here (the `except` block) is for. Seems like it would happen if `forwarded` is not a string, so it should probably be specified as an `AttributeError`. But I can't think of a case where `forwarded` would NOT be a string, so maybe the entire `except` block should be removed. Thoughts, @toolness?

toolness: Hmm, in the original code, it was taking the penultimate forwarded IP via `[-2]`, which would result in an `IndexError` if there was only zero or one comma-separated terms there. So I am guessing that `except` should have been specified as an `IndexError`?

Since you're not taking a particular index, though, it seems like you can get rid of the `try`/`except` entirely.

api/tests/test_permissions.py
toolness: Noice, gud idea.

api/permissions.py
jseppi: ah yeah, guess it could have been that also! Will remove 👍

toolness:
Noice, feel free to merge at will!

jseppi:
image

PR #1234, opened 2017-01-11
Install node modules in Docker container at /calc/node_modules; Implement airbnb react eslinting.

toolness:
This fixes #569.

However, there are currently at least two places where we've disabled the `import/no-unresolved` rule (our gulpfile and ultratest) due to the fact that prior to this, we installed our node modules in an unusual location, and eslint doesn't support that because of https://github.com/benmosher/eslint-plugin-import/issues/142.  When trying to remove these Docker-specific conditionals, eslint still failed--not because it couldn't find our third-party node modules, but because it couldn't find any of our first-party modules with `.jsx` extensions!

awgpanpgiwenigrwhniperpmehti

I fiddled with trying to get eslint to recognize our modules with `.jsx` extensions via `eslint-plugin-import`'s [`import/extensions` setting](https://www.npmjs.com/package/eslint-plugin-import#importextensions) but that didn't seem to fix it.

So this just changes the `Dockerfile`.

PR #1235, opened 2017-01-11
Use airbnb-eslint-config and fix identified errors

jseppi:
While I would prefer to not have any deviances from the default airbnb config, I did add a few to our `.eslintrc`. We can revisit them in other issues/PRs, but they are pretty minor things:

Deviances from the default config:
- ~~**comma-dangle#functions** - I don't get why you'd want dangling commas in function param lists~~ Edit: I set this one back to default.
- **no-plusplus** -- ehhhhh, I find `++` and `--` very understandable
- **import/no-named-as-default** - this rule doesn't really work with Redux, and the author suggests disabling it in that case: https://github.com/benmosher/eslint-plugin-import/issues/544#issuecomment-244976007
- **react/forbid-prop-types** - this is a documentation thing, not something that would negatively impact code, according to its own documentation: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md#when-not-to-use. We can revisit later if desired.


jseppi:
Hrm, I expected the builds to fail `eslint` because I hadn't fixed most of the errors in jsx files. Need to figure out why they didn't fail o_O

toolness:
So one thing I noticed while working on #1220 is that we don't *explicitly* run `eslint` in `.travis.yml`... I mean, I think we implicitly run it as a part of running `gulp build`, but I don't think `gulp build` exits with a nonzero exit code if eslint fails.  I think we don't explicitly run `eslint` because we could on codeclimate to run it for us in a "complain if anything is amiss" sort of way, but I'm not sure...

jseppi:
Oh, well that explains why Travis isn't failing, duh - thanks!

And CodeClimate doesn't actually use our `.eslintrc`. So I guess that means we should call `ultratest eslint` in our travis builds.

jseppi:
yay, it failed!

jseppi:
CodeClimate doesn't appear to be chiming in on this PR, unsure why...

frontend/source/js/data-explorer/components/education-level.jsx
```diff
@@ -74,34 +87,17 @@ export class EducationLevel extends React.Component {
...
-          <input
-            id={id}
-            type="checkbox"
-            value={value}
-            checked={levels.indexOf(value) >= 0}
-            onChange={this.handleCheckboxClick.bind(this, value)}
-            name="education" />
-          <label htmlFor={id}>{EDU_LABELS[value]}</label>
-        </li>
+        <EducationLevelItem
```
jseppi: Replaced with a _brand new_ component! This was necessary to get rid of the `.bind` in the `onChange` handler, as described at https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md

frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -320,11 +320,18 @@ function updateHistogram(rootEl, data, proposedPrice, showTransition) {
   stdDev.select('.label.max .stddev-text')
     .text(formatDollars(stdDevMax));
 
+  const trunc = (num) => {
```
jseppi: New function to replace use of `~~number` below. Much more readable now.

frontend/source/js/data-explorer/components/results-table/index.jsx
```diff
@@ -35,9 +35,12 @@ export class ResultsTable extends React.Component {
       .filter(r => !!priceForContractYear(this.props.contractYear, r))
       .map(result => (
         <tr key={result.id}>
-          {COLUMNS.map((col, i) => (
-            <col.DataCell key={i} sort={this.props.sort} result={result} />
-          ))}
+          {COLUMNS.map((col) => {
+            const cellKey = `${result.id}-${col.DataCell.cellKey}`;
```
jseppi: this change (and the similar one below for the header cells) was what I came up with to remove the use of the  index value from the `key` prop. Using an index in `key` is a no-no: ndex-key.md

jseppi:
Alrighty, @toolness, ready for review! Sorry it's such a large diff. Most changes were automatic via `eslint --fix`.

frontend/source/js/data-explorer/components/results-table/index.jsx
toolness: Oh, I had thought it would be OK to use the index in this particular case since our columns never actually change (at least in the sense that they never get added or removed or re-arranged), but I agree this is a good practice in general. Probably best to follow all the time, since if we ever decided that we *do* want to add/remove/re-arrange columns, we wouldn't want any bugs creeping in, eh?!

toolness:
> **comma-dangle#functions** - I don't get why you'd want dangling commas in function param lists

Does it only apply to multi-line function param lists?  If so, that'd actually make it consistent with how we deal with multi-line key-value lists in object literals... I actually frequently get confused between which kinds eslint *wants* the commas at the end of and which it doesn't.

I had also thought the rationale for dangling commas was so that you could easily add new lines at the end without having to worry about adding the comma at the beginning of the previous line, which also improves the readability of diffs, but I dunno.  I'm fine either way.

frontend/source/js/common/dap-hacks.js
```diff
@@ -65,7 +65,7 @@ export function observe(
   parentEl = window.document.documentElement,
   getInitAutoTracker = () => window[DEFAULT_INIT_AUTO_TRACKER_NAME]
 ) {
-  const observer = new window.MutationObserver(mutations => {
+  const observer = new window.MutationObserver((mutations) => {
```
toolness: Why the added parens?  I thought eslint was just telling me the opposite the other day!  Make up your mind, eslint!!!!!

frontend/source/js/data-explorer/components/app.jsx
```diff
@@ -138,18 +144,20 @@ class App extends React.Component {
...
                 <div className="download-buttons row">
                   <div className="four columns">
-                    <a className="button button-primary"
-                       id={prefixId('download-histogram') /* Selenium needs it. */}
-                       href="#"
-                       onClick={this.handleDownloadClick}>
+                    <a
+                      className="button button-primary"
+                      id={prefixId('download-histogram') /* Selenium needs it. */}
+                      href=""
```
toolness: I've noticed that some browsers won't apply the traditional "pointer" cursor to `<a>` elements that have `href` set to an empty value... will that happen here?

toolness: Nevermind, just tried this on the site and it's all gud!

frontend/source/js/data-explorer/components/histogram.jsx
toolness: whoa I had no idea that [`~` was a thing](https://www.joezimjs.com/javascript/great-mystery-of-the-tilde/).

frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -390,6 +397,14 @@ function updateHistogram(rootEl, data, proposedPrice, showTransition) {
...
 
 class Histogram extends React.Component {
+  componentDidMount() {
+    this.updateHistogram(false);
+  }
+
+  componentDidUpdate() {
+    this.updateHistogram(true);
+  }
+
```
toolness: Is there a particular reason these two methods got moved to the top of the class declaration?  Does the linter prefer lifecycle methods to be at the top of class declarations now, or was this just a weird artifact of one of the automated fixes?

frontend/source/js/data-explorer/components/results-table/contract-number-column.jsx
```diff
@@ -26,7 +28,7 @@ function createGsaAdvantageUrl(contractNumber) {
 export const DataCell = column.connectDataCell(
   ({ className, value }) => (
     <td className={className}>
-      <a target="_blank" href={createGsaAdvantageUrl(value)}>
+      <a target="_blank" rel="noopener noreferrer" href={createGsaAdvantageUrl(value)}>
```
toolness: Hmm I understand `noopener` may be needed for security reasons but why `noreferrer`?  Seems like gsa advantage should know that all its traffic is coming from CALC, muahahahaha.

frontend/source/js/data-explorer/tests/__snapshots__/description.test.jsx.snap
```diff
@@ -9,28 +9,34 @@ exports[`<Description> matches snapshot 1`] = `
    results with 
   <span
     className="filters">
-    <DescriptionFilter>
+    <DescriptionFilter
+      extraClassName={null}
```
toolness: Since `extraClassName=null` is in `defaultProps`, isn't it ok to leave it out of here?  Or are you including it to make things explicit?

toolness:
This looks good!  I noticed there was some stuff in the diff that appeared to already be in `develop` so I merged #1234 with `develop`, hoping that it'd update the diff for this too, but it ended up causing a merge conflict... hrm.

jseppi:
> **comma-dangle#functions**
>> Does it only apply to multi-line function param lists? 

I turned it back on and amended all files (almost all via `eslint --fix`). What I didn't like was that it made code like this (from gulpfile.js):

```js
    gulp.task(browserifiedBundleName, () =>
      browserifyBundle(
        path.join(dirs.src.scripts, paths[entryName]),
        dirs.dest.scripts[name],
        paths[outfileName],
      ),  // <--- node would actually fail when running code like this via gulp build
    );
```

But this reformatted version is ok:

```js
    gulp.task(browserifiedBundleName, () =>
      browserifyBundle(
        path.join(dirs.src.scripts, paths[entryName]),
        dirs.dest.scripts[name],
        paths[outfileName]));
```

I like the formatting of the first one better, but eh, the point is to not worry so much about formatting and just follow the rules, right?

frontend/source/js/data-explorer/components/app.jsx
jseppi: I suppose all these blank `href`'d anchors should really be `button`s, right? Glad the pointer styling still works!

frontend/source/js/data-explorer/components/results-table/contract-number-column.jsx
jseppi: I think `noreferrer` is necessary for older browsers. The rule enforces having both values: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md

frontend/source/js/data-explorer/tests/__snapshots__/description.test.jsx.snap
jseppi: This is just an auto-generated snapshot dawg :)

.travis.yml
```diff
@@ -23,8 +23,7 @@ install:
 - npm install
 script:
 - phantomjs --version
-- python manage.py ultratest flake8
-- python manage.py ultratest mypy
+- python manage.py ultratest flake8 eslint mypy
```
jseppi: I went ahead and combined all three ultratest tests in this one call, @toolness

PR #1237, opened 2017-01-12
Fix dangling comma in labor category description

jseppi:
Use a regex to remove an ending comma and trailing whitespace from `laborCategory` in the `Description` component.

Closes https://github.com/18F/calc/issues/1236

PR #1238, opened 2017-01-12
If sauce connect tunnel fails, fall back to phantomJS.

### LINKED ISSUE #978: Sauce on Travis timeouts happen relatively frequently
They are frustrating because they mark PRs as having failed travis builds :(

This old travis-ci thread suggests using `saucelabs.com:4445` instead of `ondemand.saucelabs.com`:

https://github.com/travis-ci/travis-ci/issues/2718#issuecomment-53583278

toolness:
This is an attempt to fix #978 by abandoning the Travis Sauce Connect add-on, but reusing its shell script to attempt to establish a tunnel when running `py.test`, and falling back to PhantomJS if that fails.

Reusing its shell script is weird, btw.  Reading the [ruby Travis add-on source](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/addons/sauce_connect.rb), it seems like Travis is `source`'ing it into a subshell and then communicating with it, so I've duplicated that kind of behavior with a handy module I found called [pexpect](https://pexpect.readthedocs.io/en/stable/index.html).


toolness:
Wait, wut?  The tests run fine but bandit is complaining about data capture stuff that's not even changed in this PR...

jseppi:
yyyyuuuupppp, see my comments and fix in https://github.com/18F/calc/pull/1237 :)

frontend/tests/sauce_connect.py
```diff
@@ -0,0 +1,118 @@
+import os
+import sys
+import subprocess
+import traceback
+import pexpect
+
+MY_DIR = os.path.abspath(os.path.dirname(__file__))
+
+# This script was taken directly from:
+# https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh
```
jseppi: maybe put this message, along with any license info, down in the actual `sauce_connect.sh` file itself.

toolness: I was thinking of doing that, except that I like keeping those vendored files as pristine as possible so we can just copy/paste new versions in their place without worrying about our changes getting clobbered... but I guess this comment should be pretty easy to see and preserve at the very top of the file.

jseppi: Yeah, and if someone does like a GitHub code search and finds that file _without_ any provenance/license info, it could be taken badly.

toolness: O A FINE POINT SIR.

toolness:
Right on, it looks like this is working.  Ready for review I suppose.

The one downside of this approach is that if our sauce tunnel ever starts failing for *legitimate* reasons, e.g. because our sauce access key became invalid, we wouldn't really know unless we looked at a travis build log and read the output.

toolness:
@jseppi hmm it says that you requested changes but there doesn't seem to be a way for me to say "yo i made the changes buddy".

toolness:
yo i made the changes buddy

jseppi:
haha, yeah, I see you did. I still have to do an actual review :P

frontend/tests/sauce_connect.py
```diff
@@ -0,0 +1,139 @@
...
+SAUCE_CONNECT_SH_PATH = os.path.join(MY_DIR, 'sauce_connect.sh')
+
+LOAD_SAUCE_CONNECT = 'source {}'.format(SAUCE_CONNECT_SH_PATH)
+
+PROMPT = "BASH IS READY > "
+
+TIMEOUT = 30  # In seconds.
+
+
+def get_last_exitcode(child):  # type: (pexpect.spawn) -> int
```
toolness: I ended up making the annotations Python 2-friendly so that we can easily factor out this module into something that can be used by the wider Python/Sauce community, if we want.

jseppi:
Alright, @toolness - let's give it a whirl!

PR #1239, opened 2017-01-12
Add mypy annotations to ultratest

toolness:
Since `ultratest` is a pretty crucial command that we don't have tests for, I thought it might be a good candidate for mypy-ification.  It actually revealed a few potential bugs in the code, which is neat.


meta/management/commands/ultratest.py
```diff
@@ -105,45 +108,31 @@ def get_coverage():
...
-        'name': 'jest',
-        'cmd': 'npm test'
-    },
+    TestType(name='flake8', cmd='flake8 --exclude=node_modules,migrations .'),
+    TestType(name='eslint', cmd=ESLINT_CMD),
+    TestType(name='bandit', cmd='bandit -r .'),
+    TestType(name='py.test',
+             cmd='py.test --cov-report xml --cov-report term --cov'),
+    TestType(name='mypy', cmd='mypy @mypy-files.txt'),
+    TestType(name='jest', cmd='npm test'),
```
toolness: There isn't a way to tell mypy about the "shape" of a `dict`, but mypy *does* have support for [named tuples](https://docs.python.org/3/library/collections.html#collections.namedtuple), which is more readable anyhow, since we don't have to use quotes and brackets all over the place (`foo.name` instead of `foo['name']`).

meta/management/commands/ultratest.py
```diff
@@ -67,34 +69,35 @@
...
         package = root.find('packages/package')
-        return float(package.attrib['branch-rate'])
+        if package is not None:
+            return float(package.attrib['branch-rate'])
     except Exception as e:
-        return None
+        return 0.0
 
-    return None
+    return 0.0
```
toolness: This was one of the bugs mypy found!  Originally I had the return type of this be `Optional[float]`, since this function returned `None` if anything weird happened.  However, mypy then rightly complained later on in the code, where we subtracted the return value of this function from another float.

I had the option of changing the other code to have some additional branching logic to see if the result of this function was `None` (mypy *does* support type narrowing, which is cool), but instead I just forced the return value of this function to always be a float.  This does mean that the coverage will appear to be 0% on rare occasions, I suppose, but at least it's better than throwing an exception because it couldn't subtract `None` from a float!

meta/management/commands/ultratest.py
```diff
@@ -1,9 +1,11 @@
 import sys
 import tempfile
 import subprocess
-import djclick as click
+from collections import namedtuple
+import djclick as click  # type: ignore
```
toolness: It's nice that mypy supports [ignoring entire imports](http://mypy.readthedocs.io/en/latest/common_issues.html#spurious-errors-and-locally-silencing-the-checker) so that we don't have to make stubs if we don't want to.  I figured it wasn't a big deal with `djclick` since its code is executed in very few places, and it's executed very often, so any weird errors will be immediately noticeable by us. I also didn't want to have to dig into how to do type annotations for decorators, lol.

(That said, it's cool that typeshed has annotations for [click on Python 3.6](https://github.com/python/typeshed/tree/master/third_party/3.6/click), but sadly nothing for *dj*click on 3.5.)

meta/management/commands/ultratest.py
```diff
@@ -67,34 +69,35 @@
...
+def get_coverage() -> float:
     COVERAGE_FILE = 'coverage.xml'
 
     try:
         tree = etree.parse(COVERAGE_FILE)
         root = tree.getroot()
         package = root.find('packages/package')
-        return float(package.attrib['branch-rate'])
+        if package is not None:
+            return float(package.attrib['branch-rate'])
```
toolness: This is sort-of a bug that mypy found; basically the type of `package` without the conditional I added is `Optional[Element]`, so mypy was rightly complaining that `package.attrib` would throw if `package` happened to be `None`.  So I added a check to see if it was `None`, and mypy's type narrowing kicked in and it was happy.

The addition of the branch might have actually rendered the try block this code is in useless, I'm not sure, but I figured I'd leave it in just in case.

meta/management/commands/ultratest.py
jseppi: Yeah, this is definitely better!

meta/management/commands/ultratest.py
jseppi: Nice catch!

PR #1242, opened 2017-01-13
Add mypy annotations for schedules/base.py

toolness:
I thought it might be nice to add mypy annotations for `BasePriceList` and its subclasses; this way we'd get mypy warnings about e.g. a subclass not presenting an interface that properly conforms to its abstract base class.


mypy-stubs/django/utils/safestring.pyi
```diff
@@ -0,0 +1,8 @@
+from typing import NewType, Union
+
+
+SafeString = NewType('SafeString', str)
```
toolness: I'm not sure if this is actually the best way to proceed, but I thought it might be nice for `SafeString` to be treated as a distinctly different type from `str` through the use of [`NewType`](http://mypy.readthedocs.io/en/latest/kinds_of_types.html?highlight=newtype#newtypes).  This could make it easier for us to distinguish between code that's supposed to return safe strings vs. unsafe strings that need escaping.

jseppi: Seems like a good idea!

PR #1243, opened 2017-01-13
add envify maybe i have no idea

### LINKED ISSUE #1225: process.env.NODE_ENV might not be set to "production" on builds to calc-dev
I just noticed that our redux logging middleware (introduced in #1075) is turned on at https://calc-dev.app.cloud.gov/, which suggests that `process.env.NODE_ENV` isn't set to `"production"` when building the front-end bundle.  (Ideally any AST branches that test on it should also be dropped at build time, to make our production bundle smaller.)

toolness:
this might fix #1225, i dunno

first i tried uglify's [`global_defs`](http://lisperator.net/uglifyjs/compress) and that didn't work, then i tried envify and it didn't work at first but then i found [this comment](https://github.com/hughsk/envify/issues/27#issuecomment-111492365) which suggested adding `{ global: true }` and that fixed things maybe


toolness:
ok i think this fixes it

jseppi:
While this does disable the logger, it seems like the `redux-logger` code still ends up in the built `index.min.js`, which is not great.

jseppi:
I'll merge this now, and open up a separate issue for figuring out how to remove `redux-logger` from the bundle.

toolness:
Er wait really?  I did a `grep` to make sure it got removed from the bundle but I guess it's possible that whatever I was grepping for got renamed by uglify or something. BLAH.

jseppi:
After building via `NODE_ENV=production npm run gulp -- build` (and also tested with just no `DEBUG` in the env), `redux-logger` I think is still in there:

screen shot 2017-01-17 at 8 39 01 am

(The `0` exit code means that the grep was successful in finding that string)

toolness:
Ah yeah, right you are.  I think that #1247 fixes it up for good.

Issue #980, opened 2016-10-28
Use bugs for status indicators


hbillings:
For the user's own price lists, bugs like the WDS uses for their "alpha" and "beta" component indicators might be really helpful: https://standards.usa.gov/labels/

jseppi:
Are "bug" and WDS's "label" the same thing?

hbillings:
LOL, I picked a terrible example page. Though we probably could use labels, depending on how we want to design it. I was thinking of these:
<img width="255" alt="screen shot 2016-10-28 at 11 48 54 am" src="https://cloud.githubusercontent.com/assets/509309/19814623/95dfdea4-9d04-11e6-8a83-fb432bbff117.png">

ericronne:
@hbillings what does a "bug" look like?

jseppi:
@ericronne I think that is TBD! What do you think a bug should look like?

ericronne:
Ah, i thought she was referring to a bug that WDS uses.

So we're talking about designing a way for data uploaders to know the status of their uploads. Could be a bug/icon, but an explicit label is generally more clear. We could also use a bug, label, and color together. I'll noodle on this as i re-style the details page.

I'm thinking we should also indicate status on a user's list of uploaded PL's. Agree?

jseppi:
Yea, I agree.

hbillings:
@ericronne I was thinking of these little labelly things that the standards site uses, but doesn't define:
<img width="283" alt="screen shot 2017-01-03 at 1 36 37 pm" src="https://cloud.githubusercontent.com/assets/509309/21622244/c529631c-d1b9-11e6-85de-0ddf73489d06.png">

But please do what makes sense to you!

ericronne:
Yep
![image](https://cloud.githubusercontent.com/assets/9259185/21622532/8df95546-d1c3-11e6-989e-abb05e9915e4.png)

ericronne:
Blew this out a little more, and added an icon for style and increased accessibility.

![image](https://cloud.githubusercontent.com/assets/9259185/21965909/d767d0e4-db2e-11e6-8d18-01b447e18fa6.png)

**Here are the [icon svg files](https://drive.google.com/open?id=0BwWYNZcEDfwabkg0RmF5Zzl4Yzg).**

**Pending**
BG: #FBEEBE
Icon and text: #825C1E

** Rejected**
BG: #FAD0D0
Icon and text: #BA1D1D

**Approved**
BG: #E6EDC2
[...]

PR #1245, opened 2017-01-13
Style error tables and implement tooltips

hbillings:
Implements @ericronne's designs for displaying errors in tables. 

Tooltips are done entirely with CSS; no JS fallback required.

Mockup:
screen shot 2017-01-13 at 3 45 36 pm

Table:
screen shot 2017-01-13 at 3 33 55 pm

Tooltip:
screen shot 2017-01-13 at 3 34 02 pm

To do:
- [x] Fix the tooltip's arrow on non-right-aligned cells (see screenshot)
screen shot 2017-01-13 at 3 44 05 pm
- [x] Fix arrow falling off the tooltip when there's more than one line in the error description.
screen shot 2017-01-13 at 4 31 54 pm


hbillings:
Aaargh, multiline cells break >_<
screen shot 2017-01-13 at 4 31 54 pm

Also, @ericronne, if an error spans multiple lines, how should it look? Currently looks like this:
screen shot 2017-01-13 at 4 32 31 pm

ericronne:
Might help if we shrunk them down to 13px (same size as the status icons in #980). Here are the svg files for the **[filled info icons](https://drive.google.com/open?id=0BwWYNZcEDfwadTFqcFdUeWpDSDg)** and the **[outlined (hover) version](https://drive.google.com/open?id=0BwWYNZcEDfwaVEMyQU54bmltM0k)** at that size.

hbillings:
Okay, this is how it looks now:
screen shot 2017-01-17 at 10 31 13 pm

There's one oddity: Because of how I've chosen to do the tooltip pointer-speech-bubble-arrow-thing, the pointer is different sizes on a cell that is one line deep vs. a cell that is two lines deep. I could use some JS to normalize it, but this is about as good as I can get it with just CSS.

hbillings:
So there was this little funky interaction that was bugging me: when you moved the mouse between the info badge and the text, the tooltip would flicker. So I rewrote the HTML, and it looks/works great except I'm not sure multiline cells look as nice now:
screen shot 2017-01-18 at 3 46 56 pm

Thoughts?

ericronne:
Can we pair on this briefly?

hbillings:
After screensharing with @ericronne, we've decided to:
- [ ] normalize the tooltip arrow heights with a little javascript
- [ ] return to underlining all of the text in the error cell instead of just the div it's in...somehow...
- [ ] tighten up spacing between text and underline and text and icon
- [ ] extend dotted line under icon

hbillings:
@ericronne whatcha think:
screen shot 2017-01-18 at 6 41 01 pm

Do those arrows look the same to you? I'm going cross-eyed, but I think I found a CSS formula that works for positioning them, no JS needed. Maybe.

toolness:
Hmm, is this implemented for the fake schedule, or just for schedule 70?  When doing it on the fake schedule, I'm getting decapitated tooltips, it's very unsettling.

FWIW, I think this could be a great place to embed an example in our fancy STYLE GUIDE.  This would make it possible for us to manually test the styling / interaction without needing to go through the PL upload flow and create our own price list that has errors in it.

I guess one option might be to reuse code in such a way that the styleguide example isn't a duplication of the markup used in the PL upload (which could then easily fall out-of-sync).  Let me know if you want me to work on this!

ericronne:
My only thoughts are A: way better! and B: if it's easy to bring the tip box closer to the text, that would be ideal. Not a deal breaker, but would look better if the boxes were slightly overlapping field edges, rather than aligning with them, like this …
image

ericronne:
BTW it's probably fine if the tip obscures the top line of two-line content, since this is just a hover state on desktop. 

PS which do you prefer: pink bg or red, like so? …

image

hbillings:
@toolness Right now it's just for S70, mostly because I'm too lazy to make changes to both sets of markup until I know which is the winning strategy. 😄 (I totally should have thought of the styleguide option for testing, tho, 'cause that would have been so much easier. Womp womp. I'll throw an example in there once the style gets a thumbs-up!)

@ericronne Lemme see what I can do about the positioning! The tooltip's bottom edge is currently positioned to hit `100%` of the table cell width, so I'm not sure how to subtract a pixel from that with CSS. I could try setting the bottom edge to be a `line-height` of the text in the table cell, though...

Re: background, I'm personally not the biggest fan of reverse type, but I'm also not the visual designer on this. 😉

hbillings:
@ericronne Okay ignore everything I said before and look at this:
screen shot 2017-01-19 at 10 03 44 am

hbillings:
@toolness Circling back around on this, it looks like all the logic is there to render the s70 table automatically in the styleguide, but I can't figure out how to call the `[to_error_table](https://github.com/18F/calc/blob/develop/data_capture/schedules/s70.py#L261)` child method in the styleguide views. Maybe we could pair?

toolness:
Sure! Sorry I couldn't make it to coworking yesterday, but I can definitely pair on it.

hbillings:
I can't figure out how to generate the tables for the styleguide automatically, and I need to move on to other things, so I hardcoded it for now. It looks like this: 
screen shot 2017-01-23 at 12 37 01 pm

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
```diff
@@ -10,32 +10,16 @@
...
-      <td class="number {% if row.price_including_iff.errors %}error{% else %}currency{% endif %}">
-        {% if row.price.errors %}
-          {{ row.price.value }}
-          {{ row.price.errors }}
-        {% else %}
-          {{ row.price.value|floatformat:2 }}
-        {% endif %}
+    {% for col in row %}
+      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price_including_iff' %}number{% endif %}">
+        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>{% if col.name == 'price_including_iff' %}${% endif %}{{ col.value }}</span>
```
jseppi: Should these spans maybe have `tabindex="0"` so that they can be tabbed to by keyboard users? Also, I think if that is done, then the tooltips will need to also show on `:focus`.

jseppi: (I honestly don't know what the best practice is for this stuff, but I just tried that locally and it seemed to work ok)

data_capture/templates/data_capture/price_list/tables/schedule_70.html
```diff
@@ -2,46 +2,25 @@
...
-        {{ row.unit_of_issue.errors }}
-      </td>
-      <td class="number {% if row.price_including_iff.errors %}error{% else %}currency{% endif %}">
-        {% if row.price_including_iff.errors %}
-          {{ row.price_including_iff.value }}
-          {{ row.price_including_iff.errors }}
-        {% else %}
-          {{ row.price_including_iff.value|floatformat:2 }}
-        {% endif %}
+    {% for col in row %}
```
jseppi: hrm, I'd be a little wary of looping the columns here but not doing the same for the headers. I could imagine the hardcoded headers getting "out of sync" with the order of the columns. Do we have any guarantee anywhere that the columns will always render in the same order as the hardcoded headers?

jseppi: Aha, I was having trouble figuring out where the order came from at all. Turns out the fields get iterated in the order they are defined (like in `Schedule70Row` for instance). So, that means this loop _is_ pretty brittle. If another field is added or if just the formatting of the `Schedule70Row` class is changed, then the table will get wrecked.

For instance:
screen shot 2017-01-23 at 3 14 38 pm

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
hbillings: @jseppi I'm not sure what best practice is either. The tooltips should be read by screenreaders, but making it work for keyboard users would be good. I think there may be a PR for focusable tooltips from way back when -- I'll see if I can find what I'm thinking about + do a little research.

data_capture/templates/data_capture/price_list/tables/schedule_70.html
hbillings: Eeeek! Would looping the headers as well solve this, or should I revert the tables to the original verbose output?

toolness: Hey, there's a new feature in Python 3.6 (see #1265) that might maybe help with this, I'm not sure.  It's called [PEP 520 -- Preserving Class Attribute Definition Order](https://www.python.org/dev/peps/pep-0520/).  I think before 3.6 there might not have been an actual *guarantee* that the fields got iterated in the order they were defined, but now there will be, or something. 🤷‍♂️

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
toolness: Yeah these should def be focusable by the keyboard-only folks!  Hmm.

I think we may have to be careful to put the `tabindex="0"` on something that makes some sense from an ARIA standpoint...  Otherwise if someone using a screen reader tabs to the tooltip, I'm not sure what kind of role it'd announce the currently focused element to have... Is it possible to put the `tabindex="0"` on the same elements that have `role="tooltip"`?  Not sure if that'd fix things but it might.

hbillings: So adding `tabindex="0"` to the table cell works beautifully, but I wanted to do some more research on the accessibility... It looks from [this article](https://www.paciellogroup.com/blog/2014/10/notes-on-fixing-incorrect-table-structure-using-aria/) like tables should have `grid` and `gridcell` roles when they're used to display tabular data. (Though that article also notes that `aria-readonly` should be `true`.) @atul, does this check out to you? If so, I think I'll leave the `tabindex` on the `td` and just fill in the appropriate roles.

ericronne:
Once this is all sussed, y'all totally need to give a talk on it.
Eg **[here](http://www.w4a.info/2016/2017/)**. Kinda far, tho :/

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
toolness: Cool!  I don't know enough about stuff to know if that's a good fix for sure or not but it seems reasonable. We could always ask bristow too.  As long as the error info is at least announced by screen-readers, and as long as the tooltips are keyboard-accessible, though, it seems like a great initial implementation that we can then update later if needed... no need to hold off on merging until the *perfect* a11y sol'n is found, imo.

data_capture/templates/data_capture/price_list/tables/schedule_70.html
jseppi: I would say I am more for being explicit than looping the fields. I wonder though if there is a way you could do something like 

`table_fields = ['sin', 'labor_category', ... , 'price_including_iff']`

and then loop through those in the template (non-working pseudo-code):

```
{% for field in table_fields %}
  <td>{{ row[field] }}</td>
{% endfor %}
```

@toolness?

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
jseppi: I _think_ [that article](https://www.paciellogroup.com/blog/2014/10/notes-on-fixing-incorrect-table-structure-using-aria/) is just talking about correcting bad/non-semantic HTML tables with ARIA roles. I'm having a hard time parsing the article, but this line seems salient: "As always the best advice is use native HTML elements whenever possible" (which we do). And their example is correcting a case where there are two tables defined successively, with the first table is used only for headers and the second table is used for content (who would do that?).

Our tables look semantically correct to me so I don't think you need to sprinkle `role`s all over.

toolness: Yeah, our tables are semantically correct for sure--my concern was just that screenreader users might be confused as to why some of the parts of the table, which are non-interactive to them, are focusable in the tab order. Normally only *interactive* things should be focusable in the tab order--and while that's true for sighted keyboard users, it's not true for screenreader users, because the tooltip content is showing by default to them (no interactivity required), at least that's my understanding.

But still, even if I am right, it's probably not enough of an issue to block this PR on it--as long as the content is still *somehow* accessible to screen reader and keyboard users, I think that's great for a first implementation, even if it might be a bit confusing or awkward.

data_capture/templates/data_capture/price_list/tables/schedule_70.html
toolness: Yeah I think that's definitely a way we could do it... passing `table_fields` in as a context variable, I assume?

jseppi: Or maybe defining it as an attribute in the `SubmittedPriceListRow` model

toolness: Hmm, imma hack on this for a bit methinks. Be back soon!

toolness: Ok I attempted to hack up a solution in #1268.  Basically we just pass an empty `Schedule70Row` instance to the template and iterate through its form fields, much in a similar way that we'd do to render an actual form, only in this case we're using the form field metadata to render the `<th>` tags instead of form fields.

PR #1246, opened 2017-01-16
Build production scripts before deploy

jseppi:
This PR `unset`s the `DEBUG` env var and runs another gulp build (which will be in "production-mode") before deployments.

To Dos:
- [x] ensure it works (React is in production mode when deployed)
- [x] change back calc-dev deploy to trigger on `develop` branch changes


toolness:
Wait, we shouldn't actually need to set `NODE_ENV` explicitly ourselves though, should we?  In #725 we made it so that the `DEBUG` setting effectively translates to a `NODE_ENV` setting in the gulpfile, so as long as `DEBUG` is false (i.e. undefined) in production, our `NODE_ENV` should already be set to production...

jseppi:
Right, but we have `DEBUG=yup` on this line of our `.travis.yml`: https://github.com/18F/calc/blob/72103af732a5394428ce1f67d0b56ad015f5c6da/.travis.yml#L39 so the assets are all built in debug mode for testing.

toolness:
OH BLERGHhhhhhhapreogkpoakgr

there's no way to un-`DEBUG=yup`ify the environment for the production use case?

jseppi:
Hrm, I think I could add `unset DEBUG` to the `before_deploy:` section.

toolness:
Yeah that would be supercool!!!!!!!!!!

jseppi:
Alright, how's this, @toolness?

jseppi:
stand by -- testing `unset` works

jseppi:
woot, it works!

screen shot 2017-01-17 at 12 29 06 pm

PR #1247, opened 2017-01-17
Use uglifyIFY to excise unused dependencies from bundle

### LINKED ISSUE #1244: Development packages (eg, redux-logger) should not be included in production bundles
As mentioned in https://github.com/18F/calc/pull/1243, even though the `redux-logger` doesn't actually get `require`'d in `explorer.js` when `NODE_ENV=='production'`, I am pretty sure its code is still included in the bundle. This is not ideal because it adds extra dead code to the bundle.

toolness:
Apparently there is a module called [`uglifyify`](https://github.com/hughsk/uglifyify) (not to be confused with `uglify`) which allows us to fix #1244.  See its [motivation/usage](https://www.npmjs.com/package/uglifyify#motivationusage) section for an example.

Its README *does* still encourage the use of uglify on its own, since apparently there are optimizations it can do that uglify*ify* cannot.

If we ever switch to webpack, this whole thing will not be nearly as confusing.  We just use its [`DefinePlugin`](https://webpack.github.io/docs/list-of-plugins.html#defineplugin), make sure minification is enabled, and we're done.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -201,6 +201,12 @@ function browserifyBundle(entryPath, outputPath, outputFile) {
     .transform(babelify.configure({ presets: ['es2015', 'react'] }));
 
   if (process.env.NODE_ENV === 'production') {
+    // Here we use uglifyify--not to be confused with uglify--to uglify
+    // each individual module by itself, which allows us to excise
+    // unused dependencies based on our build configuration. We are
+    // also passing arguments to bundler.transform() in a weird order,
+    // but it's what uglifyify's docs recommend and it seems to work,
+    // so whatever.
     bundler = bundler.transform({
       global: true,
     }, 'uglifyify');
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -200,6 +200,12 @@ function browserifyBundle(entryPath, outputPath, outputFile) {
     .transform(envify({ NODE_ENV: process.env.NODE_ENV }), { global: true })
     .transform(babelify.configure({ presets: ['es2015', 'react'] }));
 
+  if (process.env.NODE_ENV === 'production') {
+    bundler = bundler.transform({
+      global: true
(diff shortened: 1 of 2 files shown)


toolness:
Oh, to verify that the module was excised, I did:

```
cat frontend/static/frontend/built/js/data-explorer/index.min.js | grep 'next state'
```

`next state` is a string literal that appears in redux-logger, so grep will find it if it's in there.

gulpfile.js
```diff
@@ -200,6 +200,12 @@ function browserifyBundle(entryPath, outputPath, outputFile) {
     .transform(envify({ NODE_ENV: process.env.NODE_ENV }), { global: true })
     .transform(babelify.configure({ presets: ['es2015', 'react'] }));
 
+  if (process.env.NODE_ENV === 'production') {
+    bundler = bundler.transform({
+      global: true,
+    }, 'uglifyify');
```
jseppi: I'm a little puzzled by the args to `.transform` here. The browserify docs [indicate that the transform should be the first arg](https://github.com/substack/node-browserify#btransformtr-opts), but here the opts appear first (and that is indeed how `uglifyify`'s doc examples have them specified). I also looked at browserify's source, and it does some funky arg swapping if the second one is a string, so it works both ways. Weird.

Um, I guess if you want it to be more inline with browserify's docs, and look more like our other transforms, you could change this line to 

```js
bundler = bundler.transform('uglifyify', { global: true });
```
but up to you.

toolness: Yeah, I was *extremely* confused by that myself.  I am glad the other way works, I'll switch it to that so it's more consistent!

toolness:
Actually I am really tired of wrestling with this build system and don't want to deal with making sure things still work with the argument order reversed, so I am just going to add an explanatory comment and be done with it.

Issue #612, opened 2016-08-25
Update labor category filter in CALC tool to search for SIN prefix


mtorres253:
As a user of CALC, I want to look up labor categories by SIN prefix (not the entire SIN number) rather than the old schedule name, so that I can be confident that I am getting the current prices (since the recent data upload migrated the old schedules into one professional services schedule with SIN prefix replacing of the current and former schedule names). 

For example, If I want to see all labor categories in FABS (former schedule name), then I should be able to filter all labor categories with SIN prefix of 520. I should also be able to do this for SIN prefix 132/IT 70 (current schedule name)

## Acceptance Criteria

- [ ] **Scenario: Filters are adjusted so they show former (or current) schedule name, along with SIN prefix**
- Given I am a visitor to CALC (no authentication required)
- When I am on the CALC homepage (https://calc.gsa.gov/)
- Then I should see the following values for the schedule filter (pattern [SIN prefix] - [current or former schedule name]) 

> 520 - Formerly FABS
541 - Formerly AIMS
73802 - Formerly Language
871 - Formerly PES
874 - Formerly Mobis
87405 - Formerly Logistics
899 - Formerly Environmental
132 - IT 70

https://docs.google.com/a/gsa.gov/spreadsheets/d/1FgqTf-z5esSleI4vfUYBi53VIXL1Tg3uglG2rVccw54/edit?usp=sharing

- [ ] **Scenario: Filtering by SIN prefix/schedule name picks up all labor categories with that SIN prefix**
- Given I am a visitor to CALC (https://calc.gsa.gov/)
- When I select [SIN prefix] - [current or former schedule name] from the schedule filter
- Then my view should be filtered so that I see data only for vendors that have the selected SIN prefix

- [ ] **Scenario: Help text to explain SIN prefix/schedule name migration**
- When I visit https://calc.gsa.gov/about/#schedules
- Then for each schedule listed except for schedule 70, I should see a sub-header containing the current SIN prefix for each of the schedule followed by “(formally known as [former schedule name])”
- And for Schedule 70, I should see a sub-header containing the SIN prefix followed by “(Schedule 70)”

- [ ] **Scenario: Update API to return the correct information when filtering by SIN prefix**
- Given I am an authorized user of the CALC API
- When I make a request for schedule data using a valid SIN prefix
- Then the response should include data for all vendors in the schedule with that SIN prefix

- [ ] **Scenario: Update API documentation**
- When I view documentation (https://github.com/18F/calc#other-filters)
[...]

PR #1249, opened 2017-01-17
Update 'unreleased' section of changelog.

toolness:
This attempts to update the "unreleased" section of the changelog to reflect what we've done over the past several months, since there hasn't really been a release since then.

It basically boils down to "we made data capture, and a bunch of minor improvements/bugfixes to data explorer".

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,32 @@ to [Semantic Versioning](http://semver.org/).
   Data administrators can then review these price lists and add
   their data into CALC.
 
+### Changed
+
+- The "Schedule" filter in the data explorer (front page of CALC) has
+  been renamed to the "SIN / Schedule" filter, and allows rates to be
+  filtered by SIN number.
+
+- Searching for multiple words in the data explorer's search field
+  now produces expected autocompletion results.
+
+- The data explorer is now much more keyboard-accessible and
+  screenreader-friendly.
+
+- The "proposed price" field in the data explorer automatically
+  updates the histogram on a per-keypress basis; there is no longer
+  any need to click the "Go" button.
+
+- Sharing CALC links with the "Education" field filled out now works
+  properly.
+
+- The CALC banner has been redesigned to be less confusing and easier
+  to read.
+
+- Google Analytics for CALC now properly track new searches in the
+  data explorer as page views. We also track clicks on links in the
+  "Contract #" column of the data explorer results table.
+
 ## [1.2.0][] - 2015-07-21
 
 - filtering empty list items out of query building
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,12 @@ to [Semantic Versioning](http://semver.org/).
 
 ### Changed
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,32 @@ to [Semantic Versioning](http://semver.org/).
   Data administrators can then review these price lists and add
   their data into CALC.
 
+### Changed
+
+- The "Schedule" filter in the data explorer (front page of CALC) has
+  been renamed to the "SIN / Schedule" filter, and allows rates to be
+  filtered by SIN number.
```
toolness: It would be nice if we could link out to some kind of GSA blog post that explains why this was done, but I am not sure what that would be...

jseppi: Good question. I'm also not sure what the underlying reason for this was...

jseppi:
I don't see anything about Data Capture. Missing commit?

toolness:
Oh, the data capture stuff was already in changelog.md as part of its initial commit (see #1191).

jseppi:
>.< doh! I thought I had seen it somewhere...

CHANGELOG.md
```diff
@@ -14,6 +14,32 @@ to [Semantic Versioning](http://semver.org/).
   Data administrators can then review these price lists and add
```
jseppi: Maybe also add an item about how registered users can view a history of their submitted price lists and their review status.

jseppi: Oh, and another one about Region 10 administrators being able to bulk upload their pricing data from the R10 database exports.

jseppi: Should the data capture be clarified to specify that only Schedule 70 price lists can be uploaded right now?

toolness: oh good point--should it also mention the r10 bulk upload functionality too?

CHANGELOG.md
toolness: Well, according to #612:

> As a user of CALC, I want to look up labor categories by SIN prefix (not the entire SIN number) rather than the old schedule name, so that I can be confident that I am getting the current prices (since the recent data upload migrated the old schedules into one professional services schedule with SIN prefix replacing of the current and former schedule names).

So I guess this is because there is actually only *one* schedule now, the Professional Services schedule, and all the stuff that used to be other schedules is now locatable by SIN prefix?

A google search for "GSA Professional Services Schedule" gave me https://www.gsa.gov/portal/content/246403, which mentions:

> The PSS consists of seven formerly separate schedules that were consolidated in 2015 to give federal agency acquisition professionals the ability to obtain total contract solutions for their professional services requirements using one contract vehicle.

So I guess that is the reason?  There's a [Federal times article about PSS](http://www.federaltimes.com/story/government/acquisition/gsa-gwac/2015/10/08/professional-services-schedule/73605190/) but it is on a third-party site, rather than a government-owned one...

jseppi: Yeah, that sounds correct. Go ahead and link to that GSA page.

toolness:
Ok, just added the info. Does this look OK @jseppi ?

PR #1250, opened 2017-01-18
Make /updates/ friendlier to non-developers

toolness:
This attempts to make `/updates/` a bit friendlier to non-developers by removing the expository text from `CHANGELOG.md` (which mentions things like keepachangelog.com and semver) and replacing it with expository text in the `changelog.html` template:

> screen shot 2017-01-18 at 8 21 40 am

It also renames the page from "Change Log" to "Updates", following the lead of [cloud.gov/updates/](https://cloud.gov/updates/). It also just sounds less technical.

Also, if the "Unreleased" section is empty--which it should be on production deploys--then it is removed entirely from the page.

### DIFF
--- a/hourglass/templates/changelog.html
+++ b/hourglass/templates/changelog.html
@@ -1,9 +1,15 @@
 {% extends "base.html" %}
 
+{% block title %}Updates{% endblock %}
+
 {% block body %}
 <div class="container card--narrow">
   <div class="card">
       <div class="card__content">
+        <h1>Updates</h1>
+
+        <p>Curious what’s new that you might find helpful as a CALC user? Here are highlights from our updates as we’ve improved the product.</p>
+
         {{ html }}
       </div>
   </div>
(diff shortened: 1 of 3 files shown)


hourglass/templates/changelog.html
```diff
@@ -1,9 +1,15 @@
...
 
+{% block title %}Updates{% endblock %}
+
 {% block body %}
 <div class="container card--narrow">
   <div class="card">
       <div class="card__content">
+        <h1>Updates</h1>
+
+        <p>Curious what’s new that you might find helpful as a CALC user? Here are highlights from our updates as we’ve improved the product.</p>
```
jseppi: Consider simplifying the first sentence, maybe: "Curious about what's new in CALC?"

toolness: Hehe good idea. I had just cribbed the text from the top of https://cloud.gov/updates/ but I think shortening it sounds better.

Issue #1208, opened 2017-01-05
Design an email template


ericronne:

ericronne:
Keep it simple.

eg...
http://leemunroe.github.io/responsive-html-email-template/email.html

ericronne:
![](https://github.trello.services/images/mini-trello-icon.png) [Design simple html email template](https://trello.com/c/psZ8vFon/223-design-simple-html-email-template)

ericronne:
Here's a draft, let's discuss! If we roll it out before we update the home-page design and site colors, we should adjust the colors to match the current palette.

## Desktop
![email-desktop-insitu](https://cloud.githubusercontent.com/assets/9259185/22599495/c7734d26-e9fc-11e6-91bf-e9799ab95bf7.png)

## Mobile
![email-mobile-insitu-resized](https://cloud.githubusercontent.com/assets/9259185/22599515/de453de8-e9fc-11e6-9a56-8051514acb31.png)

hbillings:
I like this design overall for the email that a CO would get upon an admin taking action on one of their price lists. A couple of thoughts:
- There probably should be two separate email designs, one for admins and one for COs. Admins should get "digest" style

ericronne:
@hbillings is there a content sample approximating what admins would get?

jseppi:
@ericronne see https://github.com/18F/calc/tree/develop/data_capture/templates/data_capture/email

`approval_reminder.html`, `bulk_upload_failed.html`, and `bulk_upload_succeeded.html` are emails that go to admins.

You can generate sample emails:
1. make sure you have `docker-compose up` running in a terminal window
1. open `http://localhost:1080` in your web browser to view your local MailCatcher
1. in another terminal window run `docker-compose run app python manage.py send_example_emails`. Your MailCatcher should get populated with versions of all the emails that CALC sends.

ericronne:
Looks like we actually aren't using "digest" emails — with links to individual price lists to be reviewed. @hbillings had the thought to let the user know how many PLs they have awaiting approval. Sound useful? If so, pretty straightforward to implement @jseppi?

jseppi:
We are. The "digest" is the `approval_reminder.html`

ericronne:
So the full list will appear in the email, in place of ` https://example.com/unreviewed_price_lists `?

jseppi:
no, there is no list, just the link to the page with the unreviewed lists.

ericronne:
[...]

PR #1255, opened 2017-01-18
Send HTML email notifications with link to Price List details page

### LINKED ISSUE #1205: Email - add relevent links to templates
- all the status change emails should include a link to the associated Price List Details page
- the R10 bulk failure email should include a link back to the R10 upload
- the admin review reminder email should include a link to the unreviewed price list admin page

jseppi:
This PR adds a link back to the Price List Details page to each status-change-email-notification. It also adds a `text/html` version of the content to those email messages and some minor formatting.

Here's what the emails look like now:

HTML:
screen shot 2017-01-19 at 3 58 08 pm

Text:
screen shot 2017-01-19 at 3 58 16 pm

The extraneous line breaks in the plaintext version is a bit wonky, but I'm not too concerned about it since our users are all going to be on GSA's GMail which will render the HTML version. Keeping the text version is just as a fallback/best-practice.

Closes #1205


jseppi:
[As a CO, I should receive email notifications that have links to relevant pages.](https://trello.com/c/KeyUvkLq/216-as-a-co-i-should-receive-email-notifications-that-have-links-to-relevant-pages)

data_capture/tests/test_email.py
```diff
@@ -1,33 +1,49 @@
...
+        self.assertIn('text/html', content_types)
+
+    def assertHasDetailsLink(self, price_list, message):
+        details_link = reverse('data_capture:price_list_details',
+                               kwargs={'id': price_list.pk})
+        has_link = False
+        for (content, content_type) in message.alternatives:
+            if content_type == 'text/html':
+                has_link = details_link in content
+        self.assertTrue(has_link)
```
jseppi: Wasn't sure about best practice for creating my own assertion method. Would it have been better to just `raise` here? And maybe use `self.assertIn(details_link, content)` up in the `if` block?

jseppi:
whoops, my details links are relative, not absolute. Fixing.

screen shot 2017-01-19 at 1 56 43 pm

jseppi:
Also need to add `<p>` tags:

screen shot 2017-01-19 at 1 57 45 pm

jseppi:
Ready for review, @toolness

data_capture/tests/test_email.py
toolness: Hmm, I guess another alternative might be something like:

```py
self.assertHasHtmlAlternative(message)
html = [content for (content, content_type) in message.alternatives
        if content_type == 'text/html'][0]
self.assertIn(details_link, html)
```

That gets rid of any branching at least, I dunno.

The only thing I feel particularly strongly about is that I prefer `assertIn` over `assertTrue` simply because the former will (er, I *think* it will) provide better debugging output if the assertion fails.  `assertTrue` won't be very helpful because it'll just fail with `False is not True` whereas I *think* `assertIn` will fail with `"http://blah" is not in "Your link is http://foo"` which could potentially help the developer figure out what's wrong.

jseppi:
I had considered Markdown, but we are also going to want to have some simple styling (https://github.com/18F/calc/issues/1208) and I didn't think that would end up working well.

data_capture/tests/test_email.py
jseppi: Yeah, I also prefer `assertIn`. The potential problem with that snippet you added is only checking that the first content with `content_type=='text/html'` has the link, which is I avoided something like that...but, I don't know why we'd ever have multiple `content_type=='text/html` alternatives, so I'm probably overcomplicating things.

toolness: Yeah, I guess maybe an alternative is to ensure that there is exactly one html alternative in `assertHasHtmlAlternative`?  Or even just call it `assertHasOneHtmlAlternative`?

PR #1260, opened 2017-01-19
Send 'Cache-Control: max-age=0' by default

### LINKED ISSUE #1257: Deployment - Prod URL seems to serve cached version of index page, resulting in 404s for css and js files
screen shot 2017-01-18 at 2 12 13 pm

During today's production deployment, I noticed that after the deployment finished, when I refreshed https://calc.gsa.gov, which I already had open, I got an unstyled front page. I noticed that the front page was still referencing the previous static files (as shown in the above image). However, when I viewed the page in another browser, everything worked as expected. Also if I changed the URL by adding a query string (https://calc.gsa.gov?whatever), then I received the new version with correct static files.

I think this has to do with caching with CloudFront, but I guess it could also be something weird in Chrome, or maybe just cache headers in general.

I'm not quite sure how to determine the cause or debug at this point.

toolness:
This attempts to at least temporarily resolve #1257 by explicitly telling downstream caches to never cache our dynamic views by default.

In order to send the `Cache-Control` headers we also enable [Django's site-wide cache middleware](https://docs.djangoproject.com/en/1.10/topics/cache/#the-per-site-cache).

Because we're setting the default expiry via [`CACHE_MIDDLEWARE_SECONDS`](https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-CACHE_MIDDLEWARE_SECONDS), individual views and middlewares (like whitenoise, which serves our static files) should be able to override the default, so that we can still leverage CloudFront for resources that we actually *want* to cache.

Note, though, that because we're actively preventing our views from being cached by default, it might actually overload our server if too many folks are visiting our site at once!  If that happens we might want to consider setting `CACHE_MIDDLEWARE_SECONDS` to some low integer, so that we can at least cache our dynamic views for a short period of time.

### DIFF
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -142,6 +142,7 @@ else:
     WHITENOISE_MIDDLEWARE = 'whitenoise.middleware.WhiteNoiseMiddleware'
 
 MIDDLEWARE_CLASSES = (
+    'django.middleware.cache.UpdateCacheMiddleware',
     'hourglass.middleware.ComplianceMiddleware',
     WHITENOISE_MIDDLEWARE,
     'django.contrib.sessions.middleware.SessionMiddleware',
@@ -157,6 +158,7 @@ MIDDLEWARE_CLASSES = (
     # when the ProfilingPanel is enabled
     # http://django-debug-toolbar.readthedocs.io/en/stable/panels.html#profiling
     'hourglass.middleware.DebugOnlyDebugToolbarMiddleware',
+    'django.middleware.cache.FetchFromCacheMiddleware',
 )
 
 AUTHENTICATION_BACKENDS = (
@@ -309,6 +311,15 @@ UAA_CLIENT_SECRET = os.environ.get('UAA_CLIENT_SECRET')
 LOGIN_URL = 'uaa_client:login'
 
 LOGIN_REDIRECT_URL = '/'
+
+# We *always* want to send a Cache-Control header downstream, especially
+# in the event where we've got a reverse proxy with aggressive caching
+# defaults like Amazon CloudFront in front of us.
+#
+# For now we're just going to tell any downstream caches to never cache
+# any dynamic content we give them, to ensure that stale content never
+# gets served to end-users.
+CACHE_MIDDLEWARE_SECONDS = 0
 
 if DEBUG:
     INSTALLED_APPS += ('fake_uaa_provider',)
--- a/hourglass/tests/tests.py
+++ b/hourglass/tests/tests.py
@@ -305,6 +305,12 @@ class StaffLoginRequiredTests(DjangoTestCase):
         self.assertEqual(403, res.status_code)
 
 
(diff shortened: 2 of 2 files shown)


toolness:
Er, to clarify, I believe the reason our index page (and other dynamic views) are being aggressively cached by CloudFront is because they have no `Cache-Control` headers, which makes CF shrug and apply its aggressive defaults, which are the cause of #1257. By explicitly using a `Cache-Control` header, we can tell CloudFront to chill out.

toolness:
I guess I should also note that this PR does actually enable the site-wide cache in Django, so if one of our views *does* tell Django to cache it, then Django *will* cache it, *even in development*.  This could potentially be considered a feature rather than a bug, as it improves dev/prod parity, sort of.  Also, the Django Debug Toolbar's **Cache** panel might help us debug stuff.

jseppi:
Hrm, so after watching the [lightning talk you mentioned](https://youtu.be/fgbJJ412qRE?t=13m30s), I wonder if we should be a little finer-grained with our cache headers, and maybe follow the example from the talk:

screen shot 2017-01-19 at 4 11 59 pm

Basically, we'd be safe putting long `max-age`s on our js and css since they have the manifest suffixes, and putting a short (like 30 second) `max-age` on our html files would be pretty safe (as long as we add the `no-cache=Set-Cookie` portion also), though I'm also ok with keeping them at 0 seconds.

toolness:
So I'm pretty sure that whitenoise is already putting a long max-age on our js and css already; the [front page of the `whitenoise` docs](http://whitenoise.evans.io/en/stable/) claim that "WhiteNoise takes care of best-practices for you", including "Setting far-future cache headers on content which won’t change".  It's setting `max-age` to `0` when `DEBUG` is set, but I believe it sets a high `max-age` in production.

Just tried it out by using Web Inspector on a whitenoise-served CSS file and it's indeed set to `public, max-age=315360000`.

I don't *think* we actually need to set the `no-cache=Set-Cookie` because Django's [`SessionMiddleware`](https://docs.djangoproject.com/en/1.10/topics/http/sessions/) already adds a `Vary: Cookie` header. However, on second thought, that might actually be *bad*, since CF might interpret it to mean that it should include the cookie as part of the cache key, which would not be particularly useful...

Um, also, the session middleware gives us *anonymous* sessions, which we only take advantage of during the cloud.gov UAA handshake, but I think this means that *all* requests, even ones coming in from non-authenticated users, will have cookies associated with them, meaning that setting `no-cache=Set-Cookie` could effectively be the same as `max-age=0`.

Looking at Web Inspector again, it seems Ethnio and Google Analytics are also setting things in the user's cookie, which makes me wonder if it's possible to have CloudFront *reliably* cache anything, even if we drop anonymous sessions... arggggg caching!

toolness:
Hmm, so I guess I am gonna merge this for now, and we can discuss changes to the caching policy in a separate issue!

jseppi:
oh deer
giphy

PR #1261, opened 2017-01-19
Only display email and is_staff in user admin list

### LINKED ISSUE #1186: Admin - User list table overruns into filters
This was not really noticeable until we had a user table full of users (as we do on prod). The table on the User list admin page overruns into the filter on the right side of the page. I think an easy fix would be to remove the First and Last Name columns.

screen shot 2017-01-03 at 10 19 13 am

jseppi:
This is a simple mitigation for #1186. I customized the list display of the User model to only include `email` and `is_staff` since we don't actually use the first name, last name, and username fields.

screen shot 2017-01-19 at 5 00 47 pm

Closes #1186


data_capture/admin.py
```diff
@@ -123,6 +123,10 @@ class CustomUserAdmin(UserAdmin):
         ('Important dates', {'fields': ('last_login', 'date_joined')}),
     )
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        UserAdmin.list_display = ('email', 'is_staff')
```
cmc333333: This is modifying the (parent) class-level variable whenever an instance of the `CustomUserAdmin` is created. Is that the intent? I'd think you'd just want to set `list_display` as you do `non_superuser_fieldsets`, etc.

jseppi: heh, yeah, good catch - amended in https://github.com/18F/calc/pull/1547

Issue #1199, opened 2017-01-04
Create an automated test suite for use against production deployments


toolness:
We've had a bunch of annoying regressions when deploying to calc.gsa.gov (instead of e.g. calc-dev.app.cloud.gov, or even calc-prod.app.cloud.gov).

It might be useful for us to create a simple automated test suite that we can run from our local dev machines against the prod deployment to make sure that regressions don't occur. For instance:

* Add a field to the `/healthcheck` endpoint that includes the results of `request.build_absolute_uri()`, which the suite can use to verify that the server is at the hostname it thinks it's at (to guard against #1187).

* Issue a csrf-protected POST to the server (one that isn't hidden behind auth, such as https://calc.gsa.gov/styleguide/radio-checkbox) and make sure it works (to guard against #1198).

* Ensure the [`X-Requested-With` header](https://docs.djangoproject.com/en/1.10/ref/request-response/#django.http.HttpRequest.is_ajax) is being passed through to our origin server.  Otherwise ajaxforms in particular will explode.

* Ensure our API is delivering the proper CORS headers (#1307)

It might also be useful to surface this suite, or at least the kinds of things it's testing for, back to 18F as a whole, since it's likely that other cloud.gov-based projects will experience the same difficulties as us!

toolness:
@jseppi were there any other weird hiccups that happened when moving to calc.gsa.gov that you think we could guard against future regressions of?

jseppi:
Haven't found any others yet...

toolness:
Okydoky.

Is it possible for a service like NewRelic to regularly issue a test like this against production CALC?  It doesn't seem like the sort of thing we should do with a Travis cron job, but it'd be nice to have a test like this constantly running, I guess.

jseppi:
I don't have any experience working with this feature, but New Relic apparently has some kind of "scripted monitoring" functionality: https://docs.newrelic.com/docs/synthetics/new-relic-synthetics/scripting-monitors

toolness:
Cool!  I'm a bit confused by some of their documentation, but at a basic level it seems like their [API tests](https://docs.newrelic.com/docs/synthetics/new-relic-synthetics/scripting-monitors/write-api-tests) are nodeJS scripts that use the `request` library (which their docs keep calling `http-request` for some reason).

So maybe I should just write up these tests as simple node scripts that use `request`, so we can execute them either from our own machines or via newrelic?

jseppi:
Sounds good!

[...]

PR #1263, opened 2017-01-20
Send html versions of admin emails and include relevant links

jseppi:
This PR changes the admin emails to be sent as html (just like the status-change notification emails from #1255).

Here's the thing I don't really like: I modified the signatures of all the email methods to take a `site_base_url` arg that defaults to `'https://calc.gsa.gov'`. I had previously passed a `request` object and used its `build_absolute_uri` method to generate fully-qualified URLs. Unfortunately, with the reminder email sent in the `rq_scheduler` app, there is no `request` to pass along, and I don't think there is any other way to get the base url :\

I guess yet another env var could be used in place of hardcoding `'https://calc.gsa.gov'`...

To Do:
- [ ] File an issue to move `site_base_url` value to an environment variable setting (or other configurable approach)


jseppi:
[Admins should receive emails with links back to associated admin pages.](https://trello.com/c/JMmZdA8f/222-admins-should-receive-emails-with-links-back-to-associated-admin-pages)

data_capture/email.py
```diff
@@ -151,17 +157,22 @@ def bulk_upload_failed(upload_source, traceback):
 
 def approval_reminder(count_unreviewed):
```
jseppi: So, I can't actually dynamically generate a link back to the admin page for this email because it is sent in a scheduled task, so I have no `request` to use for `build_absolute_uri`. I could hardcode the production URL (or pass in as a default parameter), but then it seems like I should do that for all the other email methods. I guess I'm actually ok with that now that I think about it, because we don't need anything else from `request`. Thoughts, @toolness?

toolness: I suppose that is acceptable. I guess another alternative is to pull the hostname out of the [Django sites framework](https://docs.djangoproject.com/en/1.10/ref/contrib/sites/), blargh.

data_capture/email.py
```diff
@@ -32,7 +32,7 @@ def collapse_and_strip_tags(text):
     return re.sub(r'\n+', '\n', strip_tags(text))
 
 
-def send_mail(subject, body, to, html_message=None, reply_to=None):
+def send_mail(subject, to, html_message, reply_to=None):
```
jseppi: simplified this signature to just take the `html_message` arg and do the plaintextification here since now all emails will be in both text and html.

data_capture/templates/data_capture/email/approval_reminder.html
```diff
@@ -0,0 +1,9 @@
+<p>
+  This is your periodic reminder to check on and approve submitted price lists in CALC.
+</p>
+<p>
+  <b>{{ count_unreviewed }}</b> price list{{ count_unreviewed|pluralize:"s have,has" }} not been reviewed.
+</p>
+<p>
+  Visit <a href="{{ unreviewed_url }}">{{ unreviewed_url }} for a list of unreviewed price lists.
```
jseppi: Add link to the UnreviewedSubmittedPriceList changelist admin page.

data_capture/templates/data_capture/email/bulk_upload_failed.html
```diff
@@ -0,0 +1,16 @@
+{% load data_capture %}
+<p>
+  This message is in regards to Region 10 CALC data upload <b>#{{ upload_source.id }}</b>, uploaded by <b>{{ upload_source.submitter.email }}</b> on <b>{{ upload_source.created_at|tz_timestamp }}</b>.
+</p>
+<p>
+  Unfortunately, an <b>error occurred</b> while importing the data.
+</p>
+<p>
+  Please visit <a href="{{ r10_upload_link }}">{{ r10_upload_link }}</a> to try uploading again.
```
jseppi: Add link back to R10 upload Step 1

docker-compose.local.yml
```diff
@@ -17,9 +17,17 @@ gulp:
 rq_worker:
   volumes:
     - .:/calc
+  environment:
+    - EMAIL_URL=smtp://mailcatcher:25/
+  links:
+    - mailcatcher
```
jseppi: Added `mailcatcher` link and associated `EMAIL_URL` to `rq_worker` and `rq_scheduler` containers so that emails from them would also be caught.

jseppi:
blerf, ready for review, @toolness

docker-compose.local.yml
toolness: Nice, I wish we didn't have to break DRY so much in these YAML files but whatevvvs.

toolness:
Good question about the whole site hostname thing. I'm not sure what the answer is either so [I asked on #django on Slack](https://gsa-tts.slack.com/archives/django/p1485213990000002)...

> Is there an accepted best practice around defining what the "canonical" hostname for a django project is?  In CALC we usually use `request. build_absolute_uri()`, which looks at the `Host` header to figure out where it's being accessed, but we're starting to send emails from worker processes that don't have a `request` object, and we need to include absolute URLs in them that point back to the site.  It seems like one approach would be to have yet another environment variable that gets represented as a `settings` property (or perhaps it could be the sole entry in `settings.ALLOWED_HOSTS`), or I guess we could use Django's `sites` framework and pull it out of the db, but I'm not sure if there's some other obvious solution here that we're not seeing...

toolness:
It looks like the recommended best practice is to use the sites framework!  See jacobian's response in the slack thread for more details.

toolness:
Alternatively, if you want, we can just merge this as-is and we could deal with the sites framework part in a separate PR (which I'm happy to work on if you don't want to, whatever works).

jseppi:
cool, I'll look into the Sites framework.

jseppi:
I can't figure out how to effectively use this Sites framework thing. To me, it seems essentially like hardcoding the domain still, but in a much more convoluted way (ie, we define a `SITE_ID` setting, and then also have to make sure all databases (local, dev, staging, prod) has a `Site` model with that id and with `domain=https://calc.gsa.gov` (or some variant).

toolness:
Yeah, I think that while it's a bit cumbersome, it does leave open the possibility for our django project to actually serve *multiple* domains in the future without requiring a refactoring (while that's highly unlikely, it is possible).  Also, because of the fact that we're using a very well-understood Django convention, others familiar with Django should be able to easily grok it, so IMO we should just bust it up even though it does take a bit more work.

I guess a bonus is that since Django will actually have some inkling of what its hostname *should* be, we can have a self-test, perhaps as part of `/healthcheck`, that ensures that the domain in the `Host` header matches the domain in the site ID (for #1199).

Blerg though, the part about having to have site models in the DB is kind of annoying though, yeah.

I wonder if there's a way we could auto-populate it by inspecting something in the cloudfoundry environment?

jseppi:
I'll have to trust you that is a very well understood convention :) I found the docs fairly confusing personally.

We do have `application_uris` and `uris` properties (unsure of the difference - they have the same values) in `VCAP_APPLICATION`:

```json
{
 "VCAP_APPLICATION": {
  "application_id": "...",
  "application_name": "calc-prod",
  "application_uris": [
   "calc-prod.app.cloud.gov",
   "calc.gsa.gov"
  ],
  "application_version": "...",
  "cf_api": "...",
  "limits": {
  },
  "name": "calc-prod",
  "space_id": "...",
  "space_name": "prod",
  "uris": [
   "calc-prod.app.cloud.gov",
   "calc.gsa.gov"
  ],
  "users": null,
  "version": "..."
 }
}
```

toolness:
Huh, I'd first used it for a different django site and I thought the [Example usage](https://docs.djangoproject.com/en/1.10/ref/contrib/sites/#example-usage) section was particularly helpful... That said, I actually *wanted* to have multiple domains that served slightly different views of the same DB content, so it was particularly useful. In our current CALC use case it does feel a bit overkill but then again all the other options seem bleh too.

That's cool that the domains are in the `VCAP_APPLICATION` blob!  Except that I dunno if there's a predictable way to figure out what the "canonical" one is supposed to be. Just looking at that blob, it seems like it *could* be "pick the last item in the list of domains" but who knows if `calc.gsa.gov` will ever show up as the *first* item in that list... blerf

jseppi:
Well, I tried creating a custom migration that looks for `Site.objects.get(pk=1)` but it doesn't seem to exist at the time migrations are initially run. I'm out of my element on figuring this framework out. Plz help.

jseppi:
Attempt here: https://github.com/18F/calc/pull/1270

jseppi:
Alrighty, we opted to not go with the Sites framework attempt (see closed #1270 for discussion).

@toolness: can you review this as is?

PR #1264, opened 2017-01-20
Add HELP_EMAIL setting; Use it in reply-to for emails

jseppi:
work-in-progress

So, I thought this would be as simple as providing a `reply_to` arg to Django's `send_mail` method, but unfortunately, it does not accept any such argument. Instead, I had to use the underlying `EmailMultiAlternatives` class.

I also didn't want to hardcode our production email address (`calc@gsa.gov`), so I crafted a new environment variable setting, `HELP_EMAIL` (other name suggestions appreciated), and use that as the `reply-to` header AND use it in the footer template in palce of the previously hardcoded email address.

Bonus feature: I added a simple utility method to collapse the extraneous newlines that were present in the plaintext email content.

To Do:
- [x] finish tests for new `send_mail` function
- [ ] update all user-provided services


jseppi:
ready for review, @toolness 

(I still will need to update the user-provided services before merge)

jseppi:
[As a user I should be able to respond to notification emails so that I can ask questions and get clarifications.](https://trello.com/c/DpLIrWYl/224-as-a-user-i-should-be-able-to-respond-to-notification-emails-so-that-i-can-ask-questions-and-get-clarifications)

data_capture/email.py
```diff
@@ -17,6 +20,42 @@ def __init__(self, was_successful, context=None):
...
+    >>> collapse_and_strip_tags('\\n\\n<p>hi james</p>\\n\\n\\n')
+    '\\nhi james\\n'
+    '''
+    return re.sub(r'\n+', '\n', strip_tags(text))
+
+
+def send_mail(subject, body, to, html_message=None, reply_to=None):
+    '''
+    Django's convinience send_mail function does not all
+    specification of the reply-to header, so we instead use
```
toolness: Is there a word missing from the first sentence of this docstring?  "does not all specification" is confusing to me...

hourglass/settings.py
```diff
@@ -60,6 +60,8 @@
 
 DEFAULT_FROM_EMAIL = os.environ['DEFAULT_FROM_EMAIL']
 SERVER_EMAIL = os.environ['SERVER_EMAIL']
+# TODO: include HELP_EMAIL in default context for use in templates
```
toolness: Should this TODO be removed?

data_capture/email.py
jseppi: whoops. "all" -> "allow"

hourglass/settings.py
jseppi: yes

PR #1265, opened 2017-01-23
Upgrade to python 3.6

jseppi:
This PR upgrades CALC to Python 3.6 🎉


jseppi:
Currently blocked by `defusedxml` not working with Python 3.6, but there is an open PR to fix that: https://github.com/tiran/defusedxml/pull/4/files

jseppi:
The `defusedxml` maintainer is likely going to release a new version this week that will work with Python 3.6. The pre-release is already up https://github.com/tiran/defusedxml/releases/tag/v0.5.0.rc1

toolness:
Um, a weird thing I just discovered is that mypy apparently doesn't support all of 3.6 yet, such as f-strings, which is funky.

jseppi:
It _seems_ like it is something to do with the bash magic that is going on, but I haven't been able to unwind it all yet.

jseppi:
Hrm, if I'm understanding the intent of the `try/except` block at the top of `test_sauce_connect.py`, it looks like the problem during the upgrade was the exception class just needed to be changed from `SystemError` to `ModuleNotFoundError` (or `ImportError`): https://docs.python.org/3/whatsnew/3.6.html#other-language-changes

docker_django_management.py
```diff
@@ -193,7 +193,7 @@ def execute_from_command_line(argv):  # type: (List[str]) -> None
 
     try:
         from django.core.management import execute_from_command_line
-    except ImportError as e:
+    except ModuleNotFoundError as e:
```
jseppi: This change wasn't totally necessary since `ImportError` still would have worked, but I think this error class is better. (ref: https://docs.python.org/3/whatsnew/3.6.html#other-language-changes)

frontend/tests/test_sauce_connect.py
```diff
@@ -4,7 +4,7 @@
 
 try:
     from .sauce_connect import SauceConnectTunnel, maybe_run_with_tunnel
-except SystemError as e:
+except ModuleNotFoundError as e:
```
jseppi: I believe this change is necessary because `ModuleNotFoundError` is _not_ a subclass of `SystemError`, so this `except` block wasn't executed as expected.

jseppi:
Sweet, that changed work. Now just have the `mypy` stuff to fix.

jseppi:
I scrapped the change that was causing `mypy` to fail b/c I don't know how to fix it.

Here's the output from the mypy error:

```
docker_django_management.py: note: In function "execute_from_command_line":
docker_django_management.py:196: error: Name 'ModuleNotFoundError' is not defined
```

from changing the `ImportError` in `docker_django_management.py` to the new (? better ?) `ModuleNotFoundError`.

data_capture/tests/test_templatetags.py
```diff
@@ -1,27 +1,21 @@
...
 
 
 class ReadableTimestampTests(TestCase):
     def test_it_works(self):
-        utc_datetime = timezone.make_aware(datetime(2016, 2, 11, 14, 30, 5))
+        utc_datetime = datetime(2016, 2, 11, 14, 30, 5)
         val = tz_timestamp(utc_datetime)
         self.assertEqual(val, 'Feb. 11, 2016, 9:30 a.m. (EST)')
 
-    def test_it_returns_orig_value_when_not_tz_aware_datetime(self):
```
jseppi: Removed this test because PEP 495, part of Python 3.6, allows naive datetimes to have `self.astimezone()` called on them and work, as documented at https://www.python.org/dev/peps/pep-0495/#conversion-from-naive-to-aware

frontend/tests/test_sauce_connect.py
jseppi: @toolness please check on this to make sure it is working as you intended

jseppi:
Please review, @toolness!

jseppi:
[Upgrade to Python 3.6](https://trello.com/c/QGjByzUm/232-upgrade-to-python-3-6)

jseppi:
ping @toolness

toolness:
O sorry buddy.

So this is ready for review?  Or are there still mypy issues?

jseppi:
rdy4rvw

frontend/tests/test_sauce_connect.py
toolness: Ahh yeah, sorry about this, bleh.  This is happening because this doggone test suite is running itself not only as a script, but as a file passed directly to `python`, which means that a relative module import like this is going to raise a `ModuleNotFoundError`.  You can simulate this use case by running e.g.:

```
docker-compose run app python frontend/tests/test_sauce_connect.py
```

... and verifying that the `ModuleNotFound` error doesn't propagate.  Blegh, this took me a little while to reverse-engineer, and I was the one who wrote it just a month or two ago!  😞

toolness:
Ok, I just filed #1380, which should help anyone who's negatively impacted by this PR.

PR #1268, opened 2017-01-24
Generate <th> by iterating through form fields

toolness:
This is against #1245 and displays the header row by iterating through all the fields in an empty `Schedule70Row` instance.

### DIFF
--- a/data_capture/templates/data_capture/price_list/tables/schedule_70.html
+++ b/data_capture/templates/data_capture/price_list/tables/schedule_70.html
@@ -1,12 +1,12 @@
 <table class="price-list-table">
   <thead>
     <tr>
-      <th>SIN(s) proposed</th>
-      <th>Service proposed <span>(e.g. job title/task)</span></th>
-      <th>Minimum education <br/>/ certification level</th>
-      <th class="number">Minimum years <br/>of experience</th>
-      <th>Unit of issue</th>
-      <th class="number">Price offered to GSA <span>(including IFF)</span></th>
+      {% for field in header %}
+        <th {% if field.name == 'min_years_experience' or field.name == 'price_including_iff' %}class="number"{% endif %}>
+          {{ field.label }}
+          {% if field.help_text %}<span>({{ field.help_text }})</span>{% endif %}
+        </th>
+      {% endfor %}
     </tr>
   </thead>
   <tbody>
--- a/data_capture/schedules/s70.py
+++ b/data_capture/schedules/s70.py
@@ -181,7 +181,8 @@ def glean_labor_categories_from_book(book, sheet_name=DEFAULT_SHEET_NAME):
 class Schedule70Row(forms.Form):
     sin = forms.CharField(label="SIN(s) proposed")
     labor_category = forms.CharField(
-        label="Service proposed (e.g. job title/task)"
+        label="Service proposed",
+        help_text="e.g. job title/task"
     )
     education_level = forms.CharField(
         label="Minimum education / certification level",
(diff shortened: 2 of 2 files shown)


data_capture/templates/data_capture/price_list/tables/schedule_70.html
```diff
@@ -1,12 +1,12 @@
...
   <thead>
     <tr>
-      <th>SIN(s) proposed</th>
-      <th>Service proposed <span>(e.g. job title/task)</span></th>
-      <th>Minimum education <br/>/ certification level</th>
-      <th class="number">Minimum years <br/>of experience</th>
-      <th>Unit of issue</th>
-      <th class="number">Price offered to GSA <span>(including IFF)</span></th>
+      {% for field in header %}
+        <th {% if field.name == 'min_years_experience' or field.name == 'price_including_iff' %}class="number"{% endif %}>
```
toolness: Is this `class="number"` actually doing anything useful anymore?  In web inspector it seems `th.number` just has `text-align: left` but that doesn't do anything because *all* the table headers seem to be left-aligned now.  Should we just get rid of the class?

toolness: Also, I originally wanted this class setting to be determined by introspecting on the field type, e.g. to see if it was a `DecimalField` or an `IntegerField`, but it doesn't seem like there's an easy way to do that in templates, so I just hard-coded the field name. Blerg.

toolness:
One thing this PR doesn't have is the intentional `<br>`'s that exist in the table headers in #1245. I guess one way to get around that is to make the `label`s of the fields actually contain HTML and apply the `|safe` template filter on them during the rendering... Blerg.

jseppi:
I still am a little worried about naively looping through the model fields. I guess with this implementation the headers and value rows will have columns in the same order, but we still have no guarantee that either of them will render in the exact order we want for display.

toolness:
Actually, while we could ask on slack, I am pretty sure Django makes sure that looping through the model fields works in the order that they're defined in the class, because that's how it's done in a bunch of examples in Django's own documentation, e.g. [looping over the form's fields](https://docs.djangoproject.com/en/1.10/topics/forms/#looping-over-the-form-s-fields).  If the looping order was technically undefined/unpredictable I'm pretty sure that Django's docs would have actively discouraged folks from using it in templates.

jseppi:
Well, that's good then!
Should we have a test that they are defined in the order that we want our tables displayed then?

toolness:
Hmm I don't think we really need to, since it seems like a part of Django that lots of people rely on behaving this way (I've definitely written other Django apps that rely on it)... If it *does* ever manage to not show up in the order we expect, I guess a regression test would be a good idea, but I don't really think we need to write one pre-emptively... hmm.

jseppi:
I don't mean with Django not operating the way we expect - more like if we (or future devs) add a new field to `SubmittedPriceListRow` that we don't want in the table, or that is inserted in-between existing fields, then the rendered table will change in a possibly unexpected way.

toolness:
Er wait, though, it's not a `SubmittedPriceListRow` that we're displaying though, is it?  I thought it was a `Schedule70Row`, which is a `Form` (not `Model`) subclass expressly designed for the purpose of validating and displaying fields in a specific order?

We could add a docstring/comment to the class explaining that the order of the form field definitions will be reflected in the ordering of the columns of the table output, I guess?

jseppi:
Sorry, yeah, `Schedule70Row`.

I dunno, I guess it's fine if you say it's fine :) I just never would have guessed that the definition of the class's fields affected the order, but I suppose I don't really grok how python/django work still.

PR #1269, opened 2017-01-24
Add send_example_emails management command

jseppi:
Open against `html-admin-emails` (PR #1263)

This PR adds a management command to send example emails of all the email templates in the Data Capture app. It works best by having `docker-compose up` running in one terminal, and then running `docker-compose run app python manage.py send_example_emails` in another terminal. After running, you can then see the emails in your mailcatcher instance!

This actually already helped me find two different template bugs in our email templates!

To Do:
- [x] get feedback on if this implementation is acceptable
- [x] add tests
    -  Can't think of a good way to do this ~~including some kind of test that all email templates are included in the command (so that if a new email template is created we don't forget to add it to `sendexampleemails`)???~~


data_capture/management/commands/sendexampleemails.py
```diff
@@ -0,0 +1,84 @@
...
+from data_capture import email
+
+
+@click.command()
+@click.argument('to', default='user@example.com')
+def command(to):
+    '''
+    Send test emails
+    '''
+    pl_ctx = {
```
jseppi: I didn't want to actually create any `SubmittedPriceList`, `BulkUploadSource`, `User`, etc. models in this management command, so instead I leverage the fact that Django templates operate on class attributes and dictionary keys in the same way, and I just use mocked-up dictionary-based contexts when rendering templates.

toolness:
oh sweet! can u just add a smoke test that runs the cmd and makes sure it doesn't bust a nonzero exit code? i think my recent PRs for some of those changelog management cmds have examples of how to do this.

data_capture/management/commands/sendexampleemails.py
toolness: noooooice

jseppi:
Alright, @toolness, ready for final review.

jseppi:
[As a designer/developer, I should be able to view preview emails so that I know how the emails look.](https://trello.com/c/UeahxqUw/221-as-a-designer-developer-i-should-be-able-to-view-preview-emails-so-that-i-know-how-the-emails-look)

PR #1270, opened 2017-01-24
[WIP] Use Django Sites framework

jseppi:
work-in-progress


data_capture/templates/data_capture/email/approval_reminder.html
```diff
@@ -2,7 +2,7 @@
   This is your periodic reminder to check on and approve submitted price lists in CALC.
 </p>
 <p>
-  <b>{{ count_unreviewed }}</b> price list{{ count_unreviewed|pluralize:"s have,has" }} not been reviewed.
+  <b>{{ count_unreviewed }}</b> price list{{ count_unreviewed|pluralize:" has,s have" }} not been reviewed.
```
jseppi: Seems like this change is already in `html-admin-emails`, so not sure why it is showing up here o_O

meta/migrations/0001_initial.py
```diff
@@ -0,0 +1,24 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def insert_site(apps, schema_editor):
+    print("running insert_site")
+    Site = apps.get_model("sites", "Site")
+    print(Site.objects.all())
+    site = Site.objects.get(pk=1)
```
toolness: Er yeah, when you first start a django project, there's no sites in it, so I think the appropriate thing to do here would be to detect to see if there's any existing sites, and if not, bust one up?

Alternatively, since this is all editable via django-admin, we could just ignore any migrations, and count on the admins to create a site model.  The `/healthcheck` endpoint could verify that a site with `settings.SITE_ID` actually exists, and perhaps even also check that the `Host` header matches it.

toolness: Hmmmmmmmmm maybe I will tinker with some stuff and bust a PR against this one?

toolness: Oh nevermind, I see you were just following the advice in [Enabling the sites framework](https://docs.djangoproject.com/en/1.10/ref/contrib/sites/#enabling-the-sites-framework)! Neat, I didn't realize that was the thing to bust.

toolness: Uh, those docs do indeed seem to imply a chicken-and-egg problem, whereby a `Site` object is created as the result of a `post_migrate` signal, yet they then advise actually setting the site information in a data migration, which by definition occurs *before* a `post_migrate` signal is sent. Ummm.

toolness: OK FINE THIS FRAMEWORK IS CONFUSING.

toolness: I mean, one weird way to proceed would be to run `manage.py migrate sites` which would run *only* the sites framework migrations, and create the appropriate `Site` object when it was done.  Then you could run `manage.py migrate` which ought to work because at that point the `Site` object will exist.

Still, running `migrate` multiple times feels super weird.  I don't really like where this is going.

toolness: WHY IS THIS SO CONFUSING

toolness: I think we might need jacobian for this one... [Asked on slack.](https://gsa-tts.slack.com/archives/django/p1485355022000019)

jseppi: yeah, this thing is no bueno

toolness:
Ok, I just pushed to this PR branch, I hope that's ok... figured it'd be a bit weird to have a PR against a PR against a PR.

On Slack, CM suggested just going the route of checking to see if a default site exists in our migration, and creating it if not.  Because that meant adding branching logic to our migration, I decided to write tests for it, which necessitated factoring out `insert_site` into a separate module, because `0001_migration` can't directly be imported from python code since it starts with a number. lol.

jseppi:
Couple questions about this entire approach:

1. When actually using this to build URLs for request-less processes (ie, the `rq_scheduler` email sending task), we won't know the protocol, so I guess we'll have to hardcode adding `https://`. Are we ok with that level of hard-coding?
1. If we now have a dedicated migration that always sets the domain to `calc.gsa.gov`, how is this better than just hardcoding `calc.gsa.gov` into some site setting in `settings.py` (via an env var possibly)? Is it just that some future developer who wants to deploy CALC the application for multiple domains would instantly know what to do because we're using the Sites framework? Honestly that seems a little far-fetched to me, and now, instead of just changing an env-var or value in `settings.py`, they'll also have to make a new migration, right?

This seems like quite a lot of bizarre yak-shaving just to have a correct link in the `approval_reminder.html` emails. I'm probably more comfortable with just not having a link in that email at all :\

toolness:
So I think it's fine to hard-code `https` given that the world (especially the gov't, but the world is general) is moving towards an https-everywhere strategy, especially where functionality that requires some kind of login is concerned.  Even new HTML5 APIs increasingly only work with https, and given the fact that getting SSL certs is now free via services like letsencrypt, it seems like assuming a staging/production instance is running on http is a no-brainer... Hmm. That said, though, if we've got [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) on the site, it might actually be fine to just use `http` links, so long as the user has visited the site before using `https`.  I guess another alternative is to check if `DEBUG` is `True`, and if so, use `http`, but otherwise assume `https`.

The migration is really just included to set a reasonable default, I think--personally I think the "name" field is probably more useful to set a default for than the "domain" field, so long as we actually use that name value in our templates.  That way we could set the name to "CALC Staging" or something, and other governments that wanted to use the site could set it to "CALC Ohio" or whatever and the rest of the site would change.

Anyhow, the idea behind the sites framework is that its models are configurable via the Django Admin UI (kind of like how a Wordpress or Discourse instance has admin settings that are configurable via their web UIs).  So for staging and dev, we should just go into the admin UI and change the setting manually.  Similarly for if a different state/national government wanted to deploy CALC to a different place.  We'd just add docs to the setup instructions telling users to go to the sites admin and change it as needed.

toolness:
Er, to clarify, this is what the site admin view looks like, at `/admin/sites/site/1/`:

screen shot 2017-01-25 at 11 18 42 am

So any admin who wanted to customize that would do it there, rather than changing environment variables or making migrations.  Oh, and unless we actually add the permissions explicitly, COs and Data Administrators won't be able to change that.

jseppi:
Hrm, looks like the site [is not properly being set by migration](https://travis-ci.org/18F/calc/builds/195217830#L813) :(

```
=================================== FAILURES ===================================
_______________ SiteTests.test_default_site_is_set_by_migration ________________
self = <meta.tests.test_site.SiteTests testMethod=test_default_site_is_set_by_migration>
    def test_default_site_is_set_by_migration(self):
        site = Site.objects.get_current()
>       self.assertEqual(site.domain, 'calc.gsa.gov')
E       AssertionError: 'example.com' != 'calc.gsa.gov'
E       - example.com
E       + calc.gsa.gov
```

toolness:
Ummm that is super weird.  I am not able to reproduce that on my end... @jseppi can you try running it on your local machine?  Does that test pass?

jseppi:
Fails locally also. I ran:

```
docker-compose down -v
docker-compose build
docker-compose run app python manage.py syncdb
docker-compose run app python manage.py ultratest
```

toolness:
Hmm, it's still working on my end--I thought maybe I had forgotten to commit a file or something, but that doesn't seem to be the case.  I am suuuuper confused.

toolness:
Oh weird! Nevermind, I *am* able to reproduce it, but only if I run all the tests... Running just the tests for `meta`, everything works fine.  So something bizarro is going on.

jseppi:
Yeah, something very weird.

I also just redid my steps from above, it still failed, but then I `dbshell`'d and it _looks_ like it worked:

```sh
$ docker-compose run app python manage.py dbshell
psql (9.4.9, server 9.4.10)
Type "help" for help.

calc=# select * from django_site;
 id |    domain    | name
----+--------------+------
  1 | calc.gsa.gov | CALC
(1 row)
```

jseppi:
wtf. And even after doing that `dbshell`, re-running ultratest still fails!

toolness:
Ok, I think it might have something weird to do with `LiveServerTestCase`s.  Running the following, for instance, consistently fails:

```
py.test frontend/tests/test_qunit.py meta/tests/test_site.py
```

toolness:
Ugh, this is really ridiculous.  @jseppi do you mind if I just remove the offending test?  There would still be tests for `insert_site` which I think should be good enough.

toolness:
Alternatively I guess we could just remove the migration and `insert_site` entirely, and rely on the admins manually changing the default "example.com" site using the admin UI.  Blaaargh.

toolness:
actually, something as simple as putting the following above `SiteTests` in `test_site.py` does the trick:

```python
from django.test import TransactionTestCase

class AarghTests(TransactionTestCase):
    def test_blah(self):
        pass
```

Aargh.

toolness:
Ok, I *think* I fixed it, but OMG this is so friggin weird. I dislike it.

toolness:
One thing I particularly dislike about this is that if a test happens to run code that uses the sites framework, there is basically no telling whether the default site will be called `example.com` or `calc.gsa.gov`, which is completely stupid.

As far as I can tell, there doesn't seem to be a way to work around this.  Not unless we can somehow disable the sites framework's `post_migrate` thing from happening altogether.

Blah.  So I guess I recommend just rolling back most of this PR so that our code just uses the example.com site that the sites framework adds, and we have no weird data migrations that futz with sites.

toolness:
Or alternatively, we just kill this whole sites framework thing and go with an env var.

@jseppi do you have a preference?

toolness:
Well, I've just changed things so that we leave the default site set to example.com.  So I guess you can merge this if you think we should use this confusing sites framework, or close it and use a doggone env var.

Blah. I basically wasted my whole day on this. 😞

jseppi:
Sorry, dude :( Yeah, "confusing" was my takeaway from only a couple hours yesterday, but I was chalking it up to my inexperience with Django stuff.

Alternatively (to even using an env var) we can just keep the behavior as I have it in https://github.com/18F/calc/pull/1263, which defaults the `site_base_url` in the email methods to https://calc.gsa.gov. And maybe file another issue to use the env var...later.

toolness:
Hehe yeah let's just do that for now... I will close this PR and we can file an issue to pull stuff out of the env later!

PR #1275, opened 2017-01-25
Stylistic and copy edits to failed upload email

kategarklavs:
I made some stylistic and copy edits to this email template, giving it a slightly more conversational tone.

A question: What action should the user take after reading this email? (My guess is that they should reach out to the person who tried to upload data & ask them to try again, but this may not be correct.) It could be helpful to include a clear call to action so the user knows what to do next. (That said, if the user already knows what to do next, we may not need a CTA.)

### DIFF
--- a/data_capture/templates/data_capture/email/bulk_upload_failed.txt
+++ b/data_capture/templates/data_capture/email/bulk_upload_failed.txt
@@ -1,9 +1,17 @@
 {% load data_capture %}
 
-This message is in regards to Region 10 CALC data upload #{{ upload_source.id }},
+Hi [username],
+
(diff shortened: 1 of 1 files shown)


jseppi:
@kategarklavs - We actually have a currently open (but almost merged) PR to add a CTA to this template: https://github.com/18F/calc/pull/1263/files#diff-acd4303e76c77ce603b50c6de8d97de1

Also we don't really have usernames, so do you think just opening with "Hi," would be appropriate?

kategarklavs:
Ah, OK — that sounds good!

And, re: names: Yes, I think "Hi" or "Hi there" would be good, in this case. (I wasn't sure if we could use folks' actual names, but the straightforward "Hi there" will work well.)

PR #1279, opened 2017-01-25
A few copy edits

kategarklavs:
Just added a salutation & closing. Also, one question — how quickly will a CALC team member respond to a customer question? (I ask because it might be nice to add this information to the last paragraph: "If you have any questions, just reply to this email — we'll get back to you within one business day."

### DIFF
--- a/data_capture/templates/data_capture/email/price_list_approved.html
+++ b/data_capture/templates/data_capture/email/price_list_approved.html
@@ -1,3 +1,5 @@
+Hi [username],
+
 {% load data_capture %}
 <p>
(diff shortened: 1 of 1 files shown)


jseppi:
Good question. @tram: What do you think? Do we want to have even a soft-guarantee of question turn-around time?

kategarklavs:
@jseppi If not, that's also OK — if there is a guarantee, though, I always like to include it!

jseppi:
Would something like "We'll try to get back to you within one or two business days." work?

kategarklavs:
Howdy! Yep, I think that works — I might shorten it "We'll get back to you
within two business days" (only because then, if you all respond more
quickly, folks will be extra pleased at the "quick" response —
underpromise/over deliver).

PR #1280, opened 2017-01-25
V. quick copy edits

kategarklavs:
Just added a salutation/closing. And a question for you all — do we need to include a CTA here, or will recipients know what to do? Follow-up: Are there any tips/strategies we can offer to help folks from having their price lists rejected? If a tip sheet doesn't exist (/if it's something you're interested in), I'd be happy to draft one ...

### DIFF
--- a/data_capture/templates/data_capture/email/price_list_rejected.html
+++ b/data_capture/templates/data_capture/email/price_list_rejected.html
@@ -1,6 +1,8 @@
+Hi [username],
+
 {% load data_capture %}
 <p>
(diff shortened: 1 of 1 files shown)


jseppi:
Right now we don't have a great way to handle the case of when an admin rejects a price list. Ideally, we'd capture a reason and communicate that to the submitter. Once we have that functionality, we could add a CTA that asks the user to correct the issue and try resubmitting their price list.

kategarklavs:
Ah, OK — that makes sense. A note for the future, then!

PR #1282, opened 2017-01-25
Don't allow 'contains phrase' searches w/ empty string terms

toolness:
If a comma-separated list of search terms contains an empty string as one of the terms (e.g. due to a trailing comma) then the "contains phrase" search will actually return all results because every labor category contains the empty string.

This fixes that, I think.


toolness:
woooo i did something useful today lol

PR #1284, opened 2017-01-25
Region 10 Price List gleaner

jseppi:
This PR adds Region 10 price list gleaning classes and logic.

There are a number of TODO comments still in the code, primarily suggesting that duplicated code/logic from the S70 gleaning code be consolidated since so much is shared.


data_capture/schedules/s70.py
```diff
@@ -150,7 +113,6 @@ def glean_labor_categories_from_book(book, sheet_name=DEFAULT_SHEET_NAME):
         'min_years_experience': int,
         'education_level': extract_min_education,
         'unit_of_issue': extract_hour_unit_of_issue,
-        'price_excluding_iff': strip_non_numeric
```
jseppi: Took this out of the s70 gleaner because this field isn't actually used for anything

jseppi:
@toolness - is this looking like something you can start using for your price gleaning front-end?

toolness:
Yeah this is great, thanks!  I re-targeted this PR to be against #997 (the experimental price list analysis PR) instead of `develop`.

Issue #1202, opened 2017-01-05
Headers on tables are too big


hbillings:
They run into each other, force the widths of the content cells to be too great in some cases (especially numerals) and are generally a pain. They're our default `16px` currently. @ericronne, what do you think?
<img width="785" alt="screen shot 2017-01-04 at 5 44 12 pm" src="https://cloud.githubusercontent.com/assets/509309/21664696/6e8e73a2-d2a5-11e6-9224-0e275508a0e6.png">

hbillings:
@ericronne Here's what 14 looks like:
<img width="809" alt="screen shot 2017-01-06 at 1 24 12 pm" src="https://cloud.githubusercontent.com/assets/509309/21731799/74bbb00e-d413-11e6-9022-73e34b61917e.png">

Tell me more about the alignment?

ericronne:
Some of the alignment funkiness is discussed in [this slack thread] (https://gsa-tts.slack.com/archives/calc/p1484258246000547).

The biggest problem i see is with minimum years of experience. In the layout above, eg, my brain merges two columns together. I see, "5 Hour."

Here's one way things could be arranged...
![x](https://cloud.githubusercontent.com/assets/9259185/21936948/d64b27f2-d97a-11e6-91ca-3cbff12366ba.png)

And there's @jseppi's space-saving thought (#1240), which we can chat on, too

jseppi:
Will be closed by https://github.com/18F/calc/pull/1245

jseppi:
all fixed, @hbillings?

hbillings:
Done in #1285

PR #1285, opened 2017-01-27
Standardize table headers

hbillings:
After the error tables tweaks, I noticed there were some abnormalities as far as table headers across the site were implemented. These tweaks make them all the same size font, same boldness, right padding, etc.


jseppi:
Hrm, the alignment of the headers over `.number` columns (particularly the "Minimum years of experience") still looks a little weird to me:

screen shot 2017-01-27 at 8 39 16 am

(image from the `step/3/errors` table)

hbillings:
@jseppi I think that's because the data in those columns is right-aligned. The headers used to be right-aligned, also, but that got really funky visually. I suspect we probably don't really need to right-align years of experience, because at most those will be 2 digits. (What I'd really love to do is shorten the header text so we don't have a ginormous field for a 2-character value -- `Min. years exp` would fit a lot better but also wouldn't match exactly what's in the spreadsheet...sigh.)

jseppi:
Sigh indeed :(

The guidance we were originally following was from the 18F Content guide, which says:

> * When listing numbers, it’s good practice to align them to the right of their cell, with the same decimal precision (“40.50” and “1.00”) so that the numbers are easier to compare while scanning.
> * Always align column headings up with the values in the columns. For example, numeric column headings should be aligned right if the values are, too.

(from https://pages.18f.gov/content-guide/technical-and-interface-writing/)

Let's just center all of them!

hbillings:
@ericronne, can you weigh in on this? I feel like none of these options works well.

ericronne:
Would [**this**](https://github.com/18F/calc/issues/1202#issuecomment-272481565) work?

jseppi:
I'm not sure if the centered-but-right-aligned thing you have in that column is doable in html/css without some real funky markup.

hbillings:
I agree with James, though I do have one idea about how it could possibly work -- I'll try to get to that as soon as I sort out the mess that is my personnel record. 😞

hbillings:
Circling back around to this -- @ericronne @jseppi, thoughts?
screen shot 2017-02-01 at 4 03 04 pm

jseppi:
That looks very nice!

ericronne:
Dandy! Right-align the "Price offered to GSA" header, to match its content?

hbillings:
@ericronne No problem, but I thought you told me you wanted all the headers left-aligned. >_<

ericronne:
Hmmm.... nope? ...
https://github.com/18F/calc/issues/1202#issuecomment-272481565

ericronne:
p.s. So there.

hbillings:
Yeah, I know about that; I thought we had the conversation in coworking. Whatevs.

jseppi:
@hbillings - I just realized that I was wrong saying that we have this table's headers matching exactly (because of wanting copy-pasta to work). We already have minor differences in formatting/text in this table (and the regular non-error table) from what is in the spreadsheet, so I think you can safely ignore that erroneous comment I made earlier today.

hbillings:
@jseppi Okay, so...in order for me to make these textual changes, I need to change the `label` in the formfield definitions in `s70.py` -- like 
`education_level = forms.CharField(
        label="Min. education / certification level",
    )`

Is that going to mess with things? I'm scared it will...

jseppi:
Hrm... summoning @toolness!

toolness:
Hmm, I *think* the only place that label is used is when rendering the ol' table, so I think you are cool to edit it.  See if any tests break, I guess?  😁

data_capture/templates/data_capture/price_list/tables/schedule_70.html
```diff
@@ -3,7 +3,7 @@
   <thead>
     <tr>
       {% for field in header %}
-        <th {% if field.name == 'min_years_experience' or field.name == 'price_including_iff' %}class="number"{% endif %}>
+        <th {% if field.name == 'price_including_iff' %}class="number"{% endif %}>
```
jseppi: I think the `.../tables/fake_schedule.html` template might also need to have `min_years_experience` removed from the similar conditional.

hbillings:
@toolness I'm not sure what I'm doing wrong, but when I run ultratest in the docker bash shell, I get a metric ton of errors. They all seem to take the form `ERROR collecting uaa_client/tests/test_views.py` 😬

hbillings:
@jseppi It created it when I ran one of the test suites, I think? I'm not entirely sure, but since I didn't see it on `develop` I didn't want to commit it.

jseppi:
That's very weird. What is in that folder?

toolness:
I noticed a `jest` folder in my build at some point too, in a different branch.  I wonder if it's something temporary that Jest makes and deletes after a test run, but if we Ctrl-C in the middle of it, it doesn't get cleaned up, or something?

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
```diff
@@ -14,9 +14,8 @@
     {% with row_count=forloop.counter %}
     <tr>
     {% for col in row %}
-      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price' %}number{% endif %}">
-        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>
-          {% if col.name == 'price' %}
+      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'price_including_iff' %}number{% endif %}">
+        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>{% if col.name == 'price' %}
             ${{ col.value|floatformat:2 }}
```
jseppi: Hrm, I think here and in the `schedule_70.html` table you might want to check that the value is present before adding the `$`, otherwise you get "$(?)" as shown here:

screen shot 2017-02-07 at 12 04 55 pm

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
```diff
@@ -14,9 +14,8 @@
     {% with row_count=forloop.counter %}
     <tr>
     {% for col in row %}
-      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price' %}number{% endif %}">
-        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>
-          {% if col.name == 'price' %}
+      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'price_including_iff' %}number{% endif %}">
```
jseppi: Instead of `'price_including_iff'` (which doesn't exist) here in the `fake_schedule.html` table, you should use `col.name == 'price'`.  Currently looks like (no `.number` class applied):

screen shot 2017-02-07 at 12 05 11 pm

ericronne:
@hbillings could you pretty please reduce the bottom padding by about half a rem? I think that might help users connect the headers with the data in their columns.

image

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
hbillings: These should both be fixed!

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
```diff
@@ -14,11 +14,11 @@
     {% with row_count=forloop.counter %}
     <tr>
     {% for col in row %}
-      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price' %}number{% endif %}">
+      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'price_including_iff' %}number{% endif %}">
```
jseppi: This should be `{% if col.name == 'price' %}number{% endif %}` I think

jseppi:
Go ahead and merge if ready, @hbillings

PR #1289, opened 2017-01-27
Apply floatformat to price column

### LINKED ISSUE #1287: Too many digits showing in currency numbers
Just noticed while reviewing #1285. Haven't found root cause yet (but it is definitely not that PR).

screen shot 2017-01-27 at 8 34 54 am

(image from table in `/data-capture/step/4`)

jseppi:
fixes #1287

To Do:
- [x] add a test to make sure this regression doesn't happen again, man. This regression will not stand!


jseppi:
Alright, @toolness. Should be ready for final review. This is a priority to make it in, otherwise we'll have some nasty display bugs on the next release.

data_capture/templates/data_capture/price_list/tables/fake_schedule.html
```diff
@@ -14,8 +14,14 @@
     {% with row_count=forloop.counter %}
     <tr>
     {% for col in row %}
-      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price_including_iff' %}number{% endif %}">
-        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>{% if col.name == 'price_including_iff' %}${% endif %}{{ col.value }}</span>
+      <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price' %}number{% endif %}">
+        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>
+          {% if col.name == 'price' %}
```
jseppi: actually had another bug here because the previous if-statement was checking `price_including_iff`, which is not a valid field in the fake schedule row form.

jseppi:
Also, as I mentioned in slack, I think it would be better to have any custom field formatting defined back on the related `Form` class, but I don't know how to do that. I [asked in #django](https://gsa-tts.slack.com/archives/django/p1485534335000069), but no response yet, but this PR will at least get us back to where we were before the regression.

data_capture/templates/data_capture/price_list/tables/schedule_70.html
```diff
@@ -16,7 +16,13 @@
     <tr>
     {% for col in row %}
       <td class="{% if col.errors %}error{% endif %} {% if col.name == 'min_years_experience' or col.name == 'price_including_iff' %}number{% endif %}" {% if col.errors %}tabindex="0"{% endif %}>
-        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>{% if col.name == 'price_including_iff' %}${% endif %}{{ col.value }}</span>
+        <span {% if col.errors %}aria-describedby="error__row-{{ row_count}}__col-{{ forloop.counter }}"{% endif %}>
+          {% if col.name == 'price_including_iff' %}
+            ${{ col.value|floatformat:2 }}
+          {% else %}
+            {{ col.value }}
+          {% endif %}
```
toolness: ahhh looking at this code i understand why you wanted to see if it's possible to just override the `__str__` method or something. yeah, writing like this is really annoying!

PR #1299, opened 2017-01-31
add escalation_rate to admin price list views

### LINKED ISSUE #1087: Escalation rate isn't shown in admin
It seems like the escalation rate isn't shown in the admin edit details view.

jseppi:
Closes #1087 

Looks like this in admin pages for `SubmittedPriceList` details:
screen shot 2017-01-31 at 12 15 39 pm


data_capture/models.py
```diff
@@ -58,6 +58,7 @@ class SubmittedPriceList(models.Model):
         blank=True,
     )
     escalation_rate = models.FloatField(
+        verbose_name='escalation rate (%)',
```
jseppi: I wanted it to display with "(%)" in the Admin views, so I had to add a `verbose_name` here in the field constructor, which meant I also had to add a migration :\

toolness: Huh, does that affect how it's displayed in the PL upload form?

I *think* there may be some goofy way to override `verbose_name` in the admin view but if this does the job without adversely affecting the PL upload form then I'm fine with it!

jseppi: It doesn't affect the PL upload form -- I checked.

PR #1302, opened 2017-01-31
Add banner to show non-prod message

### LINKED ISSUE #1294: Use a ribbon/banner indicating non-production deployments of CALC
This would be useful for our dev/staging deployments as well as our one-off experimental ones like #997.

I actually filed an issue in the writing lab for this as https://github.com/18F/writing-lab/issues/191 back in July of last year but it didn't get resolved...

jseppi:
How's this for a showing a banner/message for our non-prod instances? Styling is pretty basic/ugly, but maybe that's good in this case :)

screen shot 2017-02-01 at 11 12 36 am

I made it by adding another config var, `NON_PROD_INSTANCE_NAME`, which, when set, will cause that yellow banner to be displayed. The value of that config var is shown in the message where "Local Development" is shown in the above screenshot.

I'm envisioning we set `NON_PROD_INSTANCE_NAME` to `'Development'` on `calc-dev`, `'Staging'` on `calc-staging`, and maybe `'Demonstration'` on @toolness's Price Analysis deployments.

Closes #1294


jseppi:
Plz review @toolness and @hbillings

jseppi:
Question: do we want to get rid of the "An official website of the U.S. government" banner when this banner is displayed?

hbillings:
yasssss I like this! I would vote for replacing the "official US gov" banner with the yellow one.

And maybe swap the US flag for an under construction gif. (kidding kidding)

hbillings:
Oh, super minor, but I think "local development" should be downcase.

jseppi:
cool, I'll make the change to get rid of the official banner.

The "Local Development" text is actually just whatever is set in the `NON_PROD_INSTANCE_NAME` environment variable, so you can change it locally to whatever you want by modifying your local `.env` file :)

jseppi:
Ok - changed it to not show the 'Official' gray banner when the `NON_PROD_INSTANCE_NAME` banner is shown. Screenshot in root comment has been updated.

Issue #1307, opened 2017-02-01
Seem to be missing CORS header on calc.gsa.gov


jseppi:
It seems that API responses (through api.data.gov), when made from the production url (`calc.gsa.gov`) are missing the `Access-Control-Allow-Origin` header, and thus being blocked by the browser.

I have no idea when or why this happened, but I noticed it tonight while deploying `v2.1.0` to production. It persisted after the deployment as well. It hopefully just happened as a result of the deployment but I'm not completely sure.

This behavior is not exhibited when visiting the "direct" address, ie https://calc-prod.app.cloud.gov (or on equivalent staging or dev instances), which leads me to believe that maybe the header is getting stripped by CloudFront, but of course, I can't tell. And I'm fairly certain this wasn't happening until recently (ie, during the deployment, otherwise I think _someone_ would have emailed us or otherwise let us know).

As a temporary fix, I've added the header in our API Umbrella configuration at api.data.gov:
<img width="508" alt="screen shot 2017-02-01 at 12 04 58 am" src="https://cloud.githubusercontent.com/assets/697848/22496750/26eb4f16-e812-11e6-99a4-8e790a12fae8.png">

cc @toolness

jseppi:
Hrm, new theory: CloudFront isn't stripping the `Access-Control-Allow-Origin` header, it isn't being added by the `django-cors-headers` middleware due to `HTTP_ORIGIN` not matching (see logic at https://github.com/ottoyiu/django-cors-headers/blob/862c68d1eb65455cf0051c4f124a4eb4341b68f2/corsheaders/middleware.py#L99)

jseppi:
I'm not even clear why we are using `django-cors-headers`. I think we could just add `Access-Control-Allow-Origin: *` to every response via an extremely simple custom middleware and get rid of that dependency all together.

jseppi:
Maybe there are security concerns with adding it everywhere -- so perhaps it should just be added to API responses.

jseppi:
Newer theory: 
* AJAX'd calls to get rates data do not have a `Cache-Control` header at all
* non-AJAX'd calls to the rates URLS do not have the `Access-Control-Allow-Origin: *` header.

So, new theory is that someone (perhaps even our New Relic monitor) visited the api endpoint directly, so it had no `ACAO` header. That response then got cached by CloudFront, and then subsequent AJAX'd requests to it were thus missing the `ACAO` header, leading to the CORS error in the browser.

jseppi:
[...]

PR #1308, opened 2017-02-01
Use simple custom CORS header middleware

jseppi:
This PR is an attempt to mitigate #1307.

I've removed the `django-cors-headers` module and replaced it with an extremely simple custom middleware called `ApiCorsMiddleware`. This new middleware adds the CORS `'Access-Control-Allow-Origin: *'` header to all responses to requests that start with the `API_HOST` prefix.


hourglass/middleware.py
```diff
@@ -52,3 +52,14 @@ def __init__(self):
...
         super().__init__()
+
+
+class ApiCorsMiddleware:
+    '''
+    Middleware to add CORS headers to all responses from requests to API_HOST
+    paths.
+    '''
+    def process_response(self, request, response):
+        if request.path.startswith(settings.API_HOST):
```
toolness: wait won't `API_HOST` be an absolute URL in production though? GArgh I'm confused

jseppi: hrm, yeah, you are right >.<

jseppi: I misunderstood what that setting was for. I remember now.

jseppi: I changed it to just have `/api` hardcoded in the check

toolness:
Er wait, so while `django-cors-headers` does seem overkill for our needs, do you think it was actively causing #1307?  

Regarding your comment in #1307:

> Hrm, new theory: CloudFront isn't stripping the `Access-Control-Allow-Origin` header, it isn't being added by the `django-cors-headers` middleware due to `HTTP_ORIGIN` not being set (see logic at https://github.com/ottoyiu/django-cors-headers/blob/862c68d1eb65455cf0051c4f124a4eb4341b68f2/corsheaders/middleware.py#L99). Maybe. I have no idea. ugh.

I believe that `HTTP_ORIGIN` (i.e. the `Origin:` header) is *always* sent by any browser that supports CORS, when it's aware that it's sending a cross-origin request.

Oh wait.  Um, I guess in this case it's really the API umbrella instance that's talking to our `/api/` endpoint, isn't it?  And *that* might not be setting an `Origin:` header... Ohhh man I am confused.

jseppi:
I actually still don't know what was causing the problem. We can hold off on this PR until we figure it out, but I wanted to have something ready in case I guess.

jseppi:
Newest theory: https://github.com/18F/calc/issues/1307#issuecomment-276713325

jseppi:
Closing because I _think_ we isolated the problem down to a (suprise) caching issue (ref https://github.com/18F/calc/issues/1307#issuecomment-276713325).

PR #1309, opened 2017-02-01
Add instructions on python debugging to README

### LINKED ISSUE #1300: Document/support how to use the python debugger w/ docker
@hbillings had to set this up last week and it looked confusing. We should document this and add any required packages to `requirements-dev.txt` so it's not hard for folks in the future!

@hbillings what did you have to do? Seppi said it had something to do with `--service-ports`...

hbillings:
Closes #1300


toolness:
Nice, thanks!

I think any local changes to your dependencies might be done quicker via adding a `RUN` instruction to the end of your `Dockerfile`, e.g. `RUN pip install my_better_debugger`, rather than adding it to `requirements-dev.txt`...  The issue with adding it to the requirements file is that because of the way Docker's layered file system works, it will basically have to re-download and re-install *all* the Python requirements, not just the one you added, whereas this isn't the case with adding a lone `RUN` command near the end of the `Dockerfile`.

In general this is a nice trick to be able to tinker with a Docker container's filesystem without having to wait a really long time for *all* your dependencies to be re-fetched and re-installed--it's what I want to do when I experiment with using a new node/python package with CALC--but it's probably a big enough topic that it deserves its own section.  I guess I can submit a separate PR that explains how to do it.

All that aside, I do also wish there was a way to allow for custom developer-specified node/python packages to be installed in the docker container without having to edit a versioned file like `requirements-dev.txt` or `Dockerfile` for it.  That just increases the likelihood that the module will accidentally be included in a PR. 😣

jseppi:
Merge if ready, @hbillings!

hbillings:
Okay, I updated this to reflect the new docs structure, and also removed the suggestion to add to `requirements-dev` (replaced with @toolness's bit verbatim). I wasn't quite sure where to throw it but I thought maybe `docker.md` made sense? Not sure; if it should be moved, lemme know!

toolness:
woot thanks!

PR #1312, opened 2017-02-02
Use cg-django-uaa for cloud.gov UAA auth.

toolness:
This excises our `fake_uaa_provider` and `uaa_client` directories in favor of bringing in [cg-django-uaa](https://github.com/18F/cg-django-uaa).

Notes:

* The `/logout` view isn't in cg-django-uaa because it's actually not specific to cloud.gov/UAA at all, so I moved it to `data_explorer`, which seems to be where the other "global" templates and views reside.

* We still have to define the `/uaa_client/login_error.html` (formerly `oauth2_error.html`) template somewhere so I decided to use `hourglass` for that, since it's where we override our `admin` templates... I guess we could move this template to `data_explorer` though, if we want to put it near the other "global" templates... Or alternatively we could move some of these "global" templates over to the `meta` package... or even `hourglass`.... Oy this is confusing.

To do:

- [x] Merge `TEMP_requirements.txt` into `requirements.txt` and get rid of references to it in `Dockerfile` and `.travis.yml`.


toolness:
@jseppi with the exception of the TODO which is purely for expediency (I didn't want to have to wait forever every time I switched to a new version of cg-django-uaa) I think this is actually good to go. Can you give it a review?

hourglass/settings.py
```diff
@@ -326,16 +326,12 @@
...
-
 if not UAA_CLIENT_SECRET:
     if DEBUG:
         # We'll be using the Fake UAA Provider.
         UAA_CLIENT_SECRET = 'fake-uaa-provider-client-secret'
-        UAA_AUTH_URL = UAA_TOKEN_URL = 'fake:'
-    else:
-        raise Exception('UAA_CLIENT_SECRET must be defined in production.')
+        if not is_running_tests():
+            UAA_AUTH_URL = UAA_TOKEN_URL = 'fake:'
```
toolness: The need for this extra conditional is annoying; for details see https://github.com/18F/cg-django-uaa/issues/18.

hourglass/settings.py
```diff
@@ -326,16 +326,12 @@
...
-if DEBUG:
-    INSTALLED_APPS += ('fake_uaa_provider',)
-
 if not UAA_CLIENT_SECRET:
     if DEBUG:
         # We'll be using the Fake UAA Provider.
         UAA_CLIENT_SECRET = 'fake-uaa-provider-client-secret'
-        UAA_AUTH_URL = UAA_TOKEN_URL = 'fake:'
-    else:
-        raise Exception('UAA_CLIENT_SECRET must be defined in production.')
```
toolness: Don't need to worry about raising this exception because cg-django-uaa has a bunch of validation built into it already; it'll raise an `ImproperlyConfigured` exception with a helpful error if things are misconfigured.

TEMP_requirements.txt
```diff
@@ -0,0 +1,6 @@
+# This is a temporary file that's being included separately just
+# so it's easier to quickly change the version of cg-django-uaa
+# being used. Before final merge, its contents will be moved to
+# requirements.txt.
+
+git+https://github.com/18F/cg-django-uaa.git@6eaa9de0e89e41b3a03ee980745e57878540d336
```
jseppi: Will cg-django-uaa eventually be published on PyPI?

jseppi:
Good point about the organization of the "global" templates. My instinct is that they should be in `hourglass/` but that should be done in another PR.

hourglass/settings.py
jseppi: Maybe put an explanatory comment for this conditional

hourglass/templates/uaa_client/login_error.html
```diff
@@ -5,7 +5,7 @@
   <div class="row card">
     <div class="card__content">
       <h1>Login Error</h1>
-      {% if error_code == 'invalid_code_or_nonexistent_user' %}
+      {% if error_code == 'authenticate_failed' %}
```
jseppi: with `cg-django-uaa`, will the `else` portion of this conditional ever happen?

toolness: Yep, the docs for `cg-django-uaa` don't list them all--they basically tell the reader to look at cg-django-uaa's source code right now--but other error codes indicate problems with the OAuth2 handshake, e.g. missing querystring args and stuff like that.

TEMP_requirements.txt
toolness: Yup yup, I just wanted to get started on this PR before making publishing the first version, especially since any problems encountered in this integration would inform the first version.

jseppi: Totally makes sense

toolness:
Ok, I just published cg-django-uaa 0.0.1 to pypi yesterday and modified this PR to use that in `requirements.txt`, so we're good to go!

PR #1316, opened 2017-02-03
Remove ethnio stuff

jseppi:


jseppi:
[Remove ethnio](https://trello.com/c/eLA4utJW/229-remove-ethnio)

toolness:
😢 

Do we want to keep around the actual functionality just in case we're able to use ethnio again in the future?  Or if others who want to use CALC want to use it?  I dunno.

jseppi:
nah, dead code!

PR #1320, opened 2017-02-03
Stop parsing S70 Price List when 'Most favored customer' is encountered

### LINKED ISSUE #1318: S70 parser doesn't always stop at the end of the pricing table
screen shot 2017-02-03 at 12 39 54 pm

(See [example price list from today in slack](https://gsa-tts.slack.com/files/trammell/F41E96GBZ/adnet_priceproposaltemplate_final_7.16.15.xls))

jseppi:
This PR mitigates the issue identified in #1318.  Basically, the problem was that if a S70 price list worksheet's prices table has no lines between it and the start of this block of cells:

>Most Favored Customer’s Discount or Discount Range (MFC) __None__.
>GSA’s Discount or Discount Range W/O IFF _2.5%___.
>Prompt payment discount _1% net 10____.

then the parser kept on going through those cells, which resulted in bizarre errors being shown to the user.

To mitigate it, I adjusted the stop logic to look at the first cell of each row to see if it starts with "Most Favored Customer" (case-insensitive), and to stop if that text was found.

Closes #1318


data_capture/schedules/s70.py
```diff
@@ -150,7 +153,6 @@ def glean_labor_categories_from_book(book, sheet_name=DEFAULT_SHEET_NAME):
         'min_years_experience': int,
         'education_level': extract_min_education,
         'unit_of_issue': extract_hour_unit_of_issue,
-        'price_excluding_iff': strip_non_numeric
```
jseppi: `price_excluding_iff` is not used anywhere, so we don't need to try to coerce/glean it. I had removed it over in https://github.com/18F/calc/pull/997, but forgot to do so in `develop`.

PR #1323, opened 2017-02-06
Create an automated test suite for use against production deployments

### LINKED ISSUE #1199: Create an automated test suite for use against production deployments

toolness:
This is an attempt to fix #1199.

Note that I am actually just writing this in Python for now, rather than node, which is what newrelic synthetics supports.

Right now I am thinking that Python might be easier because:

* It allows us to reuse our already existing knowledge of writing tests in Python
* It means the code we write can *potentially* be run by our `rq_scheduler` process, obviating the need for newrelic synthetics.  I know it's weird, to have our production instance test its own cloudfront setup, but I *think* it should work, since when our scheduler process does a DNS lookup for `calc.gsa.gov`, it will get the cloudfront-fronted IP rather than e.g. the loopback address. ~~(I've also added a test to ensure that the `X-Amz-Cf-Id`header is in the response, so the suite makes sure this is the case.)~~ We can add a separate, optional test that verifies that the target is fronted by cloudfront, but these tests are also valuable to run against dev/staging too.

Also, if execution speed becomes a concern, we can always run `py.test` in parallel with [xdist](http://pytest.org/dev/xdist.html).

~~Currently this new test suite is "hard coded" to work against calc.gsa.gov, but it's all DRY'd out such that it'd be easy to allow a different URL to be provided via e.g. environment variables or something.~~

Also, since this test suite isn't actually testing the repo's codebase, it doesn't run by default when `py.test` or `ultratest` is run; it needs to be explicitly run via `py.test production_tests`.

By default, the suite tests against `https://calc.gsa.gov` but a different origin can be specified via a command-line option like so:

```
py.test production_tests --origin=https://calc-dev.app.cloud.gov
```

(This was done by using [pytest's plugin functionality](http://pytest.org/2.2.4/plugins.html).)


toolness:
Ok @jseppi I think this is good to go!

toolness:
Oh, um, I guess if this seems like a good idea to you, we should document it in the ol' README, and/or turn it into a `manage.py` command maybe.

toolness:
Ok, just documented everything in `deploy.md`.

jseppi:
So for now, you're envisioning just running this locally?

toolness:
Yeah, I think in a subsequent PR we can try adding a periodic `rq_scheduler` job that runs it automagically.

toolness:
Woot! Filed #1331 to address that.

PR #1324, opened 2017-02-06
Upgrade python packages

jseppi:
- Upgrade all python packages except `djorm-ext-pgfulltext` since it broke tests when I did
- Remove `httmock` package (I believe this was used to test the UAAClient or Fake Cloud.gov, which have been extracted)
- Fix some minor `flake8` issues that cropped up due to updated rules after upgrading that package

Motivation is that we have to provide PDFs of all our scans/analyses for the 3-Year ATO and I wanted to have all/most packages up-to-date for the Gemnasium scan report.


toolness:
Nice!  Did you look at the release notes for `djangorestframework` in particular?  It doesn't seem to be very semver-friendly as we had to fix some backwards-incompatible changes when upgrading from 3.0 to 3.1 in https://github.com/18F/calc/pull/292...

toolness:
Ah good catch on httmock, I believe it was indeed only used by the UAA code.

toolness:
Just looked at the [3.5.x series release notes for DRF](http://www.django-rest-framework.org/topics/release-notes/#35x-series) and it doesn't look like there's any breaking changes. Woot!

toolness:
[`rq`'s changelog](https://github.com/nvie/rq/blob/master/CHANGES.md) indicates no breaking changes too.

toolness:
The only breaking change in django-cors-headers has something to do with an [undocumented `CorsModel` class being dropped](https://github.com/ottoyiu/django-cors-headers/blob/master/HISTORY.rst#200-2017-01-07), which we don't need to worry about because we didn't use it. Woot!

toolness:
It looks like upgrading whitenoise is going to break one of the production tests I'm adding in #1323, since whitenoise now adds [`immutable`](https://github.com/evansd/whitenoise/commit/b7deab1fbe75c6bb4fefbb036a5bcb22b1806653) to the `Cache-Control` header, but I can deal with that once this lands.

jseppi:
Sweet, thanks for checking out all those changelogs!

Issue #1322, opened 2017-02-06
gulp build is kinda slow (at least on my system)


toolness:
This is from the end of running `docker-compose build` on my system:

```
Step 15 : RUN gulp build
 ---> Running in 25172e667046
[12:45:46] Gulp is running in production mode
[12:45:46] Using gulpfile /calc/gulpfile.js
[12:45:46] Starting 'sass'...
[12:45:46] Starting 'lint'...
[12:45:46] Starting 'js:common:vendor'...
[12:45:46] Starting 'js:data-explorer:vendor'...
[12:45:46] Starting 'js:data-explorer'...
[12:45:46] Finished 'js:data-explorer' after 19 ms
[12:45:46] Starting 'js:data-capture'...
[12:45:46] Finished 'js:data-capture' after 7.22 ms
[12:45:46] Starting 'js:styleguide'...
[12:45:46] Finished 'js:styleguide' after 1.82 ms
[12:45:46] Starting 'js:tests'...
[12:45:46] Finished 'js:tests' after 7.03 ms
[12:45:48] Finished 'js:common:vendor' after 1.92 s
[12:45:49] Finished 'js:data-explorer:vendor' after 2.42 s
[12:45:49] Starting 'js:legacy'...
[12:45:49] Finished 'js:legacy' after 22 μs
[12:45:54] Wrote frontend/static/frontend/built/js/styleguide/index.min.js.map.
[12:45:54] Wrote frontend/static/frontend/built/js/styleguide/index.min.js.
[12:46:10] Wrote frontend/static/frontend/built/js/data-capture/index.min.js.map.
[12:46:12] Wrote frontend/static/frontend/built/js/data-capture/index.min.js.
[12:46:17] Wrote frontend/static/frontend/built/js/tests/index.min.js.map.
[12:46:17] Wrote frontend/static/frontend/built/js/tests/index.min.js.
[12:46:23] Finished 'sass' after 37 s
[12:46:23] Finished 'lint' after 37 s
[12:46:23] Starting 'js'...
[12:46:23] Finished 'js' after 14 μs
[12:46:23] Starting 'build'...
[12:46:23] Finished 'build' after 9.36 μs
[12:46:30] Wrote frontend/static/frontend/built/js/data-explorer/index.min.js.map.
[12:46:30] Wrote frontend/static/frontend/built/js/data-explorer/index.min.js.
 ---> 24ab0f652c62
Removing intermediate container 25172e667046
```

[...]

PR #1328, opened 2017-02-06
Only copy source code & run gulp for docker cloud deploys

toolness:
This helps mitigate #1322 by making things more efficient for the vast majority of use cases, and making things a bit more annoying for cloud deploys.

We do this by moving the `COPY . /calc/` and `RUN gulp build` commands out of the standard `Dockerfile`.  Instead, cloud deployers will need to create a new Dockerfile that concatenates the standard `Dockerfile` and a `Dockerfile.cloud-extras`, the latter of which contains the two missing commands, and then point to this new Dockerfile in their `docker-compose.override.yml`. This process is fully detailed in `README.md`.

Note that in order to accomplish this, I had to upgrade all our docker-compose files to [version 2](https://docs.docker.com/compose/compose-file/#/versioning).  (Uh, I also didn't realize until adding that link, that version 3 is the latest version. But apparently it's only useful for Docker Swarm stuff, which we don't use, so whatevs.)

Anyways, to upgrade to version 2 I just had to add an extra top-level layer to all the YAML files, and re-indent their existing content under a new top-level key called `services`.  **Aside from this minor change, none of the original content in the YAML files has changed.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ calc-prod_manifest.yml
 tag-message-*.txt
 release-*-instructions.txt
 release-*-instructions.html
+Dockerfile.cloud
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,13 +31,4 @@ COPY requirements-dev.txt /calc/
 
 RUN pip install -r /calc/requirements-dev.txt
 
-# The following lines set up our container for being run in a
-# cloud environment, where folder sharing is disabled. They're
-# irrelevant for a local development environment, where the /calc
-# directory will be superseded by a folder share.
-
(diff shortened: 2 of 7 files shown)


toolness:
Note that there doesn't seem to be an easier way to reuse code between multiple Dockerfiles ([The Dockerfile Explosion and the need for higher level tools](https://www.youtube.com/watch?v=IyuyA8rSBAo&list=PLkA60AVN3hh9gnrYwNO6zTb9U3i1Y9FMY&index=26) details this in-depth), which is why cloud deployers need to manually concatenate the two files.

Issue #1272, opened 2017-01-25
Modify footer design to accommodate the recent updates link more compactly


ericronne:
@jseppi added a useful "Read about recent updates to CALC" link to the footer. I'll adjust the design to attempt a more compact, shorter layout.

Layout from #1271 …
![image](https://cloud.githubusercontent.com/assets/9259185/22302177/419ae3fe-e2f3-11e6-9823-c343e8d677d1.png)

cc @hbillings

jseppi:
We could also potentially add the icons credits to the About page, maybe in a "Credits" section. And actually, after reading up on how we should be giving attribution, we should be linking the actual icon names to their associated Noun Project pages, per https://thenounproject.zendesk.com/hc/en-us/articles/200509928-How-do-I-give-creators-credit-in-my-work-

ericronne:
Ah, then we should indeed move the credits to the About page, or the attribution will get too long. Win win.

ericronne:
Any objections to deleting the contact email address and simply linking "email us"? That'd help, too.

ericronne:
@tram ^^

tram:
It would be nice to have the email address visible. If it's just in an email link, folks either have to…
1. know to roll over it to get an email address (if they are computer-savvy enough to know that's a thing they can do) and be motivated to do it when they see the link, or
2. use the browsers mailto: functionality which isn't always wired up right and/or trusted.

That said, it's not _that_ big a deal, but I'd rather cut some prose than the email address.

Also, are we using noreply-dev@calc or calc@gsa? No sweat either way. Just curious.

jseppi:
The address is filled in based on a setting. In prod, the setting value is "calc@gsa.gov". That screenshot was from my local dev, so it was a weird value.

jseppi:
This one closed @ericronne?

ericronne:
Yes! With #1333. Thx!

PR #1333, opened 2017-02-07
Widens get in touch block and floats logos

ericronne:
This will reduce text-wrapping frequency for larger viewports.


ericronne:
Will close #1272

jseppi:
Merge when you'd like, @ericronne

PR #1336, opened 2017-02-07
Tag and release v2.2.0

jseppi:
### Added

- A test suite for testing the production instance of CALC has been created
  (#1199).

### Changed

- The Schedule 70 price list parsing logic has been amended to address a bug
  where it attempted to parse rows outside of the price list table (#1318).

- Ethnio (used for recruiting users for research sessions) has been removed.

- The cloud.gov UAA authentication code has been extracted to a separate
  package called [`cg-django-uaa`](https://github.com/18F/cg-django-uaa).

- Most Python dependencies have been updated.

PR #1339, opened 2017-02-07
Mention not to squash or rebase in RELEASE.md.

toolness:
This mentions not to squash or rebase the release candidate PR into `master`, as that was the cause of the schism which prompted https://github.com/18F/calc/pull/1336#issuecomment-278130292 and #1337.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -63,8 +63,9 @@ To release version 0.0.4 of CALC:
     git push
     ```
 
-8.  Merge the PR into `master`. Once Travis CI is finished, the site
-    will be deployed to staging.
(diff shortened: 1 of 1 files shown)


toolness:
Hmm, actually, another option is to include the code to do the merge via the command-line (which would automatically close the PR).  An advantage of this is that it wouldn't be nearly as prone to human error as accidentally doing a squash/rebase via github (especially if one's defaults are set to squash/rebase, as yours are).

What do you think?

toolness:
Ehh, I'll just merge this for now and we can include the CLI instructions later if squashing continues.

jseppi:
SQUASHING WILL CONTINUE UNTIL MORALE IMPROVES

PR #1340, opened 2017-02-07
Set query when search field loses focus

### LINKED ISSUE #1334: Query string isn't always correctly built
This bug came up in a user interview. The user relies on the chart's average for price analysis, and a couple of weeks ago noticed that the chart was no longer responding properly to queries.

It looks like Search and Go don't build the proper query strings.

To reproduce the Go button issue:
- Type in Systems Administrator. Don't hit search or enter.
- Click in Proposed Price box. Enter a price (test was $53.13)
- Click Go next to proposed price field
- Chart will say $53.13 proposed, $111 average. Note that in this case, the only argument in the query string is `proposed-price`. There is no labor category in the string. 

To reproduce the Search button issue:
- Enter a labor category. Don't hit enter.
- Click Search.
- Chart will return all results. Note that no query string is formed.

toolness:
This ought to fix #1334.

### DIFF
--- a/frontend/source/js/data-explorer/autocomplete.js
+++ b/frontend/source/js/data-explorer/autocomplete.js
@@ -109,4 +109,11 @@ export function initialize(el, {
       setFieldValue(selectedInput);
     },
   });
+
+  $(el).on('blur', () => {
+    // If the user manually changed the field without actually
+    // selecting anything in the autocomplete list, it should take
+    // effect as soon as the user focuses on something else.
+    setFieldValue(el.value);
+  });
 }


jseppi:
Hrm, this is probably fine, though I could see some users finding the auto-search on blur a little confusing.

I think I would expect that if I typed something, but did not click Search, that nothing would change. The results would only change after I click the Search button (otherwise it isn't really doing anything).

toolness:
O.

I guess we could change it to work that way, though due to the fact that auto-searches occur on the change of any other field, I thought searching on blur would be the closest approximation to maintaining consistency with that sort of "always show results based on the current state of the form" approach. In some sense, the "search" button is actually almost as useless as the "Go" button next to the proposed price field.

I guess one alternative would be to change all the *other* fields such that they don't auto-submit the form whenever they change, requiring the user to always click "search" or press enter to submit the form... Or we could auto-submit the search field on a (debounced) per-keypress basis...

Hmm, looking at how Google works right now, it seems to not have a "Search" button at all, but rather displays a "Press Enter to search" tip below its autocomplete drop-down:

screen shot 2017-02-08 at 10 34 10 am

Given that our "Go" button is already useless, maybe we should just commit this quick-fix for now (it's actually much easier to implement than the alternative you suggested, due to the way our React/Redux data flow works), and wait for a more comprehensive redesign by @hbillings and @ericronne before implementing something else?

jseppi:
> I guess one alternative would be to change all the other fields such that they don't auto-submit the form whenever they change, requiring the user to always click "search" or press enter to submit the form

I don't think it needs to be taken to that extreme. And I don't think the Proposed Price Go should necessarily be a model of behavior since that was also a change made during the Great Reactification :)

My thought is that the labor category input is like the "master" field that kinda sets up a "search session," so having it get submitted in a perhaps unexpected way is more disorienting than adjusting a filter. But I don't know if that's how users perceive it.

I suppose I'm fine with merging as-is, especially given that we need a fix asap. @tram - you think this is ok?

hbillings:
I started to think about alternatives to this and basically went down a path of redesigning things. (We should probably get started on redesigning that front page sooner rather than later...)

So for this quick fix -- two thoughts.
1) What if the Search button itself reflected the fact that the data were being updated? I'm thinking grayed-out, with maybe the word "Searching" replacing the imperative verb. Then at least users who aren't expecting the search to happen automatically are given a clue as to what's happening.
2) What if we remove the "Go" button altogether?

ericronne:
Looks like we're redesigning the home page, woo!!
I'm feeling a live design chat coming on! 💥 
PING

jseppi:
Before we get to far on that we should get this or similar fix live since the bug was marked as High Priority :)

ericronne:
I think @toolness's fix is good.

tram:
Agreed! @toolness's fix sounds good to me for now. :-)

PR #1348, opened 2017-02-08
Control narrative updates for 3-year ATO

jseppi:
Work-in-progress during the ATO Sprint.

### DIFF
--- a/compliance/component.yaml
+++ b/compliance/component.yaml
@@ -25,15 +25,18 @@ satisfies:
   control_key: AC-2   # Account Management
   narrative:
     - text: >
-        Within our application (see cloud.gov for lower-level controls),
-        Contract Officer user accounts are created and managed by Data
-        Administrators. Authentication is provided by cloud.gov's User
+        Within the CALC application (see cloud.gov for lower-level controls),
+        authentication is provided by cloud.gov's User
         Account and Authentication (UAA) server, while authorization is
-        provided by the Django application; if a Data Administrator
-        hasn't explicitly created a user account for an individual and
-        associated it with the same government-issued email address that they
-        log in to cloud.gov with, the user is denied access to all parts
-        of the site requiring login.
+        provided by the Django application. Only users who successfully login
+        through the UAA and who have an associated (via @gsa.gov email address)
+        account in the CALC back end will be able to access privileged areas of
+        the application.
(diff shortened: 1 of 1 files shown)


compliance/component.yaml
```diff
@@ -112,30 +123,22 @@ satisfies:
...
-        hourglass/settings.py. Configurations which are
-        specific to one cloud.gov environment (i.e. either the dev, staging, or
-        production environment) are located in the appropriate manifest-*.yml
-        file or stored in and provided by a cloud.gov "user provided
-        service".
+        environments is located in the manifests directory of the code
+        repository. Each environment (dev, staging, production) has a separate
+        manifest file, which  is named "manifest-<environment label>.yml".
+        Application-level settings are specified in the code repository at
+        hourglass/settings.py.
```
jseppi: Should I say anything about user-provided services containing additional configuration values?

compliance/component.yaml
```diff
@@ -25,15 +25,18 @@ satisfies:
...
-        of the site requiring login.
+        provided by the Django application. Only users who successfully login
+        through the UAA and who have an associated (via @gsa.gov email address)
+        account in the CALC back end will be able to access privileged areas of
+        the application.
+        Data Administrator users can access the applications' administrative
+        back end, where they can administer, including creating new, user
+        accounts for both Data Administrators and Contract Officers.
+        Contract Officer users are not able to access to access the
+        administrative back end or perform any actions on user accounts.
```
hollyallen: This phrasing is a bit awkward. What are the different privileged parts of the application, and which users can access which parts?

compliance/component.yaml
```diff
@@ -88,20 +95,24 @@ satisfies:
 - standard_key: NIST-800-53
```
hollyallen: CA-8: Noah suggested specifying GSA Information Security team _or_ 3PAO but not both.

compliance/component.yaml
hollyallen: Sounds like a good idea. In addition, a direct link to an example file would help the testers.

compliance/component.yaml
NoahKunin: Just for the record, audio issue, other way around, let's keep the OR. :)

jseppi: Noah just corrected: keep the "and"

PR #1353, opened 2017-02-10
Modularize README and reorganize documentation

toolness:
This attempts to fix #1301 by splitting up our massive README into smaller markdown files and moving most/all of our docs into a `docs/` directory, which can be built with Sphinx (the tool used to document Python, Django, and all the nifty stuff on [readthedocs.org](https://readthedocs.org/)).

The driving philosophy of this PR is that project documentation should, for the most part, still be readable both in plain-text form and via [browsing on GitHub](https://github.com/18F/calc/tree/sphinx-docs#readme). Developers should not be *required* to visit a static site to learn how to set up CALC, though if such an option exists as an alternative, that's great.

In other words, Sphinx is an *enhancement* of our documentation, rather than a fundamental re-envisioning of it. Here's what it looks like right now:

toc

## Viewing the docs

Make sure you run `docker-compose build` first, then run `docker-compose up` and visit the `/docs/` endpoint on CALC.

## Building the docs

The docs are build as part of `gulp build` and rebuilt automatically whenever anything in `docs/` is changed, but if you need to manually rebuild them for some reason, you can do so with `gulp sphinx`.

## Notes

* There are some links to CALC's dev instance at https://calc-dev.app.cloud.gov/ in the docs, but when served through `/docs/` on the user's local dev server, they are rewritten to be against the user's local dev server. This is done using `sed`.

* I'm leaving `CHANGELOG.md` in the root directory but symlinking to it in `/docs/` so that it's both included in generated docs *and* still in the standard place recommended by http://keepachangelog.com. (Sphinx can't include docs located above the `/docs/` directory, which is why the symlink is needed.)

* There are some inconsistencies between the ways Sphinx's [recommonmark plugin](https://recommonmark.readthedocs.io/en/latest/) renders Markdown vs. the way GitHub does.  For the most part they're pretty consistent, but the following things don't seem to work with Sphinx markdown:
  * [ASCII tables](https://help.github.com/articles/organizing-information-with-tables/) don't seem to work; I had to convert the tables in our `deploy.md` to HTML tables. This makes them way less readable when read as plain text, which is a bummer.
  * Relative links to files in the repository can't be formatted in monotype text, at least not the way we've traditionally been doing it, which is weird.

[...]

Issue #1230, opened 2017-01-11
Building and changing node/python dependencies in docker is slow


toolness:
It's currently a bit of a hassle to add a single node or python dependency to our docker setup because of the fact that `docker-compose build` is running on a container that has no recollection of having installed or fetched any packages in the past--therefore it needs to re-download and re-install every single node/python package (and in the case of node, python dependencies also have to be re-downloaded/installed due to the fact that `pip install` occurs later in the `Dockerfile` than `npm install`).

This could get particularly painful once #1055 lands and we have a bunch of scientific python dependencies in addition to our existing ones; aside from updating dependencies, just *building* them could become a hassle.

Possible solutions:

1. I'm not sure if it's possible, but it'd be neat to be able to store the npm and pip caches in a Docker volume or something, so that at least re-installing the dependencies doesn't involve re-downloading them.
2. With [another open-source project I worked on](https://github.com/govlab/noi2#changing-the-dockerfile), we actually used *two* Dockerfiles: one was the "stable" base image that had most of our dependencies; it was stored on Docker Hub and made it really fast for devs to get started, since their machines could just download the Docker Hub image and not have to pip/npm install anything.  (It also made it really fast to run tests on Travis with!)  Then we had a second Dockerfile which trivially inherited from the first one and only installed the latest dependencies that were added as a result of recent PRs.  On a regular basis we'd migrate dependencies from the second Dockerfile to the first one in a way that didn't disrupt developer/deployment/CI workflows.

The second solution is one I've tried before and am pleased with, and also works nicely with Travis--thereby addressing the problem of having super slow CI builds whenever we add certain dependencies like numpy--so maybe we should just go with that. (It would obviously require changing Travis to run the *dockerized* version of our app, but doing that isn't hard.)

Also somewhat related: #569

toolness:
Regarding possible solution (1), now that we've moved all the cloud deployment-specific stuff into `Dockerfile.cloud-extras` in #1328, I think we might be able to also rejigger things such that we actually run `pip install` and `npm install` as part of the `ENTRYPOINT` script during development, but as part of `Dockerfile.cloud-extras` during deployment. This way pip/npm can be just as fast as they are during local (non-Docker) development, while still allowing for those dependencies to be part of the actual Docker image for cloud deployment.

PR #1356, opened 2017-02-10
Add requirements-temp-extras.txt for easy tinkering/WIP PRs

toolness:
In #997 and #1353, to make it easier to iterate on Python dependencies and switch between WIP PRs, I made a `requirements-<something>.txt` file and installed it at the end of my `Dockerfile` and in `.travis.yml`.  This mitigates the Python side of #1230, at least. (Afaik there isn't a similar way to mitigate the node side due to the monolithic nature of `package.json`.)

This PR is an attempt to formalize this pattern by just adding a `requirements-temp-extras.txt` to the repository, which should always be empty (save for its explanatory comments) on the `master`/`develop` branches, but which may contain modules on one's personal checkout or in WIP PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ before_install:
 - node --version
 install:
 - pip install -r requirements-dev.txt
+- pip install -r requirements-temp-extras.txt
 - npm install -g gulp-cli
 - npm install
 script:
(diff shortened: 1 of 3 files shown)


jseppi:
Maybe add it to `.gitignore` so that it won't get committed after this initial commit?

toolness:
Hmm, but we *do* actually want it to be committed to WIP PRs, is the thing... Just not to the `develop` or `master` branchez!

jseppi:
Oh right, hrm. I guess we'll just have to be careful not to commit it back into those branches.

PR #1358, opened 2017-02-10
Fix infinite loop case in autocomplete.js

### LINKED ISSUE #1355: If you enter "{{" into the labor category search, your browser/computer will run out of memory
Seems like it gets stuck in some kind of a loop. This was found during the penetration testing.

Pretty nasty!

jseppi:
Fixes #1355

In the case of `searchStr` comprising only special characters (like `"{{"`) the `while` loop in `appendHighlightedTerm` would never exit because the `sanitizedSearchString` was the empty string, so the regular expression built from it would always return a match.

The code in `autocomplete.js` is still a bit dubious, and I don't love that it is using jQuery. I bet we could could either find or write a better React autocomplete.

### DIFF
--- a/frontend/source/js/data-explorer/autocomplete.js
+++ b/frontend/source/js/data-explorer/autocomplete.js
@@ -7,7 +7,6 @@ export function appendHighlightedTerm($el, term, searchStr) {
     .trim()
     .split(/[ ]+/)
     .join('|');
-  const re = new RegExp(`(${sanitizedSearch})`, 'gi');
 
   const plainText = (start, end) => document.createTextNode(
     term.substring(start, end),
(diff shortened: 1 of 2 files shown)


frontend/source/js/data-explorer/autocomplete.js
```diff
@@ -66,6 +71,7 @@ export function initialize(el, {
       }, (error, result) => {
         autoCompReq = null;
         if (error) { return done([]); }
+        debugger;
```
toolness: I guess this should be removed?

jseppi:
I can bust up a test right quick

jseppi:
boom! also added some tests for the basic working case.

jseppi:
ready for review again, @toolness

PR #1363, opened 2017-02-13
Oops, migrate temp requirements into requirements.txt

toolness:
Oooof, I was a bit too eager to merge #1353 and did so without migrating the requirements out of `requirements-temp-extras.txt`.

This moves them into `requirements.txt`, rather than `requirements-dev.txt`, because right now we're actually serving the developer docs on production... which makes sense I guess since we're also serving the style guide on production too...

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,9 +24,3 @@ mypy-lang==0.4.6
 pytz==2016.10
 Markdown==2.6.8
 cg-django-uaa==0.0.1
-
-# These are currently needed in production because we're serving
-# the docs in production at /docs/.
-Sphinx==1.5.2
-sphinx-rtd-theme==0.1.9
-recommonmark==0.4.0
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -24,3 +24,9 @@ mypy-lang==0.4.6
 pytz==2016.10
 Markdown==2.6.8
 cg-django-uaa==0.0.1
+
+# These are currently needed in production because we're serving
```
jseppi: But the docs are _built_ during CI, which has the `requirements-dev.txt` dependencies, so should they just be in that file?

toolness: whoaaaaaaaaaaaa good point!  Will move to requirements-dev then.

toolness:
Ok just moved them @jseppi, does the comment I put above them make sense?

jseppi:
yeah, 👍

toolness:
Wooo merging!

PR #1369, opened 2017-02-14
Make updating python/node dependencies in Docker much faster and more intuitive

### LINKED ISSUE #1230: Building and changing node/python dependencies in docker is slow

toolness:
This attempts to resolve #1230 by moving the directories containing node and python dependencies out to docker volumes.

This makes updating these dependencies a *lot* more like running a local development setup: just run `pip install -r requirements-dev.txt` and `npm install` in the container and you're done.  If a single dependency changes, it's the only one that needs to be downloaded and installed.  You can even install your favorite debugging libraries this way, and the changes persist because the dependencies are in docker volumes, rather than part of the docker image itself.

It also means that there's a bit of extra setup, though: while devs likely won't have to run `docker-compose build` nearly as often, they *will* have to run `pip install` and `npm install` now.

To facilitate this, I've added an `update.sh` script that runs these, as well as `manage.py migrate` and `manage.py initgroups` all at once.  Folks developing locally (without docker) can run this, while those with Docker can run `./docker-update.sh` to get the same thing done.

I am pretty sure this should work fine with cloud-based Docker deploys, but I haven't actually tried doing it yet.

## Notes

* Due to the fact that Python doesn't encapsulate all of a project's requirements into a single directory like node does, I've made the docker setup use a virtualenv, which is the closest approximation to that. Basically, `docker_django_management.py` now detects whether the virtualenv exists--which it won't at first, because docker volumes start out empty--and if it doesn't, it creates it.  Then it transparently activates the virtualenv before running whatever command needs to be run.

* The fact that `./update.sh` runs `manage.py migrate` means that things might explode if e.g. the developer is switching between two different branches that both make migrations or something. Obviously this has always been a problem, but it might be worth it to offer some kind of advice to devs if migration fails, e.g. telling them how to easily blow away their db and start from scratch, if that's the easiest solution.

* Um, our `setup.md` could use some help. Rather than fix it now though, imma file a separate bug.

## To do

- [x] Factor out all the new CALC-specific stuff from `docker_django_management.py` somehow, so it stays nice and reusable.
- [x] We could potentially add an `update-docker.sh` script in the root dir so that docker devs can just run it instead of the more verbose `docker-compose run app ./update.sh`.  It would also allow us to run `docker-compose build` in the update script, just in case the `Dockerfile` changed.
- [x] Change `docs/docker.md` to document this new setup.


toolness:
@jseppi do you have thoughts on this approach?

update.sh
```diff
@@ -0,0 +1,15 @@
...
+pip install -r requirements-dev.txt
+
+echo "----- Updating Node Dependencies -----"
+npm install
+
+echo "----- Migrating Database -----"
+python manage.py migrate --noinput
+
+echo "----- Initializing Groups -----"
+python manage.py initgroups
```
toolness: This file obviously has some similarities with `cf.sh`... Not sure if it's worth it to refactor in such a way that code is reused or not.  Meh.

jseppi:
Approach seems solid to me. I haven't tried actually running yet since I'm on a flight and don't want to stress my in-flight wifi getting containers.

toolness:
Oh man you GOTTA get containers on a plane!  Push it to the limit man!!!!!!

jseppi:
I've had it with this containers on this plane!

toolness:
Ok @jseppi I think this is ready for review!!

docs/docker.md
```diff
@@ -57,9 +55,17 @@ if it detects that Django isn't installed.
 
 ### Updating the containers
 
-All the project's dependencies, such as those mentioned in `requirements.txt`,
-are contained in Docker container images.  Whenever these dependencies change,
-you'll want to re-run `docker-compose build` to rebuild the containers.
+Whenever you update your repository via e.g. `git pull` or
+`git checkout`, you should update your containers by running
+`./docker-update.sh`.
```
jseppi: This file wasn't executable upon checkout of this branch. Is there some git magic to check it in as executable? Maybe this: http://www.pixelninja.me/make-script-committed-to-git-executable/

jseppi: Oh, after I did `chmod +x docker-update.sh`, git is showing a file difference that can be committed, so that looks easy to fix.

toolness: Ah yeah, sorry about that, I think I just forgot to `chmod +x` it before committing!

PR #1373, opened 2017-02-14
Clear gleaned_data from session on schedule selection change

### LINKED ISSUE #1288: In PL upload, going back to step 1 and choosing a different schedule does nothing
Hmm, I chose fake schedule and got to step 4 with it, but then I went back to step 1 and chose a different schedule, and when I got to step 3 it still thought I was going to upload a fake schedule PL. Weird.

Obviously not a big deal right now since we only support one schedule in production.

jseppi:
Fixes #1288

This PR adds logic to remove `gleaned_data` from the price list upload session if a user goes back to Step 1 and changes the selected schedule to something different from the original value. I opted not to completely clear out session in this case (ie, not removing the Step 2 Form data from session), since it seems the user might want their other data to stay the same.


data_capture/tests/test_price_list_upload_views.py
```diff
@@ -98,6 +99,36 @@ def test_valid_post_does_not_overwrite_other_step_data_if_exists(self):
         self.assertEqual(session['data_capture:price_list']['step_2_POST'],
                          Step2Tests.valid_form)
 
+    @override_settings(
+        DATA_CAPTURE_SCHEDULES=[
+            FAKE_SCHEDULE,
+            'data_capture.schedules.s70.Schedule70PriceList'],
+    )
+    def test_valid_post_with_diff_schedule_removes_gleaned_data(self):
```
jseppi: This test is somewhat gross since I had to include another registered schedule in order to have another valid schedule to "select" in Step 1.

data_capture/tests/test_price_list_upload_views.py
```diff
@@ -98,6 +99,36 @@ def test_valid_post_does_not_overwrite_other_step_data_if_exists(self):
...
+        # Save gleaned data into session
+        self.set_fake_gleaned_data(rows=[{
+            'education': 'Bachelors',
+            'price': '15.00',
+            'service': 'Project Manager',
+            'sin': '132-40',
+            'years_experience': '7'
+        }])
+        session = self.client.session
+        self.assertIn('gleaned_data', session['data_capture:price_list'])
```
jseppi: sanity check

data_capture/views/price_list_upload.py
```diff
@@ -30,6 +30,9 @@
 )
 
 
+SESSION_KEY = 'data_capture:price_list'
```
jseppi: Moved this string to a variable for readability/decreased line-length in the rest of this file.

data_capture/views/price_list_upload.py
```diff
@@ -61,16 +64,27 @@ def step_1(request, step):
...
+            if SESSION_KEY in sess:
+                orig_step_1 = sess[SESSION_KEY].get(
+                    'step_1_POST')
+                if orig_step_1 and (orig_step_1['schedule']
+                                    is not request.POST['schedule']):
+                    # If a different schedule has been chosen from what was
+                    # already in session, delete 'gleaned_data' from sssion so
+                    # that the upload step will require an upload of data
+                    # from the newly selected schedule.
+                    del sess[SESSION_KEY]['gleaned_data']
```
toolness: Er, I think this might raise a `KeyError` if `gleaned_data` isn't in the dict... Is that a possibility here?

data_capture/views/price_list_upload.py
```diff
@@ -61,16 +64,27 @@ def step_1(request, step):
...
         form = forms.Step1Form(request.POST)
         if form.is_valid():
             sess = request.session
-            if 'data_capture:price_list' in sess:
-                sess['data_capture:price_list']['step_1_POST'] = request.POST
+            if SESSION_KEY in sess:
+                orig_step_1 = sess[SESSION_KEY].get(
+                    'step_1_POST')
+                if orig_step_1 and (orig_step_1['schedule']
+                                    is not request.POST['schedule']):
```
toolness: While it'd be more code, I wonder if it might be better to compare the user's chosen schedule against the schedule of the gleaned data--it's quite a silly edge case, but it'd mean that if someone  e.g. chose Schedule 70, uploaded a r10 price list, got the warning that "you chose s70 but uploaded r10", then went back to step 1 and chose r10, they wouldn't have to re-upload the same r10 price list. Again, probably unlikely to actually happen though, so it's up to you if you want to implement it.

data_capture/views/price_list_upload.py
jseppi: oh deer python

jseppi:
Alrighty, implemented your suggestion and fixed the possible `KeyError` bug.

PR #1374, opened 2017-02-14
Load Data Explorer React app in <head>

### LINKED ISSUE #1341: Visual bug: headers appear to jump as page loads (in chrome, eg)
For the backlog …

jumpyjump

Discussed in brief [on slack](https://gsa-tts.slack.com/archives/calc/p1486496535000314).

jseppi:
Fixes #1341 by moving the Data Explorer scripts back into the document `<head>`. While this will cause (hopefully unnoticeably) more time before the page is rendered, it should get rid of weird jumps that happen due to the React app even starting to load after other DOM elements.

### DIFF
--- a/data_explorer/templates/index.html
+++ b/data_explorer/templates/index.html
@@ -39,11 +39,11 @@
 
 {% block body %}
 <div data-embed-jsx-app-here class="card--half-padding">
-  <noscript>
-    <div class="card">
+  <div class="card">
+    <noscript>
       Please enable JavaScript to use this page's functionality.
-    </div>
-  </noscript>
+    </noscript>
+  </div>
 </div>
 
 <![if gt IE 8]>
--- a/data_explorer/templates/index.html
+++ b/data_explorer/templates/index.html
@@ -45,11 +45,12 @@
     </noscript>
   </div>
 </div>
-
-<![if gt IE 8]>
-{# For the list of files included in this built script, see gulpfile.js #}
-<script src= "{% static 'frontend/built/js/data-explorer/vendor.min.js' %}"></script>
-<script src= "{% static 'frontend/built/js/data-explorer/index.min.js' %}"></script>
-<![endif]>
-
 {% endblock %}
+
+{% block before_body_close %}
+  <![if gt IE 8]>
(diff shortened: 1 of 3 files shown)


ericronne:
It's better than before, but still jumps :/
It becomes more obvious when i make the window wider (on thunderbolt), because it takes a split second longer to load...

image

jseppi:
Alright want to give it another try, @ericronne?

Also, since this is more substantial now, I'm adding @toolness to review.

ericronne:
Still seeing funkiness, sry :(

jump

jseppi:
haha :)

Is that after grabbing the most commit (988cf17) to this branch? That looks to me how it was performing in the previous one (12b4643). I'm not able to recreate what you're showing in that gif, even with network throttling turned on.

jseppi:
oof, I just saw that behavior now, only happens every so often when I refresh.

I'm not sure if there is a solution here. The React app is always going to take over after some (or now, all) of the non-React DOM has loaded. I guess we could hide the rest of the page until the React app is finished, though I think that's kind of a hack. Or maybe we could show a spinner while the React app loads.

Probably the "best" way to to it is to stuff the entire page into the React app, but that has other drawbacks.

This is at least better than the nav items jumping around as reported in #1341, right?

ericronne:
@seppi i tweaked the header css file, spec’ing a height for the logo. The top stuff should stay put now. But a flash of the footer still occurs sometimes. Not a deal breaker, but if we could push down the footer or delay its load, that might do the trick. NBD tho

toolness:
Oh the bounciness could also be solved by implementing our glorious PROGRESSIVE ENHANCEMENT strategy here!!!!!  i.e., rendering the explorer view on a server and including that in the original response.

@jseppi is this ready for review or are you still working on it?

jseppi:
This is ready for review.

Yeah, an isomorphic app would be probably the most ideal thing we could do, but that is quite a large lift. I was mostly trying to get rid of the nasty nav bar bouncing shown in the gif over at #1341.

toolness:
A large but FUN lift!!!!!!!!!!!!

PR #1376, opened 2017-02-15
specify a max length for labor category query

### LINKED ISSUE #1354: Consider putting maxlength on labor category field
Otherwise the page can get flooded with a bunch of junk text. We should probably limit to something somewhat reasonable, like 128 characters.

screen shot 2017-02-10 at 10 06 10 am

### LINKED ISSUE #1372: Content Spoofing via Text Injection
From the penetration testing conducted as part of the ATO process, reported here: https://docs.google.com/document/d/1C6ZddThph2ipC2_Obni5OulVmED-joGfBDJMPCJHniM/edit#heading=h.2jxsxqh 

Content spoofing is a type of an attack that allows malicious user to send faked or modified content to the application’s user as if it was legitimate.  

During testing it was discovered that application does not properly handle user input. Malicious user can supply content to web application via text manipulation in URL parameters. This can be used to present false information to the user and for website defacement. 

Affected Devices / URLs
https://calc-staging.app.cloud.gov/?q=ANYTHINGandANYLENGTH

Solution
User input has to be properly sanitized. If application has to accept only certain content, whitelisting can be used. 

Risk Factor
This attack can lead to data spoofing and information misrepresentation. This can severely impact user’s trust. Since it is a low risk system and malicious user can only manipulate textual data, this finding classified as moderate. 

Other References
https://www.owasp.org/index.php/Content_Spoofing
https://cwe.mitre.org/data/definitions/345.html

jseppi:
Closes #1354 
Closes #1372 

This PR specifies a `MAX_QUERY_LENGTH` set to `255` (to match our back-end database column `labor_category` max length). The `q` reducer truncates the query length to that value.

I also added an `overflow: hidden;` to the `description` container to prevent this text overflow situation when there are no spaces (which is an unlikely valid query in any case):

screen shot 2017-02-15 at 11 29 38 am

to

screen shot 2017-02-15 at 11 29 31 am


jseppi:
[Malicious users should not be able to flood the front page with a page-lengthening query parameter because it looks real bad.](https://trello.com/c/bflCyrm4/234-malicious-users-should-not-be-able-to-flood-the-front-page-with-a-page-lengthening-query-parameter-because-it-looks-real-bad)

jseppi:
@hollyallen - this should clear up item 2.2.1 from the CALC ESAP report.

frontend/source/js/data-explorer/serialization.js
```diff
@@ -70,7 +71,7 @@ const stringInSet = (choices, defaultVal = '') => (val) => {
 export const serializers = {
   exclude: list => list.map(coercedString).join(','),
   education: list => list.map(coercedString).join(','),
-  q: coercedString,
+  q: s => coercedString(s).slice(0, MAX_QUERY_LENGTH),
```
toolness: Hmm, I feel like this shouldn't be doing the `slice`, because serialization is just about going from a known-good state to a string that can be plopped in a querystring.  That is, CALC's Redux store should be working in such a way that really long queries are never inside it in the first place--so it might be better to modify `reducers.js`'s processing of the `SET_QUERY` action such that it slices it there instead?

Hmm, actually, I guess this is actually a funky kind of problem.  Is `MAX_QUERY_LENGTH` there for security, or for general usability?  If it's there for security, then I guess it could go into the `serializers`, so that we could permit user  sessions to have queries longer than `MAX_QUERY_LENGTH`, but simply not allow them to be *shared* (if we do this we should also remove the `maxLength` attribute from the generated markup though).

On the other hand, if it's a general usability thing, then I think it might be more appropriate to put it in the reducer... Hmm. What do you think?

frontend/source/js/data-explorer/components/labor-category.jsx
```diff
@@ -61,6 +63,7 @@ class LaborCategory extends React.Component {
           value={this.state.value}
           onChange={this.handleChange}
           onKeyDown={handleEnter(this.handleEnter)}
+          maxLength={MAX_QUERY_LENGTH}
```
toolness: Coincidentally, if we move the `slice()` into the reducer, I don't think we need this, though it'd still be useful for a progressively-enhanced version of this page where JS isn't running (which doesn't exist yet, of course, bu WE HAVE PLANS).

jseppi: I would still think it is good to have the attribute here since it means you can't even enter/paste a string greater than the max, which I think would make it less confusing to a user that a query is getting chopped off somewhere else.

But no actual user is every going to want to enter a 255-character search.

toolness: Oh my bad, I forgot that we're actually storing the current value in `LaborCategory`'s `state`... I thought that the value was always going through the reducer data flow on a per-keypress basis before coming back to the text field, in which case it'd have the same ultimate effect as `maxLength`.

frontend/source/js/data-explorer/serialization.js
jseppi: Heh, I guess it's mostly for _satisfying_ the "security" finding, though that finding is more about usability than it is about security. I would say that it is more of a usability issue.

Maybe the reducer should _also_ enforce the max length. ?

toolness: Hmm, I'm not sure I like the redundancy in logic... One of the things I like about redux is the whole "single source of truth" aspect of it, and putting what amounts to validation logic in that `serializers` object--rather than just sticking to "take this valid value and turn it into a string suitable for inclusion in a URL"--seems counter to that philosophy.  I feel like that could start leading to a slippery slope where we're putting all kinds of validation logic in multiple places for fear that some code paths might not be hitting validation logic. 😣

jseppi: That's a very good point.

frontend/source/js/data-explorer/reducers.js
```diff
@@ -47,11 +48,12 @@ function contractYear(state = DEFAULT_CONTRACT_YEAR, action) {
   return state;
 }
 
-function q(state = '', action) {
+export function q(state = '', action) {
+  const cleanedState = state.slice(0, MAX_QUERY_LENGTH);
```
jseppi: like this, @toolness?

jseppi:
Ready for review again, @toolness

frontend/source/js/data-explorer/serialization.js
```diff
@@ -93,7 +92,7 @@ export const deserializers = {
     coercedString(list)
       .split(',')
       .filter(x => x in EDU_LABELS),
-  q: s => coercedString(s).slice(0, MAX_QUERY_LENGTH),
+  q: coercedString,
```
toolness: Oh, um, so the sort of weird thing is that `deserializers` actually *should* do sanitization, because the way that `deserializers` is used is that `history.js` calls them and passes their result to a `SET_STATE` action, which *doesn't* do any sanitization on its own.  This might be a bad idea, though--in the long-term, it might be better to either make `SET_STATE` do the sanitization itself, or have a separate `SET_STATE_FROM_QUERYDICT` that does the sanitization, or something.

For now though I think it might be easiest to leave the slicing *in* the `deserializers` dict but *out* of the `serializers`, as that's the way things are currently written.

jseppi:
Alright dudebert, check it again plz

frontend/source/js/data-explorer/reducers.js
toolness: Oh, actually you want to be doing the slicing on `action.query`, not the existing state--the existing state is already assumed to be clean and legit.  Um, that's how I understand it at least--reducers shouldn't be doing sanitization on every single iteration of the app's state, but rather they should assume a gud state and do any validation or whatever when the state changes based on an action.

jseppi: hrm, I tried that and it was...behaving quite strangely. Will try again.

jseppi: (also that was my understanding of how it should work as well, so I first tried that, but was getting really weird behavior so I thought my understanding was broken)

toolness:
Oh weird, I didn't realize that I submitted that deserializer comment as a separate review.

So um this is looking good now except for my comment about the `q()` function!

frontend/source/js/data-explorer/reducers.js
toolness: Oh hrm.  Well, now that I think about it, I was suggesting that under my incorrect assumption that a `SET_QUERY` action was busted on a per-keypress basis, rather than whenever the query is actually executed... Blarg.  I guess this validation logic will still be valuable in such a case (especially if e.g. the browser has html5 form validation turned off or something weird), but it likely won't get triggered very often, if at all.

Uggh it has been a long time since I've looked at this code, sorry I am being confusing.

jseppi:
I think I have it all good now @toolness.

jseppi:
ugh, I think the Travis pr build is failing because `develop` is now on Python 3.6, but this branch is on `3.5` so flake8 is barfing with `undefined name 'ModuleNotFoundError'`

toolness:
Ah no probs just merge `develop` back into this PR's branch amirite?!

toolness:
Nice, this looks good!  I guess since you're certain that the travis pr job is failing because of the Python lameness, feel free to just merge this into `develop` now.

Issue #1345, opened 2017-02-08
Pin all node dependencies


toolness:
Given #1343 and other problems like it that I've run into in the past, I recommend pinning our node dependencies in our `package.json`.  This will improve dev/prod parity, as well as making the development environment more consistent across our own laptops.

I suspect there is a node-based solution similar to the python-focused `requirements.in` recommended by #514.  If we do something like this, we can make regular PRs that just update the minor/patch version numbers of our node dependencies. I think this will make it a lot easier to track down regressions that are due to package upgrades.

Since node's dependency space is hierarchical rather than flat, it might even behoove us to use `npm shrinkwrap` and commit its results to the repository, but I'm not sure. It looks like this [Keeping Your npm Dependencies Immutable](https://ponyfoo.com/articles/immutable-npm-dependencies) blog post outlines some best practices (well, practices that the author thinks are best, at least).

What do you think @jseppi?

jseppi:
Yeah, I'm 👍 for pinning, and would be ok with using `shrinkwrap`.

Issue #209, opened 2015-07-01
Number of matches showing in autocomplete doesn't match search setting


theresaanna:
If you go to the site and search for "laborer", the autocomplete shows 17 results regardless of what search setting you have on, though 17 represents the exact matches for the term, and by default we have on "contains words".

toolness:
This bug still seems to be valid, and I'm pretty sure it's back-end; the front-end is sending a `query_type` parameter specifying the type of query being used, but the number of matches sent back are always the same number regardless of its setting.  I tried this with "senior engineer", where autocomplete indicates there to be 116 matches while I have "contains words" selected, when in fact there are 941 results for "contains words" (there are 116 for "exact match").

toolness:
I think actually fixing this to show the correct number of results might be hard.

If it is indeed hard, one way to fix this problem might simply be to not show the count, but still order by it.

hbillings:
We're going to circle back around to this when we start redoing the data explorer.

ericronne:
Bump! Here's a sample with visual overview …

![image](https://user-images.githubusercontent.com/9259185/42650816-0fec99d6-85d3-11e8-96b8-8e6a9fe601ab.png)

"Contains phrase" returns 429 results, btw. ¯\_(ツ)_/¯

PR #1378, opened 2017-02-15
Support synonyms/abbrs in labor category search (normalization approach)

### LINKED ISSUE #1375: Add thesaurus/synonyms for labor category terms
Part of Kelly's feedback on #997 is that it'd be nice if the price list analysis could take common labor category abbreviations into account, e.g.:

* "Jr." and "Junior" are the same thing
* "Sr." and "Senior" are the same thing
* "SME" and "Subject Matter Expert" are the same thing

and so on.

Obviously this will also be useful for standard CALC search, not just the price list analysis functionality.

The only problem is that figuring out how best to implement it is non-trivial.

Postgres actually has a [ridiculously customizeable Dictionary system](https://www.postgresql.org/docs/9.5/static/textsearch-dictionaries.html) but the problem is that customizations require changing the filesystem of the postgres server, which [AWS RDS does not support](https://forums.aws.amazon.com/message.jspa?messageID=626892).  So we might have to hack this.

Right now I am thinking one approach may be to add a `normalized_labor_category` field to the `Contract` model which will basically be the labor category post-processed by some normalization code that takes our custom terminology into account. We can then have our `search_index` be based off this normalized field instead of the raw `labor_category`.

toolness:
This is an attempt to fix #1375 by adding a `_normalized_labor_category` field to the `Contract` model and building a search index on that instead of `labor_category`.  This new "private" field takes into account the synonyms `jr`, `sr`, and `sme`, though in the future it can be enhanced to support more labor category jargon.

I've attempted to find all points where contracts models are saved, and changed them to update the `_normalized_labor_category` as needed.  I also modified the `update_search_fields` management command (by "shadowing" it in the `contracts` app) so that it first updates this new field before building the search index. This way, if we ever decide to e.g. add a new synonym, we can just run `manage.py update_search_fields contracts` to have our new dictionary logic take effect on searches.

## Notes

* ~~Right now the autocomplete search results look a bit weird, because they include results for both the "literal" and "normalized" versions of the query.  For example, if you search for "junior engineer", the autocomplete results will show entries for *both* "junior engineer" *and* "jr. engineer". What's more is that, ~~possibly~~ almost definitely due to #209, these equivalent autocomplete entries *show a different number of results*, even though selecting either shows an identical result set! Wut is up with our autocomplete code.~~

  As of 4ace007bcbac9065536b7933b23e4a18c330fc3f, we're showing the normalized labor categories in the autocomplete suggestions. This results in a bit of munging, since the normalized version e.g. lowercases everything, but it also results in more sensible autocomplete suggestions too, as e.g. "jr" is not distinct from "junior".

* If there are too many flaws/risks inherent in this approach, I can just merge it into `price-list-analysis` (#997) instead of `develop` for now, since that was the main driver for this functionality. I just figured it might be useful for production too, so I wrote it against that.

* If we decide to merge this PR, we'll need to be sure to run `manage.py update_search_fields contracts` on dev/staging/prod.  We could just do this as part of `cf.sh` so that we don't need to do it manually, though it'll obviously add a bit of extra time to deploys (albeit not a lot, given the current size of our database).

* I was worried about the speed of normalization when the algorithm changes, but I've optimized things so that it's pretty fast.  It *might* not necessarily scale to millions of rows, but since we've only got 57k right now, I don't think it's that big a deal.

## To do

- [x] Decide how we want to deal with re-indexing on dev/staging/prod (see above), if we indeed decide to merge this PR into `develop`.


contracts/models.py
```diff
@@ -74,6 +74,15 @@ class CurrentContractManager(SearchManager):
...
     # search instead of default
 
+    def update_normalized_labor_categories(self):
+        for contract in self.all():
+            contract.update_normalized_labor_category(save_if_changed=True)
+
+    def bulk_create(self, contracts, *args, **kwargs):
+        for contract in contracts:
+            contract.update_normalized_labor_category()
+        return super().bulk_create(contracts, *args, **kwargs)
```
toolness: Note that this isn't *technically* required since all our uses of `bulk_create()` are followed-up by calls to `manage.py update_search_field`, both of which update the normalized labor category field.

However, I think that this will be faster, since modifying `bulk_create` ensures that the very first write of the model to the DB already has the correct normalized LC value.  The code in `manage.py update_search_field` is written to only write the new normalized LC value to the DB if it's changed, which it won't have if `bulk_create` was just called.

If that makes sense. Oy this is a bit confusing but since we're blocked from modifying postgres' db config, I feel like this might be our only option, aside from moving everything to postgres triggers or somesuch.

toolness:
Wait... um... it might actually be possible to do this just by hacking our `convert_to_tsquery` function a bit.  It'd require me to make some changes to the price list analysis vocabulary code in #997 but it could be a nice way to do this without requiring re-indexing whenever our normalization logic changes.

toolness:
Ok, I've merged this into #997 but I'd like to see if it's worth also merging into develop--in part because it's functionality that'd be useful to data explorer users, but also because the less #997 differs from `develop`, the easier it'll be to keep it up-to-date with that branch.

toolness:

contracts/management/commands/update_search_field.py
```diff
@@ -0,0 +1,13 @@
+from djorm_pgfulltext.management.commands.update_search_field \
+    import Command as UpdateSearchFieldCommand
+
+from contracts.models import Contract
+
+
+class Command(UpdateSearchFieldCommand):
+    def handle(self, *args, **kwargs):
```
jseppi: Maybe a little docstring here that this is shadowing the djorm_pgfulltext command in order to update the normalized term?

contracts/models.py
```diff
@@ -74,6 +74,44 @@ class CurrentContractManager(SearchManager):
     # need to subclass the SearchManager we were using for postgres full text
     # search instead of default
 
+    def bulk_update_normalized_labor_categories(self):
+        '''
+        Iterate through all Contract models and update their
+        normalized labor categories.
```
jseppi: maybe add "if necessary" to the end of the sentence

contracts/tests/test_contract.py
```diff
@@ -1,13 +1,34 @@
...
+        self.assertEqual(_normalize('jr. person'), 'junior person')
+
+    def test_abbr_without_period(self):
+        self.assertEqual(_normalize('jr person'), 'junior person')
+
+    def test_abbr_at_end(self):
+        self.assertEqual(_normalize('person jr.'), 'person junior')
+
+    def test_abbr_in_middle(self):
+        self.assertEqual(_normalize('person jr. ham'), 'person junior ham')
```
jseppi: lol 🐽

jseppi:
And also broken test case for some reason

toolness:
I am suuuuuuuuuuper confused by this broken test case... It's on `api.test.js`:

```
FAIL  frontend/source/js/data-explorer/tests/api.test.js (5.194s)
  ● API get › callsback with a string on network error
    Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL.
      
      at Timeout.callback [as _onTimeout] (node_modules/jsdom/lib/jsdom/browser/Window.js:480:19)
      at tryOnTimeout (timers.js:232:11)
      at Timer.listOnTimeout (timers.js:202:5)
```

This is weird since those API tests are just testing the front-end JS code, which wasn't modified in this PR.

toolness:
So we really need to fix #1345 because `xhr` got updated and provides a different kind of error string/object than it used to. 😞 

`mock-xmlhttprequest` was hiding the error for some reason so I substituted it out for [`sinon`'s fake XHR](http://sinonjs.org/releases/v1.17.7/fake-xhr-and-server/) (we were already using `sinon` for other things, so this resulted in one less dependency total).  That revealed the true culprit.

toolness:
WOOT IT WORKS

can u code review my weird unrelated travis fixes plz @jseppi

package.json
```diff
@@ -49,7 +49,7 @@
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.1.0",
     "watchify": "^3.7.0",
-    "xhr": "^2.3.3"
+    "xhr": "^2.4.0"
```
jseppi: should we lock this down to just `2.4.0` without the `^`?

frontend/source/js/data-explorer/tests/api.test.js
```diff
@@ -27,27 +27,23 @@ describe('API constructor', () => {
 });
 
 describe('API get', () => {
-  xhr.XMLHttpRequest = MockXMLHttpRequest;
+  xhr.XMLHttpRequest = sinon.FakeXMLHttpRequest;
```
jseppi: ah, cool, didn't realize sinon had that included

frontend/source/js/data-explorer/tests/api.test.js
```diff
@@ -56,6 +52,8 @@ describe('API get', () => {
       expect(res).toMatchObject({ result: 'success' });
       done();
     });
+    expect(req.url).toEqual('/whatever');
+    req.respond(200, resHeaders, JSON.stringify({ result: 'success' }));
```
jseppi: cool, these tests are much easier to read now 👍

jseppi:
Yeah, these changes look good. Sorry I didn't know about `sinon.FakeXMLHttpRequest` previously!

package.json
toolness: Yeah, I wasn't really sure--it seems like we should take care of all those `^`'s in one fell swoop, either via using `yarn` and whatever its equivalent of `shrinkwrap` is, or by changing this `package.json`.

toolness:
Woot no probs!  Yeah we should switch to `yarn` though I guess...

PR #1384, opened 2017-02-15
Improve setup.md

toolness:
This attempts to improve our `setup.md`, which is currently kind of vague and confusing.

[You can read the new version here.](https://github.com/18F/calc/blob/improve-setup-md/docs/setup.md)

This PR also renames a few sections to use [sentence case](https://medium.com/@jsaito/making-a-case-for-letter-case-19d09f653c98) because that's what the rest of our docs use and it's awesome.

To do:

- [x] I'd like to add some tests verifying that the Python and Node versions mentioned in `setup.md` are the same as our project actually uses, but I can do that in a separate PR if needed.

### DIFF
--- a/hourglass/tests/test_configuration.py
+++ b/hourglass/tests/test_configuration.py
@@ -1,5 +1,6 @@
 import os
 from unittest import TestCase
+import json
 import yaml
 
 
@@ -34,7 +35,8 @@ class Version:
 
 class PythonVersionTests(TestCase):
     '''
-    Ensure all our configuration files specify the same Python version.
+    Ensure all our configuration files, documentation, etc. specify the
+    same Python version.
     '''
 
     version = Version('3.6.0')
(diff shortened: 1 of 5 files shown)


requirements-dev.txt
```diff
@@ -3,6 +3,7 @@
 
 # Testing and development dependencies
 bandit>=1.0,<2.0
+pyflakes==1.4.0
```
toolness: Blerg, this is a dependency of `flake8`, but flake8 implicitly gives us 1.3.0, which doesn't have https://github.com/PyCQA/pyflakes/pull/80, which fixes an incompatibility with parsing files that have f-strings in them.  So we're explicitly requiring the newer version here.

hourglass/tests/test_configuration.py
```diff
@@ -53,3 +55,26 @@ def test_travis_yml(self):
         with open(path('.travis.yml')) as f:
             data = yaml.safe_load(f)
             self.assertEqual(data['python'], [float(self.version)])
+
+    def test_docs_setup_md(self):
+        with open(path('docs', 'setup.md')) as f:
+            self.assertIn(f'Python {self.version}', f.read())
```
toolness: [AW YISS F STRINGS FTW](https://cito.github.io/blog/f-strings/) 🎉 💥 💃

requirements-dev.txt
toolness: Just filed #1394 to deal with this separately, since I don't know when this PR will get merged.

PR #1385, opened 2017-02-16
Remove and replace hourglass.js

jseppi:
`hourglass.js` had a lot of code that isn't being used any more, wasn't up to our eslint standard, and contained things that were hard to test with `jest` because they relied on jQuery (`$`) being globally present.

This PR removes all of `hourglass.js`. All functionality that was still in use is now in `api.js`. The new `API` class simplifies the old methods, and instead of using jQuery's `ajax` method, uses the tiny [`xhr` module](https://www.npmjs.com/package/xhr).


toolness:
Nice!  I had tried getting rid of a bunch of it back in #1075, but it seemed like the API object ended up using a lot more of that module's code than I'd expected, or something.

jseppi:
Working on refactoring that code a bit. It gets used a lot, but the part that is used is pretty simple.

jseppi:
[Clean up legacy hourglass.js](https://trello.com/c/2yHuv6xg/236-clean-up-legacy-hourglass-js)

frontend/source/js/data-explorer/api.js
```diff
@@ -0,0 +1,30 @@
...
+    let path = this.basePath + uri;
+    if (data) {
+      path += `?${qs.stringify(data)}`;
+    }
+    return xhr(path, {
+      json: true,
+    }, (err, res) => {
+      if (err) {
+        callback(err.toString());
+      } else if (res.statusCode >= 400) {
```
jseppi: In `xhr` (and several other ajax libs), `err` is only set when a _network_ error occurs, not when the response is an "error response," which is different from how jQuery's ajax methods work. To keep the same that was previous in `hourglass.API`, this condition checks for an error response and returns the original XMLHTTPRequest's `statusText`.

jseppi:
Ready for review, @toolness

frontend/source/js/common/hourglass.js
```diff
@@ -1,165 +0,0 @@
...
-     * });
-     *
-     * api.get("rates/", function(error, data) {
-     *   // do something with data
-     * });
-     */
-    get: function(request, callback) {
-      var url = this.url(request),
-          data = hourglass.extend({
-            format: "json"
```
jseppi: The 'format' param is something DRF uses to determine which response format to use. However, DRF will also respect request headers, which `xhr` sets to `application/json` when its [`json` param is `true`](https://www.npmjs.com/package/xhr#optionsjson), so we don't need to use this param any more.

jseppi:
Ready for review, @toolness

frontend/source/js/common/hourglass.js
toolness: BUT WILL CLOUDFRONT PASS THE HEADERS THROUGH?!!!!!!!!!!!!!!

ugh, unfortunately that is a serious question and not a joke.

frontend/source/js/data-explorer/api.js
```diff
@@ -0,0 +1,30 @@
+/* global window */
+import xhr from 'xhr';
+import qs from 'qs';
```
toolness: huh, so do we need the extra functionality [`qs`](https://www.npmjs.com/package/qs) provides over node's built-in [`querystring`](https://nodejs.org/api/querystring.html) module?  Just curious, I don't mind the extra dependency.

frontend/source/js/data-explorer/api.js
```diff
@@ -0,0 +1,30 @@
+/* global window */
+import xhr from 'xhr';
```
toolness: Ooo, I like that [`xhr`](https://www.npmjs.com/package/xhr) provides the same API as `request`!  Will make our eventual dream of progressive enhancement easier.

frontend/source/js/common/hourglass.js
jseppi: Hrm, I don't think this request will go through CloudFront. It'll go to api.data.gov, which proxies back to calc-prod.app.cloud.gov.

frontend/source/js/data-explorer/api.js
jseppi: Oh, probably not. Didn't realize there was a built-in.

frontend/source/js/data-explorer/api.js
toolness: Cool... Verifying error handling didn't regress was interesting, since the legacy behavior of the data explorer is to not show anything, but at least we do tell something to screen readers:

> aria-lol

(This was triggered by temporarily changing the route in `api/urls.py` so the Ajax request 404'd.)

👍

jseppi: Yeah, it took me a little while to realize what I had to do to make it behave the same way.

production_tests/tests.py
```diff
@@ -95,3 +95,18 @@ def test_api_supports_cors(self):
         self.assertEqual(res.status_code, 200)
         self.assertEqual(res.json(), [])
         self.assertEqual(res.headers['Access-Control-Allow-Origin'], '*')
+
+    def test_api_passes_json_accept_header(self):
```
jseppi: @toolness: like this?

toolness: Noice! Um, alternatively I guess we could just explicitly make it part of the previous test, which seems to be the basis for this one...  That's interesting that the API seems to be returning a JSON response even if we don't include the header, according to that previous test.

toolness: I suppose another way to test it out is to do *two* requests, one of which passes the JSON accept header and another which passes a `text/html` accept header, and ensures that the appropriate types of responses come back through?

jseppi: The previous test explicitly has the `format=json` query param, while this new one does not.

toolness: Ohhhh got it, I didn't notice that. NOICE.

Um would it be too much trubs to factor out the common kode into a separate function or something.

toolness: I'm thinking a `get_api_url()` function that caches its response in a global variable so multiple requests aren't needed or something i dunnnooooooooooooo

whatever i can just add it in a separate PR if it's too annoying.

toolness:
woo looks good! MERGE IT ON UP

PR #1391, opened 2017-02-16
Serve docs from /docs/ instead of /static/docs/

toolness:
This is really yak shaving, but it was bugging me that `docs/` was just a redirect to `static/docs/`, so I wrote a little [WSGI](https://www.python.org/dev/peps/pep-0333/) middleware that rewrites `docs/` internally to `static/docs`.  This "fools" the underlying Django app into thinking that all requests coming in to `docs/` are actually coming in to `static/docs/`.

This is extremely low-priority, though, so feel free to just reject it and tell me to focus on more important things. 😛 

If we do decide to merge it, though, we should add tests.

### DIFF
--- a/hourglass/tests/test_wsgi_middleware.py
+++ b/hourglass/tests/test_wsgi_middleware.py
@@ -1,0 +1,37 @@
+from unittest import TestCase
+
+from ..wsgi_middleware import static_url_rewriter
+
+
+class StaticUrlRewriterTests(TestCase):
+    def app(self, environ, start_response):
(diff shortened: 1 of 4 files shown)


jseppi:
Uh, I guess this fine. I don't have a strong feeling either way, as long as it won't cause anything to break or be weird.

toolness:
One kind of annoying thing is that your browser might be caching the redirect from `/docs/` to `/static/docs/` so you might have to clear your browser's cache or use a different browser to make sure this PR works.

PR #1392, opened 2017-02-16
Support synonyms/abbrs in labor category search (tsquery approach)

toolness:
This is an alternative fix to #1375 that adds support for synonyms/abbreviations by modifying postgres `tsquery` queries.

Advantages over #1378 (the normalization approach):

* Because we don't need to store the "normalized" version of every labor category in the DB, there's no need to run `update_search_fields` every time a synonym is added or the synonym logic changes.

  A particular advantage of this is that we could potentially move the synonym list into the database and allow Data Administrators to edit it, which would allow them to iterate and experiment with synonyms without having to go through the development team.

Disadvantages over #1378:

* In its current implementation, multi-word synonyms aren't supported. This is why the "sme" -> "subject matter expert" synonym isn't a part of this PR.

* It's going to be a bit harder to implement the labor category broadening in #997 on top of this, because of the fact that the underlying search index isn't normalized. Hopefully it shouldn't be *too* big a deal but I'm not certain.

There *might* also be a third approach which involves combining these two approaches, but I'm not sure.

@jseppi do you have a preference for one approach vs. another?


toolness:
Ok, after talking it over w/ @jseppi we think we're gonna go with the normalization approach in #1378 instead, so I'm closing this.

PR #1393, opened 2017-02-16
Add npm-shrinkwrap.json

### LINKED ISSUE #1345: Pin all node dependencies

toolness:
This is an attempt to fix #1345 by pinning all node dependencies via [`npm shrinkwrap`](https://docs.npmjs.com/cli/shrinkwrap).

@jseppi is this all we need to do?


jseppi:
Might want to add a doc note somewhere. Here's something I wrote in a previous project:

> Note that this application uses a npm-shrinkwrap.json file to freeze npm dependencies. If you'd like to update packages, you will have to do so manually and then re-run npm shrinkwrap.

toolness:
Hmm, I'm not sure if there's currently a good place to put that in the docs... And if I add it somewhere, I'm not very confident it'd be read when someone adds a dependency... Hmm.

What do you think of adding [`npm-shrinkwrap-check`](https://www.npmjs.com/package/npm-shrinkwrap-check) and making it something that `ultratest` and travis both run?  That way if someone forgets to re-run (or didn't know they were supposed to run) `npm shrinkwrap` they will be notified by the test suite.

jseppi:
I'm cool with that!

toolness:
Uhhh, running `npm-shrinkwrap-check -3` gives me:

```
package.json and npm-shrinkwrap.json out of sync
 * a found in package.json but not in npm-shrinkwrap.json
 * b found in package.json but not in npm-shrinkwrap.json
 * evil-transform found in package.json but not in npm-shrinkwrap.json
```

That is very odd because `a`, `b`, and `evil-transform` are nowhere to be found in our `package.json`.

Blegh.

jseppi:
wat

jseppi:
Maybe now we have a good place to put a note about npm-shrinkwrap as of https://github.com/18F/calc/pull/1442!

jseppi:
Or we can rejigger the whole thing to use `yarn` instead, as suggested in #1427

toolness:
Oh lol, I just noticed the top of [`npm-shrinkwrap-check`'s README says](https://github.com/redbooth/npm-shrinkwrap-check#readme):

> This module is deprecated. If you want to painlessly lock your dependencies and
check they are in sync you should use [yarn](https://github.com/yarnpkg/yarn)
and pretend `npm shrinkwrap` ( and this module ) never existed.

So yeah, I guess we should just switch to yarn. Closing this in favor of #1427.

PR #1400, opened 2017-02-17
Docker cloud fixes

### LINKED ISSUE #1396: Move "How to Deploy to your AWS Sandbox" into the repository
Now that we have fancy Sphinx docs, we should just move our one Wiki page, [How to Deploy to your AWS Sandbox](https://github.com/18F/calc/wiki/How-to-Deploy-to-your-AWS-Sandbox), into the repo.

One option may be to bring the [`create-aws-instance.sh`](https://github.com/18F/calc/blob/price-list-analysis/create-aws-instance.sh) script from the `price-list-analysis` branch into `develop` and document it.

toolness:
Oops, it looks like #1369 actually broke our docker cloud deployment setup. This fixes it and also adds the `create-aws-instance.sh` script, based on the one in #997, to make it easier to deploy to docker on an aws ec2 instance.

As such, this also basically fixes #1396.

### DIFF
--- a/Dockerfile.cloud-extras
+++ b/Dockerfile.cloud-extras
@@ -11,5 +11,3 @@
 # directory is mounted via folder share.
 
 COPY . /calc/
-
-RUN gulp build
--- a/docker-compose.cloud.yml
+++ b/docker-compose.cloud.yml
@@ -1,0 +1,20 @@
+version: '2'
+services:
+  app: &app
+    build:
+      dockerfile: Dockerfile.cloud
+    volumes:
+      - frontend:/calc/frontend/static/frontend/built/
+      - docs:/calc/docs/static/
+    environment:
+      # Feel free to change this value, or add new environment variables.
+      - DEBUG=yup
+      # Do not change the following value.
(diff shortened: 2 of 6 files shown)


Dockerfile.cloud-extras
```diff
@@ -11,5 +11,3 @@
 # directory is mounted via folder share.
 
 COPY . /calc/
-
-RUN gulp build
```
toolness: This was exploding because none of our npm/python dependencies are installed in the `Dockerfile` anymore.

docker-compose.cloud.yml
```diff
@@ -0,0 +1,20 @@
...
+      - DEBUG=yup
+      # Do not change the following value.
+      - CALC_IS_ON_DOCKER_IN_CLOUD=yup
+  rq_worker:
+    <<: *app
+  rq_scheduler:
+    <<: *app
+volumes:
+  frontend:
+  docs:
```
toolness: I fixed things by making our gulpfile's output directories docker volumes.

PR #1406, opened 2017-02-21
Add a script that creates docker-compose based "proxies" to common CLI utilities

toolness:
This is just an idea I had for creating "proxies" to CLI utilities in our `app` container, to make it so we don't have to e.g. constantly prefix everything with `docker-compose run app ` all the time.

Basically, devs who want to opt-in to this would:

1. Run `./create-docker-bin-scripts.sh` in the root of their repository. This will create a bunch of proxy scripts in `.docker-bin`, creating the directory if it doesn't already exist.
2. Add `./.docker-bin` to their `PATH`, either at the beginning or the end of it, depending on whether they want the proxies to shadow any locally-installed scripts with the same name.

Now these proxies will be accessible when the user is in the root directory of their repository. 

## Notes

* This technique was borrowed from the convention of adding `./node_modules/.bin` to the front of one's `PATH` for node developers. However, an alternative approach might be to add a virtualenv-inspired `source activate-docker-bin` mechanism that dynamically adds the absolute path to the user's `PATH` and indicates the changed environment by modifying the shell prompt in some way. This would be a particularly useful approach if the developer wants the proxies to shadow common locally-installed tools like `python`.

* The `./create-docker-bin-scripts.sh` is intended to be *optional*; some users might dislike the idea of such proxying happening transparently, and using it requires additional setup anyways, so documentation should never treat it as a necessity. That said, even though it's optional, it should be mentioned in developer documentation so that devs *know* it's an option!

* Given the optional nature of the tool, it might be nice if the `EXECUTABLES` array in particular was sourced from a git-ignored separate file, like `.docker-bin-executables.local`, so that developers have full control over what they want to proxy, without having to be worried that they accidentally commit their personalized list to git.

* I originally wanted to implement this by dynamically introspecting the contents of `/calc/node_modules/.bin` and `/venv/bin` in the container, but after looking at what was in there, I was afraid it could get confusing for users, as utilities they might expect to run on their host OS could get unexpectedly redirected to their container, so I opted for a simple whitelisting approach instead.

Anyways, this idea is just an experiment; if we don't like it we can just close the PR without merging.


jseppi:
I don't personally find typing `docker-compose run <name> <cmd>`  to be a problem, so I'm not sure the extra code/maintenance is really worth it for me. Curious what others think though.

toolness:
Ah yeah, no problem. I did post a link to this in the #docker channel on slack but nobody responded so I'm guessing nobody else is annoyed by having to type `docker-compose run app` so much.  I think I might turn this into a separate script/CLI that can be used with any docker-compose based project, so I'll close this issue for now.

Issue #1283, opened 2017-01-25
Make site base url configurable (probably via an env var)


jseppi:
Ref: https://github.com/18F/calc/pull/1263

From [this comment](https://github.com/18F/calc/pull/1263#discussion_r97120212):
> So, I can't actually dynamically generate a link back to the admin page for this email because it is sent in a scheduled task, so I have no request to use for build_absolute_uri. I could hardcode the production URL (or pass in as a default parameter), but then it seems like I should do that for all the other email methods. I guess I'm actually ok with that now that I think about it, because we don't need anything else from request.

We tried using the Django Sites Framework in #1270, but determined it was not a very good approach (mostly because that framework is confusing, poorly documented, and strangely implemented).

toolness:
Blerg I guess we could still go with the sites framework if we want. That does seem to be the recommended way to define what a site's domain is outside of a request context. The main thing to avoid is just trying to create a data migration that changes the default from being example.com.

toolness:
Oh, I guess another potential reason for using the sites framework is that I suspect third-party django apps will use it if needed.  So for instance, if we find some kind of third-party django app that, uh, sends emails outside of a `request` context, it might be built to support the `sites` framework rather than taking a base URL as a string or something. I dunno.

jseppi:
Or stop needing the domain outside of a request context and do away with the whole problem :)

toolness:
Hmm do you mean removing the links back to the site from the emails?  I think that's really useful, though...

jseppi:
The only impacted email would be the approval reminder message. The link is useful, for sure.

toolness:
Hmm, so I'm working on #1407 and realizing that the HTML email designs in #1208 might require us to embed `<img>` tags in the HTML emails, which ought to point back to the server... I suppose that doesn't actually change things, since the approval reminder email is still the only one that never has access to a `request` object, but I figured it might be worth mentioning, for some reason.

I guess that despite the fact that the sites framework is really weird, I think the biggest factor in favor of it is what I mentioned a few comments up, about third-party Django packages integrating with it.

Do dev/staging instances wipe their databases every time they're re-deployed, or do they stick around?  If they stick around, I don't think it's that big a deal to go into the django admin and set their site domains, but if they get wiped on every deploy then we'd have to set the site URL as part of the deployment script, which would be annoying.

jseppi:
Databases are not wiped on deploy

jseppi:
Closed as part of #1407

Issue #677, opened 2016-09-07
Tests should not change AUTHENTICATION_BACKENDS


toolness:
I think there's probably a way to change the way our tests log-in to the website to make it so that we don't need to replace our `UaaBackend` with Django's default `ModelBackend` in tests.  Some advantages:

* It makes tests run more like production deploys, better ensuring that everything integrates properly.
* It means we don't have to make up passwords for our fake users that freak out `bandit`.

toolness:
It also looks like [some of our tests fail when run standalone](https://github.com/18F/calc/pull/1407#discussion_r103016299) unless the `UaaBackend` is explicitly configured, as `cg-django-uaa` actually raises an `ImproperlyConfigured` error.

PR #1407, opened 2017-02-21
Implement awesomely-designed HTML emails

toolness:
This attempts to implement the badass designs in #1208:

> screen shot 2017-02-23 at 7 57 19 am

We're using Lee Munroe's [Really Simple Responsive HTML Email Template](https://github.com/leemunroe/responsive-html-email-template) and then layering on some additional SASS-generated CSS so it looks like the mockup in https://github.com/18F/calc/issues/1208#issuecomment-278721398.

## Notes

* This PR adds the `sites` framework to deal with the whole "how do we get the absolute URL to our site without a `Request` instance" problem, fixing #1283.  Also added a test to `/healthcheck` that ensures that the `Site` object's domain is properly synced with the `Request` object's (that said, this *does* mean that https://calc-prod.app.cloud.gov/healthcheck/ will return a 500, so we might want to revisit this approach if that's not desirable).

* I've renamed `manage.py sendtestemail` to `manage.py send_test_html_email`; it now uses this HTML template and includes an absolute link back to the site, to help us catch misconfigured `Site`s more easily.

* While I was at it, I also added a `<link rel="canonical">` to the base template, which uses the current `Site` object's information to populate its `href` attribute.  Aside from ensuring that search engines don't duplicate searches between calc.gsa.gov and calc-prod.app.cloud.gov, it will also make it a bit easier for us to manually ensure that the `Site` is configured properly.

* This PR adds links to the rendered emails in the style guide and exposes them via views, which should make development/debugging even easier (`send_example_emails` will still be very useful to ensure there aren't issues with the actual email-specific aspects):

  > screen shot 2017-02-21 at 5 07 12 pm

* ~~This PR also introduces [`premailer`](https://github.com/peterbe/premailer) to our `requirements.txt` and uses it to inline CSS.  (Our existing mail service, Mandrill, does do CSS inlining itself, but because there might eventually be ATO/compliance issues with Mandrill, I thought it might be better to do the inlining ourselves for now.)~~ Dropping the inliner for now because it was doing weird things and Gmail (which CO's use) [shouldn't need it anymore](https://www.campaignmonitor.com/blog/email-marketing/2016/09/responsive-email-support-in-gmail-is-coming/).

* `data_capture.email.send_email()` has been changed to take a template name and context, rather than the HTML message.  It then renders the template *twice* with slightly different contexts: one has `is_html_email` set to `True`, while the other has `is_plaintext_email` set to `True`.  Our email templates can then take advantage of these variables to tailor themselves to one format or the other.

To do:
- [x] Solidify our approach to html vs. plaintext content and enact it across all email templates.
- [x] Make Lee Munroe's template look more like the mockup in https://github.com/18F/calc/issues/1208#issuecomment-278721398.
- [x] Add documentation about configuring the default `Site` model properly.
- [x] Should probably add a smoke test for `send_test_html_email.py`.
- [x] Undo [premailer unit test change](https://github.com/18F/calc/pull/1407/files#r102286886) since we're no longer using premailer.


data_capture/email.py
```diff
@@ -43,7 +43,22 @@ def send_mail(subject, to, html_message, reply_to=None):
...
 
-    plaintext_message = collapse_and_strip_tags(html_message)
+    html_ctx = ctx.copy()
+    html_ctx['is_html_email'] = True
+    html_ctx['is_plaintext_email'] = False
+
+    html_message = render_to_string(template, html_ctx)
+    html_message = premailer.transform(html_message)
+    html_message = html_message.encode(
+        'ascii', 'xmlcharrefreplace').decode('ascii')
```
toolness: This part is weird; if I leave it out but still do the premailing, MailCatcher won't even receive the email (or at least, it won't show up in MailCatcher's inbox). Very weird. Is our email pipeline built to be OK with unicode?

data_capture/tests/test_email.py
```diff
@@ -56,7 +64,7 @@ def test_send_mail_works(self):
         self.assertEqual(message.from_email, 'hi@hi.com')
         self.assertEqual(message.body, 'test body')
         self.assertHasOneHtmlAlternative(message)
-        self.assertEqual(message.alternatives[0][0], '<p>test body</p>')
+        self.assertIn('<p>test body</p>', message.alternatives[0][0])
```
toolness: Premailer ensures the HTML is in a valid HTML page, i.e. one with a `<head>` and `<body>` element, so I had to change this assert.

data_capture/templates/data_capture/email/approval_reminder.html
```diff
@@ -1,13 +1,21 @@
...
 <p>
   Hi there,
 </p>
 <p>
-  We notice you haven't approved submitted price lists in CALC. When you can, please review and approve these lists at <a href="{{ unreviewed_url }}">{{ unreviewed_url }}</a>.
+  We notice you haven't approved submitted price lists in CALC. {% if is_plaintext_email %}When you can, please review and approve these lists at {{ unreviewed_url }}.{% endif %}
 </p>
+{% if is_html_email %}
+{% cta url=unreviewed_url text="Review submitted price lists" %}
+{% endif %}
```
toolness: An alternative to only including calls-to-action in HTML emails is to have `{% cta %}` insert different content depending on whether `is_html_email` is `True`.  If it's not, it can output text like `Visit {{ url }} to {{ text }}` or somesuch. This gives us less control over the copy for plaintext emails in exchange for making it a bit easier to write them.

data_capture/email.py
toolness: This is particularly odd given that [Django’s email framework supports Unicode transparently](https://docs.djangoproject.com/en/1.10/ref/unicode/#email).

toolness: Filed #1409 to address this.

data_capture/email.py
```diff
@@ -60,6 +65,22 @@ def send_mail(subject, to, template, ctx, reply_to=None):
         render_to_string(template, plaintext_ctx)
     )
 
+    return (plaintext_message, html_message)
+
+
+def send_mail(subject, to, template, ctx, reply_to=None):
+    '''
+    Django's convinience send_mail function does not allow
```
jseppi: `convinience` -> `convenience` (was my mistake from before)

toolness:
Hmm, so something I'm a bit confused about is whether we actually need CSS inlining *at all*.  According to a [Campaign Monitor post from September 2016](https://www.campaignmonitor.com/blog/email-marketing/2016/09/responsive-email-support-in-gmail-is-coming/), Gmail should've stopped requiring CSS inlining by the end of that month.

I'm also experiencing some weird things with `premailer` so it might be easier to just drop it. Hmm.

toolness:
It looks like Gmail doesn't require CSS inlining, so given the weirdnesses I've experienced with `premailer` and the fact that all GSA COs will be using Gmail, I've dropped `premailer`.  The emails look great on my Gmail and Mail on iOS, at the very least.

data_capture/templates/email/base.html
```diff
@@ -0,0 +1,373 @@
...
+                  <td class="content-block">
+                    {# <span class="apple-link">Company Inc, 3 Abbey Road, San Francisco CA 94102</span> #}
+
+                    {# TODO: Add an unsubscribe link. #}
+                    {# <br> Don't like these emails? <a href="#">Unsubscribe</a>. #}
+                  </td>
+                </tr>
+                <tr>
+                  <td class="content-block powered-by">
+                    Powered by <a href="http://htmlemail.io">HTMLemail</a>.
```
jseppi: Already mentioned this in email, but I'll post here for transparency: I kinda think we should remove this section since it won't mean much for our users, and it is also kind of an advertisement, which might be no bueno in any case. Since the template we're using is MIT licensed, and we have that mentioned above in a code comment, we should be a-ok to remove this line.

toolness: Sounds good, just removed it!!!

ericronne:
@toolness can we pair to make changes reflected here? …

image

toolness:
Ok @jseppi, Eric and I are going to be making some cosmetic tweaks tomorrow but this PR is pretty much ready (we can even file our tweaks as a separate PR).  Feel free to review it... it's also kinda big so I'm happy to walk you through it on screenhero if you want.

data_capture/management/commands/send_example_emails.py
```diff
@@ -1,84 +1,80 @@
...
+
+
+@click.command()
+@click.argument('to', default='user@example.com')
+def command(to):
+    '''
+    Send examples of all the email templates in the Data Capture application.
+    '''
+
+    for example in EXAMPLES:
```
jseppi: 👌

frontend/email_css.py
```diff
@@ -0,0 +1,10 @@
+import pathlib
+from django.utils.safestring import mark_safe
+
+
+CSS_PATH = (pathlib.Path(__file__).resolve().parent /
+            'static' / 'frontend' / 'built' / 'style' / 'email.min.css')
```
jseppi: whoa what is this syntax magic?

styleguide/templates/styleguide.html
```diff
@@ -575,6 +575,30 @@ <h1 id="example-dialog-title">
...
+  <ul>
+    {% for example in email_examples %}
+    <li>
+      {{ example.subject }}
+      <br>
+      <a href="{{ example.plaintext_url }}">Plain text</a> |
+       <a href="{{ example.html_url }}">HTML</a>
+    </li>
+    {% endfor %}
+  </ul>
```
jseppi: Maybe a link to where or how these examples are made would be nice (either here or in `email.md`).

data_capture/management/commands/send_test_html_email.py
```diff
@@ -0,0 +1,32 @@
...
+    missing_args_message = "You must specify some email recipients."  # NOQA
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            'email', nargs='*',
+            help='One or more email addresses to send a test email to.',
+        )
+
+    def handle(self, *args, **kwargs):
+        subject = 'Test HTML email from %s on %s' % (socket.gethostname(), timezone.now())  # NOQA
```
toolness: Heh, nah, this is the part of the code that comes from Django 1.9's `sendtestemail.py`, which I left unmodified.

data_capture/tests/test_email.py
```diff
@@ -14,6 +14,17 @@
 
 @freeze_time(datetime(2017, 1, 8, 20, 51, 0))
 @override_settings(DATA_CAPTURE_SCHEDULES=[FAKE_SCHEDULE],
+
+                   # Ugh, this is required for the cg-django-uaa to not
+                   # complain when running this test suite standalone.
+                   AUTHENTICATION_BACKENDS=(
```
toolness: Yeah, I'm not a fan of it though--the closest existing issue we have to address it is #677 so I added a note about it there.

frontend/email_css.py
toolness: Hehe it's [`pathlib`](https://docs.python.org/3/library/pathlib.html)!  The changelog code uses it too... It used to be experimental in Python 3.4 but now it's not so I figured we should use it.  Sooo much nicer than using `os.path.join`, and [recent changes in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#pep-519-adding-a-file-system-path-protocol) make it very compatible with legacy standard library code that expects filesystem paths.

PR #1410, opened 2017-02-21
Glean first integer in cell contents for min experience

jseppi:
This replaces the simple `int` coercer for `min_experience` with a new one called `extract_first_int` that instead returns the _first_ integer value found in the cell's contents.

This will allow gleaning of contents like `"At least 5 years, with 2 years management"` (-> `5`) that @tram has been seeing in research sessions. It will also handle contents like `"10+"` (-> `10`). And it will of course handle the "normal" case of just having an integer in the cell.

We discussed wanting to have "warning messages" shown for this special gleaning case (and maybe some others), since it is possible our parser could glean an unintended or surprising value, so it would be nice to notify the user, but not throw out the row. I think implementation of these warnings should be handled separately though.

### DIFF
--- a/data_capture/schedules/coercers.py
+++ b/data_capture/schedules/coercers.py
@@ -146,3 +146,28 @@ def extract_hour_unit_of_issue(text):
         return text
 
     return 'Hour'
+
+
+def extract_first_int(text):
+    '''
+    Returns the first integer found in the input text.
+
+    >>> extract_first_int('At least 12 years with 8 years management')
+    12
+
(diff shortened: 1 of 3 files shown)


jseppi:
[When a user uploads a price list with non-strictly-numeric Min Experience values, CALC tries to automatically glean an integer value.](https://trello.com/c/c3BdXS7c/238-when-a-user-uploads-a-price-list-with-non-strictly-numeric-min-experience-values-calc-tries-to-automatically-glean-an-integer-va)

PR #1418, opened 2017-02-22
Implement user "fly out" menu

### LINKED ISSUE #1413: Implement account menu (with current site look)

jseppi:
Closes #1413 

Following guidance at https://www.w3.org/WAI/tutorials/menus/flyout/

Closed:
screen shot 2017-02-22 at 1 14 55 pm

Open:
screen shot 2017-02-22 at 1 14 49 pm

Other notes:
It's apparently not really possible to make an accessible "fly over" menu with just CSS, so this PR includes some fairly light JS to make it work. In the case of no JS, the menu will degrade to just be a link to the user's activity page.

To Do:
- [x] move js to separate file
- [x] change js to not use jquery
- [x] use svg for the triangle icon after the email address
- [x] make sure it is present and works in all headers (admin, data explorer, data capture, etc)


jseppi:
[As a user I should be able to access "My activity" and "Log out" from an easily navigable location.](https://trello.com/c/JXBYBuLk/241-as-a-user-i-should-be-able-to-access-my-activity-and-log-out-from-an-easily-navigable-location)

jseppi:
arggggg we don't currently include any of our scripts in `/admin` so we don't have a bundle there. Will need to create one.

jseppi:
Alrighty. @ericronne: want to check this out to make sure I captured everything from the design?

@toolness or @hbillings: I'd also appreciate a code review from either/both of you!

frontend/source/js/common/usermenu.js
```diff
@@ -0,0 +1,20 @@
...
+
+document.addEventListener('DOMContentLoaded', () => {
+  const menu = document.querySelector('#usermenu');
+  const trigger = document.querySelector('#usermenu .usermenu-trigger');
+  const openLabel = 'Show user menu';
+  const closeLabel = 'Close user menu';
+
+  trigger.setAttribute('aria-haspopup', true);
+  trigger.setAttribute('aria-expanded', false);
+  trigger.setAttribute('aria-label', openLabel);
```
jseppi: I wasn't _quite_ sure about including `aria-label` but just announcing the user's email address didn't seem quite good enough to let them know there was also a menu that could be toggled.

data_explorer/templates/_header.html
```diff
@@ -3,7 +3,18 @@
   <a class="logo-container columns two" href="{% url 'index' %}"><img class="logo" src="{% static 'frontend/images/calc_logo.png' %}" alt="CALC: Contract Awarded Labor Category Tool"></a>
   <div class="columns ten">
     {% if user.is_authenticated %}
-    <p class="user-id">Hello, <a href="{% url 'user_account:index' %}">{{ user.email }}</a></p>
+      <div id="usermenu">
+        <p class="user-id">Hello,
+          <a class="usermenu-trigger" href="{% url 'user_account:index' %}">{{ user.email }}</a>
+          <svg class="icon-triangle" width="10" height="7" viewBox="0 0 10 7" xmlns="http://www.w3.org/2000/svg">
```
jseppi: Used an inline `svg` for styling goodness

frontend/source/js/common/usermenu.js
```diff
@@ -0,0 +1,24 @@
...
+  const openLabel = 'Show user menu';
+  const closeLabel = 'Close user menu';
+
+  if (!menu || !trigger) {
+    return;
+  }
+
+  trigger.setAttribute('aria-haspopup', true);
+  trigger.setAttribute('aria-expanded', false);
+  trigger.setAttribute('aria-label', openLabel);
```
jseppi: I wasn't quite sure about including aria-label but just announcing the user's email address didn't seem quite good enough to let them know there was also a menu that could be toggled.

frontend/source/js/common/usermenu.js
```diff
@@ -0,0 +1,24 @@
+/* global document */
+
+document.addEventListener('DOMContentLoaded', () => {
```
jseppi: This code could definitely be generalized, but we don't have any plans for additional menus like this so I don't think it is super worth it at this point. If we add another one, then it'll be an easy task to do at that point.

ericronne:
@jseppi could you drop some screenshots here for me? 😬 
thx!

jseppi:
You don't want to test all the sweet interactions on your own???

jseppi:
I'm not good at screenshots like you. Topmost PR comment has the big two screenshots, here are two more showing hover states (but I don't know how to show the mouse cursor in them).

Mouse hover, closed menu:
screen shot 2017-02-23 at 11 03 16 am

Hover item, menu open:
screen shot 2017-02-23 at 11 03 22 am

hbillings:
From an interaction perspective, would it be reasonable to make this work on hover as well as on click?

jseppi:
Hover can often be a bit whacky in menus like this. Like if moving the mouse out should also close the menu, then we'd need to set a timer or something so it doesn't instantly close (which is frustrating). I played around with it while implementing, but I ended up just going with the click to open.

ericronne:
@hbillings we should test it, but my gut is that it should open only on click. Most (all?) of these sites behave that way …

https://app.mural.ly/invitation/mural/gsa6/1486147353060?sender=ericronne&key=3701235560

hbillings:
I think if it only opens on click, we shouldn't display it as a link. When I interact with it, I'm not sure if clicking it is going to take me off the page or not. What about doing something like a background color, to make it feel a little more buttony? (I seem to remember you had some mocks like that at one point in time...)

ericronne:
@hbillings sorry i missed your comment. I figured the downward arrow was sufficient indicator. What if we also changed the type *and* arrow to white on hover (in addition to the underscore add)?

hbillings:
@ericronne y'know, I think it's actually the underline appearing on hover that's tripping me up! Maybe just change the color?

ericronne:
Borderline 508 issue, but should be ok. #doit

hbillings:
I guess part of this is also that we aren't using link-like behavior for anything but navigating away from the current page (or the current part of the page) anywhere else on the site.

jseppi:
Just pushed a minor change that allows the caret next to the email address to be included in the clickable area.

frontend/source/sass/components/_usermenu.scss
```diff
@@ -0,0 +1,87 @@
...
+    top: 1.2em;
+    right: 0;
+    width: $usermenu-width;
+    z-index: 1;
+
+    text-align: left;
+    font-size: $medium-font-size;
+    font-weight: $font-weight-bold;
+    background: $color-gray-lightest;
+    border-radius: 4px;
```
hbillings: ugh, we should be able to use the `$border-radius` variable. Opened #1426 to address.

frontend/source/sass/components/_usermenu.scss
```diff
@@ -0,0 +1,87 @@
...
+    list-style: none;
+    padding: 1.2em 1.5em;
```
hbillings: Should we use the standard spacing units here? (though they're `rem` and not `em`, which is a bit of a pain.)

frontend/source/sass/components/_usermenu.scss
```diff
@@ -0,0 +1,87 @@
...
+      }
+    }
+
+    // Menu caret
+    &:before {
+      content: '';
+      position: absolute;
+      top: -$usermenu-caret-size;
+      // aligned in center of menu
+      right: $usermenu-width/2 - $usermenu-caret-size/2;
```
hbillings: 👌

data_explorer/templates/_header.html
```diff
@@ -3,7 +3,19 @@
   <a class="logo-container columns two" href="{% url 'index' %}"><img class="logo" src="{% static 'frontend/images/calc_logo.png' %}" alt="CALC: Contract Awarded Labor Category Tool"></a>
   <div class="columns ten">
     {% if user.is_authenticated %}
-    <p class="user-id">Hello, <a href="{% url 'user_account:index' %}">{{ user.email }}</a></p>
+      <div id="usermenu">
+        <p class="user-id">Hello,
+          <a class="usermenu-trigger" href="{% url 'user_account:index' %}">
```
hbillings: I'm not entirely sure why, but when you're on the `/account/` page, it doesn't seem like the `preventDefault` call works -- clicking the link refreshes the page instead of triggering the dropdown. :-/

frontend/source/sass/components/_usermenu.scss
jseppi: Agree, but it should happen in the fix to #1426 :)

data_explorer/templates/_header.html
jseppi: good catch! looking into it...

jseppi: oof, there are no scripts included there. Might need to make a "common" bundle and put this script there, and then include that bundle in the site's `base.html`.

jseppi: okie dokie, this should work now. Had to shave a small yak (created new js bundle, `common/index.min.js`) to get it there though.

frontend/source/sass/components/_usermenu.scss
ericronne: *hold pls …*

frontend/source/sass/components/_usermenu.scss
```diff
@@ -0,0 +1,87 @@
...
+    width: $usermenu-width;
+    z-index: 1;
+
+    text-align: left;
+    font-size: $medium-font-size;
+    font-weight: $font-weight-bold;
+    background: $color-gray-lightest;
+    border-radius: 4px;
+    list-style: none;
+    padding: 1.5rem 1rem 1.7rem 2rem;
```
jseppi: @ericronne: I think @hbillings meant that we should use the spacing variables defined in `_variables.scss`, which start at line `82` (with `// Spacing units`).

ericronne: wondered. *hold pls …*

ericronne:
Is it easy to change the triangle to white on hover, like the email address now does? 
Don't know how to do that one.

jseppi:
@ericronne already done!

jseppi:
@toolness can you go ahead and check this out real quick also, especially given that I created a new bundle (`common/index.min.js`)?

jseppi:
@hbillings ready for your re-review!

ericronne:
SWOON!

frontend/source/js/common/usermenu.js
```diff
@@ -0,0 +1,26 @@
...
+  if (!menu || !trigger) {
+    return;
+  }
+
+  trigger.setAttribute('aria-haspopup', true);
+  trigger.setAttribute('aria-expanded', false);
+  trigger.setAttribute('aria-label', openLabel);
+
+  trigger.addEventListener('click', (e) => {
+    e.preventDefault();
```
toolness: Do you think it'd be OK to move this `preventDefault` to the end of the function rather than the beginning?

The reason I ask is b/c if the intervening code has any bugs--I'm thinking really old browsers that might lack `classList`, as unlikely as it sounds--then if the `preventDefault` is done at the beginning, this will mean the user doesn't see anything happen. Whereas if we save the `preventDefault` until the very end, if everything worked, then we're guaranteed to show the menu to the user, so it's safe to prevent the default navigation behavior... if that makes sense.

OTOH, if they're on a data capture page, they'll get the safe mode prompt, so I guess it's not a huge deal either way.

jseppi: yeah that's a good idea!

frontend/source/js/common/usermenu.js
toolness: NOICE it sounds splendid with Firefox + NVDA.

jseppi: Cool, thanks for checking that. VoiceOver sounded good to me as well.

jseppi:
🎉

PR #1420, opened 2017-02-22
Fix histogram label regression

### LINKED ISSUE #1402: Data Explorer chart label visual regressions
Looks like #1401 caused some regressions on the Data Explorer's labels (which also have the `.label` class):

screen shot 2017-02-17 at 10 09 10 am

Here's what they used to look like:

screen shot 2017-02-17 at 10 09 23 am

jseppi:
Fixes #1402

Fixed by changing all `.label`s to `.chart-label`s in histogram.jsx.

Looks like this again:
screen shot 2017-02-22 at 3 09 40 pm

### DIFF
--- a/frontend/source/js/data-explorer/components/histogram.jsx
+++ b/frontend/source/js/data-explorer/components/histogram.jsx
@@ -43,7 +43,7 @@ const INLINE_STYLES = `/* styles here for download graph compatibility */
   fill: #5b616b;
 }
 
-.axis .label {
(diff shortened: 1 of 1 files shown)


jseppi:
This is somewhat high priority since it fixes a styling regression currently in `develop` (that we don't want to make it into the next release), so plz review @toolness.

jseppi:
Or @hbillings you can review instead!

PR #1425, opened 2017-02-23
Show dev banner in DEBUG mode, link to styleguide/docs in it

toolness:
This shows the "non-prod" banner by default when `DEBUG` is true (unless `HIDE_DEBUG_UI` is also true) and links to the style guide and docs on it, so that devs can easy discover and access them.

> screen shot 2017-02-23 at 12 02 25 pm

(The contrast ratio at the top-right isn't great but I was lazy and figured it's not that bad since it's just on dev mode, but I can change it if needed.)

This also renders the hard-to-find "Documentation" djdt panel pretty useless so I removed it.

It also consolidates the header code into a partial so that it's not duplicated between the admin base template and the data explorer base template.


data_explorer/templates/_banner.html
```diff
@@ -0,0 +1,32 @@
+{% load staticfiles %}
```
jseppi: 👍 good idea making this a partial

PR #1428, opened 2017-02-23
Delete bulk-created Contracts when SubmittedPriceList with same Contract Number is approved

### LINKED ISSUE #1421: Remove bulk data rows that have contract number of an uploaded price list
If a CO uploads a price list whose contract number matches labor categories that were part of a bulk load and those bulk-loaded labor categories are currently being shown in Data Explorer, approval of the new price list should remove any labor categories that match the newly-approved contract number.

If we wanted to be super fancy, we would let the CO know when they submit the price list that there are labor categories currently being shown in Data Explorer that match their contract number and that they'll be removed upon price list approval.

jseppi:
When an admin approves submitted price lists, now any bulk-loaded contract data with the same contract numbers (and on the same schedule) as any of the submitted price lists will be DELETED. 

This is to prevent duplicate data showing up in the API/Data Explorer as S70 COs enter price lists through Data Capture's flow.

Closes #1421

screen shot 2017-02-23 at 3 27 45 pm

### DIFF
--- a/data_capture/admin.py
+++ b/data_capture/admin.py
@@ -10,6 +10,7 @@ from django.contrib.auth.models import User
 from django.contrib.auth.admin import UserAdmin
 from django.contrib.auth.forms import UserChangeForm
 
+from contracts.models import Contract
 from . import email
 from .schedules import registry
 from .models import SubmittedPriceList, SubmittedPriceListRow
@@ -177,6 +178,18 @@ def approve(modeladmin, request, queryset):
     not_approved = queryset.exclude(status=SubmittedPriceList.STATUS_APPROVED)
     count = not_approved.count()
     for price_list in not_approved:
+        # Remove existing Contract models that were bulk loaded
+        # and that have the same contract number and are on the same
+        # schedule
+        existing_bulk_contracts = Contract.objects.filter(
+            idv_piid__iexact=price_list.contract_number,  # same contract num
+            schedule=price_list.get_schedule_title(),  # on same schedule
+            upload_source__isnull=False,  # and that have bulk upload src
+        )
+        # TODO: keep track of num deleted (per contract num) and add to message
+        # num_existing = existing_bulk_contracts.count()
+        existing_bulk_contracts.delete()
+
         price_list.approve(request.user)
         email.price_list_approved(price_list, request.get_host())
 
--- a/data_capture/admin.py
+++ b/data_capture/admin.py
@@ -187,8 +187,7 @@ def approve(modeladmin, request, queryset):
             schedule=price_list.get_schedule_title(),  # on same schedule
             upload_source__isnull=False,  # and that have bulk upload src
         )
-        # TODO: keep track of num deleted (per contract num) and add to message
-        # num_existing = existing_bulk_contracts.count()
+        # keep track of num deleted (per contract num)
         existing_count = existing_bulk_contracts.count()
         if existing_count:
             deleted_counts[price_list.contract_number] = existing_count
--- a/data_capture/admin.py
+++ b/data_capture/admin.py
@@ -177,6 +177,7 @@ class SubmittedPriceListRowInline(admin.TabularInline):
 def approve(modeladmin, request, queryset):
     not_approved = queryset.exclude(status=SubmittedPriceList.STATUS_APPROVED)
     count = not_approved.count()
+    deleted_counts = {}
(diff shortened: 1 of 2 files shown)


jseppi:
[As a Data Admin, when I approve a price list, any bulk created rows for the same contract should be deleted so that there are no duplicates.](https://trello.com/c/udMAbN4Y/242-as-a-data-admin-when-i-approve-a-price-list-any-bulk-created-rows-for-the-same-contract-should-be-deleted-so-that-there-are-no-d)

data_capture/admin.py
```diff
@@ -188,6 +204,14 @@ def approve(modeladmin, request, queryset):
         )
     )
 
+    for del_contract_num, del_count in deleted_counts.items():
+        messages.add_message(
```
jseppi: This will add a new message for EACH contract number that has deleted bulk-loaded contracts, which could be kinda ugly/overwhelming. However, I don't think that _in practice_ this will be that bad.

jseppi: And I wanted the data to be surfaced to the admins just in case...

jseppi:
This is ready for review, @toolness

toolness:
I think the code here looks good, but I do have some concerns about the approach:

* This feels a little bit related to the (long dormant) WIP #1029, though perhaps it's just superficial, I'm not sure. I guess it's something worth noting, I don't think anything has to be done about it though.

* The original intent of the "approve" action was for it to be undoable, i.e. un-approving a price list immediately after approving it ought to result in (more or less) an unchanged db. However, this approach deletes the duplicate bulk data rows on approval, but does nothing on un-approval, which means that a data administrator who *accidentally* approves a PL that conflicts with bulk data won't be able to easily undo the action. That's not necessarily a deal-breaker but it might be worth surfacing to the data administrator, just so they know it's not easily undoable?

  Alternatively I guess we could make it undoable but that also feels like a lot of work, especially if we factor in the possibility that the bulk data could be updated after the PL is uploaded, which leads us to...

* OY! What if the bulk data is updated after the PL is uploaded, and *it* contains duplicate contract rows?  This seems like an impossible situation, since for S70 at least, the bulk data represents the "old way of doing things" while the PL upload is the new way, but maybe we should verify this just to be sure. And we should probably document it in our code just so future devs (and our future selves) understand our assumptions.

I suppose one alternate approach might be to have bulk upload actually group the data by contract number, and then create `SubmittedPriceList` models for every group... Then we could treat the situation just like we treat any other, i.e. retire the old price list and approve the new one (um I *think* that's what we do, right?).  But that might complicate the re-uploading of new bulk data, oy.

Anyways, I suspect the approach you've implemented here is probably the most practical given the circumstances, but I just wanted to make sure we considered alternatives and took weird edge cases into account before merging it.

jseppi:
Looping in @tram. See above, @toolness has lots of good points ^^

jseppi:
Thanks for giving this a lot of thought, @toolness.

I think this whole situation is pretty non-ideal. It basically boils down to, as you noted, that the bulk S70 import and the Data Capture flow were never supposed to exist together, which guided our implementation of Data Capture. I suppose my (maybe unfounded) hope is that there will not be any more S70 bulk imports going forward since we've now started the official launch of DC.

jseppi:
@tram has this use-case come up at all so far in the S70 roll-out?

hbillings:
Seems like we should revisit this now that we know the bulk uploads will continue until morale improves.

hbillings:
So I'm actually not sure how helpful this is going to be, given that bulk uploads have not ceased as we had hoped they would post-data capture launch. 😞 (In fact, bulk uploads seem to be leaned on more heavily now than they ever have been as a tool to get ever more data into CALC.)

I'd love @tadhg-ohiggins to make sure I'm reading this right, but it seems like this PR checks for duplicates in the bulk upload data when a price list is approved. Deleting dupes at that point doesn't seem like it will do much, since the original duplicate will still be in the bulk data and will be re-introduced on the next bulk upload. 

Thoughts, friends?

tadhg-ohiggins:
Given the comments from @toolness and @jseppi above, merging this at this point seems like a bad idea to me. I think we should create an issue to capture the need for something to be done about duplication, but it doesn't seem like this is the thing that should be done.

hbillings:
Updated #1421 to reflect the New CALC Realities.

PR #1435, opened 2017-02-27
Link to email template source in styleguide

toolness:
@jseppi I'm not sure if this is what you meant by https://github.com/18F/calc/pull/1407#discussion_r103044659, but I thought it'd be useful regardless. Basically it adds `{% template_tag_library %}`, `{% template_url %}`, and `{% template_link %}` template tags to the styleguide, which allows us to link to templates and template tag libraries like so:

> template_links

That way anyone who wants to edit one of these can easily find the link (though it could also be discovered by using the Django Debug Toolbar, I think this is easier).

### DIFF
--- a/styleguide/templatetags/styleguide.py
+++ b/styleguide/templatetags/styleguide.py
@@ -77,6 +77,26 @@ class WebComponentHTMLParser(HTMLParser):
                 self.extends = val
 
 
+@register.simple_tag(takes_context=True)
+def template_url(context, template_name):
+    '''
+    Return a GitHub URL to the source of the given template.
+    '''
+
+    t = context.template.engine.get_template(template_name)
+    return github_url_for_path(os.path.relpath(t.origin.name, ROOT_DIR))
+
+
+@register.simple_tag(takes_context=True)
+def template_link(context, template_name):
+    '''
+    Return a link to the source code of the given template.
+    '''
+
+    url = template_url(context, template_name)
+    return SafeString(f'<code><a href="{url}">{template_name}</a></code>')
+
+
 @register.simple_tag
 def webcomponent(html):
     '''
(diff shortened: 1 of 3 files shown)


toolness:
Urg, I think the styleguide test is failing because `TEMPLATE_DEBUG` isn't true during testing, and apparently it needs to be true in order for a template's `origin` property to not be `None`.  Blech.

styleguide/templatetags/styleguide.py
```diff
@@ -78,6 +78,71 @@ def handle_starttag(self, tag, attrs):
...
+    '''
+
+    # Note that we can't simply use the `origin` property of a Template
+    # object, because, at least in Django 1.8, it seems this property
+    # is None if TEMPLATE_DEBUG is disabled (and we want to be able to
+    # render the styleguide in non-debug instances).
+
+    candidates = []
+
+    for loader in context.template.engine.template_loaders:
```
jseppi: This looks like some _deep django_. Any way you could add a smoke test for this method to make sure it'll work in future versions or something?

jseppi: Same I guess for `template_tag_library`

toolness: It is indeed deep django--I ended up looking at Django's source code, and it looks like the template loader architecture in particular has changed a lot over the past few releases, so it will probably break as we upgrade 😣   The good news though is that it seems like it's evolved since 1.8 in a way that will make it *easier* for us to introspect in these ways rather than harder, so that's nice at least.

I figured the "render the styleguide and make sure it returns 200 OK" was enough of a smoketest, since it uses those funky tags, but I can add moar.

jseppi: Hrm, that's probably good enough actually.

toolness: TOO LATE BUDDY

PR #1436, opened 2017-02-27
Move example email metadata into emails.py

toolness:
@jseppi I'm also not sure if this is what you meant by https://github.com/18F/calc/pull/1407#discussion_r103044659, but I thought it might be a good idea to move the metadata about example emails closer to the definition of the email sending functions.

Advantages:

* It allows us to enforce DRY a bit, by not specifying the template name in two different places.
* It should also help ensure that future emails added to the system come with built-in examples (rather than not coming with examples b/c the contributor didn't know how to add them).

Disadvantages:

* I'm not sure if I'm a big fan of the decorator approach, which is responsible for passing the template name in as the first argument to the sender function. This makes it harder to understand how to use a sender function just by looking at its call signature.

  I thought about passing `template` in as a keyword argument instead, thereby allowing it to be specified last in the call signature, or even via `**kwargs` and popped-off, but neither of those seemed like very big improvements. I guess the alternative is to convert these functions into classes, but that would also require changing all of their call sites (of which there admittedly aren't many).

### DIFF
--- a/data_capture/email.py
+++ b/data_capture/email.py
@@ -1,6 +1,8 @@
 import re
+from functools import wraps
 from django.core.mail import EmailMultiAlternatives, get_connection
 from django.utils.html import strip_tags
+from django.utils import timezone
 from django.template.loader import render_to_string
 from django.contrib.auth.models import User
 from django.template.defaultfilters import pluralize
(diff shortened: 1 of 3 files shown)


data_capture/email.py
```diff
@@ -188,7 +256,14 @@ def bulk_upload_succeeded(upload_source, num_contracts, num_bad_rows):
     )
 
 
-def bulk_upload_failed(upload_source, traceback):
+@email_sender(
+    template='data_capture/email/bulk_upload_failed.html',
+    example_ctx={**bulk_upload_succeeded.example_ctx, **{
+        'r10_upload_link': 'https://example.com/r10_bulk_upload',
+        'traceback': 'error traceback'
+    }}
```
toolness: This syntax was introduced in Python 3.5 by [PEP 448](https://www.python.org/dev/peps/pep-0448/).

data_capture/email.py
```diff
@@ -93,7 +95,52 @@ def send_mail(subject, to, template, ctx, reply_to=None):
...
+    The given template name will be passed on to the decorated function
+    as its first argument.
+    '''
+
+    def wrap(func):
+        name = func.__name__.replace('_', ' ')
+
+        EXAMPLES.append({
+            'subject': f'Example {name}',
+            'template': template,
```
toolness: I thought about providing a default value for `template` by making it `f'data_capture/email/{ func.__name__ }.html'` but I kinda like the explicitness of the full template path.

data_capture/email.py
jseppi: dang, that's fancy

toolness:
Rawk!  Yeah, we can always refactor later to a klass based approach, especially if we end up wanting to add more functionality/metadata to these thingamabobs.

I like this notion of there being "example contexts" for templates.  It could be especially nice to "port" the concept to some of our more complicated templates, like the "show gleaned data, possibly with errors" step of the PL upload, so that we could easily manually test/develop those views without having to go through the whole PL upload process (and uploading just the right kind of PL to trigger whatever template logic we want).

Issue #1360, opened 2017-02-13
Write a basic front-end development guide


toolness:
There's a blurb in our setup instructions that mentions the fact that front-end assets at `/frontend/source` and built via gulpfile into `/frontend/static/frontend/built` but that isn't a lot to go on.  I think we should pull that information out into a very basic front-end guide (easily findable via the work being done in #1353) that mentions things like:

* The style guide at `/styleguide/`
* The front-end test suite at `/tests/`
* The high-level distinction between data capture's largely static, web component-based front-end vs. the data explorer's more dynamic React/Redux-based front-end

It doesn't (and *shouldn't*, for reasons of maintainability) have to be that long.

What do you think @jseppi @hbillings?

jseppi:
Great idea!

hbillings:
Yes!! we might also want to write just a little bit about philosophy (we loosely follow BEM, etc)

PR #1442, opened 2017-02-27
Add a basic front-end guide

toolness:
This is a first attempt at fixing #1360.

Thoughts @jseppi @hbillings ?

### DIFF
--- a/docs/frontend.md
+++ b/docs/frontend.md
@@ -1,0 +1,84 @@
+# Front-end
+
+In general, front-end code is in the [frontend](../frontend/) directory.
+
+This guide is about the nuts-and-bolts of developing front-end code; for
+details on how to use or style individual components, see the
+[style guide][].
+
+## The static asset generator
+
+If you haven't already, make sure you've followed the
+[setup guide](setup.md); it explains how to get the gulp-based
+static asset generator up and running. It will watch for changes to
+front-end code and re-build bundles as needed.
+
+All the static assets (SASS for CSS and ES6 JavaScript) are 
+located in the [frontend/source/](../frontend/source/) directory. Outputs
+from the static asset generator are placed in
+`frontend/static/frontend/built/`.
+
+Examine [gulpfile.js](../gulpfile.js) for details about the generator's
+pipeline.
+
+If you just want to build the assets once without watching for changes, run:
+
+```sh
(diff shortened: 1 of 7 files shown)


docs/frontend.md
```diff
@@ -0,0 +1,84 @@
...
+HTML5 Custom Elements to provide a progressively-enhanced experience.
+
+The source code is located in
+[frontend/source/js/data-capture](../frontend/source/js/data-capture/).
+
+Tests are [QUnit][]-based and are located in 
+[frontend/source/js/tests](../frontend/source/js/tests/).
+
+To run the QUnit tests, visit
+[/tests/](https://calc-dev.app.cloud.gov/tests/) on your local
```
toolness: Note that this link is actually broken because we don't expose tests on non-`DEBUG` instances, but when the docs are hosted on `/docs/` this will get re-written to be a relative URL, and it will work (well, as long as `DEBUG` is true on the deployment it's running on).

docs/frontend.md
```diff
@@ -0,0 +1,84 @@
+# Front-end
```
jseppi: While I personally prefer the hyphenated form, I believe the Front End Guild settled upon "front end" with a space as the preferred way to write that word/phrase, so it might be a good idea to maintain consistency with that decision.

docs/frontend.md
```diff
@@ -0,0 +1,84 @@
...
+```
+npm run test
+```
+
+You can also run the tests in a continuous "watch" mode, which re-runs
+tests as you change your code:
+
+```
+npm run test:watch
+```
```
hbillings: Maybe worth adding that you can do `docker-compose run app jest --coverage` like @jseppi figured out you can do to get a report on the coverage?

docs/frontend.md
hbillings: We might also want to point out that the admin page overrides are in `hourglass/templates/admin`.

docs/frontend.md
jseppi: Alternatively we could mention that `jest` can be run directly with any of its command-line args via `docker-compose run app jest` (or just `jest`). ref: https://facebook.github.io/jest/docs/cli.html

toolness:
Ok, I just added stuff based on your suggestions!

PR #1444, opened 2017-02-28
Add SlideyPanel to make education dropdown slide up/down again

### LINKED ISSUE #1227: Education drop-down no longer slides down
As of #1075, the education drop-down, while made more accessible and bug-free, no longer has that snazzy sliding animation when expanded or collapsed.

toolness:
This fixes #1227 by adding a `<SlideyPanel>` component that uses [`ReactTransitionGroup`](https://facebook.github.io/react/docs/animation.html#low-level-api-reacttransitiongroup) under the hood.

If jQuery is available, it uses `$.slideUp`/`$.slideDown` to animate the panel; otherwise the transition is instantaneous. I did this because we've still got it on the page, and it's a lot easier to use than CSS transitions, which are kinda painful to do sliding animations with.

To do:

- [x] Add tests.

### DIFF
--- a/frontend/source/js/data-explorer/tests/slidey-panel.test.jsx
+++ b/frontend/source/js/data-explorer/tests/slidey-panel.test.jsx
@@ -1,0 +1,80 @@
+import React from 'react';
+import { mount } from 'enzyme';
+
+import SlideyPanel from '../components/slidey-panel';
+
+const fake$ = (el) => {
+  const name = `<${el.nodeName.toLowerCase()}>`;
+
+  return {
+    hide() {
+      fake$.log.push(`hide ${name}`);
+      return this;
+    },
+    slideUp(speed, cb) {
+      fake$.log.push(`slideUp ${name} ${speed}`);
+      cb();
+    },
+    slideDown(speed, cb) {
+      fake$.log.push(`slideDown ${name} ${speed}`);
+      cb();
+    },
+  };
+};
+
+fake$.log = [];
+
+describe('<SlideyPanel>', () => {
+  beforeEach(() => {
+    fake$.log = [];
+  });
+
+  it('shows content if it is expanded at mount', () => {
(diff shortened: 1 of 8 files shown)


ericronne:
I'll review! How do i view @toolness??

When i load that branch, the panel defaults to opened, and i can't close it …
image

toolness:
ummmm that is very odd

toolness:
what happens if you, uh, restart the server and stuff?

ericronne:
Still stays open :/

jseppi:
`docker-compose build` ?

ericronne:
Nope. It was something weirder. I ran the ` flush ` command as a way throwing digital salt over my shoulder. Restarted (again). Works!

@toolness am i to understand that you are now <drool>styling selectors</drool>?

toolness:
Cool, glad it works now!

What do you mean by styling selectors?  The only CSS change I made was to make the dropdown list visible by default... Basically the way the "old" education drop-down worked was by being in the page at all times but hidden by default, whereas now it's only actually on the page when it's visible, which makes it redundant to style it to be hidden by default.

ericronne:
Righto. Curious about that selector; chat later

frontend/source/js/data-explorer/tests/slidey-panel.test.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+    wrapper.setProps({ expanded: false });
+    expect(wrapper.find('p').exists()).toBeFalsy();
+  });
+
+  it('slides down when expanded', () => {
+    const wrapper = mount(<SlideyPanel $={fake$}><p>hi</p></SlideyPanel>);
+
+    wrapper.setProps({ expanded: true });
+    expect(fake$.log).toEqual([
```
toolness: Yeah, I was originally going to use them, but with jQuery's method chaining (specifically `hide()` returning `this`) I wasn't sure how to adapt them, so I just ended up doing the log thingy. I kinda like the readability of the log thingy anyhow.

PR #1445, opened 2017-02-28
Adjust tz_timestamp format to be more humane

jseppi:
before: `Jan. 1, 2015, 1:55 p.m. (EST)`
after:    `Jan. 1, 2015 at 1:55 PM (EST)`


jseppi:
here you go @ericronne

ericronne:
Ooh i think unless we do small caps, "p.m." might be better than "PM" 😬

PR #1440, opened 2017-02-27
Update the text of six email templates

ericronne:
To do:
- [x] Get @jseppi’s review
- [x] Add proper weblinks
- [x] If it’s easy to do, split up date and time references to make
sentences flow better. (E.g. “The price list you submitted **at** 7:15
(EST) **on** February 20, 2017 …”
- [x] Change `AM/PM` to `a.m./p.m.`
- [x] figure out what to do about `num_bad_rows` in the bulk upload succeeded email

PR #1447, opened 2017-03-01
[WIP] Enable django.contrib.admindocs in DEBUG mode

toolness:
This enables [the Django admin documentation generator](https://docs.djangoproject.com/en/1.10/ref/contrib/admin/admindocs/) on development (`DEBUG`) instances of CALC.

It adds a "Documentation" option under the site admin:

> admindocs

And as can be seen from the above screenshot, it links to auto-generated documentation on template tags and filters, models, and views.

Since our Django admin skin wasn't created to take this kind of content into consideration, though, the styling on the individual pages isn't that great.  For example, while I *think* the tags and filters docs are supposed to have a right-hand TOC sidebar that lists all the tags and filters, it shows up at the very bottom of the page on our admin site, which isn't very helpful.

So I guess if we do decide to integrate this, we should think about:

- [ ] making at least a few styling tweaks to our admin templates so these docs don't look horrible;
- [ ] overriding [`admin_doc/index.html`](https://github.com/django/django/blob/master/django/contrib/admindocs/templates/admin_doc/index.html) so it links to our `/docs/` and `/styleguide/`, and add links from those places to these admin docs.

@jseppi @hbillings do you think this would be useful? Or nah?

### DIFF
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -367,3 +367,8 @@ DEBUG_TOOLBAR_PANELS = [
     'debug_toolbar.panels.redirects.RedirectsPanel',
     'data_capture.panels.ScheduledJobsPanel',
 ]
+
+if DEBUG:
+    INSTALLED_APPS += (
+        'django.contrib.admindocs',
+    )
--- a/hourglass/urls.py
+++ b/hourglass/urls.py
@@ -40,7 +40,9 @@ tests_url = url(r'^tests/$', TemplateView.as_view(template_name='tests.html'),
 if settings.DEBUG:
     import debug_toolbar
 
-    urlpatterns += [
+    urlpatterns = [
+        url(r'^admin/doc/', include('django.contrib.admindocs.urls')),
+    ] + urlpatterns + [
         url(r'^__debug__/', include(debug_toolbar.urls)),
         tests_url,
     ]


toolness:
FWIW, the main thing that made me think of adding this was @ericronne's recent contributions to our template code in #1440. I thought that if we have this documentation enabled (and styled a bit better), it might be easier for him to figure stuff out, but I dunno.

Even for us PRO DEVS, though, I do like the **Views** documentation in particular, as I don't think there's any other way to get a bird's-eye view of what all the routes on our site are.

toolness:
I think I might take a gander at the source code for `django.contrib.admindocs` and see how hard it'd be to reuse its existing code for inclusion in the style guide--well, at least for the template tag stuff--rather than a weird thing that requires admin access to view.

jseppi:
That would be awesome!

toolness:
Ugh, unfortunately the admindocs were not really made to do that... I tried it and it'd basically amount to forking the django app, which is a big ol' bummer.  Rigid class hierarchies ftl!

jseppi:
Bummer :\ I'm not sure it'll end up being that useful, mostly because it will be hard to find. I guess you could add a link to it in the Style Guide?

toolness:
So there have been a handful of times I've wished I had access to this myself; in particular, the view that lists all the routes on the app, and the view that lists all the templatetags, were things I wanted access to.

I guess if it is not actively *confusing* to our devs, it might be nice to merge this.  Whaddya think @jseppi ?

toolness:
Wooooot

Issue #960, opened 2016-10-24
Commas in search terms may be ambiguous


toolness:
It occurred to me when analyzing the logs that some search terms returned by the autocomplete functionality have commas in them, e.g. `Estimator/Scheduler, Jr` and `Truck Driver, Heavy Truck`.  When these queries are passed to the API, the API treats their commas as search term delimiters rather than part of the actual search terms.

For example:
1. Type `truck` into the search field.
2. One of the entries in the autocomplete list should be `Truck Driver, Heavy Truck`. Select this and press <kbd>Enter</kbd>.
3. This triggers a [search for `Truck Driver, Heavy Truck`](https://calc.gsa.gov/?sort=current_price&q=Truck+Driver%2C+Heavy+Truck%2C+&query_type=match_all&min_experience=0&max_experience=45&contract-year=current&experience_range=0%2C45). This means CALC is searching for entries containing `Truck Driver` _or_ `Heavy Truck`, and as a result it's including entries labeled `Truck Driver, Light Truck` in the results.

One way around this might be to remove commas from autocomplete results, or perhaps strip out the commas when the user presses <kbd>Enter</kbd> to add the autocomplete result to the search terms.

jseppi:
The culprit, in (api/views.py)[https://github.com/18F/calc/blob/develop/api/views.py#L76]: 
`qs = [s for s in query.split(',') if s.strip()]`

jseppi:
Kinda seems like we should use a different separator character (like `|` maybe), or we could quote individual terms.

A different separator character would break links unfortunately.

toolness:
I don't suppose there's a consistent way to normalize labor categories (via the normalization in #1378) so that they don't include commas?  For instance, if it's always in the format "X, Y" then we could normalize it to "Y X" so that "Engineer, Junior" becomes "Junior Engineer".  I dunno.

jseppi:
yeah, or even "Engineer (Junior)"

jseppi:
Something else that we could use to "version" the URLs would be to still support the legacy `q` param, but change the app to make all new URLs with something else, like `query` or `terms`. Then we could keep the old behavior when `q` is present, but do something more robust with the new param.

toolness:
[...]

PR #1454, opened 2017-03-01
[Experiment] React-based tokenized autocomplete for main search input

jseppi:
Here's the new search box, but expanded to the full width of the page content area:
screen shot 2017-03-01 at 4 26 29 pm

Here's how it would look if the width is constrained and there are more tokens present than will fit within that width:
screen shot 2017-03-01 at 4 35 17 pm

Obviously something needs to happen with the radios and with the "Search" and "Clear" buttons in either case. I think the "Clear" button should be disassociated from the labor category search input because it actually clears everything (including all filters), not just that search input.

To Do:
- [ ] visual design
- [ ] allow free-form text token entry


jseppi:
@ericronne what kind of design adjustments do you think would work for this change?

ericronne:
Think it's testable in its current form?

jseppi:
Yeah, you could get a sense of it in its current form

ericronne:
@tram would you recommend that we do some testing on this before continuing with design? My gut is that the "tag" motif is intuitive and perhaps familiar to our users (CO's use pinterest, right? :) ), but i'd be curious to see how folks interact with it in this context.

cc @kellybailey

jseppi:
I bet it is more intuitive than the comma-delimited thing we have going on now!

jseppi:
I've been working also on a fix for #960 in here, but I think I am going to extract it and finish it up in a separate PR that will work on the existing Data Explorer.

jseppi:
also maybe check out https://github.com/i-like-robots/react-tags or https://github.com/olahol/react-tagsinput

jseppi:
[As a user I should not have confusion with labor category search tokens due to commas](https://trello.com/c/GUESfUPN/243-as-a-user-i-should-not-have-confusion-with-labor-category-search-tokens-due-to-commas)

jseppi:
[As a user I should be able to search for a labor category with a comma in it (ex: Engineer, Junior)](https://trello.com/c/ncCHRIFd/245-as-a-user-i-should-be-able-to-search-for-a-labor-category-with-a-comma-in-it-ex-engineer-junior)

ericronne:
@tram would you consider this design good enough for you to put in front of folks in nyc? @hbillings and i did some style adjustments. Not perfect, but i'm hesitant to spend too much styling before testing in some fashion...

haps8wuyx1

hbillings:
I want to note here that it seems like with this library, you can only select something from the autocomplete list (if you don't select and you just hit enter, for instance, you'll automatically get the first thing in the list). This could be really confusing for folks who don't realize they've selected something other than what they entered (I typed in "engineer," hit enter, and the search term it populated was "senior engineer"). Just wanted to surface that somewhere!

jseppi:
Good catch, @hbillings. I would suspect there's a setting for allowing free-form text tokens instead of only auto-completed tokens.

toolness:
@jseppi i took an alternate approach in #1717... someone mentioned during coworking that only allowing users to enter items from the suggestion list could be considered a feature rather than a bug, since it doesn't make sense to search for something that there are no autocomplete results for.  so in #1717 i made the API results work in a way that supported that approach.

That said, now that I think about it, since we're showing the _normalized_ labor category in the autocomplete list, there may be cases where someone looking for an _exact match_ might not find their labor category in the autocomplete list.  Eh, anyways, we can talk about it later.

toolness:
I haven't tried this PR out on a screen reader, but for what it's worth, gov.uk just published a blog post called [What we learned from getting our autocomplete tested for accessibility](https://accessibility.blog.gov.uk/2018/05/15/what-we-learned-from-getting-our-autocomplete-tested-for-accessibility/).

ericronne:
Insufficient evidence that users would benefit :)

PR #1455, opened 2017-03-02
Make top-level styleguide headings sticky

toolness:
Yesterday I found out about [`position: sticky`](https://developer.mozilla.org/en-US/docs/Web/CSS/position#Sticky_positioning) and thought it might be useful for our styleguide, especially in sections that are really long, where one might forget what top-level section they're in or something.

This CSS feature seems to gracefully degrade nicely since browsers that don't support sticky positioning will just show the headings like they've always looked.

### DIFF
--- a/styleguide/static/styleguide/styleguide.css
+++ b/styleguide/static/styleguide/styleguide.css
@@ -9,6 +9,10 @@
 
 h3.styleguide-section {
   border-bottom: 1px solid lightgray;
+  position: sticky;
+  top: 0;
+  background: white;
+  z-index: 100;
 }
 
 .styleguide-section a {


jseppi:
hrm, they look kinda of funky to me:

screen shot 2017-03-02 at 4 34 41 pm

And when moused-over the "#" makes things jump around in a weird way.

screen shot 2017-03-02 at 4 34 44 pm

I assume since it is CSS there is not a way to know when a heading has been stickified so that you could maybe set the width to the full page width or something?

toolness:
Weird, what browser are you using? I'm not getting those bugs on Firefox 51 or Chrome 56...

jseppi:
Chrome 56 on OSX o_O

toolness:
Oh weird, I think this is an OS X vs. Windows bug--just tried it on OS X and I'm able to reproduce your problem.  The [Chromium `position: sticky` meta bug](https://bugs.chromium.org/p/chromium/issues/detail?id=231752) does mention that "there are several known bugs which have been discovered in 56, most of which are resolved or targeted to be resolved in 57", and I bet this is one of them--in fact I just tried out the styleguide on [Chrome Canary](https://www.google.com/chrome/browser/canary.html) on OSX, which is currently tracking version 58, and the problem is gone.

It looks like Chrome 57 will have a stable release around the middle of March, so, uh, I guess we could sit on this PR for a few weeks, or something?  Blerg.

toolness:
Oh interesting!  I guess that is why some examples I saw set `top: -1px` instead of `top: 0`.  I hadn't noticed it before but can fixit.

PR #1458, opened 2017-03-02
Ignore roman numerals in PL analysis

### LINKED ISSUE #1438: Ignore roman numerals in price list search
From Kelly:

> Is there a way to ignore any levels after the labor category? I think you had it fixed for some labor categories, but some of the other labor categories were including them. I think you got it to ignore numbers, but I think some include roman numerals.

I guess we could do this by adding a list of "stop words" to the analysis and stripping roman numerals.

Should this apply to the data explorer too, though?  i.e. should a search for "engineer II" actually just show results for "engineer"?

toolness:
Fixes #1438.

### DIFF
--- a/data_capture/tests/test_analysis.py
+++ b/data_capture/tests/test_analysis.py
@@ -18,6 +18,7 @@ from ..analysis.core import (
 from ..analysis.vocabulary import (
     Vocabulary,
     get_best_permutations,
+    broaden_query,
 )
 from ..analysis.export import AnalysisExport, COMPARABLES_NOT_FOUND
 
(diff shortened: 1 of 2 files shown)


data_capture/tests/test_analysis.py
```diff
@@ -27,13 +28,13 @@ class BaseDbTestCase(DjangoTestCase):
     def setUpTestData(cls):
         cls.contract = get_contract_recipe()
         cls.contract.make(
-            labor_category="Engineer of Doom II",
+            labor_category="Engineer of Doom ZZ",
```
jseppi: ha, maybe just leave off the `ZZ` for these tests? Or is having it there testing something that I'm missing?

toolness: eh, some of the tests specifically expect there to be a 2-character token in that labor category, so it was easiest to just do a find/replace to make sure it wasn't a stop word.

Issue #1448, opened 2017-03-01
Improve test coverage of education-level.jsx


toolness:
Running the following command:

```
docker-compose run app jest frontend/source/js/data-explorer/tests/education-level.test.jsx \
  --coverage \
  --collectCoverageFrom='["frontend/source/js/data-explorer/components/education-level.jsx"]'
```

produces the following coverage output:

```
---------------------|----------|----------|----------|----------|----------------|
File                 |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
---------------------|----------|----------|----------|----------|----------------|
All files            |    60.53 |       30 |    58.33 |    60.53 |                |
 education-level.jsx |    60.53 |       30 |    58.33 |    60.53 |... 2,80,81,169 |
---------------------|----------|----------|----------|----------|----------------|
```

It would be nice if we could improve on this. Specifically, the event handers `handleDocumentClick()` and  `handleToggleMenu()` don't seem to be exercised by the test suite.

(Coverage on line 169 isn't really a big deal since it's something that will be trivially executed by the app when it first loads. Whatevvvvvvs.)

@hbillings want to give this a shot?

PR #1462, opened 2017-03-03
Add more React tests

hbillings:
Addresses #1448 partially. Critique very welcome! Still wrestling with `handleDocumentClick`. 
image


frontend/source/js/data-explorer/tests/education-level.test.jsx
```diff
@@ -54,4 +54,16 @@ describe('<EducationLevel>', () => {
...
   });
+
+  it('fires handleToggleMenu when dropdown is clicked', () => {
+    const { wrapper } = setup();
+    const e = {
+      preventDefault: jest.fn()
+    }
+    wrapper.state().expanded = false;
+    expect(wrapper.state().expanded).toBeFalsy();
+    wrapper.find('a').simulate('click', e);
```
toolness: Huh, I would have thought that the `simulate` method passes its own fake-ish event object that has `preventDefault` already mocked out, but I'm unfamiliar with enzyme myself. Does this code explode if you remove `e` as a parameter to `simulate` here?

hbillings: Yeah, it would complain that `preventDefault was not defined. :-(

frontend/source/js/data-explorer/tests/education-level.test.jsx
```diff
@@ -54,4 +54,16 @@ describe('<EducationLevel>', () => {
     expect(props.toggleEducationLevel.mock.calls.length).toBe(1);
     expect(props.toggleEducationLevel.mock.calls[0][0]).toBe('HS');
   });
+
+  it('fires handleToggleMenu when dropdown is clicked', () => {
+    const { wrapper } = setup();
+    const e = {
+      preventDefault: jest.fn()
+    }
+    wrapper.state().expanded = false;
```
toolness: Hmm, I think that the state of `expanded` should already be falsy here? Does the test still work if you remove this line?

If not, it might be better to do `wrapper.setState({ expanded: false })`, as that will trigger any funky React lifecycle methods that need to be called.  I'm not sure if directly setting the property on the state as you're doing here will do that, but again, I'm not super familiar with the enzyme!

toolness:
This is looking good! 

I should have thought about the handleDocumentClick() business before suggesting it... testing it might be challenging because it's such an unusual thing for a React element to do (normally React elements only attach event handlers to the DOM nodes they're responsible for rendering).  Blugh.

I can think of maybe two potential ways to bust this up:

1. Calling `handleDocumentClick` directly on the component, passing it a fake event object. I think this is what you were originally trying to do and I may have foolishly dissuaded you from doing it or something.  My bad 😞 

2. You could mayyyyybe dispatch a fake click event to the JSDOM document that the component is mounted on but I have no idea if that's even possible. Blerg.

toolness:
Anyhow if the `handleDocumentClick` test proves too annoying we can always just merge just the `handleToggleMenu` test you've got, it will still improve test coverage!

frontend/source/js/data-explorer/tests/education-level.test.jsx
jseppi: Does this work https://github.com/airbnb/enzyme/issues/323#issuecomment-210039710 ?

```js
wrapper.find('a').simulate('click', { preventDefault() {} });
```

frontend/source/js/data-explorer/tests/education-level.test.jsx
toolness: Oh interesting! I wonder if that's actually a bug in our component, then... I would think that `expanded` ought to be set to `false` when they component is first created. Funky.

hbillings: Okay, so turns out it _is_ set to `false` in the component; I was trying to access state incorrectly when I thought it was undefined. Womp womp. I took that bit out!

hbillings:
I tried to do something like this for `handleDocumentClick`:
```

  it('fires handleDocumentClick to hide the dropdown when a user clicks outside of it', () => {
    const { wrapper } = setup();
    const e = {
      target: document.createElement("div"),
      preventDefault: jest.fn()
    };
    wrapper.find('a').simulate('click', e);
    expect(wrapper.state().expanded).toBeFalsy();
  });
```

Obviously this doesn't work because there isn't a `document`. I tried to figure out how to get jsdom into this to fake the document, but this set up is complex enough that I can't make it work. :-/

frontend/source/js/data-explorer/tests/education-level.test.jsx
```diff
@@ -54,4 +54,15 @@ describe('<EducationLevel>', () => {
     expect(props.toggleEducationLevel.mock.calls.length).toBe(1);
     expect(props.toggleEducationLevel.mock.calls[0][0]).toBe('HS');
   });
+
+  it('fires handleToggleMenu when dropdown is clicked', () => {
+    const { wrapper } = setup();
+    const e = {
+      preventDefault: jest.fn()
+    }
+    console.log(wrapper.state().expanded);
```
toolness: Just remove this spurious `log` and we can merge!

PR #1463, opened 2017-03-03
Switch to using yarn for node packages

### LINKED ISSUE #1345: Pin all node dependencies

### LINKED ISSUE #1427: Evaluate Yarn for managing node deps
Might alleviate #1345!

We can probably store the yarn lockfile in our node_modules docker volume.

jseppi:
Fixes #1427 
Fixes #1345 

This PR replaces our use of `npm install` with `yarn`, which has several benefits, the most important being locked node dependencies (via the `yarn.lock` file).

To Do:
- [x] update `frontend.md` and other docs
- [x] pull in changes from `develop` to fix front end tests that @toolness fixed in https://github.com/18F/calc/pull/1378 after it is merged


jseppi:
Ready for review, @toolness!

docs/docker.md
```diff
@@ -63,8 +63,8 @@ Whenever you update your repository via e.g. `git pull` or
 
 Feel free to install custom dependencies, e.g. your favorite
 debugging library, in your container via
-`docker-compose run app pip install` or 
-`docker-compose run app npm install`. Everything should work
+`docker-compose run app pip install` or
+`docker-compose run app yarn add`. Everything should work
 as expected.
```
toolness: Huh, so one thing I just noticed about `yarn add` is that it effectively seems to behave like `npm install --save`, i.e. it always modifies the `package.json`.  This is a bit different from the original intent of this paragraph, where it was assumed that `pip install` and `npm install` would add custom dependencies that the developer could use without having to worry about accidentally committing them as dependencies to the repository.

I wonder if we should recommend they use `yarn add --dev` or `yarn add --optional` or something?  Or is it still possible to even use `npm install` to install those kinds of "hey i'm just adding this for my personal  use, I don't want to inflict it on others" modules/tools?

Alternatively we could just ignore this quibble, it's probably unlikely folks will want to install custom node packages anyways.

toolness:
Hmm, I just noticed that [Yarn stores every package in a global cache in your user directory on the file system](https://yarnpkg.com/lang/en/docs/cli/cache/).

It might be nice to actually have this cache be stored in a Docker volume too, as I bet it'd make e.g. switching between different branches that have different `package.json`s a lot faster.

Hmm, yeah, just now I tried adding a new dependency to `package.json`, and re-running `yarn` took ~18 seconds... it seemed like it was re-downloading all packages.

In contrast, if I change `package.json` and re-run yarn *within the same container session* (e.g. within a single `docker-compose run app bash` invocation), the re-run only takes about 4 seconds.  So it'd be really nice if we could keep yarn's cache directory around between `docker-compose` invocations!

This is going to be a little problematic though because of the fact that the user's home directory doesn't actually exist until `docker_django_management.py` creates it, though... Blerggg.  Will have to think about this a bit--for now though I think it's OK to deal with this issue at a later time.

.travis.yml
```diff
@@ -18,8 +17,8 @@ before_install:
 - node --version
 install:
 - pip install -r requirements-dev.txt
-- npm install -g gulp-cli
-- npm install
+- npm install -g gulp-cli yarn@0.21.3
+- yarn --pure-lockfile
```
toolness: Why are we telling `yarn` not to generate a `yarn.lock` lockfile here?  Under what conditions would it actually want to change the one we've got in the repo?

frontend/source/js/data-explorer/tests/__snapshots__/experience.test.jsx.snap
```diff
@@ -21,7 +21,9 @@ exports[`<Experience> matches snapshot 1`] = `
         onChange={[Function]}
         value={5}>
         <option
-          value={0} />
+          value={0}>
+          0
+        </option>
```
toolness: What's up with this? Did the latest version of react or jest or enzyme just change in a way that broke this snapshot or something?

.travis.yml
jseppi: `yarn.lock` should be checked into the repo. In the weird case that it gets deleted or isn't included for some other reason, we don't want Travis to generate it because behavior wouldn't have been "manually verified" during dev. That's my rationale anyway.

docs/docker.md
jseppi: Ah yeah, `yarn add` will add to the lockfile. It doesn't look like either the `--dev` or `--optional` flags, or any others, will do "temporary" adds like npm install.

Hrm...perhaps the intent of this subsection isn't very clear since I didn't actually remember it was only for experimental dependencies.

frontend/source/js/data-explorer/tests/__snapshots__/experience.test.jsx.snap
jseppi: Yup, it was a bugfix in `enzyme-to-json`: https://github.com/adriantoine/enzyme-to-json/pull/45

Issue #1346, opened 2017-02-08
Upgrade to Django 1.11


toolness:
Since ~~[Django 1.11 alpha 1](https://www.djangoproject.com/weblog/2017/jan/17/django-111-alpha-1/)~~ ~~[Django 1.11 beta 1](https://www.djangoproject.com/weblog/2017/feb/20/django-111-beta-1-released/)~~ [Django 1.11 was released](https://www.djangoproject.com/weblog/2017/apr/04/django-111-released/) a few weeks ago, and since it's the next long-term-support (LTS) release after Django 1.8, *and* since Django 1.8 LTS will be end-of-life'd in a year, I thought it might be useful to start planning a migration to Django 1.11.

It looks like it will require at least the following:

- [x] ~~Either migrating our full-text search to use the native full-text search introduced in Django 1.10 *or*~~ forking djorm_pgfulltext to work with newer versions of Django, as discussed in #291.
- [ ] Ensuring our custom widgets in the `frontend` package are compatible with Django 1.11's new [template-based widget rendering](https://docs.djangoproject.com/en/1.11/releases/1.11/#template-based-widget-rendering).

toolness:
Oh hey, this was fixed by #1640.

PR #1417, opened 2017-02-22
Fixes to admin site

hbillings:
Closes #1326...
screen shot 2017-02-22 at 8 53 30 am

#1325 - This problem existed in the user/group search bar and dropdown group also, which now looks like this:
screen shot 2017-02-27 at 12 26 20 pm

#1262 - looks like this now:
screen shot 2017-02-27 at 2 36 22 pm

PR #1466, opened 2017-03-03
Upgrade to Django 1.9

toolness:
This uses [our new fork of `djorm-ext-pgfulltext`](https://github.com/18F/djorm-ext-pgfulltext), which hopefully fixes things so that it works on Django 1.9.

Note that we'll likely want to quickly move on to Django 1.10 after this, as [1.9 will only be supported through April 2017](https://www.djangoproject.com/download/#supported-versions). Ideally we should even migrate past that to the latest Django 1.11 beta, since Jacob said that beta releases are pretty solid and 1.11 is the next LTS after 1.8. (The only real barrier to that is the extent to which our third-party packages support 1.11, but I suppose this should all be discussed in #1346.)

## Notes

* To deal with Django 1.9's new `DecimalField` validation, I changed things so that we truncate prices to two decimal places when cleaning models.  I *think* this should be fine as long as price lists don't include fractional cent values or something.

* The relevant changes I had to make to our `djorm-ext-pgfulltext` fork are:
  * https://github.com/18F/djorm-ext-pgfulltext/commit/58c1c5bbcad0058ea0c1c69c804120cd4c5ef347 - This is basically a fix for #291.
  * https://github.com/18F/djorm-ext-pgfulltext/commit/1ad8bddf06201e3c89b1e4a483f819297710ec6a - This fixes a selenium test that was breaking.

## To do

- [x] Fix `django.core.exceptions.ValidationError` for some of our `Contract` `DecimalField`s (due  to [Django bug 24636](https://code.djangoproject.com/ticket/24636)).

- [x] Fix `py.test data_capture/tests/test_admin.py`.
- [x] Fix `py.test styleguide/tests.py`.
- [x] Fix `py.test frontend/tests/test_selenium.py -k test_query_type_matches_words`. Looks like there's something else in `djorm-ext-pgfulltext` we need to fix:

  ```pytb
  File "/venv/lib/python3.6/site-packages/djorm_pgfulltext/models.py", line 315, in search
    "%s('%s', %s)" % (function, config, adapt(query))
  File "/venv/lib/python3.6/site-packages/djorm_pgfulltext/utils.py", line 9, in adapt
    a.prepare(connection.connection)
  TypeError: argument 1 must be psycopg2.extensions.connection, not None
  ```

- [x] Make sure there aren't any horrible visual regressions in the admin part of the site, since we're overriding Django's default admin CSS.

- [x] Fix anything else not covered by tests that appears to be broken.

- [x] Deal with any `RemovedInDjango110Warning`s.


toolness:
For future reference, to develop on our fork of `djorm-ext-pgfulltext` while also integrating it with CALC to make sure my changes worked, I did the following:

1. I ran `git clone https://github.com/18F/djorm-ext-pgfulltext` in the directory immediately *above* my CALC checkout (i.e., I did `cd ..` from the root of the repo).  I didn't want to clone it in the CALC directory because that would've made all our linters, py.test, etc delve into that dir and raise false positives.

2. I edited my `docker-compose.local.yml` like so:

```diff
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -3,6 +3,7 @@ services:
   app:
     volumes:
       - .:/calc
+      - "../djorm-ext-pgfulltext:/djorm-ext-pgfulltext"
     command: python manage.py runserver 0.0.0.0:${DOCKER_EXPOSED_PORT}
     environment:
       - DEFAULT_DEBUG_EMAIL_URL=smtp://mailcatcher:25/
```

This will mount `djorm-ext-pgfulltext`'s repo at `/djorm-ext-pgfulltext` in the `app` container.

3. I replaced the line for `djorm-ext-pgfulltext` in `requirements.txt` with:
  
```
-e /djorm-ext-pgfulltext
```

This will make `pip` install the directory in edit mode, so that any changes I make to the `djorm-ext-pgfulltext` module immediately take effect, without me having to re-run `pip` or anything.

4. I ran `pip install -r requirements.txt` for the above changes to take effect (alternatively, I could've run `./docker-update.sh`).

toolness:
Ok, I think this is good to go, though I think @hbillings should give the admin UI a once-over to make sure nothing is super broken there.  I did notice a few wonky things but I think they may be present in `develop` and fixed by the incoming #1417.

From my cursory glance it *seems* like the Django 1.9 admin UI overhaul was one that just changed the CSS rule definitions, rather than changing the actual class names or anything, but I could be wrong. It might be a good thing to go over during coworking.

jseppi:
> as long as price lists don't include fractional cent values or something

Hrm, actually I think they _can_ be fractional since often the price cells are calculated from other fields, so they end up actually having values like "50.08725"

jseppi:
Found at least one regression in the admin site:

screen shot 2017-03-14 at 11 25 25 am

From `/admin/auth/user/1/change/` but I assume it will be like that on all model `change` views

hbillings:
Filed #1492 to merge develop and fix a few things.

toolness:
Cool @jseppi!  Do you think the `CashField` is taking care of the decimal issue?  All it's really doing is ensuring that the values we ultimately end up writing to the db only have two decimal places, but intermediate values can be floats...

Just merged in some of @hbillings' CSS fixes for the admin, so barring the `CashField` stuff, I guess this is ready to merge?

PR #1467, opened 2017-03-03
[WIP] Support quote-delimited query strings parts

jseppi:
To Do:
- [x] adjust api to parse csv-like strings
- [x] tests for api adjustment
- [ ] adjust React app to handle csv-like labor category query strings
- [ ] more/better tests for `<DescriptionLaborCat>`

Issue #1459, opened 2017-03-02
Commas in labor category search terms produce weird results


jseppi:
Related: #960 

Searching for "Civil Engineer, Senior" (which populates from the autocomplete):

<img width="589" alt="screen shot 2017-03-02 at 9 49 09 am" src="https://cloud.githubusercontent.com/assets/697848/23514637/85b9ba96-ff2d-11e6-8ee9-9d5d797ce63d.png">

yields results like:

<img width="346" alt="screen shot 2017-03-02 at 9 49 13 am" src="https://cloud.githubusercontent.com/assets/697848/23514648/8d5fdd66-ff2d-11e6-8bb9-c24cdaf3d3f7.png">

jseppi:
![](https://github.trello.services/images/mini-trello-icon.png) [As a user I should be able to search for a labor category with a comma in it (ex: Engineer, Junior)](https://trello.com/c/ncCHRIFd/245-as-a-user-i-should-be-able-to-search-for-a-labor-category-with-a-comma-in-it-ex-engineer-junior)

PR #1468, opened 2017-03-03
Add API support for quote-delimited labor category param (`q`)

jseppi:
Pulled this from #1467.

This adjusts up our api to parse the `q` param in "csv-style," i.e. allowing quotation-mark-delimited individual terms within the comma-separated list. For example, the api can now handle a query like `q="Driver, Heavy Truck", Engineer` as two distinct labor category terms: `"Driver, Heavy Truck"` and `"Engineer"`.

Ref #960 and #1459

### DIFF
--- a/api/views.py
+++ b/api/views.py
@@ -1,3 +1,5 @@
+import csv
+
 from django.http import HttpResponse
 from django.db.models import Avg, Max, Min, Count, Q, StdDev
 from decimal import Decimal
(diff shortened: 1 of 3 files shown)


jseppi:
What do you think about this @toolness? It won't impact any existing front end behavior, but it sets things up for improvements there.

api/views.py
```diff
@@ -10,7 +12,26 @@
...
+        ['jane', 'jim', 'jacky', 'joe']
+
+        >>> parse_csv_style_string('carrot, beet  , sunchoke')
+        ['carrot', 'beet', 'sunchoke']
+
+        >>> parse_csv_style_string('turkey, "hog, wild", cow')
+        ['turkey', 'hog, wild', 'cow']
+    '''
+    reader = csv.reader([s], skipinitialspace=True)
+    return [qq.strip() for qq in list(reader)[0] if qq.strip()]
```
toolness: I like the re-use of the CSV reader here--how does one embed a quote inside double quotes with it?  Like, uh, if I wanted to search for a **"cloud" engineer**, with the quotes around the first word...  I guess it's probably a nonexistent use case, huh.

jseppi: Hrm, I bet it works with escaped quotation marks `"\"cloud\" engineer"`

Issue #1028, opened 2016-11-09
Rename (or remove) load_data command and associated CSV files


toolness:
I just realized that the `load_data` command, which basically loads region 10 bulk CSV data (I think?), has an extremely similar name to Django's built-in [`loaddata`](https://docs.djangoproject.com/en/1.10/ref/django-admin/#django-admin-loaddata) command. We might want to rename it to `load_r10` (short for "load region 10") or remove the command entirely if that's easier (as I *think* the functionality has been superseded by the region 10 bulk upload feature).

jseppi:
I am generally 👍  on removing `load_data` and `load_s70`. Though I could see a potential use case for pre-populating the initial database with the current csv data, I guess.

jseppi:
related (and actually also calls for removing these commands): https://github.com/18F/calc/issues/840

toolness:
Yeah, I do like the ability to instantly seed the db with a bunch of real (or at least real-ish, as time passes and the contracts go out of date) data.  We could potentially make a totally different command that uses `model_mommy` to seed the db with random data, but it might be easier to just leave `load_data` around, but rename it to `seed_db` or somesuch?

toolness:
How about this:

* We get rid of all files in `contracts/docs` except `s70_data.csv` and `hourly_prices.csv`. Alternatively, we could keep only the files that end in `_sample.csv`; they're much smaller.  But since the big files are already in the git repo it seems like we might as well use those, I dunno.
* Rename `contracts/docs` to `contracts/example_data`
* Rename the `load_data` management command to `load_example_r10_data`
* Rename the `load_data` management command to `load_example_r10_data`
* Update the documentation accordingly

The only alternative I can think of is to get rid of _all_ of it but replace it with a `seed_db` command that uses `model_mommy` to recreate the same kind of records, but that feels like a lot of work.

toolness:
Oh wait, I just thought of a different solution.  In #1714 I'm adding a `process_bulk_upload` command that is basically the same thing as `load_data`, only it works with XLSX files, but unlike `load_data`, it follows the exact same code path that's used during bulk upload (as its name implies).  So we could potentially just add Kelly's latest R10 XLSX file to the repo (or add it via Git LFS if we don't want to bloat the repo), which would then allow us to get rid of `load_data`, `load_s70` and all related code.

jseppi:
I like the idea of directing new folks to download the production data and load it into to their local dev instance, definitely more so than including the Excel file.

[...]

PR #1469, opened 2017-03-06
Add load_data test.

toolness:
I know we might be getting rid of the `load_data` command (see #1028), but since it's still around, I figured we might as well test it.  I noticed its code didn't seem to be called by the test suite when working on #1466.

This adds a simple test that runs the command on sample data, but I'm confused about something: there are 100 rows, and when running, the command says that it processed 100 rows and skipped none, yet only 79 `Contract` models have been added to the database. Why is that?

At first I thought it might be because some of the contracts might be expired, so perhaps those aren't being added to the db, but after using `freezegun` to alter the current date, it didn't seem to affect things.

@jseppi any ideas?

### DIFF
--- a/contracts/tests/test_load_data.py
+++ b/contracts/tests/test_load_data.py
@@ -1,0 +1,18 @@
+import pathlib
+from django.core.management import call_command
+from django.test import TestCase
+
+from contracts.models import Contract
+
+MY_DIR = pathlib.Path(__file__).resolve().parent
+
+
+class LoadS70TestCase(TestCase):
+    sample_filename = MY_DIR.parent / 'docs' / 'hourly_prices_sample.csv'
+
+    def test_loads_sample(self):
+        call_command(
+            'load_data',
+            filename=self.sample_filename
+        )
+        self.assertEquals(Contract.objects.count(), 79)


jseppi:
Huh, I don't know why that would happen :\

jseppi:
So, I agree that having this test is a good idea, but I think we should get to the bottom of why not all the contract models are created.

jseppi:
ok, just took another look at this. It looks like the rows that don't end up coming back are rows that have blank `CurrentYearPricing` in the source csv.

What's weird is that even individually calling `.save()` on all the contracts _appears_ like it is working, but then fetching the resultant contracts by `id` throws an exception:

```py
# file: contracts/management/commands/load_data.py
# I replaced Contract.objects.bulk_create(contracts) with this junk:

log.info("Inserting records")
for c in contracts:
    c.save()
    print(f"saved {c.id}")

print("--------------")

for c in contracts:
    try:
        c2 = Contract.objects.get(id=c.id)
        print(f"got {c2.id}")
    except:
        print(f"missed {c.id} - {c.idv_piid}")
```

jseppi:
aha! This `get_queryset` method is why: https://github.com/18F/calc/blob/develop/contracts/models.py#L121

```py
    def get_queryset(self):
        return ContractsQuerySet(self.model, using=self._db)\
            .filter(current_price__gt=0)\
            .exclude(current_price__isnull=True)
```

toolness:
Ahh cool thanks for looking into that sir!

PR #1472, opened 2017-03-06
Add yarn cache volume for faster docker yarning

### LINKED ISSUE #1471: Include yarn cache in docker volume
As @toolness noted in https://github.com/18F/calc/pull/1463#issuecomment-284418080:

> Hmm, I just noticed that [Yarn stores every package in a global cache in your user directory on the file system](https://yarnpkg.com/lang/en/docs/cli/cache/).
> 
> It might be nice to actually have this cache be stored in a Docker volume too, as I bet it'd make e.g. switching between different branches that have different `package.json`s a lot faster.
> 
> Hmm, yeah, just now I tried adding a new dependency to `package.json`, and re-running `yarn` took ~18 seconds... it seemed like it was re-downloading all packages.
> 
> In contrast, if I change `package.json` and re-run yarn *within the same container session* (e.g. within a single `docker-compose run app bash` invocation), the re-run only takes about 4 seconds.  So it'd be really nice if we could keep yarn's cache directory around between `docker-compose` invocations!
> 
> This is going to be a little problematic though because of the fact that the user's home directory doesn't actually exist until `docker_django_management.py` creates it, though... Blerggg.  Will have to think about this a bit--for now though I think it's OK to deal with this issue at a later time.
>

jseppi:
Closes #1471

To Do:
- [ ] instead of venv and a special yarn cache, just make `/home/calc_user` a docker volume /cc @toolness

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,6 +22,7 @@ RUN pip install virtualenv
 WORKDIR /calc
 
 RUN npm install -g yarn
+RUN yarn config set cache-folder /calc/.yarn-cache
 
 ENV PATH /calc/node_modules/.bin:$PATH
 ENV DDM_IS_RUNNING_IN_DOCKER yup
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,3 +38,4 @@ services:
 volumes:
   node-modules:
   python-venv:
+  yarn-cache:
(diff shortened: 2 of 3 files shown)


jseppi:
How's this look, @toolness? Feel free to push changes to this branch since I'm a little out of my element with the docker stuff.

toolness:
Oops, I ended up filing #1473... I think it might be a bit nicer because it doesn't require changing yarn's default path, and it will also work with `pip`'s caching mechanism, and likely any other future tools we decide to use that put cached stuff in the user's home directory...

jseppi:
Sweet, no worries

PR #1473, opened 2017-03-06
Store /home in a Docker volume

### LINKED ISSUE #1471: Include yarn cache in docker volume

toolness:
This fixes #1471 by storing `/home` in a Docker volume. It also means that `pip`'s cache should be saved across `docker-compose` runs too, which is nice.

I made `/home` a docker volume instead of `/home/calc_user` because we actually want `useradd` to create the actual home directory in `docker_django_management.py`, as it will automatically copy the skeleton files into it (e.g. `.profile`, `.bashrc`), which can be helpful.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -38,3 +38,4 @@ services:
 volumes:
   node-modules:
   python-venv:
+  home:
--- a/docker-services.yml
+++ b/docker-services.yml
@@ -10,6 +10,7 @@ services:
       # other non-Linux native code compiled on the Docker Host.
       - python-venv:/venv/
       - node-modules:/calc/node_modules/
+      - home:/home/
     environment:
       - DDM_VENV_DIR=/venv
       - DDM_USER_OWNED_DIRS=/venv:/calc/node_modules
--- a/docker_django_management.py
+++ b/docker_django_management.py
@@ -25,6 +25,11 @@
 
     * Friendly warnings are logged if anything seems to be misconfigured.
 
+    Requirements:
+
+    * The docker container this runs in should support the
+      `useradd` command.
+
     To use it:
 
     * Place this script in the same directory as your Django project's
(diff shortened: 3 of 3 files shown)


jseppi:
Hrm, I'm not seeing a yarn cache when I `docker-compose run app bash` into the the container. Are you?

On my real machine, I have `~/.yarn-cache/` which is populated with a bunch of packages.

When I `bash` into the `app` container, I do see `~/.cache/yarn/`, but it is essentially empty.

jseppi:
I also just found there is a command to look at the cache contents: `yarn cache ls`. It also reports the `app` container's cache being empty:

screen shot 2017-03-06 at 2 37 54 pm

jseppi:
aha! It was because my `node_modules/` volume was already populated, so yarn didn't actually have to install anything. I manually cleared my `node_modules/` and now the container's yarn cache is populated 🎉

PR #1478, opened 2017-03-07
Use querystring module in place of custom functions

jseppi:
Replaces our custom `parseQuery` and `joinQuery` with node's `querystring` module methods `parse` and `stringify`.

### DIFF
--- a/frontend/source/js/data-explorer/components/export-data.jsx
+++ b/frontend/source/js/data-explorer/components/export-data.jsx
@@ -1,7 +1,7 @@
+import * as qs from 'querystring';
+
 import React from 'react';
 import { connect } from 'react-redux';
-
-import { joinQuery } from '../util';
 
 import { getRatesParameters } from '../rates-request';
 
@@ -25,7 +25,7 @@ ExportData.propTypes = {
 
 function mapStateToProps(state) {
   return {
-    querystring: joinQuery(getRatesParameters(state)),
+    querystring: `?${qs.stringify(getRatesParameters(state))}`,
   };
 }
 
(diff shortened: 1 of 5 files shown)


toolness:
This is cool, my one concern though is how multiple querystring args with the same name are handled. `util.parseQuery()` ensures that only one of them is used, while [`querystring.parse()`](https://nodejs.org/api/querystring.html#querystring_querystring_parse_str_sep_eq_options) turns the value into an array of strings, which I don't *think* our code is built to handle right now?

It's not that big a deal because we never intentionally use multiple querystring args with the same name, but I guess it could be used to make our app crash in potentially confusing ways.

jseppi:
Hrm, good catch! I just manually checked duplicating most of our query string params and the app doesn't crash or error, it just ignores any of the params that get array-ified due to duplicate params.

We could guard against the array-ified params by either deleting them or picking the first one in the parsed `qsFields`.

jseppi:
```js
   // delete any array-ified fields
    Object.keys(qsFields).forEach((k) => {
      if (typeof qsFields[k] !== 'string') {
        delete qsFields[k];
      }
    });
```
or
```js
    // get first value from array-ified fields
    Object.keys(qsFields).forEach((k) => {
      if (Array.isArray(qsFields[k])) {
        qsFields[k] = qsFields[k].length ? qsFields[k][0] : null;
      }
    });
```

jseppi:
or not worry about it since this case would just happen from the user doing weird stuff

toolness:
Yeah, I like the "get first value from array-ified fields" option!  Do u mind busting that up.  I know it's highly unlikely that it will happen but I figure it also protects us just in case our *own* code accidentally generates URLs with multiple querystring args... I dunno.

I kinda wish the node API was consistent about it and *always* set the values to arrays, even if they only had one entry, rather than having them be strings or arrays.

jseppi:
ok d00d, there you go

PR #1479, opened 2017-03-08
Switch to react-based slider

jseppi:
This PR removes our vendor'd, old version of noUiSlider in favor of a pure React slider component.

I also got rid of `eyeglass`, which I previously thought would be a good solution to using sass from `npm install`'d modules due to a reference to it in the `bourbon-neat` [docs](https://www.npmjs.com/package/bourbon-neat#installing-with-npm-and-using-a-node-based-asset-pipeline).  Instead, I've just changed our `gulpfile.js` to directly include bourbon-neat's `includePaths` and our `node_modules` in sass's `includePaths`.

Benefits:
- easier testing
- code is easier to understand
- removes unmaintained noUiSlider stuff

Currently looks like:

screen shot 2017-03-08 at 12 31 46 pm

screen shot 2017-03-08 at 1 25 29 pm

To do:
- [x] style more like the old one, maybe
- [x] either un-vendor the `rc-slider.css` or include proper attribution at the top of it


jseppi:
@ericronne any styling desires from you on this one, or should I just make it look like the current production one?

The current slider in production looks like this:
screen shot 2017-03-08 at 11 52 00 am

The default styling for this new slider implementation is:
screen shot 2017-03-08 at 11 51 09 am

ericronne:
something along these lines?
(from my wip)

image

jseppi:
Ok, updated the styling to look like:

screen shot 2017-03-08 at 12 31 46 pm

How's that, @ericronne?

jseppi:
We'll definitely be able to make it look like the new design/colors when we make that switch.

jseppi:
Alright, this is ready for review, @toolness

ericronne:
better, more accessible color, thx

ericronne:
Looks good. Found two minor bugs, but it looks like they aren't new! We can address later, in another issue:
- The number fields seem to be resizing slightly when you start playing with the widget, which makes everything below it shift downward a couple of pixels
- The arrows are too close to the numbers

PR #1480, opened 2017-03-08
Use d3 v4 npm modules instead of global d3

jseppi:
This PR gets rid of our global, vendor'd d3.min.js in favor of separately installed d3 modules.

 d3 v4 switched to using individual modules, so it was relatively straightforward to get this working. There were a few minor d3 api changes that I had to make to get everything working, but they were fairly minimal for our usage. For a list of all d3 v3 -> v4 changes, check out https://github.com/d3/d3/blob/master/CHANGES.md.

This change brings our `data-explorer/index.min.js` size down from `357K` to `270K`.


frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -1,15 +1,28 @@
...
+import { extent as _extent } from 'd3-array';
+
 import {
   templatize,
   formatCommas,
   formatPrice,
   formatFriendlyPrice,
 } from '../util';
 
+const d3 = {
```
jseppi: Might want to put this in a separate local module, called "customd3" or something, especially if we end up using d3 elsewhare. I think it is fine here for now though.

data_explorer/templates/base.html
```diff
@@ -75,9 +75,7 @@
       <![if gt IE 8]>
       <script {{ csp_nonce }}>
         if (typeof aight === 'object') {
-          d3.select('body')
-            .classed('ie', true)
-            .classed('ie' + aight.browser.ie, true);
+          document.body.className += ' ie ie' + aight.browser.ie;
```
jseppi: didn't use `classList` since the whole point of this is for old versions of IE.

frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -143,10 +156,8 @@ function updateHistogram(rootEl, data, proposedPrice, showTransition) {
         .attr('class', d => `chart-label ${d.type}`);
     stdDevLabels.append('line')
       .attr('class', 'label-rule')
-      .attr({
```
jseppi: this way of calling `.attr` is gone in d3 v4

frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -349,8 +362,7 @@ function updateHistogram(rootEl, data, proposedPrice, showTransition) {
   const ticks = bins.map(d => d.min)
     .concat([data.maximum]);
 
-  const xa = d3.svg.axis()
-    .orient('bottom')
+  const xa = d3.axisBottom()
```
jseppi: minor d3 v4 api change

frontend/source/js/data-explorer/components/histogram.jsx
```diff
@@ -379,9 +391,8 @@ function updateHistogram(rootEl, data, proposedPrice, showTransition) {
     .text('Ceiling price (hourly rate)');
 
   const yd = d3.extent(heightScale.domain());
-  const ya = d3.svg.axis()
-    .orient('left')
-    .scale(d3.scale.linear()
+  const ya = d3.axisLeft()
+    .scale(d3.scaleLinear()
```
jseppi: minor d3 v4 api changes

jseppi:
@toolness: when you review, please give it a good manual exercising as well to make sure I found and accounted for all the d3 api changes.

PR #1481, opened 2017-03-09
Use webpack for data-explorer JavaScript bundle

### LINKED ISSUE #1322: gulp build is kinda slow (at least on my system)

jseppi:
This should make @toolness happy :)

This PR changes our `gulp` process to use Webpack 2 via `webpack-stream` to build the Data Explorer React app.

I didn't measure, but it seems **much** faster to me, both on on initial build and on subsequent watch-triggered builds. File size of the resultant bundle is also slightly smaller. 

Closes #1322

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -27,7 +27,8 @@ const watchify = require('watchify');
 const envify = require('envify/custom');
 const babelify = require('babelify');
 const del = require('del');
-const webpack = require('webpack-stream');
+const webpackStream = require('webpack-stream');
+const webpack = require('webpack');
 
 const BUILT_FRONTEND_DIR = 'frontend/static/frontend/built';
 
@@ -66,8 +67,9 @@ const bundles = {
       'vendor/jquery.nouislider.all.min.js',
     ],
   },
-  // CALC 1.0 scripts
+  // Data Explorer scripts
   dataExplorer: {
+    noBrowserify: true,
     dirName: 'data-explorer',
     vendor: [
       'vendor/rgbcolor.js',
@@ -78,13 +80,13 @@ const bundles = {
       'vendor/jquery.auto-complete.min.js',
     ],
   },
-  // CALC 2.0 Data Capture scripts
+  // Data Capture scripts
   dataCapture: {
     dirName: 'data-capture',
   },
   // Styleguide scripts
   styleguide: {},
-  // CALC 2.0 Tests
+  // Test scripts
   tests: {},
   // Common scripts
   shared: {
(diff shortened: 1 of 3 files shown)


jseppi:
ready for review, @toolness

toolness:
Nice, this looks great!

I noticed that our trusty ol' `browserifyBundle` function has the following code in it:

```js
  // Some modules are referenced in the source code for
  // Enzyme, but they're never actually loaded because they're only
  // needed for older versions of React, so we'll explicitly tell
  // Browserify to ignore them here.
  bundler = bundler
    .exclude('react/addons')
    .exclude('react/lib/ReactContext')
    .exclude('react/lib/ExecutionEnvironment');
```

Should we go ahead and remove that since browserify isn't building any React-related code?

Also, do you eventually see the data capture bundle being webpackified?  It seems like it could be useful, especially since webpack is so much faster... Would also make the asset pipeline simpler if we only used one bundler too. But not really something we need to do for this PR, methinks.

toolness:
Hmm, when running `gulp js:data-explorer` on its own, I noticed this is displayed:

```
(node:1) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```

Is that something we can control, or is it coming from one of our dependencies?

toolness:
Hmm, it seems like the uglifier is no longer doing dead code removal--specifically, the string `redux-logger` is showing up in the *production* `data-explorer/index.min.js`, which is NO BUENO because I think it means that `redux-logger` is being included in the bundle, right?  Whereas the browserified bundler *does* remove the dead code as per #1247.

jseppi:
- Webpack for Data Capture: Yeah, that'll be cool! Let's do it separately though :)
- Deprecation warning: That warning is meant for loader authors, not consumers (ref https://github.com/webpack/loader-utils/issues/56). It has been fixed in babel-loader as of https://github.com/babel/babel-loader/issues/392 but it's not in a real release yet.
- Production dead code removal: daggit, will look into this.

toolness:
Sweet! No need to look into it yourself, I just fixed it in #1487.

Once that is merged in I think this is 👍 !

toolness:
arrrg missing trailing comma piece of @#%%@

PR #1487, opened 2017-03-10
Define NODE_ENV properly in webpack pipeline

toolness:
This fixes the problem mentioned in https://github.com/18F/calc/pull/1481#issuecomment-285788426.


jseppi:
haha, doh >.<

PR #1489, opened 2017-03-13
Use webpack for all JS bundles

### LINKED ISSUE #1488: Use webpack for all JS bundles
Now that we have webpack going for the data-explorer bundle (https://github.com/18F/calc/pull/1481), we should probably get it setup for data-capture and any other bundles!

jseppi:
Closes #1488

This PR removes `browserify` and instead uses `webpack` to generate all JS bundles 🎉 

Instead of bundles named `path/to/data-capture/index.min.js` they are named according to the last folder in the path to the "entry" file (or the `dirName` property from the corresponding `bundle` object). So, for the example here, it would be `data-capture.js` (I also took out the `.min`).
 
Similarly, the vendor bundles have been renamed like `<name>.vendor.js`, e.g. `data-explorer.vendor.js`.

To Do:
- [x] figure out eslint and (probably) related stream issues


jseppi:
Ready for review, @toolness

package.json
```diff
@@ -46,11 +42,8 @@
     "react-redux": "^4.4.6",
     "redux": "^3.6.0",
     "redux-logger": "^2.7.4",
-    "sinon": "1.17.5",
-    "uglifyify": "^3.0.4",
-    "vinyl-buffer": "^1.0.0",
-    "vinyl-source-stream": "^1.1.0",
-    "watchify": "^3.7.0",
+    "sinon": "next",
```
jseppi: blech, this is to fix an issue with `sinon` not playing well with `webpack`.

When using `1.17.5`, we'd get an error when running `js:webpack`:

```
WARNING in ./~/sinon/lib/sinon.js
40:25-32 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

It looks like the problem is due to how `sinon` does its AMD/UMD checking, ref: https://github.com/webpack/webpack/issues/304#issuecomment-193387024

The solution to use the `"next"` version of sinon is from https://github.com/webpack/webpack/issues/304#issuecomment-193387024

package.json
```diff
@@ -46,11 +42,8 @@
     "react-redux": "^4.4.6",
     "redux": "^3.6.0",
     "redux-logger": "^2.7.4",
-    "sinon": "1.17.5",
-    "uglifyify": "^3.0.4",
-    "vinyl-buffer": "^1.0.0",
-    "vinyl-source-stream": "^1.1.0",
-    "watchify": "^3.7.0",
+    "sinon": "2.0.0-pre.6",
```
jseppi: blech, this is to fix an issue with `sinon` not playing well with `webpack`.

When using `1.17.5`, we'd get an error when running `js:webpack`:

```
WARNING in ./~/sinon/lib/sinon.js
40:25-32 Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
```

It looks like the problem is due to how `sinon` does its AMD/UMD checking, ref: https://github.com/webpack/webpack/issues/304#issuecomment-193387024

The solution to use the `"next"` version of sinon was from https://github.com/webpack/webpack/issues/304#issuecomment-193387024, but I didn't feel super comfortable with that so I specified the most recent 2.0.0 prerelease version.

gulpfile.js
```diff
@@ -284,13 +198,22 @@ gulp.task('js:data-explorer', () => {
...
+      const lastFolderName = parts[parts.length - 2];
+      return lastFolderName;
+    }))
     .pipe(webpackStream({
       watch: isWatching,
       resolve: {
         extensions: ['.js', '.jsx'],
       },
+      stats: {},
       devtool: 'source-map',
```
jseppi: There are a number of options to specify how source maps are made, but I really don't understand them and haven't found a good guide on what all the options mean :\

Here's the section in webpack's docs: https://webpack.github.io/docs/configuration.html#devtool

jseppi: `cg-documentation` has this in their webpack config:

```js
devtool: PRODUCTION ? 'cheap-source-map' : 'eval-source-map',
```

jseppi: ah, the new Webpack docs have more explanation: https://webpack.js.org/configuration/devtool/

It looks like what we want is:

```js
devtool: isProd ? 'source-map' : 'eval-source-map',
```

gulpfile.js
```diff
@@ -203,95 +175,48 @@ function concatAndMapSources(name, sources, dest) {
...
   const plugins = [
     new webpack.DefinePlugin({
       'process.env.NODE_ENV': isProd ? '"production"' : '""',
     }),
   ];
 
   if (isProd) {
-    plugins.push(new webpack.optimize.UglifyJsPlugin());
+    plugins.push(new webpack.optimize.UglifyJsPlugin({
+      sourceMap: true,
```
jseppi: needed to have source maps in production build (which are very nice to have. New Relic's JS error catcher can load them so you know what caused the errors!).

gulpfile.js
```diff
@@ -17,18 +17,12 @@ const concat = require('gulp-concat');
...
 const eslint = require('gulp-eslint');
-const gulpif = require('gulp-if');
 const gutil = require('gulp-util');
-const uglify = require('gulp-uglify');
-const source = require('vinyl-source-stream');
-const buffer = require('vinyl-buffer');
-const browserify = require('browserify');
-const watchify = require('watchify');
-const envify = require('envify/custom');
-const babelify = require('babelify');
```
toolness: O man so great to get rid of all these dependencies!!!!! 🎉

toolness:
Hmm, so one thing I just noticed is that the auto-linking to web components in the styleguide seems to break with this PR.  For instance, the `<upload-widget>` link should go to [`/frontend/source/js/data-capture/upload.js`](https://github.com/18F/calc/blob/develop/frontend/source/js/data-capture/upload.js) on the develop branch, but instead it seems to go to `/index.js`, on that branch, which does not exist.  This suggests that `__filename` may not be set to the correct value in `upload.js` for some reason.

toolness:
Ugh this GitHub review thing is so weird and confusing to me still.  I've queued up a bunch of comments that pointed at specific lines of code, but then when I want to make a comment that isn't specific to a line of code, it doesn't get queued up?  Bleh.

toolness:
Interesting, I'm noticing that `eslint` doesn't seem to be running when files change anymore.  I actually prefer it this way, though, to be honest--while I really appreciate some of eslint's warnings, like "hey, this symbol you're importing doesn't seem to exist", the vast majority of them are trivial and I've actually been finding it hard to differentiate between purely stylistic linting warnings and actual compilation failures in our current browserify-based system.  Having *only* webpack errors show up is actually much easier for me to parse.

And anyways, hopefully we will be moving to prettier soon (#1493) and even the trivial eslint stuff will be a thing of the past. 😁

toolness:
[Wait wut](https://github.com/webpack/webpack/issues/5#issuecomment-7957736)

> By design no path info should be exposed. `__filename` and `__dirname` is not supported.

😢 😢 😢

toolness:
Oh nevermind, just found http://stackoverflow.com/a/25565222. Will bust a PR to this PR shortly.

jseppi:
wat

https://codeclimate.com/github/18F/calc/pull/1489#

screen shot 2017-03-20 at 10 51 29 am

jseppi:
Also getting an error during sphinx build:

screen shot 2017-03-20 at 10 53 40 am

jseppi:
gulp error is fixed after the merge back from `develop`. Still no clue what CodeClimate is upset about.

jseppi:
Oof, also I think something funky is going on with the gulp streams. They don't seem to be resolving completely until after the second file change or something. Looking into it...

toolness:
So one of the things that surprised me about running the jest tests was that `import` actually works... It seems like jest isn't running our webpack'd bundles, it's somehow just running our code in node, and I thought that node didn't support `import`... So given that I have no idea how those tests are actually being run, I'm not very surprised that CodeClimate has some level of confusion about it too, lol.

jseppi:
`jest` is actually using `babel` via `babel-jest` (ref https://facebook.github.io/jest/docs/getting-started.html#using-babel).

toolness:
O funkadelic.

I wonder if using `require()` would somehow quell CodeClimate?  Or can we just `# NOQA` that line or something?

.codeclimate.yml
```diff
@@ -16,6 +16,8 @@ engines:
     checks:
       import/no-unresolved:
         enabled: false
+      import/extensions:
+        enabled: false
```
jseppi: Uh, I just disabled this to get rid of the CodeClimate complaint, which was happening for an unknown reason. We don't actually want any extensions on our imports, as per our eslint config.

jseppi:
Ok, @toolness. Got the stream weirdness sorted. The problem, I think, is that `webpack-stream` never actually returns its stream when it is in `watch`-mode. The solution was to not return the gulp stream in the `'webpack'` task. Sorry, buddy, this means eslint is now working again :)

PR #1490, opened 2017-03-13
Support fuzzy column titles

toolness:
This fixes some problems with the health IT price list that was uploaded earlier today, which has a header row that looks like this:

> screen shot 2017-03-13 at 6 28 20 pm

It does this by adding a new class called `ColumnTitle` which allows us to specify the canonical name for a column heading (in case we need to display it to the user) as well as possible alternatives, which could be strings or compiled regular expressions.

Other things:

* It also allows the unit of issue to be **Hours** (this is what it is in the health IT price list).

* If the header row can't be found, it will now raise a validation error complaining about not being able to find the `SIN(s) PROPOSED` column header, instead of `Could not find Labor Categories price table`, which is what the health IT price list was producing earlier, and which is not very helpful as it doesn't give users any clue as to how they might move forward.

Stuff to do later (not done in this PR):

* The uploaded price list specifies "associate" as a degree type, which we complain about.  It'd be nice if we could support that.

### DIFF
--- a/data_capture/schedules/base.py
+++ b/data_capture/schedules/base.py
@@ -20,7 +20,7 @@ min_price_validator = MinValueValidator(
             '(${0:.2f})'.format(FEDERAL_MIN_CONTRACT_RATE))
 
 
-hour_regex = re.compile(r'^hour(ly)?$', flags=re.IGNORECASE)
+hour_regex = re.compile(r'^hour(ly|s)?$', flags=re.IGNORECASE)
 
 hourly_rates_only_validator = RegexValidator(
     hour_regex, 'Value must be "Hour" or "Hourly"')
--- a/data_capture/schedules/s70.py
+++ b/data_capture/schedules/s70.py
@@ -9,7 +9,8 @@ from django.template.loader import render_to_string
 
 from .base import (BasePriceList, min_price_validator,
                    hourly_rates_only_validator)
-from .spreadsheet_utils import generate_column_index_map, safe_cell_str_value
+from .spreadsheet_utils import (generate_column_index_map,
+                                safe_cell_str_value, ColumnTitle)
 from .coercers import (strip_non_numeric, extract_min_education,
                        extract_hour_unit_of_issue, extract_first_int)
 from contracts.models import EDUCATION_CHOICES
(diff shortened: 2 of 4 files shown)


data_capture/schedules/s70.py
```diff
@@ -9,28 +9,56 @@
...
 
 DEFAULT_SHEET_NAME = '(3)Labor Categories'
 
+COLUMN_TITLES = {
+    'sin': ColumnTitle(
+        canonical_name=r'SIN(s) PROPOSED',
+        alternatives=[
+            re.compile(r'SIN.*')
+        ]
+    ),
```
toolness: @tram, this provides a structure which you can later update if you run into other alternatives! I can explain it to you at coworking.

data_capture/schedules/base.py
```diff
@@ -20,7 +20,7 @@
             '(${0:.2f})'.format(FEDERAL_MIN_CONTRACT_RATE))
 
 
-hour_regex = re.compile(r'^hour(ly)?$', flags=re.IGNORECASE)
+hour_regex = re.compile(r'^hour(ly|s)?$', flags=re.IGNORECASE)
```
jseppi: I guess this is ok, but it is kinda weird to me. We want the price per hour not per hours :\ Not sure where to draw the line with this validation though.

data_capture/schedules/spreadsheet_utils.py
```diff
@@ -3,6 +3,33 @@
 from django.core.exceptions import ValidationError
 
 
+class ColumnTitle:
```
jseppi: Cool, nice implementation!

data_capture/schedules/base.py
toolness: Do you mean that someone could be specifying a payment based on e.g. "per 5 hours of work" or something? i.e., if we accept "hours" then we might accidentally be accepting work that isn't on an "every *one* hour" unit of issue?

If that's the case, I'm cool with reverting this change, and essentially telling the COs uploading these price lists to fix it manually on their end.

tram: Yeah… it seems like when we have seen (or can come up with) cases that aren't beyond belief and would accept data that would be invalid (like this case), forcing the CO to fix it makes sense.

PR #1492, opened 2017-03-14
Django 1.9 admin fixes

hbillings:
Fixes the regression @jseppi noticed:
screen shot 2017-03-14 at 2 57 33 pm

Also adjusts some input box alignment and background icons.

EDIT: Looks like I got a whole bunch of commit history from merging `develop` into this. The last couple are all you gotta worry about -- there were no conflicts when I merged `develop`!


toolness:
Woot thanks!  Um, I ended up cherry-picking your two commits into `django-1.9` so consider this merged!

PR #1493, opened 2017-03-15
[WIP] [Experiment] Trying prettier

jseppi:
Trying out `prettier` alongside of the Airbnb `eslint` config.

To Do:
- [ ] run `prettier --list-different` in Travis to fail if files are not in `prettier` style. Tricky thing is going to be skipping `vendor` and jest snapshot folders.


jseppi:
I do like `prettier` but I'm reluctant to start using it in this codebase for a few reasons:

1) We'll have to do a big commit to reformat all existing JS. That's kinda nasty for our git history.
2) The response of the front end guild when we talked about last week was fairly lukewarm.
3) It has some differences from the `eslint-config-airbnb` that we as an organization have decided to standardize on, so we'll lose the benefits of that standardization (mostly other developers being able to hopefully easily jump in and understand the JS code, though I'm not totally convinced that's 100% valid as a concern).

We should be democratic about it and put it to a vote! @hbillings and @toolness?

toolness:
I am 👍, albeit mostly out of a hatred for eslint rather than a liking of prettier, which I haven't actually tried yet. 😝

jseppi:
We'll still (IMO) want to have some eslint, like for the React stuff I have in https://github.com/18F/calc/pull/1493/files

toolness:
Still, wouldn't prettier reduce the pain of eslint by automatically formatting code to obey most of eslint's rules?

jseppi:
It would reformat it to follow its own rules, but yeah, I suppose it would reduce the pain you've been having 💊

jseppi:
This is way out of date and I don't personally feel like this switch is necessary, so closing.

PR #1495, opened 2017-03-16
Call strip on text in extract_hour_unit_of_issue

### LINKED ISSUE #1494: Unit of Issue validation will fail if there are extraneous spaces
@tram was sent a spreadsheet and screenshots that showed many invalid rows being detected. 

The root cause turned out to be that the failing Unit of Issue fields had extra spaces at the end (eg, `"Hour  "`).

Confusingly, I was not initially able to duplicate the problem in the current codebase from our `develop` branch. It turns out that Django 1.9, which we recently upgraded to 
(https://github.com/18F/calc/pull/1466), automatically calls `strip()` on `CharField`s. Ref: https://docs.djangoproject.com/en/1.10/releases/1.9/#forms

However, the `extract_hour_unit_of_issue` coercer should still `strip()` on its own in order to allow the other variations of "hour" to pass validation when they have extra spaces (ie, `"hourly "` or `"hours "`).

jseppi:
Fixes #1494

### DIFF
--- a/data_capture/schedules/coercers.py
+++ b/data_capture/schedules/coercers.py
@@ -24,11 +24,11 @@ def strip_non_numeric(text):
     return re.sub("[^\d\.]", "", text)
 
 
-def strip_and_lower(s):
+def strip_punctuation_and_lower(s):
     '''
     Helper to remove punctuation and lowercase the given input
 
-    >>> strip_and_lower('@!.HELLO, Friend!!!!')
+    >>> strip_punctuation_and_lower('@!.HELLO, Friend!!!!')
     'hello friend'
     '''
     exclude = set(string.punctuation)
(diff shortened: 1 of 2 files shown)


data_capture/schedules/coercers.py
```diff
@@ -24,11 +24,11 @@ def strip_non_numeric(text):
     return re.sub("[^\d\.]", "", text)
 
 
-def strip_and_lower(s):
+def strip_punctuation_and_lower(s):
```
jseppi: I renamed this method to better reflect what it does

data_capture/schedules/spreadsheet_utils.py
```diff
@@ -38,11 +38,15 @@ def safe_cell_str_value(sheet, rownum, colnum, coercer=None):
...
 
+    print(f"val before: '{val}'")
     if coercer is not None:
         try:
             val = coercer(val)
+            print(f"val after: '{val}'")
         except ValueError:
+            print(f"Failed on '{val}'")
             pass
+    print("-----------------------")
```
toolness: I assume this is leftover debugging code?

jseppi: whoops!

Issue #1498, opened 2017-03-17
Dev & Staging v2.5.0 - Exceptions thrown when calling API methods


jseppi:
Looks like a problem with `djorm_pgfulltext`. Logs below:

```txt
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR   File "/home/vcap/app/api/views.py", line 173, in get
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR     contracts_all = self.get_queryset(request.query_params, wage_field)
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR   File "/home/vcap/app/api/views.py", line 205, in get_queryset
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR     return get_contracts_queryset(request, wage_field)
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR   File "/home/vcap/app/api/views.py", line 100, in get_contracts_queryset
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR     contracts = contracts.multi_phrase_search(qs)
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR   File "/home/vcap/app/contracts/models.py", line 136, in multi_phrase_search
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR     return self.search(convert_to_tsquery_union(queries), raw=True)
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR   File "/app/.heroku/python/lib/python3.6/site-packages/djorm_pgfulltext/models.py", line 315, in search
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR     "%s('%s', %s)" % (function, config, adapt(query))
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR   File "/app/.heroku/python/lib/python3.6/site-packages/djorm_pgfulltext/utils.py", line 10, in adapt
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR     a.prepare(connection.connection)
2017-03-16T22:54:03.13-0500 [APP/PROC/WEB/0]ERR TypeError: argument 1 must be psycopg2.extensions.connection, not ConnectionWrapper

[...]

PR #1500, opened 2017-03-17
Alternative fix for #1498 (direct encoding approach)

toolness:
Most of the fix is actually contained in our `djorm-ext-pgfulltext` fork, which this PR references in its `requirements.txt`. For more details, see https://github.com/18F/djorm-ext-pgfulltext/pull/1.

This PR also contains regression tests to ensure that the fix works.


toolness:
Note that even if we don't end up using this PR, we might want to merge just its test, as this will help protect against regressions once we switch to Django 1.10's native full-text search.

toolness:
Ended up going with #1502 instead!

PR #1501, opened 2017-03-17
Normalize Postgres version to 9.4.7, add to healthcheck, and test against it

### LINKED ISSUE #1388: Normalize postgres version and test against it
It'd be nice to have tests in [`hourglass/tests/test_configuration.py`](https://github.com/18F/calc/blob/develop/hourglass/tests/test_configuration.py) that ensure travis, dev/prod/staging, and our docker-compose setup all use the same version of postgres, but right now they're actually in conflict:

* `.travis.yml` specifies postgres 9.3
* `docker-compose.yml` specifies 9.4
* I have no idea what we use on dev/staging/prod

If there's no way to express what version of postgres CloudFoundry/AWS should use in our repo (the way  we specify the python version via `runtime.txt`) then it might be good to add a check in `/healthcheck` that queries the DB for its version and makes sure it's something expected.

This seems increasingly important as we improve CALC's performance and functionality by using features introduced in postgres' more recent versions.

jseppi:
Closes #1388

I used `9.4.7` as the version since that is what cloud.gov (through RDS) is giving us.

### DIFF
--- a/hourglass/healthcheck.py
+++ b/hourglass/healthcheck.py
@@ -1,27 +1,44 @@
+import re
+import django_rq
+
+from semantic_version import Version
 from django.http import JsonResponse
 from django.db.migrations.executor import MigrationExecutor
 from django.db import connections, DEFAULT_DB_ALIAS
-import django_rq
 
 from hourglass import __version__
 from hourglass.site_utils import get_canonical_url
 
 
-def is_database_synchronized(database=DEFAULT_DB_ALIAS):
+def parse_pg_version(pg_version):
+    # pg_version should be an integer that looks like 90410
+    # returns a semantic_version.Version
+    result = re.match(r'(\d)(\d\d)(\d\d)', str(pg_version))
+    return Version(f"{result[1]}.{int(result[2])}.{int(result[3])}")
+
+
+def get_database_info(database=DEFAULT_DB_ALIAS):
     connection = connections[database]
     connection.prepare_database()
     executor = MigrationExecutor(connection)
     targets = executor.loader.graph.leaf_nodes()
-    return False if executor.migration_plan(targets) else True
+    is_synchronized = False if executor.migration_plan(targets) else True
+    return {
+        'is_synchronized': is_synchronized,
+        'pg_version': str(parse_pg_version(connection.pg_version)),
+    }
 
 
 def healthcheck(request):
     canonical_url = get_canonical_url(request)
     request_url = request.build_absolute_uri()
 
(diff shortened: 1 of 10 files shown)


hourglass/settings.py
```diff
@@ -282,6 +282,7 @@
 
 DATABASES = {}
 DATABASES['default'] = dj_database_url.config()
+POSTGRES_VERSION = '9.4.7'
```
jseppi: Wasn't exactly sure on best place to put this. I had previously put it in our `version.py` but that felt weird so I moved it here to `settings.py`.

requirements.txt
```diff
@@ -24,3 +24,4 @@ mypy-lang==0.4.6
 pytz==2016.10
 Markdown==2.6.8
 cg-django-uaa==0.0.1
+semantic_version==2.6.0
```
jseppi: Moved this package to the production requirements since it is now used as part of the healthcheck code.

jseppi:
Ready for review, @toolness

jseppi:
Oh whoops, hold up. Travis is failing because it has a different patch version (9.4.10) from what I'm testing against...

jseppi:
Ok, all good now.

hourglass/healthcheck.py
```diff
@@ -1,27 +1,44 @@
...
     canonical_url = get_canonical_url(request)
     request_url = request.build_absolute_uri()
 
+    db_info = get_database_info()
+
     results = {
         'version': __version__,
-        'is_database_synchronized': is_database_synchronized(),
+        'postgres_version': db_info['pg_version'],
+        'is_database_synchronized': db_info['is_synchronized'],
```
toolness: O I think you can use python 3.6's dictionary splatting here or whatever it's called, e.g.:

```python
results = {
    'version': __version__,
    **get_database_info()
}
```

hourglass/settings.py
toolness: Huh interesting.  I guess since it's not actually used for *configuration* per se, like most settings are, it does feel a *little* weird, but I guess it's coolio.

hourglass/healthcheck.py
```diff
@@ -1,27 +1,44 @@
+import re
+import django_rq
+
```
toolness: I guess this newline doesn't really need to be here--I only put newlines in between third-party imports and our own module imports, but it's up to you.

jseppi: o, I kinda like them between groups of `import`s and `from`s

PR #1503, opened 2017-03-17
Fix webpack web component link, add webpack-related gulp tests

toolness:
**This PR is against #1489, not `develop`.**

This makes web component links in the styleguide work again.

It also adds tests for the webpack-related functionality in the gulpfile.


frontend/source/js/styleguide/web-component-link.js
```diff
@@ -23,7 +23,7 @@ class WebComponentLink extends window.HTMLAnchorElement {
         'has no SOURCE_FILENAME property!',
       );
     }
-    this.setAttribute('href', this.href + proto.SOURCE_FILENAME.slice(1));
+    this.setAttribute('href', this.href + proto.SOURCE_FILENAME);
```
toolness: Browserify set `__filename` to an absolute path starting with a `/`, which we stripped off, while webpack sets it to a relative path assumed to be at the context root (the location of our gulpfile, in our case).

toolness:
Um, so I ended up adding tests for some of the webpack-related functionality, since they are things I could easily see breaking in the future without anyone noticing for a really long time.

I also figured it might be a good idea to start a test suite for gulp-related code, as this will lower the barrier to adding tests for future gulp modifications.  I think it is especially important to add tests for conditional code that branches depending on `NODE_ENV === "production"`, since we rarely enable production mode on our development machines.

frontend/gulp/tests/webpack-util.test.js
```diff
@@ -0,0 +1,75 @@
...
+});
+
+function webpackify(filename, options = {}) {
+  return new Promise((resolve) => {
+    const src = gulp.src(path.join(__dirname, filename))
+      .pipe(webpackUtil.webpackify(Object.assign({
+        isWatching: false,
+        isProd: false,
+      }, options)));
+    src.on('data', resolve);
```
toolness: When `isProd` is truthy, I *think* that the vinyl stream might also emit a separate source map file.  I was worried that there might be a race condition where the source map is emitted first, but after running the test suite a bunch of times, this never seemed to happen, so hopefully the webpack code always emits the source maps after the original file.

jseppi:
Cool! Good idea with the gulp tests. I also hadn't seen the `vm` module before. Looks handy!

PR #1509, opened 2017-03-21
Limit autocomplete results on server-side

jseppi:
This PR limits the server-side autocomplete results to a maximum of 20 results. It also adds a few tests for the autocomplete (`/api/search`) endpoint since there were no tests at all. This will result in speed/efficiency increases for our servers and for all clients.

I don't know how to do "real" profiling, but a quick test using `"en"` as the query reduced network request time on my LOCAL dev environment from about `1.7 seconds` down to `400ms`.

Background:

While looking around through our [New Relic transaction monitoring](https://rpm.newrelic.com/accounts/987071/applications/5976077/transactions), I noticed that the autocomplete methods take kind of a long time:

screen shot 2017-03-21 at 11 35 37 am

I took a quick look at the code and noticed that the our autocomplete API method (at `/api/search`) returns ALL results it finds, but the client-side JavaScript code limits the results shown to just 20. This is pretty inefficient, from the server-side, through the wire, to the client. For example, an autocomplete search for `"en"` would return a payload containing over 7000 results, which then have to be JSON parsed (a blocking operation) by the client.

### DIFF
--- a/frontend/source/js/data-explorer/autocomplete.js
+++ b/frontend/source/js/data-explorer/autocomplete.js
@@ -46,7 +46,7 @@ export function processResults(result) {
   if (!result || !result.length) {
     return [];
   }
-  const categories = result.slice(0, 20).map(d => ({
+  const categories = result.map(d => ({
     term: d.labor_category,
     count: d.count,
   }));
--- a/api/tests/test_autocomplete_api.py
+++ b/api/tests/test_autocomplete_api.py
@@ -1,0 +1,55 @@
+from itertools import cycle
+from django.test import TestCase
+
+from ..views import GetAutocomplete
+from contracts.mommy_recipes import get_contract_recipe
+
+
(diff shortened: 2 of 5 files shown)


frontend/source/js/data-explorer/autocomplete.js
```diff
@@ -46,7 +46,7 @@ export function processResults(result) {
   if (!result || !result.length) {
     return [];
   }
-  const categories = result.slice(0, 20).map(d => ({
```
jseppi: I went ahead and removed this since we can now just rely on the server-side result limiting, and I didn't think it made sense to have it in both places.

jseppi:
Just did a little spot check comparing calc-dev and prod:

Autocomplete for "te" -
dev: 1019 bytes, 612ms
prod: 26 **K**b, 747ms

PR #1513, opened 2017-03-22
[WIP] Investigate react-storybook

toolness:
I saw a [React Conf 2017 talk](https://www.youtube.com/watch?v=PF0Vi-iIyoo) on [React Storybook](https://getstorybook.io/) and thought it might be a nice companion to our styleguide, for our React-based components, so I'm trying it out here.

> 

## Thoughts

So far, I really like this.  As with [Jest](https://facebook.github.io/jest/), storybook has its own built-in server that runs webpack/babel and re-runs it whenever files change, auto-reloading stuff in the browser as needed. (I think it even uses hot module reloading so it doesn't have to reload the whole page.)

* When I make syntax errors, they show up really nicely in the browser--basically the component area in the storybook is replaced with a console-like panel that shows the error output.

* The stories are really easy to write and serve nicely as example code for component usage.

* Embedding the `<EducationLevel>` component has interesting results: while the component is interactive, clicking on an education level doesn't *actually* toggle it, because that logic is actually done by the redux store and then passed back into the component via props. Instead, though, a message shows up in the **Action logger** panel, with details on the arguments passed to the callback. This may or may not be confusing to new developers, I'm unsure.

## Things to explore

* Right now I've only tried using the storybook dev server, but storybook also has the ability to export itself as a static site. This would be nice for staging/dev deployments; perhaps our styleguide and/or sphinx docs can have a link out to it, to make it nice and discoverable.

* Try out the [info addon](https://github.com/storybooks/react-storybook-addon-info), the [README addon](https://github.com/tuchk4/storybook-readme) and/or other addons to integrate documentation into the storybook. This will help contextualize things better.

* I think there might be an aXe-core storybook addon out there somewhere.

* It'd be nice to integrate the storybook into our automated test suite somehow--at least an initial render of every story, ensuring that none of them throw any obvious errors, would be good for smoke testing. Perhaps simply exporting the storybook as a static site will be enough for this.

### DIFF
--- a/docker-compose.local.yml
+++ b/docker-compose.local.yml
@@ -15,6 +15,16 @@ services:
     volumes:
       - .:/calc
     command: gulp
+  storybook:
+    extends:
+      file: docker-services.yml
+      service: base_app
+    volumes:
+      - .:/calc
+    stop_signal: SIGKILL
+    command: start-storybook -p 9001 -c .storybook -s ./frontend/static
+    ports:
+      - 9001:9001
   rq_worker:
     volumes:
       - .:/calc
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,0 +1,7 @@
+import { configure } from '@kadira/storybook';
+
+function loadStories() {
+  require('../frontend/source/js/data-explorer/stories/index.jsx');
+}
+
+configure(loadStories, module);
--- a/.storybook/head.html
+++ b/.storybook/head.html
@@ -1,0 +1,4 @@
+<link rel="stylesheet" href="/frontend/built/style/main.min.css">
+<link rel="stylesheet" href="/frontend/built/style/data_explorer.min.css">
+<script src="/frontend/built/js/common.vendor.js"></script>
+<script src="/frontend/built/js/data-explorer.vendor.js"></script>
--- a/frontend/source/js/data-explorer/stories/index.jsx
+++ b/frontend/source/js/data-explorer/stories/index.jsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import { storiesOf, action } from '@kadira/storybook';
+
+import * as constants from '../constants';
+import { EducationLevel } from '../components/education-level';
+import Tooltip from '../components/tooltip';
+
+storiesOf('EducationLevel', module)
+  .add('with none selected', () => (
+    <EducationLevel
+      levels={[]}
+      toggleEducationLevel={action('toggled')}
+    />
+  ))
+  .add('with one selected', () => (
+    <EducationLevel
+      levels={[constants.EDU_HIGH_SCHOOL]}
+      toggleEducationLevel={action('toggled')}
+    />
+  ));
+
+storiesOf('Tooltip', module)
+  .add('showing on focus/hover', () => (
+    <Tooltip text="Hello, I am a tooltip!">
+      <a
+        href=""
+        aria-label="Hello, I am a tooltip!"
+        onClick={(e) => { e.preventDefault(); }}
+      >
+        I am a keyboard-focusable tooltip.
+      </a>
+    </Tooltip>
+  ))
+  .add('always showing', () => (
+    <Tooltip text="I am always showing!" show>
+      This text will have a permanently visible tooltip.
+    </Tooltip>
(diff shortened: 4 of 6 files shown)


toolness:
@jseppi @hbillings uhhh I guess this is maybe ready to be merged actually. I can show it to u over screenhero if u want, it seems spiffy.

jseppi:
I do like the idea of this since it should make development of individual components a lot easier, but I'm a bit concerned by the behavior you described with the `<EducationLevel>` component's storybook. Will that be the case with most of our components since we're using Redux? Not being able to really see behavior of components seems like it would make many of the stories not all that useful.

docker-compose.local.yml
```diff
@@ -15,6 +16,16 @@ services:
     volumes:
       - .:/calc
     command: gulp
+  storybook:
+    extends:
+      file: docker-services.yml
+      service: base_app
+    volumes:
+      - .:/calc
+    stop_signal: SIGKILL
```
jseppi: whoa, that's a new one for me

.gitignore
```diff
@@ -36,3 +36,4 @@ jest/*
 Dockerfile.cloud
 docs/_build
 docs/static
+.storybook/static
```
toolness: Well, the `.storybook` directory (minus the `static` subdirectory) is what was recommended in the docs, and I guess that makes sense because it stores configuration data--so it's like `.eslintrc` or `.travis.yml` but it's a directory instead of an individual file.

The `static` subdirectory is sort of a hack because I needed a place to put the statically-generated version of the site, and it seemed like `.storybook` worked as well as anything else.

toolness:
Hmm, I guess we could maybe wrap the component stories in a `<Provider>` (and their redux component wrapper) if they really don't make much sense on their own?

Alternatively we could provide a wrapper that provides minimal scaffolding for the component to make sense in the context of the story.  For the `<EducationLevel>` component that could mean providing a toggle callback that actually modifies the props passed back into the component... hmmmmm.

toolness:
In a way this might actually point to a weakness of the way we're writing components, but I'm not sure. Oy.

toolness:
So I'm thinking two things right now:

* Ultimately this `<EducationLevel>` component is a bit wonky, most likely because it was created in the great React migration of #1075. Ideally I think we should have a `<SlideyCheckboxes>` component or something, which isn't so tightly coupled to the notion of education levels, and `<EducationLevel>` should really just be a redux-connected adapter on top of it.  The stories should be for `<SlideyCheckboxes>`, not `<EducationLevel>`... But I also wanted this PR to focus on the react storybook integration rather than refactoring components, so I didn't do any refactoring.

* It's hard for me to tell right now whether my testing instincts are good or bad when it comes to storybook, but I'm leaning more towards providing minimal scaffolding to our components to make them meaningful in their stories, rather than connecting them to our entire redux store. The latter feels a lot more like an integration test, while the former feels more like an isolated unit test... and storybook stories feel kind of like visual versions of unit tests. Again, not sure if this actually the right choice, but it's where I'm leaning right now.

jseppi:
Yeah, something like `<SlideyCheckboxes>` would be more general, but if it's only used in one place, then I don't think it makes a ton of great sense to overgeneralize everything. Also I think that whole component is kinda bizarre from a UX standpoint.

I agree with your notion that the stories feel more like they should be unit tests. 

Hrmmm...

hbillings:
I can't speak to storybook, but I can say I'd really love to do away with the slidey education bar altogether, especially since analytics show us people look for 5-year intervals mostly.

jseppi:
heh, that's the Experience slider!

I actually kinda like that one, but it does seem maybe weird to have it
_and_ the related select inputs.

jseppi:
Looks like a similar project: http://react-styleguidist.js.org/

toolness:
Yeah, I just noticed Dan Abramov tweet about it... I wonder if storybook has been around longer/has a bigger community?  I dunno.

toolness:
@hbillings can we show this to you at coworking sometime?  It'd be nice to get your input on whether we should actually move forward with using this thingy.

jseppi:
heh, yeah, same place I saw it!

PR #1514, opened 2017-03-22
Upgrade to mypy 0.501

toolness:
This upgrades us to mypy 0.501.

* `silent_imports` has been replaced by `ignore_missing_imports=True; follow_imports=skip`
* Previously, annotations specified in comments auto-imported built-in typings like `Dict` and `List`, but this is no longer the case, so we have to import them explicitly.

The biggest hassle here is actually just the fact that mypy changed its pypi entry from `mypy-lang` to `mypy`: this means that when you run `pip install -r requirements.txt` (or `./update.sh`) it will actually install `mypy` *on top of* `mypy-lang`.  By itself, this isn't horrible, because the `mypy` CLI will overwrite the command of the same name provided by `mypy-lang`.

However, what *does* suck is that if you switch back to a different branch that *doesn't* use this new version of mypy, running `pip install` won't re-install the old version of mypy, because it thinks it's still installed. So it won't do anything, and then your mypy tests will fail because it's running the new version of mypy but without the fixes to files contained in this PR.

*Phew*. Anyways, super annoying. If you want to switch back to a different branch after merging this you might want to run `pip uninstall mypy mypy-lang` followed by `pip install -r requirements.txt` just to make sure you've got the right version of mypy for your branch.

I guess once we merge this in, we should just advise folks with WIP PRs to merge `develop` back into their PRs ASAP or they will run into annoying problems with this.

### DIFF
--- a/data_capture/schedules/base.py
+++ b/data_capture/schedules/base.py
@@ -13,6 +13,7 @@ from ..models import SubmittedPriceList
 
 if False:
     from django.forms import Form  # NOQA
+    from typing import List  # NOQA
 
 min_price_validator = MinValueValidator(
     FEDERAL_MIN_CONTRACT_RATE,
--- a/docker_django_management.py
+++ b/docker_django_management.py
@@ -98,7 +98,7 @@ if sys.platform != 'win32':
 
 if False:
     # This is just needed so mypy will work; it's never executed.
-    from typing import Iterator, Any  # NOQA
+    from typing import Iterator, Any, List  # NOQA
 
 MY_DIR = os.path.abspath(os.path.dirname(__file__))
 HOST_UID = os.stat(MY_DIR).st_uid
--- a/frontend/tests/sauce_connect.py
+++ b/frontend/tests/sauce_connect.py
@@ -6,7 +6,7 @@ import pexpect
 
 if False:
     # This is just needed so mypy will work; it's never executed.
-    from typing import Optional  # NOQA
+    from typing import Optional, Dict, List  # NOQA
 
 MY_DIR = os.path.abspath(os.path.dirname(__file__))
 
--- a/mypy.ini
+++ b/mypy.ini
@@ -5,29 +5,8 @@ strict_optional = True
 check_untyped_defs = True
 warn_incomplete_stub = True
 warn_unused_ignores = True
-disallow_untyped_defs = True
+disallow_untyped_defs = False
 disallow_untyped_calls = True
-
-[mypy-decimal]
-disallow_untyped_defs = False
-
-[mypy-csv]
-disallow_untyped_defs = False
-
-[mypy-time]
-# Weird, apparently typeshed's own annotations for `time`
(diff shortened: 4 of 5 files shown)


toolness:
So we can mitigate this in one direction by adding the following to the top of `./update.sh`:

```sh
# For details on why we're doing this, see:
# https://github.com/18F/calc/pull/1514
export IS_OLD_MYPY_INSTALLED=$(pip freeze | grep mypy-lang)
if [ ! -z "$IS_OLD_MYPY_INSTALLED" ]
then
  pip uninstall -y mypy-lang
fi
```

However, it's not going to address the more common case (in the very near-term) of someone having an old branch but having `mypy` mapped to the *new* `mypy` instead of the old one. Blerg.

toolness:
[Urg, travis is unhappy](https://travis-ci.org/18F/calc/builds/214021044#L739-L747) and it's something I can't reproduce locally. Seems to be a bunch of variations on the following:

```
/home/travis/virtualenv/python3.6.0/lib/mypy/typeshed/stdlib/2and3/xml/__init__.pyi:6: error: Function is missing a type annotation
```

toolness:
Weird, [`xml/__init__.pyi`](https://github.com/python/typeshed/blob/master/stdlib/2and3/xml/__init__.pyi) didn't actually exist in my container's virtualenv, but after uninstalling and re-installing `mypy` it did. How odd.

Anyways, I fixed the error, but then I actually just disabled our rule that was causing it, `disallow_untyped_defs = True`. This is normally `False` by default and a bunch of mypy's own annotations for the standard library lack them, so it's just easier to disable this for now.  Besides, I think our default settings might raise the bar for adding new typings a bit *too* high.

toolness:
Ok, here's a script I wrote that will unscrew mypy for whatever branch you happen to be on.

If you're having problems once this PR has been merged, you can copy the contents of this script to `unscrew-mypy.sh` in your repo's root directory and then run `./unscrew-mypy.sh` to hopefully fix things.

```sh
#! /bin/bash

set -e

# If you are not using Docker, change this to just 'pip'.
export PIP='docker-compose run app pip'

echo "Preparing to unscrew mypy: https://github.com/18F/calc/pull/1514"

export IS_OLD_MYPY_INSTALLED=$($PIP freeze | grep mypy-lang==)
if [ ! -z "$IS_OLD_MYPY_INSTALLED" ]
then
  echo "Old mypy-lang seems to be installed; uninstalling it."
  $PIP uninstall -y mypy-lang
fi

export IS_NEW_MYPY_INSTALLED=$($PIP freeze | grep mypy==)
if [ ! -z "$IS_NEW_MYPY_INSTALLED" ]
then
  echo "New mypy seems to be installed; uninstalling it."
  $PIP uninstall -y mypy
fi

echo "Now re-installing all dependencies, which should give you "
echo "the correct version of mypy for your branch."

$PIP install -r requirements-dev.txt

echo "Ok, hopefully mypy is unscrewed now."
```

PR #1523, opened 2017-03-27
Add DEBUG_HTTPS setting

toolness:
This adds a new setting, `DEBUG_HTTPS`, which allows the site to be served in debug mode over HTTPS.

It's primarily of importance to the site's notion of its "canonical url" when no `request` object is present: until now, we've always assumed that if `DEBUG` is true, the protocol for the canonical URL should be `http`, but setting `DEBUG_HTTPS` essentially forces it to be `https`.

I needed to add support for this to #997 because we're now putting the price list analysis deployment behind cloud.gov's SSL-terminating reverse proxy, and I figured this kind of functionality might be useful for local development too.

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -16,6 +16,11 @@ string), the boolean is true; otherwise, it's false.
 * `DEBUG` is a boolean value that indicates whether debugging is enabled
   (this should always be false in production).
 
+* `DEBUG_HTTPS` is a boolean value that indicates whether the
+  site should consider itself to be served over HTTPS while
+  debugging is enabled. This can be useful if you want to develop
+  with SSL enabled.
+
 * `HIDE_DEBUG_UI` is a boolean value that indicates whether to hide
   various development and debugging affordances in the UI, such as the
   [Django Debug Toolbar][]. This can be useful when demoing or user testing
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -32,7 +32,7 @@ load_redis_url_from_vcap_services('calc-redis')
 # SECURITY WARNING: don't run with debug turned on in production!
 DEBUG = 'DEBUG' in os.environ
 
-DEBUG_HTTPS = 'DEBUG_HTTPS' in os.environ
+DEBUG_HTTPS = 'DEBUG_HTTPS' in os.environ and not is_running_tests()
 
 HIDE_DEBUG_UI = 'HIDE_DEBUG_UI' in os.environ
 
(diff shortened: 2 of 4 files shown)


toolness:
Blerg, I am not a huge fan of introducing yet *another* setting but whatevs.

I guess alternatively we could change things so that `DEBUG` also assumes HTTPS unless `FORCE_DISABLE_SECURE_SSL_REDIRECT` is truthy, just like prod mode does... This would basically require us to change the `.env.sample` to include `FORCE_DISABLE_SECURE_SSL_REDIRECT=yup` and then tell all devs to add that line to their `.env`.

jseppi:
Hrm, do you ever use `FORCE_DISABLE_SECURE_SSL_REDIRECT`?

toolness:
ummmmmmmmmmmmmm i think i might have when we first created it, but i don't anymore............

jseppi:
heh, so maybe simplifying down to one such setting would be best :)

toolness:
hehe sure will do. want me to modify this PR to do that, or deal w/ it in a separate PR?

jseppi:
I guess just here if it isn't too urgent to get the new functionality of `DEBUG_HTTPS` into `develop`

toolness:
Yup sounds good sir!

toolness:
Blorg, while none of us actually use `FORCE_DISABLE_SECURE_SSL_REDIRECT` in practive, it seems a number of our *tests* do, so untangling them from that will take a little work.

jseppi:
I...can't seem to find it use in any code:

```
jamesseppi at XCOH2J-0TMFVH8 in ~/CODE/CALC (develop●)
$ ag FORCE_DISABLE_SECURE_SSL_REDIRECT
docs/environment.md
62:* `FORCE_DISABLE_SECURE_SSL_REDIRECT` is a boolean value that indicates

hourglass/settings.py
289:if 'FORCE_DISABLE_SECURE_SSL_REDIRECT' in os.environ:
```

toolness:
Oh, `FORCE_DISABLE_SECURE_SSL_REDIRECT` isn't used in code, but [`SECURE_SSL_REDIRECT`](https://docs.djangoproject.com/en/1.10/ref/settings/#std:setting-SECURE_SSL_REDIRECT) is... I forget why I thought that was relevant to removing `FORCE_DISABLE_SECURE_SSL_REDIRECT` though...

jseppi:
on 🦌

toolness:
Blorg I will just merge this. Filed #1532 to address the other thing.

PR #1528, opened 2017-03-29
Use redux store in react-storybook

toolness:
**This is a PR against #1513, not `develop`.**

This implements the first solution proposed in https://github.com/18F/calc/pull/1513#issuecomment-289552196 to help make the stories a bit more meaningful/useful, by making it possible to embed stories in a react-redux `<Provider>`.


toolness:
@jseppi what do you think of this? should I merge it into the react-storybook PR?

frontend/source/js/data-explorer/stories/index.jsx
```diff
@@ -1,29 +1,29 @@
...
-import * as constants from '../constants';
-import { EducationLevel } from '../components/education-level';
+import appReducer from '../reducers';
+import EducationLevel from '../components/education-level';
 import Tooltip from '../components/tooltip';
 
+const actionLoggerMiddleware = () => next => (reduxAction) => {
+  action(reduxAction.type)(reduxAction);
+  return next(reduxAction);
+};
```
toolness: This basically just makes the action show up in the "action logger" panel of the storybook:

screen shot 2017-03-29 at 11 38 32

PR #1529, opened 2017-03-29
Don't eslint on file change if ESLINT_CHILL_OUT is set

### LINKED ISSUE #1511: Consider disabling linting in gulp watch mode
As I mentioned in https://github.com/18F/calc/pull/1489#issuecomment-287480342, I think that having `gulp watch` *not* run eslint every time a file changes is actually more helpful:

* eslint errors tend to be minor stylistic nits; when I'm in the midst of implementing or fixing something, I just want to know if there are any syntax errors or warnings about the correctness of my code, rather than style issues. When the two are merged into one stream (as they are in the console), it becomes very hard to distinguish one from the other.

* on my Mac with `dinghy`, modifying a single file causes eslint to run for upwards of **six seconds** before it prints anything out. I don't know if it's because it's re-linting the whole codebase or what, but that kind of delay makes its feedback particularly unhelpful, as I've already moved on to making other changes once it finally gets back to me.

At the very least, can we add an environment variable like `DISABLE_ESLINT_DURING_WATCH=yup` that allows devs like me to opt-out of it?

toolness:
This fixes #1511 by not running eslint every time a file changes if `ESLINT_CHILL_OUT` is set.

### DIFF
--- a/docs/environment.md
+++ b/docs/environment.md
@@ -110,6 +110,10 @@ string), the boolean is true; otherwise, it's false.
   will be sent messages whenever certain kinds of
   [events](../slackbot/signals.py) occur in the app.
 
+* `ESLINT_CHILL_OUT` is a boolean; if true, it will change the behavior
+  of gulp's watch mode such that it doesn't run `eslint` every time a
+  file changes.
+
 [RoboBrowser]: http://robobrowser.readthedocs.io/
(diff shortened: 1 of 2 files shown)


toolness:
Sure... I hesitated at first because it's for development only, and our `environment.md` is getting kind of big, but I guess the solution to that is breaking it up into subsections rather than not documenting environment variables at all, eh?

PR #1533, opened 2017-03-30
Upgrade to jquery 1.12.4

jseppi:
Gemnasium was reporting (for other projects) that jquery 1.11 has critical security advisory related to  XSS:

> When text/javascript responses are received from cross-origin ajax requests not containing the option dataType, the result is executed in jQuery.globalEval potentially allowing an attacker to execute arbitrary code on the origin.

Ref: https://gemnasium.com/npms/jquery/versions/1.11.0

I don't see anywhere in our code that we're doing cross-origin requests with jquery so we are not affected by this security problem, though upgrading is still wise.

I've removed `jquery.xdomainrequest`, which is a plugin for supporting CORS for IE8 and 9 with jquery. But we aren't doing any cross-domain jquery requests since we switched to `xhr`, so it's not necessary. The `xhr` lib does issue cross-domain requests to our `api.data.gov` endpoint, but according to its docs it supports IE+ out of the box.

I've also merged in #1535


jseppi:
The reason gemnasium wasn't reporting for us was because we have jquery vendored. I'm going to see if I can get it installed and used from npm real quick also.

jseppi:
[Put jquery in package.json so we can track security releases](https://trello.com/c/f3akdjni/243-put-jquery-in-package-json-so-we-can-track-security-releases)

PR #1535, opened 2017-03-30
Move jquery to package.json

jseppi:
This PR is against https://github.com/18F/calc/pull/1533

It moves `jquery` to be listed in our `package.json` which will allow easier upgrades, and for vulnerabilities to be tracked by Gemnasium.


jseppi:
What do you think about this @toolness? 

It would of course be best to have everything installed from `npm` but that'll take a little more work since the versions of some of our libs are pretty old, and some of them require IE to test, and some of them have CSS that needs to also come along.

jseppi:
I'm just going to merge this so that there is only one PR to review :)

jseppi:
[Put jquery in package.json so we can track security releases](https://trello.com/c/f3akdjni/243-put-jquery-in-package-json-so-we-can-track-security-releases)

PR #1542, opened 2017-04-04
Upgrade to jquery 3.2.1

### LINKED ISSUE #1539: Upgrade jquery to >= 3.0
I misinterpreted the (confusing) gemnasium jquery XSS alert that I thought was fixed in jquery 1.12.4  (ref #1533). It actually looks like only jquery v3 has the fix (ref https://github.com/jquery/jquery/issues/2432#issuecomment-290983196 and https://nodesecurity.io/advisories/328).

I'll repeat here that we aren't using jquery to do any cross-origin ajax (we switched to `xhr` for the Data Explorer, which does cross-origin ajax to our own api.data.gov proxy), but it would be good to move to a maintained version of jquery. Unfortunately this would be jumping two major versions, so there is a potential for upgrade problems.

cc @toolness

jseppi:
Closes #1539

To Do:
- [x] confirm no breaking changes


jseppi:
Test failure looks to be with call to `event.stopPropagation()` in `upload.js`

jseppi:
Should be good, @toolness, but please give it a run-through!

PR #1544, opened 2017-04-13
Prepare CALC for migration to Django 1.10

toolness:
This PR *prepares* CALC for migration to Django 1.10--by which I mean that it makes low-impact, backwards-compatible changes to ensure that CALC's test suite passes on Django 1.10, but it doesn't actually *upgrade* CALC's version of Django. This is because we're about to put CALC on hiatus and might not be able to address any problems that might arise from actually migrating. But we do want to make it *easy* to migrate if we (or another team that inherits this project) need to upgrade once Django 1.9 reaches end-of-life. This will make it easier to eventually make our way to 1.11 (#1346), which is the next long-term support (LTS) release.

This is coupled with the changes made in our fork of djorm-ext-pgfulltext at https://github.com/18F/djorm-ext-pgfulltext/pull/3, so if we merge this PR, we should also merge that one. (If you need to make local edits to that fork and test them with CALC, you can follow the instructions at https://github.com/18F/calc/pull/1466#issuecomment-284104266.)

### DIFF
--- a/styleguide/views.py
+++ b/styleguide/views.py
@@ -20,12 +20,12 @@ def get_existing_filename_upload_form():
         js = forms.FileField(widget=UploadWidget(
             required=False,
             existing_filename='boop.csv',
-        ))
+        ), required=False)
         no_js = forms.FileField(widget=UploadWidget(
             degraded=True,
             required=False,
             existing_filename='boop.csv',
-        ))
+        ), required=False)
 
     return MyForm()
 
--- a/hourglass/templates/admin/base.html
+++ b/hourglass/templates/admin/base.html
@@ -24,13 +24,13 @@
 
 <body class="{% if is_popup %}popup {% endif %}card--wide{% block bodyclass %}{% endblock %}">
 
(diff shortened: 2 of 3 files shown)


styleguide/views.py
```diff
@@ -20,12 +20,12 @@ class MyForm(forms.Form):
         js = forms.FileField(widget=UploadWidget(
             required=False,
             existing_filename='boop.csv',
-        ))
+        ), required=False)
```
jseppi: Does it need this both here and up at line 21?

toolness: To be compatible with both Django 1.9 *and* 1.10, yes.  Once we migrate to 1.10, I think we can get rid of the `required` in the `UploadWidget`.  Note that I *think* we're already doing this in the actual upload form (outside of the styleguide) because the `FileField` arg dictates the behavior of server-side validation while the `UploadWidget` arg dictates the client-side rendering of the widget (i.e. the HTML5 `required` attribute).

PR #1547, opened 2017-04-13
Fix update_search_field args

### LINKED ISSUE #1546: manage.py update_search_field is not working
Noticed this was happening locally with `docker-update.sh` (which calls `update.sh`) and I checked the logs from the last deployment to `calc-dev` and saw it happened there also:

```
2017-04-13T10:05:34.92-0500 [APP/PROC/WEB/0]ERR usage: manage.py update_search_field [-h] [--version] [-v {0,1,2,3}]
2017-04-13T10:05:34.92-0500 [APP/PROC/WEB/0]ERR manage.py update_search_field: error: the following arguments are required: model
```

Looks like `model` needs to be specified as a param to the shadowed `update_search_field`

cc @toolness

jseppi:
Fixes #1546 

- Updates all `update_search_field` calls to include `Contract` as the `"model"` arg
- Adds `set -e` to `cf.sh` so that it will properly fail deployments when a subcommand has an error
- Unrelated change: Minor code cleanup suggested by @cmc333333 to move the `list_display` fields of `CustomUserAdmin` in `data_capture/admin.py` up to the class definition instead of in the constructor. Don't know what I was thinking there 😝 

Note that I thought about modifying our shadowed `update_search_field` to itself add the model param into the `kwargs` that get passed to the original `update_search_field`, but thought that was a little too whacky.

### DIFF
--- a/cf.sh
+++ b/cf.sh
@@ -1,11 +1,14 @@
 #!/bin/bash
+
+set -e
+
 if [ $CF_INSTANCE_INDEX = "0" ]; then
     echo "----- Migrating Database -----"
     python manage.py migrate --noinput
 
     echo "----- Updating search field -----"
     # The '-W ignore is to suppress https://github.com/18F/calc/issues/291.
-    python -W ignore manage.py update_search_field contracts
+    python -W ignore manage.py update_search_field contracts Contract
 
     echo "----- Initializing Groups -----"
     python manage.py initgroups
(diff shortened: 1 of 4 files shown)


toolness:
Cool!  I also fixed the underlying cause of #1546 in #1548 so you can merge that too.

My only concern about specifying the `Contract` model is that *technically* it does change the ouput of the `update_search_field` command:

```
Processing model <class 'contracts.models.Contract'>...Done
Processing model <class 'contracts.models.Contract_Deferred_business_size_contract_end_con3c5191c64f6bb1853aab160e8583eb47'>...Done
```

That second line isn't displayed if we specify `Contract` as a second argument. However, since we don't actually index anything off that model, I am pretty sure it doesn't matter.

PR #1549, opened 2017-04-14
Prepare CALC for migration to Django 1.11

toolness:
This prepares us for #1346 by making all our tests pass on Django 1.11.  As with #1544, it doesn't actually _upgrade_ us to Django 1.11, it just makes changes required for tests to pass under the new version, without breaking our current use of Django 1.9.

However, unlike #1544, the breaking changes in Django 1.11 related to [template-based widget rendering](https://docs.djangoproject.com/en/1.11/releases/1.11/#template-based-widget-rendering) are backwards-incompatible with earlier versions of Django. So I had to add branching logic that detects what version of Django we're on. 😞 

So to fully test both conditional branches, you need to run the test suite, then manually change `requirements.txt` to use Django 1.11, and run the tests again.

I'm not sure how I feel about the branching logic--or rather, I'm not sure whether we should just close this PR unmerged and later crib from it when we're ready to actually migrate to Django 1.11.

One advantage of keeping the branching logic, though, is that it better prepares our USWDS widgets for being reused by other Django projects. If we were to factor out the USWDS widgets into a separate third-party package, for example, it would probably be a good idea to keep the branching logic, so that clients could use it regardless of their Django version. As first-party CALC code, though, I'm not sure how much sense it makes to keep the branching logic.

### DIFF
--- a/frontend/widgets.py
+++ b/frontend/widgets.py
@@ -1,43 +1,57 @@
 from django import forms
 from django.utils.html import format_html
 
+from . import USE_NEW_FORM_API
 
-class LabelInputSiblingRenderer():
-    def render(self, name=None, value=None, attrs=None):
-        # This is mostly just a copy-paste of our superclass method, it
-        # just tweaks the HTML structure to be USWDS-friendly.
-        if self.id_for_label:
-            label_for = format_html(' for="{}"', self.id_for_label)
-        else:
-            raise ValueError('USWDS-style inputs must have "id" attributes')
-        attrs = dict(self.attrs, **attrs) if attrs else self.attrs
-        return format_html(
-            '{}<label{}>{}</label>', self.tag(attrs), label_for,
-            self.choice_label,
-        )
+if not USE_NEW_FORM_API:
+    class LabelInputSiblingRenderer():
+        def render(self, name=None, value=None, attrs=None):
+            # This is mostly just a copy-paste of our superclass method, it
+            # just tweaks the HTML structure to be USWDS-friendly.
+            if self.id_for_label:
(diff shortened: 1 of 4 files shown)


frontend/widgets.py
```diff
@@ -1,43 +1,57 @@
...
-
-class UswdsCheckboxFieldRenderer(forms.widgets.ChoiceFieldRenderer):
-    choice_input_class = UswdsCheckboxInput
-
-
-class UswdsCheckbox(forms.widgets.CheckboxSelectMultiple):
-    renderer = UswdsCheckboxFieldRenderer
+from . import USE_NEW_FORM_API
+
+if not USE_NEW_FORM_API:
```
toolness: Everything in this branch is unchanged from the old code, it's just indented an additional level.

jseppi:
Hrm, yeah, I think I'm more comfortable keeping this not-merged for now if that's cool with you.

toolness:
Update!

I've spent the past week making [django-uswds-forms](http://django-uswds-forms.readthedocs.io/en/latest/), which factors out our form stuff and makes it reusable. It also improves on it a bunch because Django 1.11's form rendering API allows for lots of customization that wasn't possible before.

Notes on migrating over to this:

* It doesn't use our `<uswds-date>` custom element *but* the template it uses for USWDS dates is quite flexible, so it will be easy for us to override.

* We'll be able to get rid of some of our CSS that makes Django's default classes--`errorlist`, `helptext`, etc--look like USWDS. This is because django-uswds-forms is able to tell Django 1.11 to use the official USWDS classes instead.

toolness:
Ok, I am closing this in favor of #1640.

PR #1550, opened 2017-04-14
Remove model arg in calls to update_search_field

jseppi:
PR against #1548 that removes specification of the `model` arg to calls to `update_search_field`, which was added in #1547.

Note that even with this change I am still seeing this output when running `manage.py update_search_field contract`, as in `update.sh`:

```
----- Updating search field -----
Updating normalized labor categories...
Updating full-text search indexes...
Processing model <class 'contracts.models.Contract'>...Done
Processing model <class 'contracts.models.Contract_Deferred_business_size_contract_end_con3c5191c64f6bb1853aab160e8583eb47'>...Done
```


jseppi:
Closing, not necessary.

PR #1554, opened 2017-04-14
Run aXe on styleguide during tests, fix violations

toolness:
Here u go.


frontend/templatetags/frontend.py
```diff
@@ -10,6 +10,7 @@
 # that screen-readers contextualize them properly.
 LEGEND_WIDGETS = (
     date.SplitDateWidget,
+    forms.CheckboxSelectMultiple,
     forms.RadioSelect
```
toolness: We were adding a `<legend>` to group radio boxes and the split date widget, but *not* to group related checkboxes, which aXe noticed, so this fixes that.

frontend/tests/axe.py
```diff
@@ -43,26 +43,31 @@ def run_and_validate(webdriver):
...
         var callback = arguments[arguments.length - 1];
-        axe.run(function(err, results) {
+        axe.run(options, function(err, results) {
           callback({
             error: err && err.toString(),
             results: results
           });
         });
-        ''')
+        ''', options)
```
toolness: I moved the rule-excluding logic to the JS rather than the post-processing because there were, uh, 976 color contrast violations on the styleguide and it was making the test runner time out.

frontend/upload.py
```diff
@@ -61,12 +61,15 @@ def render(self, name, value, attrs=None):
...
+        label_txt = 'Choose file'
+        final_attrs['aria-label'] = label_txt
+
         return "\n".join([
             '<upload-widget%s>' % flatatt(widget_attrs),
             '  <span class="nojs-preamble">%s</span>' % nojs_preamble,
             '  %s' % super().render(name, value, final_attrs),
             '  <div class="upload-chooser">',
-            '    <label for="%s">Choose file</label>' % id_for_label,
+            '    <label for="%s">%s</label>' % (id_for_label, label_txt),
```
toolness: This one is kind of weird--aXe thinks our upload widget doesn't have a `<label>`.  I'm not sure if that is because the label isn't a sibling of the input element or something else, but explicitly adding an `aria-label` with same value as the actual label fixed things.  (Aside from pleasing aXe, I figure this could also fix some screenreaders.)

styleguide/templates/styleguide.html
```diff
@@ -416,7 +416,7 @@ <h3 class="example-heading">Example</h3>
   <label for="input-type-text">Text input label</label>
   <input id="input-type-text" name="input-type-text" type="text">
 
-  <label for="input-has_focus">Text input focused</label>
+  <label for="input-focus">Text input focused</label>
   <input class="input-focus" id="input-focus" name="input-focus" type="text">
```
toolness: This was just a typo.

styleguide/views.py
```diff
@@ -12,7 +12,7 @@ def get_degraded_upload_widget():
     class MyForm(forms.Form):
         file = forms.FileField(widget=UploadWidget(degraded=True))
 
-    return MyForm()['file']
+    return MyForm(auto_id='degraded_upload_widget_%s')['file']
```
toolness: There were multiple instances of this form's upload widget on the style guide (one degraded, the other non-degraded) which meant that `id` attributes conflicted, which aXe caught.  This ensures the degraded version has a different `id` than the non-degraded one.

toolness:
@jseppi do you think you can review this one? it is kind of ancient but now that I've resolved merge conflicts, it should be good to go.

PR #1560, opened 2017-05-17
adding to update.sh so that things will work with 3.6

EricSchles:
I was having trouble following calc's installation instructions with python 3.6.  So I am updating the install script from:

pip install -r requirements.txt

to

python -m pip install -r requirements.txt 

So that I can install things.


toolness:
Looks good, I will merge once travis gives the 👍 !

PR #1563, opened 2017-08-28
Migrate to CircleCI

jseppi:
This PR migrates CALC from TravisCI, which we are no longer permitted to use, to CircleCI.

To Do:
- [x] change (almost) all references to Travis to Circle
- [x] remove Travis from configured GitHub integrations
- [x] create new deployer credentials for each space and save in Circle:
  - [x] dev
  - [x] staging
  - [x] prod

After merging:
- [x] file issue for not having Sauce/webdriver tests working with ie (https://github.com/18F/calc/issues/1566)
- [x] file issue to get nightly cron builds running at some point when Circle has them (https://github.com/18F/calc/issues/1565)
- [ ] turn off CALC in Travis
- [ ] mark CALC migration as complete in the Trello board
- [ ] delete old service deployer accounts in each cloud.gov space
- [ ] make Circle a required PR check on `develop` and `master`


jseppi:
How's this look, @toolness?

jseppi:
Builds are over at https://circleci.com/gh/18F/calc/tree/circleci

Not sure where the GitHub status report went...it used to be here o_O

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+          DATABASE_URL: postgres://postgres@localhost/hourglass
+
+          # Some of our integration tests either take a really long time to run or
+          # are very error-prone via Selenium, so prefer RoboBrowser for now.
+          TEST_WITH_ROBOBROWSER: yup
+
+          # These environment variables affect the running of Webdriver/Selenium
+          # tests. Note that if we're not in a trusted environment (e.g. a PR
+          # is being issued from an untrusted fork), we will default to using
+          # PhantomJS locally, rather than a real-world browser via Sauce Labs.
```
toolness: So does CircleCI basically ensure that this is still the case, e.g. that `SAUCE_ACCESS_KEY` or whatever isn't defined if the PR is from an untrusted user?

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+          DJANGO_LIVE_TEST_SERVER_ADDRESS: 127.0.0.1:8000
+
+          # TODO: Add this back in once we have time to get Sauce running in Circle
+          # WD_TESTING_BROWSER: "internet explorer"
+          # WD_TESTING_BROWSER_VERSION: 11.103
+          # WD_JOB_VISIBILITY: public
+          # WD_SOCKET_TIMEOUT: 60
```
toolness: I guess we should file an issue to make sure this isn't forgotten about once this PR is merged?

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+      - image: circleci/postgres:9.5
+        environment:
+          POSTGRES_USER: postgres
+          POSTGRES_DB: hourglass
+      - image: redis:3.0.7-alpine
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: v1-py-cache-{{ .Branch }}-{{ checksum "requirements-dev.txt" }}
```
toolness: Hmm, since `requirements-dev.txt` has `-r requirements.txt` in it, perhaps the cache key should also include the checksum of `requirements.txt` in it? Because if we change a non-dev requirement, that won't change `requirements-dev.txt`, which could result in CircleCI using a stale cache.

CHANGELOG.md
```diff
@@ -7,6 +7,29 @@ to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
+## [2.7.0][] - 2017-04-14
```
toolness: Huh, why is this part of this PR?  Or rather, why wasn't this in `CHANGELOG.md` before?

docs/monitoring.md
```diff
@@ -43,7 +43,4 @@ For more details on the `/healthcheck/` endpoint, see
 CALC has a suite of tests in [production_tests](../production_tests/) for
 checking that the production site is properly configured.
 
-These tests are run as part of our daily [Travis cron][travis-cron] build.
-
 [rq]: http://python-rq.org/
-[travis-cron]: https://docs.travis-ci.com/user/cron-jobs/
```
toolness: Hmm, I guess we should file an issue for restoring these lines (as well as uncommenting the lines in the circle YML) once CircleCI starts supporting scheduled builds, too?

.circleci/config.yml
jseppi: yes, you're right

jseppi:
You can run the `deploy-circle.sh` script locally by just setting the needed environment variables. I ran it from Circle in a few builds, setting the target branch to this branch (`circleci`) to make sure it ran. For example, see the "deploy to dev" step of [this build](https://circleci.com/gh/18F/calc/65). It failed because of the python version, but got far enough along to convince me that it works.

CHANGELOG.md
jseppi: I can't figure out what is going on there o_O It is in `master`, but it is not in `develop` for some reason.

Maybe I merged master into dev locally at some point in the past, but it never happened in the GitHub repo.

.circleci/config.yml
jseppi: That's right, from the [Advanced Settings](https://circleci.com/gh/18F/calc/edit#advanced-settings) screen:

screen shot 2017-08-30 at 10 15 03 am

jseppi:
Changes made, @toolness!

toolness:
Nice! Um, I am confused--how do I see just the changes you made since my last review?  Did you squash your commit history and force push to this repo or something? I'm only seeing one relevant commit, 41c06e663d098204624cca40d5a2144a64e14e3d, but it seems to contain the majority of the PR content, rather than just the changes made as a result of my review...

jseppi:
Yes, I squashed my changes. I've gotten into that habit per recommendations from @cm.

toolness:
wat

i don't like that recommendation :(  i get merging a PR by telling it to squash the changes into a single commit, but doing it within a PR makes it really hard to see what's changed since the previous review... 😣

jseppi:
It's what we do on Federalist also. I dunno, maybe I'm doing it wrong.

toolness:
Yeah, I disagree strongly with this approach to be honest. Aside from making the code review process harder, it also makes it difficult to follow the PR as a narrative in the future--for instance, if I come back to this PR to see what went on, I can see the tiny snippet of the old version of `circleci.yml` that I commented on regarding `requirements.txt`, but clicking on the filename to gain additional context takes me to https://github.com/18F/calc/pull/1563/files/56432dcfeb73f2624ceb2c8570658b53891fa38a#diff-1d37e48f9ceff6d8030570cd36286a61, which is actually a broken link (github says "we went looking everywhere, but couldn't find those commits") because the history got overwritten.

jseppi:
I got the Circle status reporter to either come back on or start back up. It was a setting in the Circle dashboard:

screen shot 2017-08-30 at 11 21 36 am

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+      - image: circleci/postgres:9.5
+        environment:
+          POSTGRES_USER: postgres
+          POSTGRES_DB: hourglass
+      - image: redis:3.0.7-alpine
+    steps:
+      - checkout
+
+      - restore_cache:
+          key: v1-py-cache-{{ .Branch }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "requirements.txt" }}
```
jseppi: added `-{{ checksum "requirements.txt" }}`

.circleci/config.yml
```diff
@@ -0,0 +1,164 @@
...
+          command: |
+            python3 -m venv ~/venv
+
+            echo ". ~/venv/bin/activate" >> $BASH_ENV
+            source $BASH_ENV
+
+            pip install -r requirements-dev.txt
+      - save_cache:
+          name: Save python deps cache
+          key: v1-py-cache-{{ .Branch }}-{{ checksum "requirements-dev.txt" }}-{{ checksum "requirements.txt" }}
```
jseppi: added `-{{ checksum "requirements.txt" }}`

docs/monitoring.md
```diff
@@ -43,7 +43,10 @@ For more details on the `/healthcheck/` endpoint, see
 CALC has a suite of tests in [production_tests](../production_tests/) for
 checking that the production site is properly configured.
 
-These tests are run as part of our daily [Travis cron][travis-cron] build.
+<!--
+TODO: Restore this information once cron jobs are supported and setup in
```
jseppi: added this TODO comment

jseppi:
I called out the changed lines @toolness

toolness:
Nice! This looks good to me... um, since this includes commits that are in `master` but not `dev`, I recommend just merging *without* squashing here, as squashing could muck with git's understanding of history and make our next merge from `dev` to `master` harder to understand... i dunno, it's just a gut feeling though, i could be wrong.

PR #1567, opened 2017-08-31
Prefix excel formula chars with apostrophe in query csv export text

jseppi:
The raw query string is written to the CSV export of price search results.
If the query starts with excel formula characters, then when the CSV
is opened in Excel, lots of (potentially bad) stuff can happen as Excel
processes the formula. This PR mitigates this issue by prefixing the query
string with an apostrophe if it starts with an Excel formular character.
That way, Excel will treat the cell content as plain text rather than a formula.

ref: https://github.com/18F/vuln-reports-private/issues/8

### DIFF
--- a/api/tests/test_rates_csv.py
+++ b/api/tests/test_rates_csv.py
@@ -5,14 +5,15 @@ from .test_rates_api import GetRatesTests
 
 RATES_CSV_PATH = '/api/rates/csv'
 
+
 # TODO: These tests need to be more fully fleshed out
 class GetRatesCSVTests(TestCase):
-  def setUp(self):
-    GetRatesTests.make_test_set()
-    self.path = RATES_CSV_PATH
+    def setUp(self):
+        GetRatesTests.make_test_set()
+        self.path = RATES_CSV_PATH
 
-  def test_prefixes_excel_formula_chars(self):
-    excel_formula_prefixes = ['@', '-', '=', '|', '%']
-    for char in excel_formula_prefixes:
-      resp = self.client.get(f'{self.path}/?q={char}query')
-      self.assertContains(resp, f"'{char}query")
+    def test_prefixes_excel_formula_chars(self):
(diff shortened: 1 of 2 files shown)


toolness:
Oof, looks like flake8 doesn't like your changes!

jseppi:
Yeah, incoming. Forgot to setup flake8 integration in vscode

api/views.py
```diff
@@ -223,6 +223,13 @@ def get(self, request, format=None):
         contracts_all = get_contracts_queryset(request.GET, wage_field)
 
         q = request.query_params.get('q', 'None')
+
+        # if the query starts with special chars that could be interpreted
+        # as parts of a formula by Excel, then prefix the query with
+        # an apostrophe so that Excel instead treats it as plain text
+        if q.startswith(('@', '-', '=', '|', '%')):
+            q = "'" + q
```
toolness: Is there a web page anywhere that verifies that this is a secure solution to the problem?  It's ok if not, just thought it'd be nice to link to if possible, since I don't know much about how Excel interprets data, and I suspect lots of other devs won't either...

jseppi: Good idea. Added a link.

api/tests/test_rates_csv.py
```diff
@@ -0,0 +1,21 @@
...
+class GetRatesCSVTests(TestCase):
+    def setUp(self):
+        GetRatesTests.make_test_set()
+        self.path = RATES_CSV_PATH
+
+    def test_prefixes_excel_formula_chars_in_query(self):
+        excel_formula_prefixes = ['@', '+', '-', '=', '|', '%']
+        for char in excel_formula_prefixes:
+            encoded_query = urllib.parse.quote(f'{char}query')
+            resp = self.client.get(f'{self.path}/?q={encoded_query}')
```
jseppi: I had to urlencode the query to get the test to work as expected when using a `+` character. If it wasn't encoded, then the plus character never made it to the view function.

PR #1571, opened 2017-09-19
[WIP] Fix prod deployments on tag

jseppi:
Unfortunately the way we were attempting to do production deploys in Circle like we were doing in Travis won't work as easily. I think we have to switch to a `workflows`-based build process.

Before merge:
- [ ] change repo settings to make `ci/circleci: build_and_test` a required status check and remove the `ci/circleci` status check


jseppi:
not a great path to go down right now, closing

PR #1579, opened 2017-09-20
improve pricelist status logging and add tests

### LINKED ISSUE #1578: Price list status log text should display the status text instead of code
screen shot 2017-09-20 at 8 56 04 am

In the above screenshot, `3` should instead be something like `REJECTED`. There also needs to be a space in between `3` and `by`.

jseppi:
- fix #1578
- show status text instead of code
- add tests to check that logging is happening when pl status changes


jseppi:
Will merge into `develop` but won't be able to promote this (or other code) up to production until after the current pentest is finished.

PR #1587, opened 2017-10-02
Fix PR link in docs/release.md

toolness:
I think this is due to the changes we recently made when switching to a branch release model instead of a tag-based one. Basically, the text says to issue a PR against `staging` but the PR link actually creates a PR against `master`.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -85,7 +85,7 @@ To release version 0.0.4 of CALC:
     git push https://github.com/18F/calc.git v0.0.4
     ```
 
-12. In GitHub, open a PR from `staging` to `master` called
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -108,5 +108,6 @@ Hooray, you're done!
 
 [unreleased]: https://github.com/18F/calc/blob/develop/CHANGELOG.md#unreleased
 [pr]: https://github.com/18F/calc/compare/staging...v0.0.4-rc
+[pr2]: https://github.com/18F/calc/compare/master...staging
```
jseppi: 👍

Issue #1599, opened 2017-10-10
Worker threads need static files


toolness:
Thanks to @jseppi's detective work, it appears that our worker processes, which process region 10 bulk upload data, are crashing because they need to send emails which link to our static assets:

```
2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR Traceback (most recent call last):
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/deps/0/python/lib/python3.6/site-packages/rq/worker.py", line 700, in perform_job
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR     rv = job.perform()
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/deps/0/python/lib/python3.6/site-packages/rq/job.py", line 500, in perform
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR     self._result = self.func(*self.args, **self.kwargs)
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/app/data_capture/jobs.py", line 77, in process_bulk_upload_and_send_email
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR     email.bulk_upload_failed(upload_source, tb)
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/app/data_capture/email.py", line 123, in wrapped
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR     return func(template, *args, **kwargs)
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/app/data_capture/email.py", line 288, in bulk_upload_failed
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR     to=[upload_source.submitter.email],
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/app/data_capture/email.py", line 84, in send_mail
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR     plaintext_message, html_message = render_mail(template, ctx)
   2017-10-06T06:32:12.47-0700 [APP/PROC/WEB/0] ERR   File "/home/vcap/app/data_capture/email.py", line 54, in render_mail
[...]

Issue #1594, opened 2017-10-04
Process bulk upload fails if Region10Loader.make_contract() returns None


toolness:
The bulk upload processing code in `data_capture/jobs.py` calls `Region10Loader.make_contract()` (in `contracts/loaders/region_10.py`) and expects it to always return a contract object or throw an exception. However, looking at the implementation of that function, there is an edge case where it actually returns `None`.  This later causes an exception to be raised, when code further down the pipeline attempts to call `update_normalized_labor_category` on `None`.

I think we can fix this by having the edge case raise a `ValueError` or a `ValidationError` instead of returning `None`.

PR #1601, opened 2017-10-12
[EXPERIMENT] Figure out why r10 bulk uploads require lots of memory

toolness:
After fixing #1594 and #1599, we found that the region 10 bulk upload was _still_ failing on Kelly's latest XLSX because, well, it seemed to just be running out of memory. We tried upgrading the dev instance's `rqworker` from 256mb to 512mb, which didn't fix things, but further upgrading to 1024mb *did* fix things.

So this PR is an investigation into why so much memory is being taken up, and whether there's anything we can do to alleviate it.

I'm starting with adding a `process_bulk_upload` management command and using Python's [`memory_profiler`](https://pypi.python.org/pypi/memory_profiler) to profile the amount of memory that the job is using:

```
python -m memory_profiler manage.py process_bulk_upload data_capture/static/data_capture/r10_export_sample.xlsx
```

(That just runs the bulk upload on the tiny r10 export sample in the repo; substitute the filename for a massive XLSX to test out the real thing.)

Here's the output when running the command on the real thing, though:

```
Line #    Mem usage    Increment   Line Contents
================================================
    18    122.6 MiB      0.0 MiB   @transaction.atomic
    19                             @profile
    20                             def _process_bulk_upload(upload_source):
    21    122.6 MiB      0.0 MiB       file = ContentFile(upload_source.original_file)
    22    122.6 MiB      0.0 MiB       converter = Region10SpreadsheetConverter(file)
    23
    24    122.6 MiB      0.0 MiB       contracts_logger.info("Deleting contract objects related to region 10.")
    25
    26                                 # Delete existing contracts identified by the same
    27                                 # procurement_center
    28    122.6 MiB      0.0 MiB       Contract.objects.filter(
    29    122.8 MiB      0.1 MiB           upload_source__procurement_center=BulkUploadContractSource.REGION_10
    30                                 ).delete()
    31
    32    122.8 MiB      0.0 MiB       contracts = []
    33    122.8 MiB      0.0 MiB       bad_rows = []
    34
    35    122.8 MiB      0.0 MiB       contracts_logger.info("Generating new contract objects.")
    36
    37    200.8 MiB     78.0 MiB       for row in converter.convert_next():
    38    200.8 MiB      0.0 MiB           try:
    39    200.8 MiB      0.0 MiB               c = Region10Loader.make_contract(row, upload_source=upload_source)
    40    200.8 MiB      0.0 MiB               contracts.append(c)
    41    191.3 MiB     -9.5 MiB           except (ValueError, ValidationError) as e:
    42    191.3 MiB      0.0 MiB               bad_rows.append(row)
    43
    44    200.8 MiB      9.5 MiB       contracts_logger.info("Saving new contract objects.")
    45
    46                                 # Save new contracts
    47    117.7 MiB    -83.0 MiB       Contract.objects.bulk_create(contracts)
    48
    49    119.2 MiB      1.5 MiB       contracts_logger.info("Updating full-text search indexes.")
    50
    51                                 # Update search field on Contract models
    52    103.6 MiB    -15.6 MiB       Contract._fts_manager.update_search_field()
    53
    54                                 # Update the upload_source
    55    103.6 MiB      0.0 MiB       upload_source.has_been_loaded = True
    56    154.6 MiB     51.0 MiB       upload_source.save()
    57
    58    154.6 MiB      0.0 MiB       return len(contracts), len(bad_rows)
```

So it's not using a *ton* of memory anywhere, but perhaps one of these functions it's calling is allocating a ton of memory but releasing it before returning?  I'll have to do some more digging.


toolness:
I may have alleviated the memory problem a bit in #1675.  Additional notes from that PR:

> If we really want to reduce memory usage _and_ speed things up, we might want to require Kelly to upload a CSV instead of an XLSX: my guess is that whatever Microsoft tool he's using on his end will have no problem doing that, and it will certainly reduce a lot of resource consumption on our end.

toolness:
I think that #1714 has basically fixed the memory issue, so I'm closing this.

PR #1631, opened 2018-03-19
Specify Python at minor version instead of patch

### LINKED ISSUE #1630: Pin to minor version of Python runtime
As @jmcarp [pointed out](https://github.com/18F/calc/commit/c42203a900e4ccf5358873b8d475f4c810c818a9#r28162037), we can pin to `python-3.6.x` in `runtime.txt` instead of to a specific patch version.

This seems like a great idea to me!

jseppi:
Closes #1630. See issue for details.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6.4
+FROM python:3.6
 
 ENV PHANTOMJS_VERSION 1.9.7
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.6.4
 
 ENV PHANTOMJS_VERSION 1.9.7
 
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -9,7 +9,7 @@ you may want to check out our alternative [Docker instructions](docker.md).
 You'll need the following tools and services installed to run CALC
 locally:
 
-* [Python 3.6.4](https://www.python.org/)
+* [Python 3.6](https://www.python.org/)
 * [Node 6.0](https://nodejs.org/)
 * [yarn](https://yarnpkg.com)
   * Install globally via `npm install -g yarn`
--- a/hourglass/tests/test_configuration.py
+++ b/hourglass/tests/test_configuration.py
@@ -31,6 +31,8 @@ class PythonVersionTests(TestCase):
 
     def test_dockerfile(self):
         with open(path('Dockerfile')) as f:
+            # We specify the patch version in our Dockerfile, but this
+            # test will just check down to the minor version level
             self.assertIn(f'FROM python:{self.python_version}', f.read())
 
     def test_circle_yml(self):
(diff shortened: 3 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6.4
```
toolness: Hmm, see the comment I made [here](https://github.com/18F/calc/issues/1630):

> To be honest I noticed there are some pretty significant changes between the first version of 3.6 and 3.6.2, so I wouldn't mind us making it explicit in Docker-related stuff, lest we use new features introduced in a patch release and then someone's dev environment breaks because they're on the older version of Python but there's no way for docker to know the minor version changed.

In other words, say 3.6.9 comes out and adds a very minor new feature that we take advantage of.  As soon as we push this to cloud.gov, cloud.gov will be fine because it knows a new version of Python has been released and will use the latest version.

HOWEVER, when a dev runs `./docker-update.sh` (or `docker-compose build`), the new version of python will _not_ be pulled from Docker Hub because docker will see the older version of 3.6 in its cache and just use that.  While there might be some fancy way to tell Docker to _always_ consult Docker Hub to see if a new patch release has come out, we should also realize this can slow things down, especially if a dev happens to be tethering... eh.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.6
+FROM python:3.6.4
```
jseppi: Changed this back to `3.6.4`, @toolness, and added a comment in the associated test.

Dockerfile
jseppi: Does Python really add features in patch version updates? Looks all like patches/bugfixes in https://docs.python.org/3/whatsnew/changelog.html

toolness: Yeah, this is where the line between "feature" and "bugfix" becomes really blurry: in my case, I was working on a personal project a few months ago and one of my machines had 3.6.0 while the other had 3.6.2.  I wrote my code on the one with 3.6.2, and unknowingly took advantage of the [zipfile does not support pathlib](https://bugs.python.org/issue28231) fix.  Then when I ran my code on the other machine, I got an exception, which actually took a little while to debug--I kept assuming the problem was occurring because of other environmental differences in the two computers' setups, and it wasn't after a while that I tried upgrading my Python version.

I saw this as a new feature, but I guess it could be interpreted as a bugfix because 3.6 was _supposed_ to support the use of `Path` objects instead of raw strings across the whole standard library, but apparently they forgot about the `zipfile` module. (This kind of thing is, I think, the basis of the arguments of some vociferous critics of semver.)

Anyways, I realize that updating the minor version number is kind of annoying but IMO it is still less annoying than running into the bug I ran into, because the one I ran into took more time to diagnose and fix 😫

jseppi:
Great, thanks for the enumeration of potential issues. I think the likelihood of those cases happening are sufficiently low to proceed.

One kind of big issue that this PR does fix is when the Cloud Foundry Python buildpack drops support for older Python versions, our deploys break (as I noticed today with the `3.6.2` version we were on). Given that the patch releases can and typically do contain security fixes, I think making this process more streamlined (ie, just re-running the last good build+deploy from Circle) makes this change worth it.

toolness:
Ah good point about the breakage!  Noice, sounds good.

Issue #1566, opened 2017-08-30
Add back sauce/web-driver tests to CI builds


jseppi:
During the migration to CircleCI (https://github.com/18F/calc/pull/1563) time was too short to get our tests running in WebDriver IE via Sauce like we had in Travis (ref: https://github.com/18F/calc/blob/302cc668e7325aa7a7752bf1ee80d9b4c1e83bb6/.travis.yml#L29 )

Refs:
- https://github.com/18F/calc/pull/1563/files#1d37e48f9ceff6d8030570cd36286a61R25
- https://github.com/18F/calc/pull/1563/files#1d37e48f9ceff6d8030570cd36286a61R135

toolness:
Fixed by #1669!

PR #1634, opened 2018-03-20
Actually upgrade to Django 1.10

### LINKED ISSUE #1633: Actually upgrade to Django 1.10
Um, so I was just going over our old issues from last year, and apparently we _prepared_ for migration to Django 1.10 in #1544, but at that time we were low on funding so we punted on actually upgrading (in case any breakages surfaced in production).

Now that we have some funding, we should probably upgrade to Django 1.10, and then bust forth to 1.11 as per #1346.

toolness:
Fixes #1633.

I _think_ this is all that's needed, since the real work was done in #1544?


jseppi:
hrm, a test is failing due to an already allocated port. Seems weird...

toolness:
@jseppi ok, I added fallback ports for Django to use, and that seems to have fixed things.  _However_, while grepping for `DJANGO_LIVE_TEST_SERVER_ADDRESS` I noticed that `test_selenium.py` is using it--and assuming it is a single `host:port` value, which is no longer the case.  But it's using it within a conditional that tests to see if `TRAVIS` is truthy, which can no longer be the case, because we're not on TRAVIS.  So ... maybe I should file an issue to clean that up later, and we should just merge this now?

jseppi:
Yeah, please file an issue to clean up that Travis+Selenium stuff. Here's a related issue from the Travis->Circle migration: https://github.com/18F/calc/issues/1566

PR #1640, opened 2018-03-21
Upgrade to Django 1.11, use uswds_forms.

toolness:
This upgrades us to Django 1.11 and also moves us to using [django-uswds-forms](http://django-uswds-forms.readthedocs.io/), since Django 1.11 breaks our custom widgets.

This PR supersedes #1549.

## Notes

* We override [uswds-forms' date template](https://github.com/18F/django-uswds-forms/blob/master/uswds_forms/templates/uswds_forms/date.html) so we can use our own `<uswds-date>` custom element. Our override is largely a copy-paste of the original with a few alterations.

* Our date widget uses USWDS classes that aren't prefixed with `usa-`, which is weird.  It's fine, since in order to use `<uswds-date>` I had to provide a custom template anyways, but it's still weird.  I think there might be a reason behind this that I'm simply not remembering... Ohhh, I guess it's because of [this](https://github.com/18F/calc/tree/develop/frontend/source/sass/uswds#readme).

* The input fields in our date widget used to use the `input-inline` class, and now they use the `usa-input-inline` class, but this seems to make no visual difference because we didn't have any CSS rules for `input-inline` to begin with (and we don't have any for `usa-input-inline` either).  Since switching to `input-inline` would take extra work, I decided not to pursue it, as it's apparently meaningless.

## To do

We can potentially punt some of these to issues.

- [x] Fix display errors w/ dates.
- [x] Get rid of any templates that are no longer used.
- [x] Ensure dates are using our `<uswds-date>` custom element.
- [x] Get rid of `frontend/widgets.py`. 
- [x] Get rid of `frontend/date.py`.
- [x] Add tests to ensure our custom date template works OK.
- [x] Update the style guide as needed, so that it doesn't refer to non-existent widgets.
- [ ] Consider getting rid of some of our CSS that makes Django's default classes--`errorlist`, `helptext`, etc--look like USWDS. (This is because django-uswds-forms is able to tell Django 1.11 to use the official USWDS classes instead.)
- [ ] Consider replacing our custom `{% fieldset %}` template tag with [`{% fieldset %}` from uswds-forms](http://django-uswds-forms.readthedocs.io/en/latest/reference.html#fieldset).
- [ ] Consider using [`uswds_forms.UswdsForm`](http://django-uswds-forms.readthedocs.io/en/latest/reference.html#uswds_forms.UswdsForm) and/or [`uswds_forms.UswdsErrorList`](http://django-uswds-forms.readthedocs.io/en/latest/reference.html#uswds_forms.UswdsErrorList), if they'd simplify our codebase.
- [x] Manually test stuff and make sure it all works okay.


toolness:
@jseppi I think this is good to go (I'd like to split out the last unfinished task into a separate ticket, if that's ok).

jseppi:
Yup, definitely ok to split out to separate task!

jseppi:
Minor thing: on the data widgets, the hint text used to have a slight gray color to it.

See screenshot - left is this PR, right is current styling:

screen shot 2018-03-21 at 9 41 27 am

toolness:
Oh good catch, just fixed that!

toolness:
Yay thanks!  We are now good to go with 1.10 until at least April 2020!

PR #1642, opened 2018-03-21
Upgrade to Redis 3.2 service

### LINKED ISSUE #1568: Upgrade cloud.gov redis service
From cloud.gov newsletter on 9/13/17:

> Redis 3.2, including both standard and high availability versions, are now available! If you use Redis, we encourage you to upgrade your application to use these beta version 3.2 services instead of the older alpha version 2.8 service.

jseppi:
Closes #1568.

This is the pre-work for upgrading our cloud.gov Redis service to version 3.2.

Before deploying this to each space, make sure that each space has a `calc-redis32` service created and bound, as outlined in https://github.com/18F/calc/issues/1568#issuecomment-374945736

New service created:
- [x] `dev`
- [ ] `staging`
- [ ] `prod`


jseppi:
We were in fact already using a `3.x` version in our tests and local environments because `2.8` was not available on docker hub, if I recall correctly.

PR #1650, opened 2018-03-21
Add app-level service bindings in each manifest

jseppi:
This should make it so our service instances are automatically bound to each Cloud Foundry app in each space.

Previously, service instances were manually bound and would have to be manually re-bound if they were disconnected. This change makes it so the binding is enforced with each deployment.

### DIFF
--- a/manifests/manifest-dev.yml
+++ b/manifests/manifest-dev.yml
@@ -1,5 +1,5 @@
 ---
-services:
+services: &services
 - calc-db
 - calc-env
 - calc-redis32
@@ -8,6 +8,7 @@ applications:
   instances: 1
   memory: 1G
   disk_quota: 1024M
+  services: *services
   routes:
     - route: calc-dev.app.cloud.gov
   buildpack: python_buildpack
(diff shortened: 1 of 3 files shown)


manifests/manifest-dev.yml
```diff
@@ -1,5 +1,5 @@
 ---
-services:
+services: &services
```
jseppi: I used this whacky YAML tag and merge syntax so that the array would only have to be defined once.

jseppi: You can try it out at a site like http://yaml-online-parser.appspot.com/

toolness: ah yeah, the ol' anchor syntax. CM originally used that a bunch in his dockerfilez.

jseppi: it feels so awkward to me >.<

toolness: yeah, it is weird. and another big downside is that it doesn't play very well with docker-compose's built-in way of extending compose files.

PR #1652, opened 2018-03-22
Remove djorm-exp-pgfulltext

### LINKED ISSUE #1645: Migrate off djorm-ext-pgfulltext to native Django full-text search
Now that we're on Django 1.10, we should move off djorm-ext-pgfulltext, as mentioned in https://github.com/18F/calc/issues/1603#issuecomment-336582208 !

To do this, it looks like we'll need to:

* Replace [`VectorField`](https://github.com/18F/djorm-ext-pgfulltext/blob/master/djorm_pgfulltext/fields.py#L15) with [`SearchVectorField`](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/search/#searchvectorfield).  It looks like these two fields are largely the same; looking at their source, they both have a `db_type()` method that returns `tsvector`, though `VectorField` has some additional methods.

* Make sure that the SQL queries sent to the database stay the same. For instance, right now doing a search for `cool engineer` on the front page results in a query containing the term:

    ```
    ( ("contracts_contract"."search_index") @@
      (to_tsquery('pg_catalog.english', 'cool:* & engineer:*')))
    ```

    We should make sure that stays the same when we migrate.

toolness:
This fixes #1645.

Implementing the fix involved some reading of [`djorm_pgfulltext`'s source code](https://github.com/18F/djorm-ext-pgfulltext/), as well as of [`django.contrib.postgres.search`](https://docs.djangoproject.com/en/2.0/_modules/django/contrib/postgres/search/).  A few things djorm-ext-pgfulltext made easy aren't supported out-of-the-box by Django's full-text-search, so I had to implement a solution.

## Notes

* `manage.py update_search_field` no longer delegates to a `djorm_pgfulltext` command.  It was easier to just hard-code the auto-updating of the `Contracts` model, so I removed the `<app-name> <model-name>` arguments to the command.

* Doing this work required examining the underlying SQL Django was sending to the database.  Hacking this into `settings.py` was cumbersome, so I added a boolean environment variable called `DEBUG_LOG_SQL` that can be used to toggle it, which should make future debugging easier too.

* Django's [`SearchQuery`](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/search/#searchquery) full-text search queries only translate to SQL as [`plainto_tsquery`](https://www.postgresql.org/docs/8.3/static/textsearch-controls.html#TEXTSEARCH-PARSING-QUERIES), whereas our search code has always been using `to_tsquery`.  As far as I can tell, `plainto_tsquery` doesn't let us specify to search for words that are _prefixes_ of labor category names.  For instance, searching for `eng` via a `plainto_tsquery` will only find results like "senior eng", but not "senior engineer".  So I had to implement a [custom `Value` subclass](https://docs.djangoproject.com/en/2.0/ref/models/expressions/#value-expressions) that used `to_tsquery` instead.

* There's no easy built-in way to tell Django to auto-update [`SearchVectorField`](https://docs.djangoproject.com/en/2.0/ref/contrib/postgres/search/#searchvectorfield)s on model save (in contrast, djorm-ext-pgfulltext did this kind of thing automatically for us).  In fact, Django's documentation tells us to just install a postgres trigger, which is kind of annoying, since we're not really creating new models very often for that to be a necessity.  Anyways, I dug into how djorm-ext-pgfulltext did things, which was via a post-save signal, and replicated the strategy in our own code.

* I made modifications to the initial migration of the `Contracts` model instead of adding a custom migration; this is because the initial migration included an import of `djorm_pgfulltext`, which I had to remove in order to decouple us from that package. And it's OK that I did this, because the underlying postgres database field remains unchanged: it's still just a `tsvector` under-the-hood.

### DIFF
--- a/contracts/management/commands/update_search_field.py
+++ b/contracts/management/commands/update_search_field.py
@@ -1,5 +1,4 @@
 from django.core.management.base import BaseCommand
-from django.contrib.postgres.search import SearchVector
 
 from contracts.models import Contract
 
@@ -15,5 +14,4 @@ class Command(BaseCommand):
         Contract.objects.bulk_update_normalized_labor_categories()
 
         print("Updating full-text search indexes...")
-        Contract.objects.update(
-            search_index=SearchVector('_normalized_labor_category'))
+        Contract.objects.update_search_index()
--- a/contracts/management/commands/update_search_field.py
+++ b/contracts/management/commands/update_search_field.py
@@ -1,19 +1,19 @@
-from djorm_pgfulltext.management.commands.update_search_field \
-    import Command as UpdateSearchFieldCommand
+from django.core.management.base import BaseCommand
+from django.contrib.postgres.search import SearchVector
 
 from contracts.models import Contract
 
 
-class Command(UpdateSearchFieldCommand):
+class Command(BaseCommand):
     def handle(self, *args, **kwargs):
         '''
-        This just shadows the djorm_pgfulltext command of the same
-        name, modifying it so that we normalize the labor categories
-        before updating the full-text search indexes.
+        This normalizes Contract labor categories
+        and then updates the full-text search indexes.
         '''
 
         print("Updating normalized labor categories...")
         Contract.objects.bulk_update_normalized_labor_categories()
 
         print("Updating full-text search indexes...")
-        super().handle(*args, **kwargs)
+        Contract.objects.update(
+            search_index=SearchVector('_normalized_labor_category'))
--- a/contracts/migrations/0010_auto_20150529_1956.py
+++ b/contracts/migrations/0010_auto_20150529_1956.py
@@ -2,7 +2,6 @@
 from __future__ import unicode_literals
 
 from django.db import models, migrations
-import djorm_pgfulltext.fields
 
 
 class Migration(migrations.Migration):
@@ -17,12 +16,6 @@ class Migration(migrations.Migration):
             name='contract_year',
             field=models.DecimalField(
                 null=True, max_digits=1, decimal_places=0, blank=True),
-            preserve_default=True,
-        ),
-        migrations.AlterField(
-            model_name='contract',
-            name='search_index',
-            field=djorm_pgfulltext.fields.VectorField(),
             preserve_default=True,
         ),
     ]
--- a/contracts/models.py
+++ b/contracts/models.py
@@ -283,7 +283,7 @@ class Contract(models.Model):
 
     _normalized_labor_category = models.TextField(db_index=True, blank=True)
 
-    search_index = SearchVectorField(null=True, db_index=True)
+    search_index = SearchVectorField(null=True, db_index=True, editable=False)
 
     upload_source = models.ForeignKey(
         BulkUploadContractSource,
--- a/contracts/models.py
+++ b/contracts/models.py
@@ -6,7 +6,7 @@ from decimal import Decimal
 from django.db import models, connection
 from django.contrib.auth.models import User
 from django.db.models.expressions import Value
-from django.contrib.postgres.search import SearchVectorField
+from django.contrib.postgres.search import SearchVectorField, SearchVector
 
 
 EDUCATION_CHOICES = (
@@ -119,6 +119,10 @@ class CurrentContractManager(models.Manager):
     def search(self, *args, **kwargs):
         return self.get_queryset().search(*args, **kwargs)
 
+    def update_search_index(self):
+        return self.update(
+            search_index=SearchVector('_normalized_labor_category'))
+
     def get_queryset(self):
         return ContractsQuerySet(self.model, using=self._db)\
             .filter(current_price__gt=0)\
@@ -147,6 +151,10 @@ class ContractsQuerySet(models.QuerySet):
 
     def search(self, query):
         return self.filter(search_index=query)
+
+    def update_search_index(self):
+        return self.update(
+            search_index=SearchVector('_normalized_labor_category'))
 
     def order_by(self, *args, **kwargs):
         edu_sort_sql = """
--- a/contracts/models.py
+++ b/contracts/models.py
@@ -5,8 +5,9 @@ from decimal import Decimal
 
 from django.db import models, connection
 from django.contrib.auth.models import User
-from djorm_pgfulltext.models import SearchManager, SearchQuerySet
-from djorm_pgfulltext.fields import VectorField
+from django.db.models.expressions import Value
+from django.contrib.postgres.search import SearchVectorField
+
 
 EDUCATION_CHOICES = (
     ('HS', 'High School'),
@@ -73,10 +74,7 @@ def convert_to_tsquery_union(queries):
     return " | ".join(queries)
 
 
-class CurrentContractManager(SearchManager):
-    # need to subclass the SearchManager we were using for postgres full text
-    # search instead of default
-
+class CurrentContractManager(models.Manager):
     def bulk_update_normalized_labor_categories(self):
         '''
         Iterate through all Contract models and update their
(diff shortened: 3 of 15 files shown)


hourglass/settings.py
```diff
@@ -236,7 +236,7 @@
     ),
 }
 
-LOGGING = {
+LOGGING: Dict[str, Any] = {
```
toolness: Had to appease mypy here: it was inferring `LOGGING` to be a dictionary mapping strings to _objects_, which are non-indexable.

Also, code climate seems to think this type annotation is the same thing as putting two statements on one line, which is super weird.

contracts/models.py
```diff
@@ -118,22 +116,45 @@ def bulk_create(self, contracts, *args, **kwargs):
...
+        return f"to_tsquery('pg_catalog.english', %s)", [tsquery]
+
+
+class ContractsQuerySet(models.QuerySet):
+
+    def multi_phrase_search(self, queries):
+        return self.search(MultiPhraseSearchQuery(queries))
+
+    def search(self, query):
+        return self.filter(search_index=query)
```
toolness: `search()` was a method on the manager provided by `djorm_pgfulltext`, and it was easiest to just tack it on to our manager.  I guess we could just remove it and inline this function's body if we think that makes things more readable, though.

contracts/migrations/0010_auto_20150529_1956.py
```diff
@@ -19,10 +18,4 @@ class Migration(migrations.Migration):
                 null=True, max_digits=1, decimal_places=0, blank=True),
             preserve_default=True,
         ),
-        migrations.AlterField(
-            model_name='contract',
-            name='search_index',
-            field=djorm_pgfulltext.fields.VectorField(),
-            preserve_default=True,
-        ),
```
toolness: I believe this `AlterField` was solely due to an upgrade of `djorm_pgfulltext` that made a bunch of formerly-explicit constructor arguments implicit.  In other words, this `AlterField` doesn't actually represent any kind of change to the underlying database schema--similar to how merely changing a field's `help_text` often triggers a migration, even though `help_text` doesn't have anything to do with the underlying database field.

contracts/signals.py
```diff
@@ -0,0 +1,10 @@
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+
+from .models import Contract
+
+
+@receiver(post_save, sender=Contract)
+def on_contract_save(sender, instance=None, **kwargs):
+    if instance:
+        Contract.objects.filter(pk=instance.id).update_search_index()
```
toolness: According to Django's documentation on [updating multiple objects at once](https://docs.djangoproject.com/en/2.0/topics/db/queries/#updating-multiple-objects-at-once), `update()` (which is called by `update_search_index()`) "doesn’t run any `save()` methods on your models, or emit the `pre_save` or `post_save` signals", so this won't erroneously cause an infinite loop.

contracts/migrations/0010_auto_20150529_1956.py
toolness: Note that actually tracing the cause for these lines is non-trivial; they were first added in #170, which doesn't make any changes to the `search_index` field at all.  Furthermore, from that time to the beginning of the project, `djorm_pgfulltext`'s entry in `requirements.txt` didn't include a version number, so it was hard to tell what version was being used at what time.

The only other evidence we have is that the change to `VectorField`'s constructor defaults was made in https://github.com/linuxlewis/djorm-ext-pgfulltext/commit/db8c220f0476bb6823204b9074909b33e844d9a4, which was in the [0.9.3 release](https://github.com/linuxlewis/djorm-ext-pgfulltext/releases/tag/0.9.3). which came out on March 31, 2015 (a few months before this migration was made).

Furthermore, the migration immediately preceding this one was made in #45 on January 9, 2015.  So given that 0.9.3 shipped between the preceding migration and this one, it makes sense that this one would reflect the changes to `VectorField`'s constructor.

Phew! So in other words, I think my hunch that this `AlterField` doesn't actually represent a change in the underlying db schema is correct.

toolness: Wait, there's something I'm _really_ confused about here... `null` defaulted to `True` prior to djorm-ext-pgfulltext 0.9.3, but it seems like that default changed to `False` in 0.9.3.  So maybe the underlying schema _did_ change in this migration... Arrgh so confusing.

toolness: Yup, this is super duper weird.  If I port the `test_schema_is_what_we_expect` test  I added in this PR over to `develop`, it fails, because the underlying `is_nullable` field in postgres is `NO` in `develop`, but `YES` in this PR.

Yet if I change `null=False` in this PR, a bunch of tests start failing with database constraint violations, claiming that the search index can't be set to `NULL`... ohh man this is weird.

toolness: Oops nevermind. When I was changing `null=False` to this PR's initial migration, I wasn't making a corresponding `null=False` modification to the `Contract` model.  Once I did that in 6be8f40a7ad4ca4c65a542d36f0d1abb6cc8a2ab everything seems to have parity with `develop`, I think.

toolness: So technically, if we really want to reflect past history with 100% fidelity, we could change the initial migration's `null` to `True`, and then re-add this `AlterField` in this migration, setting `null` to `False`, as that would properly reflect the (utterly non-obvious and probably unintentional) changes in this migration.  However, since it's not like we actually have any instances from 2015 that will actually need to run this migration to properly have its `is_nullable` flipped, I don't think it's a big deal.

contracts/models.py
```diff
@@ -118,22 +116,45 @@ def bulk_create(self, contracts, *args, **kwargs):
...
 
-    def multi_phrase_search(self, queries):
         if isinstance(queries, str):
             queries = [queries]
         queries = [
             Contract.normalize_labor_category(q)
             for q in queries
         ]
-        return self.search(convert_to_tsquery_union(queries), raw=True)
+        tsquery = convert_to_tsquery_union(queries)
```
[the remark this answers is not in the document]
jseppi: I'm having a bit of a brain-block on how to combine the terms best, but this is along the lines of what I'm thinking:

```py
def multi_phrase_search(self, queries):
    if isinstance(queries, str):
        queries = [queries]
    queries = [
        SearchQuery(Contract.normalize_labor_category(q))
        for q in queries
    ]

   # There has to be a better/more-pythonic way to do this combination
    combined_search_query = queries[0]
    for q in queries[1:]:
        combined_search_query |= q

    return self.search(combined_search_query)
```

then we could get rid of all our custom tsquery code.
Maybe best to handle in another PR though!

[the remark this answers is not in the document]
jseppi: :doh: This was your third bullet point in the PR write-up. I got too much into reading the code and totally missed that you already explored this option.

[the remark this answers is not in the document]
jseppi: I think that is what this Django feature request is about: https://code.djangoproject.com/ticket/28041

contracts/management/commands/update_search_field.py
```diff
@@ -1,19 +1,17 @@
-from djorm_pgfulltext.management.commands.update_search_field \
-    import Command as UpdateSearchFieldCommand
+from django.core.management.base import BaseCommand
 
 from contracts.models import Contract
 
 
-class Command(UpdateSearchFieldCommand):
+class Command(BaseCommand):
```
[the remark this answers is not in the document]
toolness: Hmm, I was thinking we could keep it the same, just in case we do start adding more indexed fields, we could bring back the previous behavior of the old command?

contracts/models.py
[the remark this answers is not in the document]
toolness: Oh cool, didn't realize there was a feature request.  That could be a fun way to contribute  upstream to Django... anyways, yeah, I originally tried exactly what you suggested but ALAS.

toolness:
Wooo thanks for reviewing this!

PR #1653, opened 2018-03-22
Combine jest and python coverage reporting

### LINKED ISSUE #1439: Report JS coverage in CodeClimate
For the Data Explorer React app at least.  `jest --coverage` will generate a coverage report, then we'd have to send the report to Code Climate through the `codeclimate-test-reporter`. Will need to investigate how to do this for JS in addition to what we already have for Python.

jseppi:
This PR adds combined test coverage reporting of our Jest tests (for the Data Explorer React code) with our py.test tests.

- upgrade `jest` version
  - also update all component snapshots since they render slightly differently (basically just formatting) with this new version
- add `coverage/` to ignore-dotfiles
- add jest config to `package.json`

Closes #1439

### DIFF
--- a/frontend/source/js/data-explorer/tests/__snapshots__/business-size.test.jsx.snap
+++ b/frontend/source/js/data-explorer/tests/__snapshots__/business-size.test.jsx.snap
@@ -1,8 +1,12 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
 exports[`<BusinessSize> matches snapshot 1`] = `
 <div
-  className="filter filter-business_size">
+  className="filter filter-business_size"
+>
   <label
-    htmlFor="zzz_business_size">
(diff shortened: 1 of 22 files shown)


frontend/source/js/data-explorer/tests/__snapshots__/business-size.test.jsx.snap
```diff
@@ -1,26 +1,34 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
 exports[`<BusinessSize> matches snapshot 1`] = `
 <div
-  className="filter filter-business_size">
+  className="filter filter-business_size"
+>
```
toolness: Oh weird, is the reason for this just b/c of the Jest upgrade?  I guess it's probably doing this so future diffs based on attribute changes are easier to read, huh...

PR #1654, opened 2018-03-22
Setup CircleCI workflows and nightly production tests

### LINKED ISSUE #1565: Add cron-type run of production tests once CircleCI supports it
Due to the migration to CircleCI (https://github.com/18F/calc/pull/1563), we no longer have daily cron builds to run our suite of production tests. CircleCI is apparently adding this type of functionality. We should use it once it is available.

jseppi:
This PR modifies our CircleCI config to add a workflow for running our production tests (which test various things at the production calc.gsa.gov URL) every night at 2:11AM UTC (10:11 PM ET).

closes #1565

Before merging, we should modify our GitHub branch config to remove the required `ci/circleci` check and replace it with a required `ci/circleci: build` check. This change is because now our main CircleCI "job" has a name due to changing to using CircleCI "workflows" (which is necessary for setting up scheduled builds). The old `ci/circleci` check will be gone so it appears like it is still waiting.


jseppi:
I did test the scheduled job by doing a commit (that I later rebased away) that had the cron time specified to a minute after the push, and it worked! Note the `/ nightly` workflow success in the screenshot below:

screen shot 2018-03-22 at 3 48 33 pm

.circleci/config.yml
```diff
@@ -168,3 +193,21 @@ jobs:
...
+  build-test-and-deploy:
+    jobs:
+      - build
+  nightly:
+    jobs:
+      - prod-test
+    triggers:
+      - schedule:
+          # every night at 2:11 AM UTC (10:11 PM ET)
+          cron: "11 02 * * *"
```
toolness: Thanks for adding the comment here, b/c cron syntax is so friggin' obtuse to me.

PR #1658, opened 2018-03-23
Ensure sphinx docs build on Windows without errors, and rebuild docs when changelog changes

toolness:
Right now our `docs` directory contains `changelog.md`, which is a symlink to the changelog in our repo's root directory.  However, symlinks aren't easily supported across current versions of Windows, which means the docs build, but not with the changelog in them (and with loads of annoying error output complaining about it).

This fixes things so that the changelog gets copied over to the `docs` directory at build time.  It also watches the changelog in the root directory, so that if that changes, the Sphinx docs are rebuilt (I don't think the watch logic accounts for this).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,6 +35,7 @@ release-*-instructions.txt
 release-*-instructions.html
 jest/*
 Dockerfile.cloud
+docs/changelog.md
 docs/_build
 docs/static
 data_capture_uploaded_files/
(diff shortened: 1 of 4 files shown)


toolness:
Nice! Yeah let's deal with moving it into a docs directory later. I thought changelogs were supposed to be in the root dir to make them easy to find (kind of like readmes) but yeah let's talk about it later.

PR #1660, opened 2018-03-23
Exclude all tests from Code Climate analysis

jseppi:
Adjusts our Code Climate config to remove all tests from Code Climate analysis. We had a lot of "issues" in CC about similar/duplicated code in tests, which is not necessarily a maintainability issue, and CC is overly aggressive with this check IMO.
I checked several other 18F projects and `test/` dirs are often excluded from CC analysis, so there is also precedence for this exclusion.


jseppi:
The 4 issues that Code Climate has for some reason identified here seem invalid to me. They are "similar code" issues on code that wasn't even touched here, so I'm also unsure why it is bringing them up as part of this change. Suggest merging without fixing those, pending approval from a PR reviewer.

PR #1663, opened 2018-03-23
Uglify vendored JS in production builds

### LINKED ISSUE #1555: Minify data-explorer.vendor.js in production
I just noticed that the Data Explorer's vendor bundle is not minified in production. Doing so will reduce its size.

Code needs to be amended at https://github.com/18F/calc/blob/develop/gulpfile.js#L79

jseppi:
When run locally, this reduced my data explorer vendor bundle from `143 Kb` to `77.9 Kb`. Pretty good savings!

Closes #1555.


jseppi:
On calc-dev, the bundle size decreased from `33.2 Kb` down to `21.9 Kb`

Issue #1639, opened 2018-03-21
Clean up travis-related code from test_selenium.py


toolness:
As mentioned in https://github.com/18F/calc/pull/1634#issuecomment-374762181, `test_selenium.py` has a bunch of logic at the beginning that checks to see if it's running under Travis, but we moved from Travis to CircleCI in #1563, so that code will never get called anymore. We should clean it up.

This is also related to #1566.

jseppi:
Also related to #1661

Issue #858, opened 2016-10-04
Travis CI should integrate w/ Sauce Labs using an account associated w/ CALC


toolness:
Currently we're logging into Sauce Labs with my GSA account, but if I leave 18F then the integration will explode.

So we should eventually move it to an account that's associated with CALC and not a particular person, if possible.

toolness:
18F has abandoned Sauce Labs, so I'm closing this.

Issue #857, opened 2016-10-04
Sauce Labs-based Selenium tests take a long time to run


toolness:
As mentioned in https://github.com/18F/calc/pull/841#issuecomment-251503470, running our Selenium tests on Sauce Labs takes about 84 seconds (compared to 15 seconds via PhantomJS) on AWS + Docker infrastructure. I haven't tried running them specifically on Travis yet, but I wouldn't expect them to be much faster, so for now we're just leaving them disabled. (We _are_ running QUnit tests via Selenium + Sauce Labs on Travis CI though, since that is still quite speedy.)

toolness:
18F has abandoned Sauce Labs, so I'm closing this.

PR #1667, opened 2018-03-26
Cleanup selenium tests

### LINKED ISSUE #1069: Add a Travis cron job to run nightly tests w/ sauce labs
As mentioned in #1052, it's difficult to use sauce labs on every push to the repository:

* The sauce labs tunnel frequently fails to start up, which, aside from leading to lots of false positives, also interferes with our deployments that are triggered on pushes.

* As mentioned in https://github.com/18F/calc/pull/1049#issuecomment-261621886, sauce labs takes a *really* long time to run selenium tests due to the back-and-forth that needs to occur when using sauce connect tunnels.

So it'd be nice if we could skip the selenium tests (or run them via `RoboBrowser` instead where possible) on normal pushes, but configure a nightly [cron job](https://docs.travis-ci.com/user/cron-jobs/) to run the tests using selenium on sauce labs. This way we'd get the best of both worlds.

It looks like setting up a Travis cron job involves emailing support@travis-ci.com to request that cron job support be activated for our repository.

### LINKED ISSUE #1661: Many disabled ("xtest_") tests in test_selenium.py - are they even needed?
I just noticed that in the old (from CALC 1 days) `test_selenium.py` tests, there are a lot of [disabled tests](https://github.com/18F/calc/blob/develop/frontend/tests/test_selenium.py#L296-L507) and a few notes about their fragility.

I'm wondering if these tests are even that useful and/or should be fixed at all. Upon first look, I don't think they have much value because the things they look to be testing are already covered by either Jest tests or by the backend python tests.

We might want to decommission that entire file and associated code to run it. Thoughts, @toolness or @hbillings?

toolness:
This fixes #1661 and #1639 by removing all `xtest_*` tests from `test_selenium.py`, and also removing all code referencing Travis and Sauce Labs (since Sauce Labs is no longer supported by 18F/TTS).

Since it gets rid of all Sauce Labs code, we should probably close #1069, #858, and #857 once this is merged.


toolness:
Oh good catch, somehow my grep tool missed those.  Just fixed the mentions of sauce.

PR #1669, opened 2018-03-26
Attempt to run Selenium tests on CircleCI.

toolness:
This is an experiment to see if CircleCI can reliably run selenium tests with PhantomJS.  As far as I can tell, it was only Travis that had problems in #330 (I was unable to reproduce the errors locally), so we might have better luck with Circle.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,8 +142,7 @@ jobs:
           name: Run python tests
           command: |
             # Note that this MUST come after building the front end assets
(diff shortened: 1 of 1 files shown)


toolness:
Seems to work at least this one time.  @jseppi , do you think we should merge this and see if #330 resurfaces over the next few days?

toolness:
If we merge this, we should probably close #330 and #1566, but if it starts intermittently failing, we should re-open them, I guess.

PR #1670, opened 2018-03-26
Use TypeScript's checkJs functionality on whitelisted files

### LINKED ISSUE #1527: Consider using Flow or TypeScript
So we've been using mypy for our python code and I feel like it's been useful.  It's especially nice because it's all standard Python 3.6, so it's not some "superset" of the language that we need to learn to use it effectively.  It also means there isn't an extra build step involved in compiling and running the code.

As much as I like TypeScript, it doesn't have these advantages, because it's a superset of JS that requires extra work for developers and it introduces an annoying build step.

An alternative to TypeScript that I haven't explored yet is Flow.  One of the interesting advantages it has over TypeScript is that it's possible to annotate code using *comments* instead of Flow's JS superset:

https://flowtype.org/en/docs/types/comments/

This could allow us to sprinkle in strong typing on our JS much like we've been doing for our python.  What do you think @jseppi ?

toolness:
This is an attempt to fix #1527 by using [TypeScript's `--checkJs` mode](https://github.com/Microsoft/TypeScript/wiki/Type-Checking-JavaScript-Files), which allows us to type-check JS files without needing to integrate TypeScript into our static asset build pipeline (to strip out type information).  It's basically a much more powerful alternative to eslint (well except for the part of eslint that overlaps with prettier).

There are reasons this could be a bad idea, though, so feel free to criticize and close this PR unmerged.

Right now we just type-check two JS files, `dap-hacks.js` and `usermenu.js`.

## Notes

* Use of TypeScript's `--checkJs` mode doesn't prohibit us from actually using "real" TypeScript in the future. I just thought this might be a quick, easy way to dip our toes into TypeScript.

* The original `dap-hacks.js` had a few workarounds to make eslint happy with it.  But because eslint and TypeScript are doing similar things, and because both linters really have different methods of appeasement for the weird things `dap-hacks.js` does, I've just forcibly told eslint to ignore this file for now. It's not a great solution but meh.

* A new file, `dap-hacks.d.ts` has been added. The `.d.ts` extension means it's a **TypeScript Declaration File** that just adds type annotations for pre-existing third-party code.  In this case, it's adding type annotations for the DAP script.  I don't think we should have to normally add `.d.ts` files to our repo, the DAP hacks script is just a really weird edge case.

## Notes about doing this with Visual Studio Code

* Not only are the JSDoc comments we add to our files parsed by TypeScript and used as type annotations, but their actual documentation shows up in tooltips in VS Code, which is nice:

    > 

* One kind of cool thing about VS Code's TypeScript (or perhaps just its built-in JS) add-on is that, if you have an existing function and start writing a comment above it, as long as you start the comment with `/**`, VS Code will automatically parse the function's declaration and include JSDoc comments for all the parameters.  This made adding JSDoc docs to the functions pretty easy!

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,3 +5,6 @@ htmlcov/
 vendor/
 
 frontend/static/frontend/built/
+
+# TypeScript checks these files, so ignore them.
+frontend/source/js/common/dap-hacks.js
--- a/.eslintignore
+++ b/.eslintignore
@@ -6,6 +6,3 @@ vendor/
 coverage/
 
 frontend/static/frontend/built/
-
-# TypeScript checks these files, so ignore them.
-frontend/source/js/common/dap-hacks.js
--- a/frontend/source/js/common/usermenu.js
+++ b/frontend/source/js/common/usermenu.js
@@ -10,14 +10,14 @@ function enableUsermenu() {
     return;
   }
 
-  trigger.setAttribute('aria-haspopup', true);
-  trigger.setAttribute('aria-expanded', false);
+  trigger.setAttribute('aria-haspopup', true.toString());
+  trigger.setAttribute('aria-expanded', false.toString());
   trigger.setAttribute('aria-label', openLabel);
 
   trigger.addEventListener('click', (e) => {
     menu.classList.toggle('is-open');
     const isOpen = menu.classList.contains('is-open');
-    trigger.setAttribute('aria-expanded', isOpen);
+    trigger.setAttribute('aria-expanded', isOpen.toString());
     trigger.setAttribute('aria-label', isOpen ? closeLabel : openLabel);
     e.preventDefault();
   });
--- a/frontend/source/js/common/usermenu.js
+++ b/frontend/source/js/common/usermenu.js
@@ -1,3 +1,4 @@
+// @ts-check
 /* global document */
 
 function enableUsermenu() {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -124,7 +124,7 @@ jobs:
       - run:
           name: Run linters and static analysis
           command: |
-            python manage.py ultratest flake8 eslint mypy bandit
+            python manage.py ultratest flake8 eslint mypy bandit typescript
       - run:
           name: Build front end and run front end tests
           command: |
--- a/frontend/source/js/common/dap-hacks.d.ts
+++ b/frontend/source/js/common/dap-hacks.d.ts
@@ -1,0 +1,10 @@
+interface Window {
+  /**
+   * This is a private function used by the DAP script. It
+   * happens to be attached to the global `window` object,
+   * which allows us to call it.
+   * 
+   * Yes, this is horrible.
+   */
+  _initAutoTracker: Function|undefined
+}
--- a/frontend/source/js/common/dap-hacks.js
+++ b/frontend/source/js/common/dap-hacks.js
@@ -2,13 +2,19 @@
 
 export const IS_SUPPORTED = 'MutationObserver' in window;
 
-const DEFAULT_INIT_AUTO_TRACKER_NAME = '_initAutoTracker';
-
 /**
  * Here we notify DAP about new links we've added to the page, so that
  * it can auto-track them. For more background, see:
  *
  *   https://github.com/digital-analytics-program/gov-wide-code/pull/47
+ */
+
+/**
+ * This returns all the <a> elements in a NodeList. It also searches
+ * through the descendants of nodes in the list and returns them.
+ * 
+ * @param {NodeList} list The list of nodes to filter/descend through.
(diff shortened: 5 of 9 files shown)


frontend/source/js/common/usermenu.js
```diff
@@ -10,14 +10,14 @@ function enableUsermenu() {
...
-  trigger.setAttribute('aria-expanded', false);
+  trigger.setAttribute('aria-haspopup', true.toString());
+  trigger.setAttribute('aria-expanded', false.toString());
   trigger.setAttribute('aria-label', openLabel);
 
   trigger.addEventListener('click', (e) => {
     menu.classList.toggle('is-open');
     const isOpen = menu.classList.contains('is-open');
-    trigger.setAttribute('aria-expanded', isOpen);
+    trigger.setAttribute('aria-expanded', isOpen.toString());
```
toolness: TypeScript really wants `setAttribute` to be passed a string, so we'll explicitly convert here, instead of assuming the browser will do it for us.  [According to MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Boolean/toString), a Boolean's `toString()` always returns either `"true"` or `"false"`, which is what we want.

.eslintignore
```diff
@@ -5,3 +5,6 @@ htmlcov/
 vendor/
 
 frontend/static/frontend/built/
+
+# TypeScript checks these files, so ignore them.
+frontend/source/js/common/dap-hacks.js
```
jseppi: hrm, I think we'd still want our JavaScript code style enforced on these files, if possible.

toolness: So the challenging thing about eslint is that a bunch of its non-aesthetic checks are sometimes at odds with TypeScript's, so this is going to result in us writing comments to tell either of them to ignore certain lines of code, which is kind of nasty.

Hmm, I'm looking at #1493 now, and wondering... since it exclusively deals with aesthetics in a way that doesn't make me want to pull my hair out, and since TypeScript exclusively deals with semantics, maybe we could have the whitelisted files use TypeScript and prettier, but not eslint?

Erg that sounds kind of complicated.

toolness:
@jseppi aside from the eslint issue, what do you think of this PR?  Should we just skip to straight-up TypeScript, or go with this kind of `--checkJs` solution for now, or neither?

.eslintignore
jseppi: Yeah, I don't think having multiple (mostly) style-checkers/formatters is a great idea. That'll be confusing to people, especially future people.

jseppi:
hrm, I don't have a great feel for how useful this will be vs. being a burden/barrier for whoever the project eventually ends up with, especially when applied/implemented piecemeal. Like it seems it would be fine and beneficial if all JS was checked, but it also seems like it'll be kind of a large effort to bring existing code into compliance.

What do you think the effort to get a clean result from `--checkJs` vs migrating totally to TypeScript would be?

I do really like the function hinting from the JSDoc comments and think we should should adopt them throughout our JavaScript code.

toolness:
Ok, so I've changed things around a bit for our use case:

* We now use TypeScript's `// @ts-check` comment at the top of any file that we want to opt-in to JS checking.  This is a lot easier than figuring out the file's path and adding it to a JSON file at the root of the repo.

* I re-introduced eslint into `dap-hacks.js`, it wasn't so bad really.  The main takeaway was to use [ESLint environments](https://eslint.org/docs/user-guide/configuring#specifying-environments) to specify the globals that were lying around; previously, we'd just define a `/* global window */` and then namespace everything off `window`, but TypeScript doesn't like that (e.g., it's OK with `HTMLElement` being a global, but not with `window.HTMLElement`--it might have something to do with `window` being _both_ a WebIDL interface _and_ the global object in browsers, but I'm not sure).

    In any case, even if we stick with ESLint and don't use TypeScript at all, I think specifying the browser environment and using its globals directly is probably better, since it will ensure that ESLint catches any typos--for instance, ESLint will catch `HTMLElment` in the browser environment, but it won't catch `window.HTMLElment`.

I _think_ this might be good enough for our needs?  ESlint runs on the exact same files as before, it's super easy to add new files for TypeScript to check, and if we ever decide that TypeScript just complicates things and gets in the way, we can simply remove it from our CI regimen (and ideally from `package.json` etc too) and we are good to go.

toolness:
Er, also, to answer your earlier question:

> What do you think the effort to get a clean result from `--checkJs` vs migrating totally to TypeScript would be?

Hmm, right now I think they'd probably be pretty similar amounts of work, honestly.  The workload could maybe be reduced by using very a loose TypeScript configuration, e.g. disabling `noImplicitAny`, which sort of parallels our mypy approach, I guess.  I could look into that.

toolness:
Oops you replied while I was typing hehe. Ok cool!  I will merge this but I can still look into loosening our TypeScript config and running `--checkJs` on all our code, if we want.

PR #1672, opened 2018-03-27
Use WDS

### LINKED ISSUE #1657: Replace current WDS implementation
The way we're using the WDS right now is not updateable. Basically, I copied the bits of the system we needed at the time and pasted them into correspondingly named files in `frontend/source/sass/uswds/`. My thinking was that we're actually using very little of the WDS and I didn't want to glob a huge bunch of unnecessary code onto the site. (Also, as @toolness reminds me, building on to the existing bespoke code was tricky at the time.) But it's also impossible to update, and it's a pain to extend or to take advantage of parts of the WDS that aren't already here. Also, none of the current styles have the `usa` prefix, which brings up issues mentioned in #1646.

I'm thinking we should rip out the current implementation and replace it with an `npm install` version of the system. I still don't like the bloat, but I guess it seems better to have the bloat than a fragile system. What say you all?

hbillings:
Closes #1657.

Historically, there were reasons we couldn't just implement the WDS (namely, legacy usage of Skeleton conflicted with WDS' grid and styles, and that was out of scope for us to address in the last go-round of CALC). But for maintenance reasons, it would be much easier to have the WDS available in a way that's easy to keep updated. So here we are.


hbillings:
Circle is complaining about a couple things that mystify me. 
1) That it can't find Source Sans Pro. ```The CSS file 'frontend/built/style/main.min.css' references a file which could not be found:
E                     frontend/built/fonts/sourcesanspro-light-webfont.eot```
I can't find anywhere that references this folder. `frontend/built/fonts` doesn't exist. 

2) Axe errors? `Exception: axe.run() found violations: link-name` but I'm not sure how to tell what it's looking at that is a violation.
3) `ValueError: /home/circleci/calc/frontend/source/sass/uswds/_forms.scss does not exist` No, no it does not. Why is it looking for something that I removed?

toolness:
Doh!

Regarding the aXe errors, unfortunately the error feedback isn't particularly friendly, as it's just a JSON dump of whatever aXe has to say in the "captured stdout" section below the error text... In this case there's a `html` key in there that mentions `<a href="mailto:"></a>`... does that seem familiar? I guess that needs some link text or something.

jseppi:
Hrm, I am also missing various fonts when I run locally:

screen shot 2018-03-28 at 8 24 39 am

I'll try to figure out what the deal is.

toolness:
With regards to Source Sans Pro, my _guess_ is that it's complaining because the fonts referenced by the USWDS aren't being copied over to the final destination directory where static assets will be served from.

For example, in my [`uswds-simple-sass`](https://github.com/18F/uswds-simple-sass) repo that I made several months ago, I had a [step in the gulpfile](https://github.com/18F/uswds-simple-sass/blob/master/gulpfile.js#L14-L17) that copied over all of USWDS's fonts and images from their `node_modules` locations to the place where my static assets were supposed to be:

```js
gulp.task('copy-uswds-assets', () => {
  return gulp.src(`${USWDS_DIST}/@(js|fonts|img)/**/**`)
  .pipe(gulp.dest('./vendor/uswds'));
});
```

My guess is that we need to add something like this to our own gulpfile.

toolness:
With regards to the `_forms.scss` not existing, that's an error being raised by our styleguide test to prevent documentation rot: something in our style guide is mentioning `_forms.scss`, and that test is making sure it exists, so that we don't have stale references in our docs to non-existent artifacts in our codebase.  But it doesn't exist, which is why it's complaining.

Looking at the styleguide now, I'm betting it's this:

> forms

so we should fix that part of the docs to reference whatever file the SCSS has been moved to, or remove it entirely.

jseppi:
Yup, that looks right to me, @toolness.

As for @hbillings item 3, that's because the styleguide references `_forms.scss`: https://github.com/18F/calc/blob/hjb-use-uswds/styleguide/templates/styleguide.html#L411

jseppi:
ha, jinx 😆@toolness

jseppi:
Also, @hbillings, remember we have our sweet test suite `ULTRATEST` that can be run via `docker-compose run app python manage.py ultratest`. It should report all the same things that you're seeing in CircleCI.

hbillings:
This is so helpful; thank you both! I think I've fixed everything except the axe error -- that's being caused by this line:
```
 We're always looking for feedback on CALC!<br>
To share your thoughts, email us at<a href="mailto:{{ HELP_EMAIL }}">{{ HELP_EMAIL }}</a>
```
I guess `HELP_EMAIL` isn't defined when it runs the test or something? I could throw the whole thing in an `{% if HELP_EMAIL %}` statement, but I'm not sure that's the way to go. Thoughts?

toolness:
Oh weird... I'm confused as to why `HELP_EMAIL` (or `DEFAULT_FROM_EMAIL`, which is what the setting falls back to) wouldn't be defined on your branch, but would be defined in all our other branches... I think wrapping it in an `{% if HELP_EMAIL %}` would make the test pass but we should figure out what's causing that to be empty.  I'll try investigating a bit.

toolness:
Umm I just looked at your PR and am quite flummoxed as to why `HELP_EMAIL` is empty on this branch but not others.

toolness:
@hbillings when you run `ultratest` locally, does the aXe error come up?

hbillings:
@toolness Weirdly, this last time Circle ran, it did not surface the same aXe error. (I also don't see it locally. Actually, I have completely different test errors locally than on Circle. Wat.)

jseppi:
I think this is still a WIP, but since I was tagged for review I did a quick visual check. Some issues I noticed:

Account dropdown placement is a little too high:
image

Default `:visited` coloration on Download link:
image

Data Explorer main data table is now bordered and does not have alternating-row coloration:
image

Links disappear on hover:
disappearing_links

Main logo is squashed/stretched:
image

Tables in Admin interface also all have borders now:
image

hbillings:
Ugh. So. I added the snippet that @toolness linked to for compiling the WDS, except it seems like `collectstatic.py` is looking for the assets at `frontend/built/fonts/` and the code snippet copies WDS to `frontend/built/vendor/uswds`. The latter seems like a better pattern to me (specifically namespacing WDS assets to a WDS folder), but I'm not sure how to make that work with collectstatic...

toolness:
Oh, so I think this is actually a problem with where the USWDS CSS rules end up relative to the fonts, rather than a problem with `collectstatic.py` (or rather, I think `collectstatic.py` is probably raising the error because it's following the `url()` links in the CSS).  

IIRC, I _think_ the USWDS might define some kind of SASS variable that lets us change the relative paths to static assets referenced by the CSS...

Ooo yeah!  Check it out, I think it's called `$asset-path`:

https://github.com/uswds/uswds/blob/develop/src/stylesheets/core/_variables.scss#L123-L124

If we could override the value of that in our own SASS, then I _think_ we can put our USWDS assets in a nicer place.

hbillings:
@toolness & @jseppi, I think this is ready for review again. Thank you both so much for all your help!

jseppi:
Some more visual review issues I noticed. I think the biggest one is the date widget not fitting its numbers.

Double :focus styling (glow and dashed border) on all inputs/checkboxes/selects/buttons/etc. Example:

image

Price List Upload - input is squared, select is rounded:

image

Price List Upload step 1 - margin on contract number heading looks a bit large. This happens on other Price List upload "cards" as well:

image

Date widgets - numbers don't fit in the boxes:

image

[Site admin](http://localhost:8000/admin/data_capture/unreviewedpricelist/) - rounded select doesn't fit well with button:

image

Site admin table "checkbox" column heading underline is off by a couple pixels:

image

For the others, maybe it would be best to handle them in a separate issue/PR?

frontend/source/sass/components/_alerts.scss
```diff
@@ -79,13 +105,13 @@ alerts-widget:focus {
 
 /* for Django admin */
 .messagelist li {
-  @extend .alert;
+  @extend .usa-alert;
```
jseppi: I think this didn't quite work. I see a `.messagelist > li.info` when I approve a price list in the admin interface, but the styling is just gray. I _think_ it used to have some color and look like our other alerts:

screen shot 2018-03-29 at 2 39 10 pm

screen shot 2018-03-29 at 2 40 33 pm

frontend/source/sass/components/_inputs.scss
```diff
@@ -35,6 +34,7 @@ select {
 
 select {
   border-radius: $border-radius;
+  background-image: none !important;
```
jseppi: maybe leave a comment about this being a WDS override

hbillings:
@jseppi I think I've addressed almost everything you mentioned, save the alerts. I noticed some more issues with those and opened #1686 to address further.

jseppi:
Whoops, sorry about the `yarn.lock` conflict, @hbillings. It turns out new versions of `yarn` can actually do the merge conflict resolution for you! 

You'll just have to do the merge of `develop` -> `hjb-use-uswds` locally, run `yarn` while in git's merge-conflict-resolution flow, commit, and push. I think something like this:

```sh
git checkout develop
git pull
git checkout hjb-use-uswds
git merge develop
# you should get a git merge conflict message here
yarn
# yarn will say it is fixing the merge conflict
git add yarn.lock
git commit
git push
```

PR #1677, opened 2018-03-27
[WIP] General style updates

hbillings:


hbillings:
I think we should look at breaking this into multiple pull requests. The first one would have to be the general style updates that the PR mentions (color, typography, that sort of thing). But then the header could be its own thing -- and the login in the header is a whole discussion we need to have, and that shouldn't hold up other things from getting merged. The card structure could be another pull request, and the admin nav a third.

hbillings:
closed in favor of more granular pull requests.

PR #1678, opened 2018-03-28
Upgrade `dj-email-url` for easy TLS email connections

jseppi:
`dj-email-url` v0.1.0 [(changelog)](https://github.com/migonzalvar/dj-email-url#010---2018-03-24) automatically sets `EMAIL_USE_TLS` setting when using the `submission:` "protocol" in the email connection string. This setting is required for `STARTTLS`, which is what we need for use with Amazon SES.

### DIFF
--- a/hourglass/settings.py
+++ b/hourglass/settings.py
@@ -73,6 +73,9 @@ DEFAULT_FROM_EMAIL = os.environ['DEFAULT_FROM_EMAIL']
 SERVER_EMAIL = os.environ['SERVER_EMAIL']
 HELP_EMAIL = os.environ.get('HELP_EMAIL', DEFAULT_FROM_EMAIL)
 
+# Use a secure TLS connection for SMTP email sending
+EMAIL_USE_TLS = True
+
 API_HOST = os.environ.get('API_HOST', '/api/')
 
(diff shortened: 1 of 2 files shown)


toolness:
wait, don’t we use https://github.com/migonzalvar/dj-email-url ?  i think we can specify thst config param in the url?

jseppi:
ah, good catch. Didn't think to look there. Apparently we can use `submission:` instead of `smtp:` in the config url: https://github.com/migonzalvar/dj-email-url#smtp-backend

jseppi:
hrm, just tried and it didn't work without the `EMAIL_USE_TLS` explicitly set in `settings.py`...

jseppi:
Oh! Looks like that feature was just added in the version of dj-djang-url released 4 days ago: https://github.com/migonzalvar/dj-email-url/releases

jseppi:
There you go, @toolness!

Note that the `EMAIL_URL` setting will need to start with `submission:` instead of `smtp:`, e.g.:

```sh
EMAIL_URL=submission://<AWS_ID_KEY>:<AWS_SECRET_KEY>@email-smtp.us-east-1.amazonaws.com
```

PR #1679, opened 2018-03-28
Use `yarn` instead of `npm run` everywhere

### LINKED ISSUE #1541: Use yarn to run npm scripts
Just noticed that with yarn you can run any of your npm scripts, and it leaves out all the unhelpful npm cruft.

We should update `.travis.yml`, docs, etc to recommend using `yarn <script-name>` instead of `npm run <script-name>`.

jseppi:
We switched to `yarn` a while back but didn't update various docs and tooling to use it in place of `npm` or `npm run`.

Closes #1541.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -117,9 +117,9 @@ jobs:
           name: Display versions
           command: |
             echo $(node --version)
+            echo $(npm --version)
             echo $(yarn --version)
             echo $(phantomjs --version)
-            echo $(npm --version)
             echo $(pwd)
       - run:
           name: Run linters and static analysis
--- a/docs/frontend.md
+++ b/docs/frontend.md
@@ -24,7 +24,7 @@ pipeline.
 If you just want to build the assets once without watching for changes, run:
 
 ```sh
-npm run gulp -- build
+yarn gulp build
 ```
 
 ## Developing the front end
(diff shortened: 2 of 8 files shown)


.circleci/config.yml
```diff
@@ -117,9 +117,9 @@ jobs:
           name: Display versions
           command: |
             echo $(node --version)
+            echo $(npm --version)
```
jseppi: just moved this to group it a little better with `node`

docs/frontend.md
```diff
@@ -49,19 +49,19 @@ The explorer's test suite uses [Jest][], and the tests are located in
...
 
 ```
-npm run test:watch
+yarn test:watch
 ```
 
 Finally, you can also run `jest` directly. If you're using Docker,
 this can be done via `docker-compose run app jest`; otherwise you can
-use `npm run test --`, followed by any
+use `yarn test`, followed by any
```
jseppi: `yarn`  doesn't need the `--` argument demarcation

docs/setup.md
```diff
@@ -13,7 +13,7 @@ locally:
   * If not using the [Docker-based development workflow](docker.md), make sure your system `python` command runs Python 3. You can check this by running `python --version`.
 * [Node 6.0](https://nodejs.org/)
 * [yarn](https://yarnpkg.com)
-  * Install globally via `npm install -g yarn`
+  * Install globally via `npm install -g yarn` or via another method described at [https://yarnpkg.com/lang/en/docs/install/](https://yarnpkg.com/lang/en/docs/install/).
```
jseppi: add link to other installation options

PR #1687, opened 2018-03-30
[WIP] Upgrade babel and webpack

jseppi:
**work-in-progress, open against #1685 currently**

also switch to babel-preset-env, which will target the latest 2 browser versions

TODO:
- [ ] figure out why test is failing
- [ ] do some manual testing/investigation to make sure everything looks good

### DIFF
--- a/frontend/gulp/webpack-util.js
+++ b/frontend/gulp/webpack-util.js
@@ -28,13 +28,8 @@ exports.webpackify = ({ isWatching, isProd }) => {
     }),
   ];
 
-  if (isProd) {
-    plugins.push(new webpack.optimize.UglifyJsPlugin({
-      sourceMap: true,
-    }));
-  }
-
   return webpackStream({
+    mode: isProd ? 'production' : 'development',
     watch: isWatching,
     watchOptions: USE_POLLING ? {
       aggregateTimeout: 300,
@@ -55,7 +50,7 @@ exports.webpackify = ({ isWatching, isProd }) => {
           exclude: /node_modules/,
           loader: 'babel-loader',
           options: {
-            presets: ['es2015', 'react'],
+            presets: ['env', 'react'],
             cacheDirectory: true,
           },
         },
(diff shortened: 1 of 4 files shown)


frontend/gulp/tests/webpack-util.test.js
```diff
@@ -1,3 +1,7 @@
+/**
+ * @jest-environment node
```
jseppi: required, otherwise `jest` uses `jsdom`, which doesn't have the same `setTimeout` capabilities, apparently.

toolness: Heh, I learned too, this while using TypeScript and expecting `setTimeout()` to return a number but instead getting this funky [`Timeout` class](https://nodejs.org/api/timers.html#timers_class_timeout).  Wacky!

frontend/gulp/webpack-util.js
```diff
@@ -28,13 +28,8 @@ exports.webpackify = ({ isWatching, isProd }) => {
...
   ];
 
-  if (isProd) {
-    plugins.push(new webpack.optimize.UglifyJsPlugin({
-      sourceMap: true,
-    }));
-  }
-
   return webpackStream({
+    mode: isProd ? 'production' : 'development',
```
toolness: Interesting, I [read up on `mode`](https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a) and it looks like it might allow us to get rid of more of our code that branches based on `isProd`:

* Apparently `process.env.NODE_ENV` is now automatically set to either `"development"` or `"production"` based on `mode`.  While we currently use the `DefinePlugin` to set it to `""` or `"production"`, we might want to consider removing that logic entirely, or adding a `TODO` that mentions it may be redundant.

* `mode` has its own defaults for `devtool`, but our overrides are different and seem to work well for us, so I don't think we should worry about it.

jseppi: Good catch! Just removed `DefinePlugin` and made a minor adjustment to our webpack tests.

PR #1699, opened 2018-04-03
Add new color palette and update header styling

ericronne:
Implements new styles in header, updates some colors, etc.


data_explorer/templates/index.html
```diff
@@ -20,31 +20,20 @@
...
-      <p>Fully burdened costs</p>
-    </div>
-    <div class="icon icon-professional-services">
-      <p>Services data</p>
-    </div>
-    <div class="icon icon-worldwide-rates">
-      <p>Worldwide rates</p>
-    </div>
+  <div class="hero-callout four columns">
+    <p><b>The Contract-Awarded Labor Category</b> (CALC) tool helps federal contracting officers and others find awarded prices to use in negotiations for labor contracts. It offers ceiling prices, fully burdened costs, services data, and worldwide rates.
```
jseppi: We could instead use the `<abbr>` element here maybe:

```html
 <abbr title="Contract-Awarded Labor Category">CALC</abbr>
```

data_explorer/templates/index.html
```diff
@@ -20,31 +20,20 @@
...
+    </p>
   </div>
 </div>
 {% endblock header_extension %}
 
 {% block body %}
-<div data-embed-jsx-app-here class="card--half-padding">
+<div data-embed-jsx-app-here class="card--narrow"> <!--change to card--wide-->
```
jseppi: Is the comment here a To-do? Does it still need to be done?

frontend/source/sass/base/_variables.scss
```diff
@@ -4,51 +4,52 @@
 // Set the base path for assets (used for font and image paths)
 $asset-path: '../vendor/uswds/' !default;
 
-/* #044364: used for header background fallback */
+/* #092d39: used for header background fallback */
```
jseppi: could this comment and the one below just be removed? seem like they could easily become doc-rot.

data_explorer/templates/index.html
jseppi: just a suggestion, not necessary obviously

frontend/source/sass/base/_variables.scss
```diff
@@ -63,6 +64,8 @@ $heading-line-height: 1.3;
 $small-line-height: 1.3;
 
 $font-family-main: 'Libre Franklin', 'Helvetica', 'Arial', sans-serif;
+$font-sans: $font-family-main;
+$font-serif: $font-family-main;
```
jseppi: Both the `sans` and `serif` fonts have serifs. Is that intended?

jseppi: also are the `$font-sans` and `$font-serif` variable overrides from the WDS? Maybe we should batch all the WDS-overrides into a grouped section for easier finding.

frontend/source/sass/components/_footer.scss
```diff
@@ -27,7 +27,7 @@ footer {
 }
 
 .divider_footer {
-  border-left: 1px solid $color-gray-light;
+  border-left: 4px solid #a0ad64; //add a variable for this color
```
jseppi: to-do?

data_explorer/templates/index.html
hbillings: I didn't even know that was a thing!

frontend/source/sass/base/_variables.scss
hbillings: We should override the `h` font families instead (I think that by default they inherit sans styles and we want serif, or vice versa). 

Most of the variables in this section are WDS overrides, but there are some key ones that aren't (spacing units). Separating them into two files seems like the right move. I'm inclined to keep WDS ones in this file (since WDS names its variables file the same thing) and move our custom variables to something like `_calc_variables.scss`

ericronne: I defer to y'all to decide upon the best way to override, but yes, the design uses no serif fonts.

data_explorer/templates/index.html
ericronne: ooh. ok thx!

data_explorer/templates/index.html
ericronne: yeah a to-do, will be a part of #1702

frontend/source/sass/base/_variables.scss
ericronne: i don't know what function that ever played, if any
likely none

frontend/source/sass/components/_footer.scss
ericronne: yep will fix with #1705

frontend/source/sass/base/_variables.scss
hbillings: Originally, because the header fallback color wasn't used anywhere else, this was just a comment to track that that color existed.

frontend/source/sass/base/_variables.scss
hbillings: Pushed a fix for this.

data_explorer/templates/index.html
ericronne: `<abbr>` is cool. Could be useful at some point, but it introduces some odd formatting:
image

Also i do want to explicitly spell out what CALC means here, since it's the first instance of the acronym.

PR #1700, opened 2018-04-03
Adds footer styling

ericronne:
Adds footer styles and deletes uswds _buttons.scss file.


frontend/source/sass/components/_footer.scss
```diff
@@ -1,20 +1,16 @@
 @import '../base/variables';
 
 footer {
-  color: $color-white;
-  background: $color-gray-medium-dark;
+  background: #cddc86; //add a variable for this
```
jseppi: looks like a to-do comment was not to-done

frontend/source/sass/components/_footer.scss
```diff
@@ -27,7 +23,7 @@ footer {
 }
 
 .divider_footer {
-  border-left: 1px solid $color-gray-light;
+  border-left: 1px solid #a0ad64; //add a variable for this
```
jseppi: to-do?

ericronne:
Correct. Further palette work is coming in a day or two; planning to handle it then. I wouldn't let that block merging here, since it works.

frontend/source/sass/components/_footer.scss
ericronne: Doing in #1705

frontend/source/sass/components/_footer.scss
ericronne: Doing in #1705

PR #1701, opened 2018-04-03
Adds login functionality to header: discuss! (/test?)

ericronne:
Duplicates the login function which was previously only in the footer, far down on the home page. 

We should discuss the pros and cons of including it in this high-profile location, and keep or kill it accordingly.

screen shot 2018-04-04 at 12 48 08 pm


hbillings:
One of the things I think we need to discuss is what this would do for users who don't have an account. (We talked about that a bit previously, but we never came to a conclusion.)

ericronne:
Totes. A previous version had a "request access" link in the footer, never implemented …
image

ericronne:
Let's put this on a research branch, and test with certain users.

ericronne:
Insufficient evidence that users want it, and users who have login rights generally bookmark the page after logging in.

PR #1709, opened 2018-04-04
Adds footer styling, take 2

toolness:
This is a funky attempt to resurrect #1700 so we can merge it into `develop` instead of the other branch. It might end up being the exact same thing as #1707, I'm not sure...


toolness:
Ok yeah, this seems to be identical to #1707 so I am just gonna close this, and we'll merge #1707!

PR #1711, opened 2018-04-04
Updates colors

### LINKED ISSUE #1705: Finish updating color variables
… and make names more meaningful / useful

ericronne:
Should bring colors much closer to a consistent, usable palette.
Closes #1705


ericronne:
@hbillings this might be a good time to pair, to walk you through the changes and get some pointers on ironing out nitpicky bits!

ericronne:
Sure let's!
_except for the blasted CircleCI complaints_ @toolness ;)

ericronne:
@jseppi / @toolness any ideas why this is failing the test?
i merged dev into the branch — which fixed a similar-looking error on another branch last week — but that didn't do the trick this time

frontend/source/sass/base/_variables.scss
```diff
@@ -5,41 +5,46 @@
...
+$color-green-darkest:        #61701c; // CALC-specific variable
+$color-green-dark:           #a0ad64; // CALC-specific variable
+$color-green:                #bbcb70; // Modified from USWDS
+$color-green-light:          #ccdc86; // Modified from USWDS
+$color-green-lighter:        #dae6a3; // Modified from USWDS
+$color-green-bright:         #ccde61; // CALC-specific variable
+
+$color-primary-alt:          $color-green-bright; // Modified from USWDS
+$color-secondary:            $color-green-bright; // Modified from USWDS
+// Solve later: disconnect between ^this^ and other "secondary" colors
```
hbillings: Let's open an issue for this and remove the comment.

frontend/source/sass/base/_variables.scss
```diff
@@ -5,41 +5,46 @@
...
-
-$color-green:                #61701c;
-$color-green-light:          $color-green; // for wds compat
-$color-green-lightest:       #e6edc2;
-
-$color-focus:                $color-primary-alt;
-$color-visited:              #4c2c92;
-
-$color-error-bg: #fad0d0;
+$color-primary:              #2D6580; // Modified from USWDS
```
hbillings: I think for clarity we should preface this entire list of colors with a comment. Maybe something along the lines of:
```
/* These variables are primarily overrides for U.S. Web Design System color variables.
 * However, some variables are custom to this project. They follow the general naming scheme
 * that the USWDS uses, but are noted as "CALC-specific variables."
 * There are also some variables that have been brought over directly from the USWDS and have
 * retained their original values; these are noted as "No change; including for visual reference." */
```

This has the added benefit of allowing us to strip out all of the `Modified from USWDS` comments.

frontend/source/sass/base/_variables.scss
```diff
@@ -5,41 +5,46 @@
...
+$color-gray:                 #547d8c; // Modified from USWDS
+$color-gray-medium:          #7da1b0; // Modified from USWDS
+$color-gray-light:           #c5d6de; // Modified from USWDS
+$color-gray-lighter:         #dfe7ee; // Modified from USWDS
+$color-gray-lightest:        #f5f8fa; // Modified from USWDS
+
+$color-gold:                 #fdb81e; // No change from USWDS, including for visual reference
+$color-gold-darkest:         #825c1e; // CALC-specific variable
+$color-gold-lightest:        #fbeebe; // Modified from USWDS
+$color-visited:              #4c2c92; // No change from USWDS, including for visual reference
```
hbillings: You did WERK here and it looks amazing!

frontend/source/sass/components/_alerts.scss
```diff
@@ -1,5 +1,5 @@
 @import '../base/variables';
-$color-error-bg: #fef5f5;
+$color-red-lightest: #fef5f5;
```
hbillings: I know originally this was my hack for getting around only using this color in one place; does it still make sense or can we use the `$color-red-lightest` from the variables file? If not, does it make sense to keep it as `$color-error-bg` instead of overriding a variable that already exists? Just wondering if that will be confusing to anyone reading the source in the future.

frontend/source/sass/base/_variables.scss
ericronne: 👍

jseppi:
@ericronne: The test failures on Circle are because the admin pages are referencing a file that now does not exist. See in the CircleCI build logs the errors all eventually say something like 

```
frontend.crotchety.CrotchetyFileNotFoundError: Cannot find frontend/built/style/admin/overrides.min.css"
```

You should get the same results locally when running `docker-compose run app python manage.py ultratest`.


This is because `frontend/built/style/admin/overrides.min.css`, which is created from `frontend/source/sass/admin/overrides.scss`, no longer exists but is still referenced somewhere.

Removing the line that references `overrides.min.css` from **`hourglass/templates/admin/base.html`** should fix the broke tests.

frontend/source/sass/components/_alerts.scss
ericronne: Your call, i can't say i'm following 100% here 😬

jseppi:
Ok, my previous assessment was wrong. The file _should_ still exist, but doesn't due to a sass error during the `Build front end and run front end tests` step of the CircleCI build:

```
Error in plugin 'sass'
Message:
    frontend/source/sass/admin/overrides.scss
Error: Undefined variable: "$color-gray-medium-dark".
        on line 72 of frontend/source/sass/admin/overrides.scss
>>        border-bottom: 2px solid $color-gray-medium-dark;
   --------------------------------^
```

Basically `$color-gray-medium-dark` is not defined anywhere. It should probably be in `frontend/source/sass/base/_variables.scss`.

I'm not sure why the build did not break on that `sass` error. That's a bug in the build process. I'll open another issue for that.

ericronne:
@jseppi good catch!

ericronne:
🎉

frontend/source/sass/components/_alerts.scss
jseppi: Good catch, @hbillings. I'm not very familiar with sass variable scoping rules. Does setting it here end up overriding it only for this file, or for everywhere? Either way, it is confusing to be overriding a variable that already exists in `_variables.scss`

hbillings: @jseppi It will only override it for this file, I think. All of the other files import `_variables.scss`, so that should reset it locally for each file. Mostly I'm concerned about two things: 1) the confusion it could cause to override this -- it's easy to miss that this variable is redefined in this file. 2) the fact that we already have a `color-red-lightest` and it's off from this variable juuuuust a little. It'd be nice if we just used a standard color here. @ericronne does that make more sense, and if so, what are your Opinions?

ericronne: Ah. `$color-red-lightest` *is* the correct color for warning backgrounds. I killed the override.

PR #1719, opened 2018-04-05
[WIP] Replace .card classes with body classes

hbillings:
**NOTE** This PR is against `er-nav-and-card-gridding`, not `develop`.

The `.card` model was originally only supposed to apply to data capture, but ended up needing to be used on the whole site. Because the original model only applied to content widths, the header and footer would have needed to be brought in line manually. @ericronne ran into this while implementing his new styles. This rejiggers the original thing into something more reusable.


ericronne:
Whoa awesome!!!

Only the nav should shift, btw. The logo, user email, and footer should always be at the full width.

hbillings:
@jseppi I would love to have the eyes of someone who is not me review this, plz!

PR #1714, opened 2018-04-05
Process bulk uploads in batches

toolness:
This potentially fixes #1712 by chopping up the insertion of new contracts into batches of 5000 labor categories each.  It also reduces the lifetime of the `Region10SpreadsheetConverter` instance, so that it won't be around anymore when search indexes are updated.

While I wasn't able to definitively profile memory usage in #1601, I'm hoping that this sufficiently reduces memory usage to the point that our rq_workers don't crash anymore.

This PR also brings in the `process_bulk_upload` management command from #1601 to make future manual testing and profiling easier.

Note that this PR doesn't have ideal test coverage.  I don't want to write tests if this doesn't actually fix #1712, so I'm going to hold off on writing them until we actually try this out on cloudfoundry.

Here's example log output of what running `manage.py process_bulk_upload` with Kelly's latest R10 export looks like:

```
[05/Apr/2018 12:07:45] INFO [contracts:45] Deleting contract objects related to region 10.
[05/Apr/2018 12:07:49] INFO [contracts:53] Generating new contract objects.
[05/Apr/2018 12:08:04] INFO [contracts:63] Saved 5000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:07] INFO [contracts:63] Saved 10000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:10] INFO [contracts:63] Saved 15000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:13] INFO [contracts:63] Saved 20000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:16] INFO [contracts:63] Saved 25000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:19] INFO [contracts:63] Saved 30000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:22] INFO [contracts:63] Saved 35000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:25] INFO [contracts:63] Saved 40000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:29] INFO [contracts:63] Saved 45000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:32] INFO [contracts:63] Saved 50000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:35] INFO [contracts:63] Saved 55000 contracts so far (0 bad rows found).
[05/Apr/2018 12:08:36] INFO [contracts:63] Saved 55672 contracts so far (0 bad rows found).
[...]

PR #1729, opened 2018-04-06
Tag and release v2.8.6 to staging

toolness:
- Fixed bulk upload to use less memory (#1714).


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@ to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
 
+- Fixed bulk upload to use less memory (#1714).
```
jseppi: usually these are under a subheading, like `### Changed` but I think it's fine without

toolness: yeah it's an EMERGENCY HOTFIX so let's leave the heading out to indicate URGENCY

toolness: actually nvm i just added the heading in 605f2b9.

Issue #1716, opened 2018-04-05
Icon credits are no longer needed (provided Kelly approves the header redo)


ericronne:
- [x] Get Kelly's 👌 on the header redesign, in particular the intro paragraph which replaces the current set of "icon bulleted" text …
![image](https://user-images.githubusercontent.com/9259185/38378716-29c0b01a-38c4-11e8-9925-27755efd5092.png)

- [x] Delete the icon credits from the about page …
![image](https://user-images.githubusercontent.com/9259185/38378746-442a4c36-38c4-11e8-8224-f11ad8c15273.png)

KellyBailey:
I like it. Let's go with it.

Kelly Bailey,MBA,PMP

Technical Product Manager,

CALC.GSA.GOV

Office of Professional Services

and Human Capital Categories(PSHC)

Federal Acquisition Service (FAS)

U.S. General Services Administration (GSA)

Phone: 253-347-9702

jseppi:
credits were removed in https://github.com/18F/calc/pull/1730. All done!

PR #1730, opened 2018-04-06
Removes the (now) superfluous icon credits card from About

ericronne:


ericronne:
This will close #1716

Issue #1692, opened 2018-04-03
"next" parameter in API results is wrong in production, and also undocumented


toolness:
```
curl 'https://api.data.gov/gsa/calc/rates/?q=engineer&query_type=match_all'
```

(Aside: I'm surprised this even works--I thought an `api_key` had to be provided here, but I guess not?)

The JSON response will include a `next` key whose value is:

```
https://calc-prod.app.cloud.gov/api/rates/?page=2&q=engineer&query_type=match_all
```

However, running `curl` on it returns a 403 with the response `{"detail":"Authentication credentials were not provided."}`.

I assume this is because `calc-prod.app.cloud.gov`'s `/api/` endpoint filters by IP address, since it's only supposed to be accessed by the API umbrella reverse proxy at api.data.gov.

Ideally, the `next` argument should point at `api.data.gov` and the correct path.

As a workaround, though, clients can just parse the `page` querystring argument from `next` and construct the next URL manually by taking their original request and adding the page to its querystring.

Aside from that, I also noticed that we don't seem to document `next` in our API docs.

toolness:
Thanks for the clarification on `api_key`!

I think this ultimately has the same root problem as #1694.  Basically, anything being served from `https://api.data.gov/gsa/calc/` thinks it's coming from `https://calc-prod.app.cloud.gov/api/`, which is going to massively mess with any URLs generated in the response.

PR #1734, opened 2018-04-09
Add a load_api_data management command

toolness:
This adds a `load_api_data` management command, as discussed in #1028.

Here's the command-line help, with the universal options (e.g. verbosity) removed:

```
usage: manage.py load_api_data [-s START_PAGE]
                               [-e END_PAGE] [--append] [-u URL]

Load rate data from the API of another CALC instance.

optional arguments:
  -s START_PAGE, --start-page START_PAGE
                        start page (default is 1)
  -e END_PAGE, --end-page END_PAGE
                        end page (default is to read all pages)
  --append              append to existing rates (instead of deleting them
                        first)
  -u URL, --url URL     URL of CALC API (default is
                        https://api.data.gov/gsa/calc/rates/)
```

## Example usage

This will load the first 5 pages of API data (given that our pagination size is currently hard-coded to 200, it will retrieve 1000 rates):

```
python manage.py load_api_data -e 5
```

This will append the next five pages to the database, without wiping out the previously loaded data:

```
python manage.py load_api_data -s 6 -e 10 --append
```

## Notes

* Right now I don't think it's possible to ask the CALC API to change its pagination size.  Enabling this is pretty easy and could make this management command run quite a bit faster, for when developers want to transfer the whole database.

* The command isn't wrapped in an atomic transaction. Since the command is intended primarily for development, I figured it'd be OK if devs just ran the command with no arguments and then pressed <kbd>Ctrl</kbd> + <kbd>C</kbd> once they got tired of waiting.  This would result in some data in the database, at least.  But if this is unexpected, and we really want the database to be rolled back, then we can do that.  Alternatively, if we really want, we could prompt the user, asking them if they want to commit or rollback, but this would require adding additional tests.

* Right now one disadvantage of this command in comparison to `load_data` is that because the data is remote, it's not quite as easy to re-populate it if one has to wipe their DB.  I guess one could use the built-in `dumpdata` command to dump their data and re-load it later, though.  Hmm.

## To do

- [x] Add tests.
- [x] Do we want to rename the command? I wasn't sure if `load_api_data`, `load_from_api`, `loadapi`, or something else was best.


api/serializers.py
```diff
@@ -2,10 +2,36 @@
 from rest_framework import serializers
 
 
-class ContractSerializer(serializers.ModelSerializer):
+class ContractSerializer(serializers.ListSerializer):
+    def create(self, validated_data):
+        return Contract.objects.bulk_create([
+            Contract(**item) for item in validated_data
+        ])
```
toolness: By default, `ListSerializer` doesn't bulk create, so this subclass speeds things up a lot.

contracts/management/commands/load_api_data.py
```diff
@@ -0,0 +1,88 @@
...
+from api.serializers import ContractSerializer
+
+
+def iter_api_pages(url, start_page=1, end_page=None):
+    page = start_page
+    i = 0
+    while True:
+        if end_page is not None and page > end_page:
+            break
+        res = requests.get(f"{url}?page={page}").json()
```
toolness: Ideally we would just use `res['next']` for this value, but alas for #1692.  It also means that devs couldn't add to the querystring in the `--url` argument to the command, e.g. to load their CALC with the results of a certain search query, though I don't even know if that's a desired use case.  But still, just pointing it out.

toolness:
@jseppi @hbillings what do you think about this so far?  If it looks 👍 then I will add tests but otherwise we can talk about how to improve it.

jseppi:
Looks pretty good to me!

toolness:
@jseppi ok cool!  so the plan is...

1. Add tests for this, review and merge this PR.
2. Update developer docs to have developers run this new command instead of `load_data` and `load_s70`.
3. Remove `load_data` and `load_s70` from the repository.

Sounds reasonable?

toolness:
uggh selenium

toolness:
@jseppi do you think we should move this new `load_api_data` command from `contracts` to `api`?  I don't _think_ `contracts` currently has any dependencies on API, but putting this command in it will make that the case...

Anyways, aside from that I think this is finally ready for review.

.circleci/config.yml
```diff
@@ -142,7 +142,8 @@ jobs:
           name: Run python tests
           command: |
             # Note that this MUST come after building the front end assets
-            py.test --cov=. --cov-report xml:./coverage/python/coverage.xml
+            # For details on why we're ignoring selenium tests, see https://github.com/18F/calc/issues/330
```
jseppi: 😭

api/serializers.py
jseppi: That's some good pythoning right there

api/serializers.py
```diff
@@ -2,10 +2,36 @@
...
+        if not edu_code:
+            raise serializers.ValidationError(
+                f'Invalid education level: {data}'
+            )
+
+        return edu_code
+
+
+class ContractSerializer(serializers.ModelSerializer):
+    # TODO: Figure out whether we actually support a null value for
```
jseppi: Want to make a separate issue for this so we don't have a dangling `TODO`?

contracts/management/commands/load_api_data.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def process_pages(self, pages, dry_run):
+        num_rates = 0
+        num_pages = 0
+        pbar = None
+
+        try:
+            for rates, total_pages in pages:
+                if pbar is None:
+                    pbar = tqdm(total=total_pages)
```
jseppi: These two names together are kinda funny. maybe `progress_bar` instead?

contracts/tests/test_load_api_data.py
```diff
@@ -0,0 +1,158 @@
...
+
+from contracts.mommy_recipes import get_contract_recipe
+from contracts.models import Contract
+from contracts.management.commands.load_api_data import (
+    iter_api_pages,
+    Command,
+)
+
+
+# These tests are a bit funky because we run the command against its
```
jseppi: APICEPTION!

jseppi:
It kinda makes sense to the in the `api` app, but I don't feel strongly about it.

toolness:
> That's some good pythoning right there

Hah, it was actually [copypasta'd from the DRF docs](http://www.django-rest-framework.org/api-guide/serializers/#customizing-multiple-create). :)

Ok I think your suggestions are good, I will implement them and then merge this!

Issue #1724, opened 2018-04-06
Data Explorer js bundle is large


jseppi:
From webpack's prod build output:

```
WARNING in asset size limit: The following asset(s) exceed the recommended size limit (244 KiB).
This can impact web performance.
Assets: 
  tests.js (706 KiB)
  data-explorer.js (402 KiB)

WARNING in entrypoint size limit: The following entrypoint(s) combined asset size exceeds the recommended limit (244 KiB). This can impact web performance.
Entrypoints:
  data-explorer (402 KiB)
      data-explorer.js
  tests (706 KiB)
      tests.js

WARNING in webpack performance recommendations: 
You can limit the size of your bundles by using import() or require.ensure to lazy load some parts of your application.
For more info visit https://webpack.js.org/guides/code-splitting/
```

Suggest we first analyze the `data-explorer.js` bundle with something like https://www.npmjs.com/package/webpack-bundle-analyzer to see if there are any obvious large dependencies that could be removed/replaced.

If not, then we should look into code-splitting, maybe using something like https://github.com/jamiebuilds/react-loadable

jseppi:
Snapshot of the tree map from webpack-bundle-analyzer:

<img width="1239" alt="screen shot 2018-04-06 at 1 26 38 pm" src="https://user-images.githubusercontent.com/697848/38437800-37740ce0-399e-11e8-8a92-5a59499bbfee.png">

`rc-slider` is much bigger than I expected. Nearly the same size as our entire app code 😬

jseppi:
hrm, that image is showing `stat` sizes, which are input sizes without transforms like minification applied. I think we want to rely on `parsed` sizes instead.

ref: https://www.npmjs.com/package/webpack-bundle-analyzer#size-definitions

jseppi:
Here we go for the `parsed` sizes:

<img width="1244" alt="screen shot 2018-04-06 at 3 45 39 pm" src="https://user-images.githubusercontent.com/697848/38443631-fec696b0-39b1-11e8-98fc-d1a8a04b4e76.png">

To get this I just had to tell webpack-bundle-analyzer to put `stats.json` into the `./frontend/static/frontend/built/js/` directory along with all the bundles, and then run the webpack-bundle-analyzer server from the CLI instead of from the `gulp` build:

```sh
webpack-bundle-analyzer ./frontend/static/frontend/built/js/stats.json
```

jseppi:
closed by https://github.com/18F/calc/pull/1736

PR #1736, opened 2018-04-09
Add React-Loadable and analyze webpack

jseppi:
ref #1724.

This PR adds support for using [`react-loadable`](https://github.com/jamiebuilds/react-loadable), and uses that code to make the Data Explorer's `OptionalFilters` "loadable" as a new higher-order component called `LoadableOptionalFilters`. It also adds `webpack-bundle-analyzer` and an `npm` script for visualizing our bundle sizes.

`react-loadable` makes it pretty easy to generate webpack chunks/bundles based that are split at the component-level. Then, those chunks can be loaded asynchronously instead of having to everything in a single bundle. In practice, this means the initial load of the Data Explorer app will be faster (due to a smaller initial bundle size).

Unfortunately `webpack-bundle-analyzer` seems to have a problem with analyzing the fully-transformed chunks that are generated by React-Loadable and `import` (ref https://github.com/webpack-contrib/webpack-bundle-analyzer/issues/166) so we can only see the `stat` size of our bundle modules. The `stat` size, while not what actually will get transmitted to a client over the wire, appears to at least be representative of the relative size of modules to each other, so it should still help identify large modules/dependencies.


jseppi:
hrm, our selenium tests failed in the last CircleCI build. Hoping a restart will correct.

.eslintrc
```diff
@@ -1,5 +1,6 @@
 {
   "extends": "airbnb",
+  "parser": "babel-eslint",
```
jseppi: required so that `eslint` doesn't barf on not-yet-implemented syntax (like `import`) that gets transpiled by `babel`.

.babelrc
```diff
@@ -1,3 +1,6 @@
 {
-  "presets": ["env", "react"]
+  "presets": ["env", "react"],
+  "plugins": [
+    "syntax-dynamic-import"
```
jseppi: allows use of dynamic `import`: https://babeljs.io/docs/plugins/syntax-dynamic-import/

frontend/source/js/data-explorer/components/loadable-wrapper.jsx
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Loadable from 'react-loadable';
+
+export const Loading = ({ error, pastDelay }) => {
+  if (error) {
+    return (
+      <div className="usa-alert usa-alert-error" role="alert">
```
jseppi: This looks maybe a little weird, but I was unsure of what would be better, and didn't really want to introduce additional styling rules. This error case _should_ never happen.

jseppi:
I'm at a loss for why the selenium tests are failing on CircleCI. They all pass locally.

jseppi:
Welp, they passed after the last build re-trigger 🎉!

docs/frontend.md
```diff
@@ -39,30 +39,45 @@ Read the [yarn workflow docs][] if you are not familiar with how to use it.
...
-[frontend/source/js/data-explorer/tests](../frontend/source/js/data-explorer/tests/).
+#### Testing
+
+The data explorer's test suite uses [Jest][], and the tests are located in [frontend/source/js/data-explorer/tests](../frontend/source/js/data-explorer/tests/).
 
 To run all the tests, run:
 
-```
+```sh
 yarn test
```
hbillings: Is there a way to run this from Docker, also?

docs/frontend.md
```diff
@@ -39,30 +39,45 @@ Read the [yarn workflow docs][] if you are not familiar with how to use it.
...
+```
+
+If not using Docker, run:
+
+```sh
+yarn gulp build
+yarn analyze-webpack
+```
+
+Then visit http://localhost:8888 to explore the modules that comprise each bundle.
```
hbillings: 💯 on the correct use of `comprise`!

frontend/source/js/data-explorer/components/loadable-wrapper.jsx
hbillings: Does this wrap around the entire Loadable container, or each filter? If the former, I think it's fine; if the latter, we have some form error classes we could apply that would probably look less weird.

docs/frontend.md
jseppi: oh, yeah, all commands can be prefixed with `docker-compose run app`. It's a little cumbersome to keep up with two ways of doing everything. I think we should standardize on the `docker-compose` way, personally.

docs/frontend.md
jseppi: 🤓

ericronne: `which` comprise, perhaps? 🤓

jseppi:
For whatever reason I can't comment on your question about the error message. It looks like this:

screen shot 2018-04-10 at 9 03 21 am

docs/frontend.md
hbillings: Agreed, especially since none of us are running it the vanilla way.

jseppi: Filed an issue: https://github.com/18F/calc/issues/1743

hbillings: 👯

PR #1738, opened 2018-04-10
Adds logged-in design to footer

ericronne:
The logged-in and logged-out designs are now more consistent with each other, and links on both sides of the (now thicker) dividing border are formatted identically. Also adds a link to add price data.

"Log in" button formatting needs to be fixed, but as it entails a global button change, will do elsewhere.


data_explorer/templates/_footer.html
```diff
@@ -9,8 +9,12 @@ <h2>Log in to add price lists</h2>
               <p>Now contracting officers and other acquisition professionals can upload price lists to CALC.</p>
               <a href="{% url 'uaa_client:login' %}" class="usa-button button-footer">Log in</a>
             {% elif user.is_authenticated %}
-              <p>You are logged in as {{ user.email }}</p>
-              <p><a href="{% url 'user_account:index' %}">View account</a> | <a href="{% url 'logout' %}">Log out</a></p>
+            <h2>Welcome!</h2>
+            <p>You are logged in to CALC as {{ user.email }}, and can now <a href="{% url 'data_capture:step_1' %}">add price lists</a>.</p>
+            <div class="footer_extra">
```
hbillings: We should stick with [BEM](https://frontend.18f.gov/css/naming/) here. `footer_extra` should be something like `footer__user-links`, I think.

frontend/source/sass/components/_footer.scss
```diff
@@ -26,7 +26,7 @@ footer {
 }
 
 .divider_footer {
```
hbillings: Same here; BEM naming conventions would be `footer__divider`.

frontend/source/sass/components/_footer.scss
```diff
@@ -26,7 +26,7 @@ footer {
...
-  border-left: 1px solid #a0ad64; //add a variable for this
+  border-left: 4px solid #a0ad64; // This will be $color-green-darker once `er-palatte-a-go-go` is merged
```
hbillings: Let's take the comment out, otherwise it will stay even when the other branch is merged in.

data_explorer/templates/_footer.html
ericronne: Class had also been applied to the right side, so i renamed to `footer__bottom-links`

ericronne:
@hbillings cropped the divider, all done! 🤞

Issue #1743, opened 2018-04-10
Developer docs - standardize on using docker-compose


jseppi:
Currently our docs generally (but don't always) give both the "vanilla" and the `docker-compose`-way of running various commands. This is cumbersome to write and to read.

Everyone on the team has switched to docker-based development, so I think we should just standardize on that for our documentation as well.

Related https://github.com/18F/calc/issues/1698

toolness:
This sounds reasonable to me sir!

PR #1744, opened 2018-04-10
Add card examples to styleguide

### LINKED ISSUE #1739: Add card markup to styleguide
The `.card` markup we use on pages is a little hard to follow from just the styles in `_page.scss`. There are a few options, too. We should document how to use it and what the options are in the styleguide.

hbillings:
Closes #1739 and #1761 

To do:
- [x] Document basic card layout
- [x] Document using `card__footer`
- [x] Document using `content--skinny`
- [x] Document using `card--no-header`
- [ ] Document using `card--half-padding`-- NOT DOING THIS

After assessing some headers, `card--no-header` is now `card--collapse-header` and `card--half-padding` has been removed in favor of smarter typography selectors.


hbillings:
Will address the second two checkboxes in #1761 -- I'm not entirely sure we need those classes.

styleguide/templates/styleguide.html
```diff
@@ -61,6 +61,96 @@ <h6>Heading 6</h6>
...
+          <div class="content">
+            <p>Thrilling, isn't it?</p>
+          </div>
+        </div>
+      </div>
+    </main>
+  </div>
+  {% endexample %}
+
+  <p>The data capture <code>step.html</code> template is set up to expect step headings to occur in their own card; however, the designs for some steps (specifically 3 and 5) call for a single-card look. To create this look, use the <code>card--collapse-header</code> class on <code>main</code>, which will remove the bottom border of the heading card and reduce the amount of padding applied to the second card.</p>
```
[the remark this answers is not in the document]
toolness: Also, what do you think of changing `step.html` to `{% pathname "data_capture\templates\data_capture\step.html" %}`?  This will hyperlink to the GitHub source code for that template, which will be useful for readers acquainting themselves with the code, and it will also cause our test suite to fail if that link ever breaks, preventing doc rot.

docs/frontend.md
```diff
@@ -61,7 +61,27 @@ You can run `jest` directly, too. If you're using Docker, this can be done via `
...
 #### Analyzing bundles
 
-To see a visual representation the bundles generated by Webpack, you can use `webpack-bundle-analyzer`.
+To see a visual representation of the bundles generated by Webpack, you can use `webpack-bundle-analyzer`.
+
+If you are using Docker, run:
+
+```sh
+docker-compose run app yarn gulp build
+docker-compose run --service-ports app yarn analyze-webpack
```
toolness: Uh, a fairly huge caveat here is that the developer shouldn't have `docker-compose up` running in a separate tab; if they do, `docker-compose` will get confused because `docker-compose up` is already exposing the ports that `--service-ports` is trying to expose here.  It will result in a fairly inscrutable error like `Bind for 0.0.0.0:8888 failed: port is already allocated`.

A better solution would be to change things so that port 8888 isn't exposed by `docker-compose up`, but it is exposed when developers choose to run `webpack-analyzer`.  To do this, we could:

1. Change the above command-line to:

    ```
    docker-compose run -p 8888:8888 app yarn analyze-webpack
    ```

2. *Remove* the following line from our `docker-compose.yml`:

    ```diff
    -       - "8888:8888" # for analyze-webpack
    ```

docs/frontend.md
```diff
@@ -61,7 +61,27 @@ You can run `jest` directly, too. If you're using Docker, this can be done via `
...
+```
+
+If not using Docker, run:
+
+```sh
+yarn gulp build
+yarn analyze-webpack
+```
+
+Then visit http://localhost:8888 to explore the modules that comprise each bundle.
```
toolness: In light of #1743, we might want to consider just removing this part about not using Docker, but it's up to you. Alternatively someone else can just remove it when submitting a PR to fix that issue!

docs/frontend.md
jseppi: These changes are from me in https://github.com/18F/calc/pull/1736. I'm not sure why they're in this diff, but I'll file a separate issue to handle the requested changes.

jseppi: https://github.com/18F/calc/issues/1772

docs/frontend.md
jseppi: also in https://github.com/18F/calc/issues/1772

styleguide/templates/styleguide.html
```diff
@@ -61,6 +61,96 @@ <h6>Heading 6</h6>
...
+        </div>
+      </div>
+      <div class="card__footer">
+        <button type="submit" class="usa-button-link" name="cancel" data-a11y-dialog-show="cancel-dialog" formnovalidate="">I've changed my mind; abort!</button>
+      </div>
+    </div>
+  </main>
+  {% endexample %}
+
+  <p>By default, the cards are <code>980px</code> wide. This is too wide for sections like About and most of the price list upload, so you may need to use the <code>content--skinny</code> body class.</p>
```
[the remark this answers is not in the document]
hbillings: Good call; I like the `too wide` language that is less specific. Re: mentioning specific pages, the reason I did that was mostly because I myself was trying to figure out what the logic behind using one card one place and another another place was. BUT maybe it's better for me to talk about that logic rather than saying, "You can find an example of this on X page."

styleguide/templates/styleguide.html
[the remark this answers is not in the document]
hbillings: I think I mostly addressed the reasoning behind this above. Also, I didn't realize that template tag did that! Will add.

hbillings:
@jseppi @toolness this is ready for review!

frontend/source/sass/components/_page.scss
```diff
@@ -75,6 +37,8 @@ footer {
   }
 }
 
+
+// A top-level body class. Used on About.
```
jseppi: This is used in lots of places I think. Maybe remove the comment?

styleguide/templates/styleguide.html
```diff
@@ -61,6 +61,96 @@ <h6>Heading 6</h6>
...
+          <div class="content">
+            <p>Thrilling, isn't it?</p>
+          </div>
+        </div>
+      </div>
+    </main>
+  </div>
+  {% endexample %}
+
+  <p>The data capture {% pathname "data_capture\templates\data_capture\step.html" %} template is set up to expect step headings to occur in their own card; however, the designs for some steps call for a single-card look. To create this look, use the <code>card--collapse-header</code> class on <code>main</code>, which will remove the bottom border of the heading card and reduce the amount of padding applied to the second card.</p>
```
jseppi: I think the slashes are the wrong direction in the `pathname` directive. Should be 

```
{% pathname 'data_capture/templates/data_capture/step.html' %}
```

frontend/source/sass/components/_page.scss
toolness: I agree, I am a bit fearful of the ol' doc rot here too. 😣

styleguide/templates/styleguide.html
toolness: Yeah! What are you, some kind of weirdo Windows user?! 😆

toolness:
> Looks like some kind of aXe violation on a link, and maybe a mis-targeted path to the `step.html` template. For the aXe violation, the Chrome plugin is pretty useful for actually finding the error since the text-based output is hard to follow.

I agree, the [aXe Chrome extension](https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US) is really good, and a much better way to understand its error output than the JSON blob.

PR #1745, opened 2018-04-11
Visual adjustments to hero

ericronne:
… and bumps up global nav size by 1px


ericronne:
@hbillings this includes a forced line break in the headline. Apologies if this is not the optimal method for it. If so, perhaps we address when we dive into making the site overall more responsive?

ericronne:
@hbillings i know you fixed this giant-logo-focus issue previously … is it on another branch?
image

data_explorer/templates/_banner.html
```diff
@@ -1,29 +1,4 @@
 {% load staticfiles %}
-{% if NON_PROD_INSTANCE_NAME or show_debug_ui %}
```
hbillings: I don't think you meant to commit this. We definitely want to keep this code in the repo!

frontend/source/sass/components/_header.scss
```diff
@@ -61,7 +61,7 @@ header {
 
 .hero-callout {
   border-left: 4px solid $color-gray;
-  padding-left: $space-3x;
+  padding: 0 $space-3x;
```
hbillings: Preferable to add `padding-right: $space-3x` instead of using the shorthand with a `0` rule, unless you specifically want to set the top and bottom padding to `0`.

frontend/source/sass/components/_header.scss
```diff
@@ -89,6 +89,7 @@ header {
   font-size: $display-font-size;
   margin-top: 0;
   margin-bottom: .3em;
+  padding-right: 1em;
```
hbillings: Why is the padding (and the margin, for that matter, though I don't think you set the margin) in `em` instead of `rem`?

frontend/source/sass/components/_header.scss
```diff
@@ -171,7 +172,7 @@ nav {
     margin-bottom: 0;
 
     li {
-      font-size: 1.5rem;
+      font-size: $base-font-size;
```
hbillings: yassss

frontend/source/sass/components/_header.scss
ericronne: I do want top and bottom padding to be zero, to insure that the vertical rule is properly sized.

ericronne:
@hbillings backatcha! ☝️

frontend/source/sass/components/_header.scss
hbillings: Is it not zero to begin with?

toolness: i feel like the phrase "from zero to hero" could be used here

hbillings: ping on this -- unless you're explicitly overriding something, the padding on the top and bottom should already be `0` and we don't need to override.

PR #1758, opened 2018-04-12
[WIP] Replace test_selenium with Puppeteer-based tests

jseppi:
This is an attempt to use [Puppeteer](https://github.com/GoogleChrome/puppeteer) instead of Selenium for browser-based tests.

Really, this just ports our aXe tests, which used to be run via selenium, to be run with Puppeteer. This is because upon looking closely at `test_selenium.py`, nearly all the other tests are things already covered by either our React Jest tests, or by backend Python tests. It does not seem worth it to port these tests, but I'm willing to accept some pushback here.

As currently implemented, these are the steps to run the Puppeteer tests:

1. Build and CALC as normal (`./docker-update.sh`, `docker-compose up`) in one terminal
1. In another, run the tests:
    1. Make sure you have node 8 installed and in use : `nvm install v8` or `nvm use v8`
    1. `cd browser_tests`
    1. Install dependencies `yarn`
    1. Run the tests: `yarn jest`

### TODO

- [ ] Fix detected aXe violations. There are a few aXe violations that the tests found on both the Data Explorer and on the Styleguide. See below for screenshots from the aXe Chrome plugin, which detected the same problems.
- [ ] Figure out a good way to run this:
  - [ ] locally
  - [ ] in CI
- [ ] Delete the selenium test and support files/directories.

screen shot 2018-04-12 at 5 29 39 pm

screen shot 2018-04-12 at 5 30 41 pm

### Alternative strategy

Since this really ended up boiling down to just running aXe, we could 

We instead use [`axe-cli`](https://github.com/dequelabs/axe-cli) and run it against our production instance as part of our nightly `production_tests` that get run in CircleCI.

Or maybe even better, add another nightly cron to run `axe-cli` against `calc-dev.app.cloud.gov` so we catch a11y errors before they wind up in production.


toolness:
should we close this since #1767 landed?

jseppi:
yep!

PR #1765, opened 2018-04-13
[WIP] Attempt to operationalize puppeteer

toolness:
This is an attempt to operationalize #1758, maybe.

## Notes

* I was originally going to run [headless chrome in a separate Docker container](https://github.com/yukinying/chrome-headless-browser-docker) but ran into problems.  Since I last did this, it seems like chrome's debugging protocol changed in a way that makes it hard to have long-lived headless instances that accept requests from puppeteer; instead, it seems like headless chrome really wants the test runner to spin up a separate instance of it for each test run, which massively complicates running headless chrome in a separate docker container.  So instead, I just opted to follow the same approach that we currently follow with PhantomJS, which is to simply include support for launching headless chrome in the app container itself. This massively simplifies networking anyways, so I think we should be ok.

* I scavenged parts from [Running Puppeteer in Docker](https://github.com/GoogleChrome/puppeteer/blob/master/docs/troubleshooting.md#running-puppeteer-in-docker) and integrated them into CALC's `Dockerfile`, replacing the PhantomJS jank.  I removed a few dependencies we don't need, like non-English fonts.

* I upgraded CALC's version of node to 8.x, since we want to use all the nice stuff, and we might as well anyways since eventually 6.x will reach end-of-life right?!

## To do

- [ ] Add funky jest-puppeteer tests from #1758. This will involve making it possible to run both the existing jest tests and puppeteer tests from the same container, which i hope is possible?!  In the worst case I guess we could go with the separate `package.json` that #1758 does.
- [ ] Add a Django `LiveServerTestCase` to run the jest-puppeteer tests against the live server.  Ideally perhaps it should still be possible to run the jest-puppeteer tests manually against the currently-running server hosted by `docker-compose up`, just to make things super fast during development.
- [ ] Remove all remaining PhantomJS and Selenium jank from our codebase, including developer documentation et cetera.


toolness:
I'm going to close this in favor of #1767 for now, though we can re-open this if Selenium proves to be a bust yet again.

Issue #1747, opened 2018-04-12
Stop using api.data.gov as the primary API endpoint for CALC


toolness:
We discussed this recently on a call but I don't think we have an issue for it, so I thought I'd add one here.

Currently, on production/staging/development ~~(but not staging or development)~~ we restrict the IP addresses of clients of `/api/` to the [API Umbrella](https://apiumbrella.io/) instance at [api.data.gov](https://api.data.gov/).  This, in turn, proxies everything at `/gsa/calc/` through to the production instance's `/api/` path.  However, we don't do this on local dev instances.

I _think_ this was originally done so that api.data.gov could take care of things like API keys, rate limiting, and analytics without CALC having to worry about it.  However, in practice, we don't use API keys and haven't had to care about rate limiting.  And the analytics could, I think, just as easily be ensured by simply ensuring that all API requests are logged, which would then ensure that they're in our ELK stack at [logs.fr.cloud.gov](https://logs.fr.cloud.gov), where all our other logging data is (in fact, this may already be the case).

There's a number of drawbacks to this strategy:

* It complicates the code and documentation.  For example, we've got an [`API_HOST` constant](https://calc.gsa.gov/docs/api.html#access-from-calc-front-end-code) in the front-end code which needs to be documented and understood by anyone who on-boards to CALC.  Additional tests need to be written which take the proxying into account, and there are even outstanding issues (like #1692 and #1694) that exist because of this setup and haven't yet been resolved.

* It massively deviates from [dev/prod parity](https://12factor.net/dev-prod-parity).  Because the proxy only operates on the production/staging/dev instances~~, we can't even see if bugs exist on staging, to say nothing of our development systems~~. This means that bugs are very easy to creep in and go unnoticed; see e.g. #1692 and #1694, which have probably been around for years but we've only just noticed them recently.

* For the above complications, it doesn't seem to provide much benefit. As mentioned earlier, we don't currently need rate limiting or API keys, and the analytics could be done via our ELK stack.  It could potentially be argued that migrating away from api.data.gov would _simplify_ how we do analytics while also making our analytics more powerful: API Umbrella doesn't seem to offer very advanced (or even particularly user-friendly) analytics functionality, while Kibana offers all kinds of ways to analyze/visualize data.

Given all this, it might behoove us to migrate away from api.data.gov and just have production/staging/dev CALC access its API at `/api/`~~, just like dev and staging do~~.

jseppi:
I'm 👍 to this. 

[...]

PR #1766, opened 2018-04-13
Remove API_HOST, WHITELISTED_IPS, associated code, and documentation

### LINKED ISSUE #1692: "next" parameter in API results is wrong in production, and also undocumented

### LINKED ISSUE #1694: Production site API pages are not loading CSS/JS/image static assets
When visiting https://api.data.gov/gsa/calc/rates/, we should see a nice Django Rest Framework page that looks something like:

screen shot 2018-04-03 at 9 11 37 am

However, we get an unstyled page that looks like:

screen shot 2018-04-03 at 9 09 31 am

Here's what the console shows:

screen shot 2018-04-03 at 9 13 43 am

jseppi:
As described in #1747, we don't really see any benefit in using the api.data.gov API Umbrella instance to proxy requests to our own app, and it causes a few other bugs.

This PR removes the `API_HOST` and `WHITELISTED_IPS` environment variables and the associated code for proxying API requests and whitelisting only specific IPs (those of api.data.gov) to the "real" API path.  Documentation around those settings and the usage of api.data.gov is also removed.

While they will be harmless, it would be good to remove the `API_HOST` and `WHITELISTED_IPS` vars from `calc-env` in each of our cloud.gov spaces, but **this should only be done upon promotion of these code changes** to each space.

closes #1692 and closes #1694


frontend/source/js/data-explorer/constants.js
```diff
@@ -124,7 +124,3 @@ export const QUERY_TYPE_LABELS = {
 };
 
 export const MAX_QUERY_LENGTH = 255;
-
-export const API_HOST = window.API_HOST;
-
-export const API_RATES_CSV = `${API_HOST}rates/csv/`;
```
toolness: So the one thing I like about keeping these routes in their own separate file is just that they're all in one place, so maintainers (hopefully) know where to look if they decide to change any routes on the back-end, rather than relying on grepping through the source code or something.  Hmmmmmmmmm.  What do you think of adding a separate module called `backend-routes.js` that has either constants or functions (if the route requires parameters) that return route strings?  Or is that going overboard?

docs/api.md
```diff
@@ -2,34 +2,25 @@
...
+The following documentation assumes you're trying to access the API from the production instance via a tool like `curl` at `https://calc.gsa.gov/api/`.
+In development, use `http://localhost:8000/api/`.
 
 ### `/rates/`
 
 You can access labor rate information at `/rates/`.
 
 ```
-https://api.data.gov/gsa/calc/rates/
+https://calc.gsa.gov/api/rates/
```
toolness: Something this is making me realize is that API requests will now go through CloudFront on production.  I don't think this currently impacts us in a negative way, but if e.g. we eventually decide to use API tokens or something, we'd have to remember to have CloudFront pass any fancy headers we use on to the origin server.

frontend/source/js/data-explorer/constants.js
jseppi: yeah, I'll add them back. Good point!

jseppi: I ended up adding the constants to `api.js` instead of to `constants.js` or another new file. Seemed like a logical place for them to me.

jseppi:
Added dependency on https://github.com/18F/calc/issues/1773

frontend/source/js/data-explorer/constants.js
toolness: Ah nice, good idea!

jseppi:
Verification of analytics capabilities confimed in #1773!

Issue #1553, opened 2017-04-14
Fix color contrast issues on data explorer


toolness:
When I added aXe to our test suite in #1552, I disabled the color contrast check because it brought up a number of violations on the data explorer.  I think most of the violations were extremely minor, e.g. a contrast ratio of 4.5:1 was required but ours was 4.45:1.

Anyways, it'd be good to get them out of the way and re-enable the contrast check.  Or we can just wait until we refresh the data explorer w/ the new color scheme that @ericronne made.

ericronne:
How do i view the violations?

afeld:
Some tools that can be used for this: https://accessibility.18f.gov/tools/

hbillings:
Closing because we have the new color schemes!

PR #1767, opened 2018-04-13
Use headless chrome with selenium in docker

toolness:
This is an alternative to #1758 and #1765 which attempts to just use headless chrome + chromedriver (instead of phantomjs) with selenium.

As I mentioned in Slack:

> uh... i feel kinda bad bringing this up ... but yet another option is for us to actually just use chromedriver to make our selenium tests run against headless chrome instead of phantom.  from what i could tell it seemed like the real problem w/ our selenium tests was phantomjs' implementation of the webdriver protocol, not selenium itself.  blerg.

> The one big advantage I still see with Selenium is the fact that it's a full-on integration test that allows us to easily switch between the front-end browser (via selenium) and the back-end db (via the `LiveServerTestCase` that runs the test).  That ensures that our back-end and front-end are cooperating in the ways we expect, which we can't really get from the standard jest tests.

## To do

- [x] Remove support for phantomjs (both in code and docs).
- [x] Make CircleCI log the version of Chrome and chromedriver in use.
- [x] Either ensure that `pytest data_capture/tests/test_integration.py` works with headless chrome, or remove support for selenium and just run those tests using robobrowser.
- [x] ~~Now that this works, I am extremely confused as to why our aXe tests seem to pass here, but don't on #1758. We should investigate this to ensure that the aXe tests are actually running properly.~~ aXe tests work here because `frontend.tests.axe.IGNORED_VIOLATIONS` contains `color-contrast`, whereas I think #1758 paid attention to all violations.  If we remove `color-contrast` then our aXe tests fail with contrast violations, so things are definitely working as expected here.  The color contrast exception was even filed as #1553 so we don't need to create any new issues for this.
- [x] Ensure CircleCI is running selenium tests against headless chrome
- [x] Add docs for new environment variables/settings/etc:
  - [x] Specifically, document `WD_CHROME_ARGS`.
  - [x] Consider just removing the `Example: Docker on OS X` and `Running the tests on your machine's Chrome browser` sections from `docs/selenium.md`.  The first doesn't make sense since we're already using headless Chrome by default now, and the second doesn't really make sense in light of #1743.


toolness:
@jseppi this is basically done, but I could use some advice on what to do about `selenium.md`.  Given that testing on Chrome is a pretty good default, and that we don't even support Sauce Labs anymore, I'm tempted to just remove support for testing on any non-Chrome browser, as it would simplify our documentation and testing implementation.  (That said, in the future we could potentially add support for [headless Firefox](https://developer.mozilla.org/en-US/Firefox/Headless_mode) so that CircleCI runs selenium tests against both Chrome and Firefox.)

What do you think?

frontend/tests/test_selenium.py
```diff
@@ -34,29 +35,28 @@
...
-        driver.command_executor.set_timeout(PHANTOMJS_TIMEOUT)
-        return driver
     raise Exception('No such webdriver: "%s"' % name)
 
 
 class SeleniumTestCase(StaticLiveServerTestCase):
     connect = None
     driver = None
-    screenshot_filename = 'selenium_tests/screenshot.png'
+    screenshot_filename = 'selenium_tests_screenshot.png'
```
toolness: I'm not sure what behavior was like before, but right now specifying a non-existent directory (`selenium_tests`) doesn't make that directory be created, so the screenshot doesn't get written.  By making this screenshot just a filename, we ensure the screenshot is actually written.

data_capture/tests/browsers/selenium.py
```diff
@@ -20,7 +23,9 @@ def set_file(self, name, path):
         self.form.find_element_by_name(name).send_keys(path)
 
     def submit(self):
+        old_page = self.driver.find_element_by_tag_name('html')
         self.form.submit()
+        WebDriverWait(self.driver, 10).until(staleness_of(old_page))
```
toolness: This basically assumes that there's a full page refresh when a form is submitted, i.e. we don't do some fancy ajax thing.  I _thought_ this was the default behavior of a selenium form submission but I guess not.

frontend/tests/test_selenium.py
jseppi: Is this functionality even needed any more? It doesn't look like it would be that helpful, so maybe even consider removing it.

jseppi:
Re: Your `selenium.md` question - I think making it simpler (ie, only supporting chromium for now) is a good course of action.

.circleci/config.yml
```diff
@@ -2,11 +2,11 @@ version: 2.0
 
 default-docker-config: &default-docker-config
   docker:
-    - image: circleci/python:3.6
+    - image: circleci/python:3.6-browsers
```
jseppi: Whoa, didn't know about these [CircleCI docker variants](https://circleci.com/docs/2.0/circleci-images/#python). Very useful!

frontend/tests/test_selenium.py
toolness: Hmm, I think it's useful to see what a browser is "seeing" when a test fails.  It was actually useful when I was debugging the failure [here](https://github.com/18F/calc/pull/1767#discussion_r181704445) in this very PR.

data_capture/tests/test_integration.py
```diff
@@ -26,7 +25,14 @@ def autologin(request):
     if user is None:
         raise Exception('Unable to authenticate')
     django.contrib.auth.login(request, user)
-    request.session[safe_mode.SESSION_KEY] = True
+
+    # We used to enable safe mode to reduce the chance of JS race conditions
```
jseppi: Do you think we'll re-enable this at some point? If not, maybe just strike this comment and commented-out code since it probably won't make much sense to future developers.

frontend/tests/test_selenium.py
jseppi: Ah, ok. Not a problem to keep it :)

data_capture/tests/test_integration.py
toolness: I think we might want to... The original reason I enabled it was because I was worried that Web Components' asynchronous upgrading mechanism might mess with the tests, and I also figured it'd be useful to explicitly ensure that the flow worked with JS disabled.  Ah well.

toolness:
@jseppi ok I think this is good to go!

docs/testing.md
```diff
@@ -21,17 +21,31 @@ For more information on running only specific tests, see
...
-a real-world browser. This can be done via Selenium/WebDriver. The
-trade-off is that this requires configuration.
+* `WD_SOCKET_TIMEOUT` specifies the default socket timeout to use for
+  requests before giving up, in seconds.
 
-For details on how to do this, see our in-depth [Selenium guide](selenium.md).
+* `WD_CHROME_ARGS` are command-line flags to pass to Chrome,
+  e.g. `--headless --no-sandbox --disable-setuid-sandbox`.
+
+* `TEST_WITH_ROBOBROWSER` is a boolean that indicates whether to run
```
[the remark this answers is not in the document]
toolness: Can we do that in a separate PR?  There's actually a fair amount of abstraction and goop in there that I'd rather just do in a separate PR...  Harumph.

The other possibility is to actually bust this up using _both_ Selenium _and_ RoboBrowser.  The former case--especially since we're no longer enabling safe mode--will ensure that things work in a browser with JavaScript, while the latter will ensure that things still work with JS disabled... ehh.

frontend/tests/test_selenium.py
```diff
@@ -15,6 +15,7 @@
 from django.contrib.staticfiles.testing import StaticLiveServerTestCase
 
 from selenium import webdriver
+from selenium.webdriver.chrome.options import Options as ChromeOptions
```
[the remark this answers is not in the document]
toolness: oh sure might as well remove em.

docs/testing.md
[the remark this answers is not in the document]
jseppi: Sure!

Issue #1750, opened 2018-04-12
Prototype & test a homepage to show only the search bar on load


hbillings:
The data displayed on the homepage on load is at best useless and at worst confusing. It slows down the page load and gives the user too much to look at.

Instead, @jseppi suggested we use React-Loadable to take the current homepage and split it into two main components: search and results. We could load the search first, and then when the user has entered a term, load the results. While this doesn't address any of the issues with the readability of the homepage, it would allow us to very quickly prototype and test a search-first approach to see if it's worth pursuing.

jseppi:
I'd be willing to spike on this. Or better, to pair spike on it!

hbillings:
@jseppi I'd love to pair with you to get my head around react-loadable a little more!

jseppi:
related: https://github.com/18F/calc/issues/295

ericronne:
Blocked pending research

tram:
Using sketches from design studio in upcoming sessions.

tram:
Prepare to spin this branch up as needed for research sessions.

abrouilette:
Based on last week's testing we should revisit.

PR #1768, opened 2018-04-13
[Experiment] Show only the search bar on Data Explorer load

jseppi:
**work-in-progress**

Spike for #1750 - Only showing the search bar on load.


frontend/source/js/data-explorer/reducers.js
```diff
@@ -119,7 +120,7 @@ function rates(state = {
     if (d && d.results && d.results.length) {
       return d;
     }
-    return EMPTY_RATES_DATA;
+    return NO_MATCHING_RATES_DATA;
```
hbillings: I'm not sure if this is the right pattern or if this is weird. I almost think maybe `EMPTY_RATES_DATA` isn't actually necessary any more, since in this PR, nothing will be loaded unless the user has made a successful query?

frontend/source/js/data-explorer/components/app/index.jsx
```diff
@@ -118,7 +117,7 @@ class App extends React.Component {
 
         <LoadingIndicator />
 
-        {(!ratesInProgress && hasRates) &&
+        {((!ratesInProgress && hasRates) || !ratesReturned) &&
```
hbillings: This is kind of a funky check that seems _almost_ duplicative of the check that happens within `LoadableSearchResults`. The only condition in which we don't load `LoadableSearchResults` is on initial page load.

frontend/source/js/data-explorer/components/app/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+
+import LoadableSearchResults from './loadable-search-results';
+
+import { autobind } from '../../util';
+
+
+/**
+ * TODO:
+ *   - add a no results component/alert to search-results.jsx
+ *   - Think about/improve the hasRates prop
```
jseppi: Perhaps by changing the design of the `store`'s `rates` object. Maybe it could be changed to something along the lines of:

```js
rates: {
  isLoading: true | false,
  error: null | 'some error message',
  data: <null if error> | [ array of rates if no error ],
}
```

frontend/source/js/data-explorer/components/app/index.jsx
```diff
@@ -0,0 +1,153 @@
...
+ *   - Probably extract the #search section to its own component
+ *   - Think about/improve where redux state is connected
```
jseppi: For instance, the `LoadingIndicator` doesn't really need to be connected to the redux store. Instead, it could just be conditionally rendered by a parent component, and that parent component could do the checking.

```jsx
// in some parent component
render() {
  return (
    <div>
       {isLoading && <LoadingIndicator />}
       {!isLoading && <AllTheLoadedData /> }
    </div>
  );
}

Comment (jseppi) [reply frontend/source/js/data-explorer/components/app/index.jsx]: This might be better handled in another issue. It's a bit of an opinion-based decision though (ie, deciding upon which components should be/need to be connected to the redux store directly).

File: frontend/source/js/data-explorer/rates-request.js [hunk 5]
```diff
@@ -76,7 +77,8 @@ export default class StoreRatesAutoRequester {
         field => newState[field] !== oldState[field],
       );
 
-      if (updated || newState.rates.stale) {
+      // TODO: improve this
+      if (newState.q !== '' && (updated || newState.rates.stale)) {
```
jseppi: I think I was thinking this check _might_ be better handled in the `rates` reducer. Like maybe checking for `q` could instead factor into the decision to set `stale`.

data_explorer/templates/index.html
```diff
@@ -33,7 +33,7 @@ <h1 class="description-main">Search awarded ceiling rates for labor categories</
 {% endblock header_extension %}
 
 {% block body %}
-<div data-embed-jsx-app-here style="background-color: #fff; border-bottom-color: #c5d6de; margin-bottom: 8rem;">
+<div data-embed-jsx-app-here style="background-color: #fff; border-bottom-color: #c5d6de; margin-bottom: 8rem; padding: 2rem 1rem;">
```
tbaxter-18f: Why do we have the embedded styles here rather than in the site CSS?

hbillings: We're hacking this to look like our `.card` classes, but because this is all going to change drastically, any styles are just temporary at best. The goal is to, when we rework it, make it be something we can use the existing classes on!

ericronne:
Closing pending design

Issue #840, opened 2016-09-30
Review and/or remove updating_data.md


jseppi:
Since Data Capture will now be part of the application, bulk updates of data should no longer be necessary. We want to avoid document rot, so we should review, and potentially remove, `updating_data.md`.

Related:
- remove old data files from `contracts/docs/`
- review/remove `load_data` and `load_s70` management commands

toolness:
I agree!  It looks like there's also one reference to `updating_data.md` in `README.md`, so if we get rid of it, we should be sure to remove it from the readme too.

jseppi:
Closed by #1774

PR #1774, opened 2018-04-16
Remove load_data and load_s70 commands

### LINKED ISSUE #1028: Rename (or remove) load_data command and associated CSV files

toolness:
Fixes #1028.


toolness:
Er, I'm just removing `updating_data.md` (see #840) because it's so out of date that it mentions the CSVs a bunch and keeping it around raises a bunch of test failures, because the `contracts/docs` directory it refers to doesn't exist anymore.

PR #1779, opened 2018-04-16
Keep a long-lived WebDriver instance through the whole test run

toolness:
For some reason, on local dev instances (but strangely not CircleCI) instantiating a WebDriver instance more than once during a test run caused a socket timeout.  This appears to fix things, at least on my machine, by keeping one long-lived WebDriver instance around for the whole test run, and shutting it down when the process exits.

To try it out without running the whole test suite, try:

```
pytest frontend/tests/test_qunit.py data_capture/tests/test_integration.py frontend/tests/test_selenium.py
```

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -28,9 +28,6 @@ ensure that the back-end and front-end integrate properly. The
 following environment variables may be useful when configuring these
 tests:
 
-* `WD_SOCKET_TIMEOUT` specifies the default socket timeout to use for
-  requests before giving up, in seconds.
-
 * `WD_CHROME_ARGS` are command-line flags to pass to Chrome,
   e.g. `--headless --no-sandbox --disable-setuid-sandbox`.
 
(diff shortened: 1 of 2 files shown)


jseppi:
I pulled this branch and ran `ultratest` locally. The first time I got a test failure:

```
E       AssertionError: 'Software Engineer I' != 'Software Engineer'
E       - Software Engineer I
E       ?                  --
E       + Software Engineer
E        : bad cell text: "Software Engineer I"

frontend/tests/test_selenium.py:312: AssertionError
```

then I immediately ran `ultratest` again and it worked. I think there might be some kind of timing issue :\

toolness:
Um... that is super weird, although strangely not very unfamiliar-sounding. I feel like I may have run into something similar a few weeks ago but just re-rain ultratest with my fingers crossed and it worked?

Looking at that particular test, `test_query_type_matches_exact`, I can't imagine _why_ it would intermittently fail in the way that it does.  It is just super bizarro.

In any case, I don't think that the change in this PR is responsible for that test failure, and this PR seems like a step forward, since `ultratest` fails reliably on developer machines without it.

PR #1784, opened 2018-04-17
[WIP EXPERIMENT] Add fullpage_example template tag

### LINKED ISSUE #1771: Styleguide should support snippets that include <body> tags
Currently we have some snippets in our styleguide that are like this:

```html
<div class="content--skinny"><!-- goes on BODY in actuality -->
```

We should support using the actual `<body>` tag, somehow.  And actually render it, probably in an iframe, so that it's easy to see different page styles within the styleguide.

toolness:
This is an attempt to fix #1771 by adding a new `{% fullpage_example %}` template tag.

## Instructions

The basic idea is that when you want to include a full-page example in the style guide, you will:

1. Come up with a slug that identifies the example, e.g. `content-skinny`.

2. Add an HTML file for the example at:

    ```
    styleguide/templates/styleguide_fullpage_examples/<slug>.html
    ```

    The HTML file should just include the raw HTML for the example; it's what will be shown to readers of the styleguide as the example HTML.  So it should only include the `<body>` tag and everything inside it--no `<head>` or `<html>`.

3. In the styleguide, when you want to include the full-page example, use the following:

    ```
    {% fullpage_example "<slug>" %}
    ```

## Example

Here's an example of one of the snippets from #1744 converted to use the `{% fullpage_example %}` template tag:

> fullpage_example

## To do

- [ ] Add tests.
- [ ] Ensure `{% fullpage_example "<slug>" %}` raises an exception if `<slug>` doesn't exist.
- [ ] Ensure the `styleguide:fullpage_example` route returns 404 for non-existent examples.
- [ ] Clean up the iframe styling so it's not all inlined
- [ ] Consider using DJDT's [`SHOW_TOOLBAR_CALLBACK`](https://django-debug-toolbar.readthedocs.io/en/stable/configuration.html#debug-toolbar-config) to not render the debug toolbar on full-page example iframes, rather than hiding the toolbar.

### DIFF
--- a/styleguide/fullpage_example.py
+++ b/styleguide/fullpage_example.py
@@ -1,0 +1,27 @@
+from pathlib import Path
+from django.utils.safestring import SafeString
+from django.shortcuts import render
+from django.urls import reverse
+
+
+MY_DIR = Path(__file__).parent.resolve()
+
+EXAMPLES_DIR = MY_DIR / 'templates' / 'styleguide_fullpage_examples'
+
+REGEX = r'([0-9A-Za-z_\-]+)'
+
+
+def get_html_source(name):
+    filename = EXAMPLES_DIR / f'{name}.html'
+    return SafeString(filename.read_text())
+
+
+def get_url(name):
+    return reverse('styleguide:fullpage_example', args=(name,))
+
+
+def view(request, name):
+    html = get_html_source(name)
+    return render(request, 'styleguide_fullpage_example.html', {
+        'html': html,
+    })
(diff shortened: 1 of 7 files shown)


hbillings:
I like this! I can see being able to build out other page components (like header and footer, or particular admin things, or the `formrowbuttons` set of components) as being useful.

toolness:
Ok this is good to go!  @hbillings I haven't actually fullpage'd the other examples in #1744, do you want to do that in a separate PR?  If not I'm happy to, just lmk.

Issue #1777, opened 2018-04-16
Sass portion of Gulp build does not fail on error


jseppi:
In CircleCI, during the `Build front end and run front end tests` step of the build, individual sass bundle builds can fail silently. 

For example [this build](https://circleci.com/gh/18F/calc/588) contained the following error:

```
Error in plugin 'sass'
Message:
    frontend/source/sass/admin/overrides.scss
Error: Undefined variable: "$color-gray-medium-dark".
        on line 72 of frontend/source/sass/admin/overrides.scss
>>        border-bottom: 2px solid $color-gray-medium-dark;
   --------------------------------^
```

which ideally would break during `yarn gulp build`. Instead, the build in question did not fail until later when `test_admin.py` was run and could not find the compiled `overrides.min.css` (which did not exist since it's sass build failed).

https://github.com/18F/calc/pull/1711#issuecomment-381702697

jseppi:
Sass not breaking the gulp build is due to https://github.com/18F/calc/blob/develop/gulpfile.js#L173, which instead just logs the error. This is kinda nice during development since you don't have to restart the gulp process on a broken sass build and can just fix the problem and have it get re-compiled.

During CI builds, however, we _do_ want failures to break the gulp build task. We maybe could just something to the `on('error', ...)` handler of the sass task:

```js
if (isProd) { throw err; }
```

PR #1785, opened 2018-04-17
break non-watched builds on sass error; upgrade gulp-sass

jseppi:
ref #1777 

Adds a stream `'error'` event handler to the `sass` task that will cause the build to exit with a non-zero status code when a sass error occurs during a non-watched build.

This is nice for CI because builds will fail faster and the cause of the failure will be more easily visible in the CircleCI logs.


jseppi:
@toolness: oh, good catch. Actually it won't be until the deployment step, but we should break the build before that. What do you think about instead checking `process.env.CI`?

jseppi:
I totally forgot we also have an `isWatching` var. That might be the best thing to use. Will change to that.

toolness:
Ooo ya I like using `isWatching`!

PR #1787, opened 2018-04-17
Email style mods

ericronne:


ericronne:
@hbillings i've dropped in an svg logo here. Does that present compatibility / implementation issues?

data_capture/templates/email/base.html
```diff
@@ -298,8 +298,8 @@
                   <table border="0" cellpadding="0" cellspacing="0">
                     <tr>
                       <td>
-                        <div class="calc-logo-wrapper align-center">
-                          <img src="{% absolute_static 'frontend/images/calc_logo_blue.png' %}" alt="" width="180" height="73">
+                        <div class="calc-logo-wrapper align-left">
+                          <img src="{% absolute_static 'frontend/images/calc_logo_black.svg' %}" alt="" width="180">
```
jseppi: I browsers (especially IE) can have problems with SVG when `width` and `height` are not both specified

> If you use an  to embed your SVG, setting height and width will make the SVG scale predictably in most browsers, but not in Internet Explorer. With CSS like img { width: 100%; height: auto; }, IE will auto-scale the image area to keep the width:height aspect ratio constant, but it won't scale the actual drawing to match the scale of the image dimensions.

ref: https://css-tricks.com/scale-svg/

ericronne: Cool thx

ericronne:
@hbillings (et al) do we have access to Litmus, or another email emulator for testing?

ericronne:
that tool is MWAH

toolness:
What does MWAH mean?  Is it [the paticular sound that a fretless bass guitar makes when played](https://www.urbandictionary.com/define.php?term=mwah)?

ericronne:
whoa who knew?
i am MWAH over the sound of a fretless bass as well

### "MWAH!"
giphy

ericronne:
svg is misbehaving 😠 will debug

image

frontend/source/sass/email.scss
```diff
@@ -39,54 +39,58 @@ a {
...
 
 table.body .btn table {
+  margin: auto;
+  max-width: 300px;
   width: 100%;
 }
 
 table.body .btn a {
+  margin: auto;
+  max-width: 300px;
```
hbillings: Is this going to be wide enough to not look funky on mobile?

ericronne: i _think_ so??
this is @400 wide, but it's from desktop; don't know how to actually preview on mobile

image

hbillings: I _think_ you might be able to preview this on iOS simulator: https://gist.github.com/shawnbot/5527355

ericronne: unauthorized app :(

ericronne: ooh installed xcode!
but …
image

maybe we should find a pc somewhere

hbillings: You're viewing the source; what does the `HTML` tab show?

ericronne: hrm that was the source view, here's with the styles.
we really need to test with an email client.

image

ericronne:
Ok simulator is cool, but not a great way for us to QA these emails.

Per @jseppi:
Authorized users can upload a dummy price list to [calc-dev.app.cloud.gov](http://calc-dev.app.cloud.gov/), and then reject it in the Admin section. This will generate an email to the user.

ericronne:
Closed #1782

PR #1793, opened 2018-04-17
Default to dockerized project setup

### LINKED ISSUE #1743: Developer docs - standardize on using docker-compose

jseppi:
Closes #1743.

This modifies all our developer documentation to assume that Docker/Docker Compose are used for running a CALC development environment.

### DIFF
--- a/docs/setup.md
+++ b/docs/setup.md
@@ -1,125 +1,77 @@
 ## Setup
 
-Setting up CALC locally without Docker is possible, but it requires
-quite a few prerequisites. If the following instructions seem daunting,
-you may want to check out our alternative [Docker instructions](docker.md).
+CALC is primarily a [Django] project.
+
+CALC uses [Docker][] and [Docker Compose][] to make setting up a development environment easier.
+
+If you are unfamiliar with Docker or Docker Compose, read their docs as well as the [18F Docker guide][].
 
 ### Prerequisites
 
-You'll need the following tools and services installed to run CALC
-locally:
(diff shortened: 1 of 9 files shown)


docs/setup.md
```diff
@@ -1,125 +1,77 @@
...
+you can http://localhost:8000/ to see your local CALC instance.
 
-[redis-ubuntu]: https://www.digitalocean.com/community/tutorials/how-to-install-and-configure-redis-on-ubuntu-16-04
+### More information
 
-[redis-osx]: https://medium.com/@petehouston/install-and-config-redis-on-mac-os-x-via-homebrew-eb8df9a4f298#.fa2s6i1my
+For more information on interacting with CALC's dockerized development environment, see the [Docker section](docker.md) of our docs.
 
 [Django]: https://www.djangoproject.com/
+[18F Docker guide]: https://pages.18f.gov/dev-environment-standardization/virtualization/docker/
```
toolness: Blerarg, this 404's now.  I think CM wrote a new guide at some point but I'm not sure where it is...

jseppi: Good catch. I’ll track it down or remove the reference.

PR #1794, opened 2018-04-17
Fix admin user search

### LINKED ISSUE #1726: Admin site - user search does not appear to do anything
Steps to reproduce:

1. Go to https://calc.gsa.gov/admin/auth/user
1. Input some text into the Search input and click the Search button

Actual Result:

The page will refresh but not results will appear. The browser URL will now include `?e=1`, which I think indicates some kind of error, but no visual error message is shown.

Expected Result:

The list of users shown is filtered to those who somehow match the entered search screen.

toolness:
Fixes #1726.

It took me a _really_ long time to figure out what was wrong.  At first I started comparing our `search_form.html` to [Django 1.11's](https://github.com/django/django/blob/stable/1.11.x/django/contrib/admin/templates/admin/search_form.html) and didn't notice anything different.  Then I started fiddling with our custom `UserAdmin` subclass, to no avail.  Then I just started deleting stuff, slowly reverting our custom Django admin back to the default one.

The user search field started working again when I _deleted_ `search_form.html`, which made Django fall back to the default one.  Which was weird, because that was the first thing that I'd looked at.  But it wasn't until I looked _really_ closely that I noticed that our search field had _two_ `name` attributes!  Oof.

### DIFF
--- a/hourglass/templates/admin/search_form.html
+++ b/hourglass/templates/admin/search_form.html
@@ -5,7 +5,7 @@
 <div id="toolbar"><form id="changelist-search" method="get">
 <div role="search" class="form__inline"><!-- DIV needed for valid HTML -->
   <label for="searchbar">Search</label>
-  <input type="search" name="search" size="40" name="{{ search_var }}" value="{{ cl.query }}" id="searchbar" autofocus />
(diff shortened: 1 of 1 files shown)


jseppi:
whoa!

hbillings:
_wat_

toolness:
haha woot!

Issue #1698, opened 2018-04-03
Review and update developer documentation


jseppi:
Our `/docs/` folder has lots of developer documentation. It hasn't been holistically reviewed in a while. Having good documentation will be important for transition.

- [x] api.md - #1703 
- [x] auth.md - #1733
- [x] changelog.md
- [ ] deploy.md - #1795
- [x] docker.md - #1904
- [x] email.md
- [ ] environment.md - #1902, maybe
- [x] frontend.md - #1903
- [x] index.rst
- [ ] monitoring.md - #1913
- [ ] release.md - #1795
- [x] ~~selenium.md~~ This file was destroyed as part of the Great Selenium Cleansing of #1767.
- [ ] setup.md - #1901
- [x] testing.md
- [x] ~~updating_data.md~~ This file has been removed and largely replaced by the ultra-awesome admin help in #1844.

jseppi:
![](https://github.trello.services/images/mini-trello-icon.png) [As a CALC developer, I need to review the developer docs to ensure they are up to date and to identify additional documentation tasks for transitioning CALC.](https://trello.com/c/SD64mBiL/252-as-a-calc-developer-i-need-to-review-the-developer-docs-to-ensure-they-are-up-to-date-and-to-identify-additional-documentation-t)

jseppi:
related: #840

toolness:
**Update:** we've now got DRF-generated API documentation (and interactivity) done in #1797 but more work needs to be done on actually connecting it with (or using it to completely replace) `docs/api.md`.  I've filed #1802 to address that.

toolness:
Ok, at this point all the remaining items have been spun off into separate PRs, so I'm closing this issue.

PR #1795, opened 2018-04-18
Remove staging tier of deployments

### LINKED ISSUE #1637: Revise release/deployment procedures to a 2-tier system
The current release procedure is a bit over-complicated and we did not historically make much use out of the `staging` branch/deployment. I think it would be a good idea to switch to just a two-tier system (with `develop` and `production` branches).

This change would involve:

- updating documentation
- removing unused steps in CircleCI

jseppi:
Closes #1637

This sets us up for moving to a two-tier deployment strategy (ie, just `dev` and `production`). It also updates our `deploy.md` docs to include information about usage of service accounts and revises the production deployment section to reflect reality (ref #1698).

To-dos, after this is merged and promoted:

- [ ] Remove the `staging` space on cloud.gov. Might have to stop and delete and all its apps and services first.
- [ ] Delete the staging space credentials from CircleCI settings.
- [ ] Ask cloud.gov to reduce the `fas-calc` memory quota since we won't need as much any more.
- [ ] Update the system security plan to reflect the now two-tiered deployment strategy.
- [ ] Let Akisha (our ISSO) know that `calc-staging.app.cloud.gov` is gone and no longer needs to be scanned


toolness:
Just talked to @KellyBailey and he's good w/ following through on this!

PR #1797, opened 2018-04-18
Attempt to document API via Django REST Framework / Core API

toolness:
This attempts to implement the solution suggested by @jseppi in https://github.com/18F/calc/issues/1698#issuecomment-378401583, whereby we use [DRF's built-in support for API documentation](http://www.django-rest-framework.org/topics/documenting-your-api/).

Specifically, it adds an endpoint at `/api/docs/` that looks like this:

> api-docs

Clicking the "INTERACT" button brings up a modal that allows users to tinker with a specific API endpoint:

> interact

## Notes

* It seems like DRF has embraced something called [Core API](http://www.coreapi.org/), which is "a format-independent Document Object Model for representing Web APIs".  Sadly, it seems somewhat scant on documentation; for instance, [some code examples in the DRF docs](http://www.django-rest-framework.org/api-guide/schemas/#autoschema) use a module called `coreschema` which has [a githubo repo with no README](https://github.com/core-api/python-coreschema).

    Based on a cursory look, it seems like the plumbing provided by this API could be useful for automatic parameter validation, but given the apparent lack of maturity, I'm hesitant to start using it for anything other than documentation right now.

* I was having some problems running `flake8`.  Apparently the version we use is supposed to [require `pycodestyle` 2.3.0](https://gitlab.com/pycqa/flake8/merge_requests/212) but running `pip install -r requirements-dev.txt` left mine at 2.2.0, which was breaking `flake8`.  So for now I've just added `pycodestyle==2.3.0` to our dev requirements.

* It seems DRF has a whole [fancy filtering system](http://www.django-rest-framework.org/api-guide/filtering/).  Reworking our APIs to use this could make it so that DRF's `AutoSchema` automatically pulls querystring arg docs by introspecting our code, rather than requiring us to manually define fields as I'm doing in this PR.  This would help keep our code nice and DRY.

* As far as I can tell, it doesn't seem like coreapi / the auto-generated documentation provides any way to tell readers about the kind of data that an API endpoint _returns_, which is a bummer.  Although that said, it seems like a third-party library called [drf-yasg](https://github.com/axnsan12/drf-yasg/) _does_ provide response schemas, so in the future we could use that, maybe...

* I am not a big fan of how DRF is auto-naming all our GET methods "list", and using that as the heading for our INTERACT modals, but I can't figure out if there's a way to change that default.

## To do

- [ ] Consider mounting these docs at `/api/` instead of `/api/docs/`.  Currently `/api/` just 404s, and it's more easy to type and find that way.
- [ ] Link to `/api/docs/` from the Sphinx docs.
- [ ] Consider making the docs a bit more robust. Right now they're fairly minimal and some fields are a bit vague (e.g. worksite).
- [ ] Consider either merging the existing `API.md` into these API docs or getting rid of them entirely.
- [ ] It'd be nice if these API docs were linked-to from (or even embedded in) the actual HTML views of the API endpoints.  So that e.g. if someone went to `/api/rates/` with their browser, aside from seeing the HTML-rendered version of the response, they'd also see a link to `/api/docs/` or something.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -5,6 +5,7 @@
 bandit>=1.0,<2.0
 pyflakes==1.6.0
 flake8==3.5.0
+pycodestyle==2.3.0
 pytest==3.0.6
 pytest-cov==2.4.0
 pytest-django==3.1.2
--- a/.bandit
+++ b/.bandit
@@ -1,2 +1,2 @@
 [bandit]
-exclude: node_modules,frontend/tests,api/tests,meta/management/commands,docker_django_management.py,data_capture/tests
+exclude: node_modules,frontend/tests,api/tests,meta/management/commands,docker_django_management.py,data_capture/tests,venv
--- a/.eslintignore
+++ b/.eslintignore
@@ -5,5 +5,6 @@
 htmlcov/
 vendor/
 coverage/
+venv/
 
 frontend/static/frontend/built/
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,5 @@
+[flake8]
+exclude = 
+    */migrations/*.py,
+    node_modules,
+    venv
--- a/api/pagination.py
+++ b/api/pagination.py
@@ -6,8 +6,10 @@ from django.conf import settings
 
 class ContractPagination(pagination.PageNumberPagination):
 
-    def __init__(self, context):
+    def __init__(self, context=None):
         super().__init__()
+        if context is None:
+            context = {}
         self.context = context
         self.page_size = settings.PAGINATION
 
--- a/api/tests/test_docs.py
+++ b/api/tests/test_docs.py
@@ -1,0 +1,7 @@
+from django.test import TestCase
+
+
+class DocsTests(TestCase):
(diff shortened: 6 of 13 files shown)


toolness:
@jseppi what do you think of this jibberjabber so far?  I feel like we could just merge it as-is and then improve it through subsequent PRs...

requirements-dev.txt
```diff
@@ -5,6 +5,7 @@
 bandit>=1.0,<2.0
 pyflakes==1.6.0
 flake8==3.5.0
```
jseppi: Weird that this is required. I just upgraded our flake8 and pyflakes versions and didn’t see any warnings when I did that...or maybe I just missed them? 🤷‍♂️ Doesn’t really matter though.

toolness: Yeah, something might just be messed up with my local configuration. I am confused.

toolness:
Woot thanks!  I don't think the funky special DRF docstrings help--they're just useful if our view class supports more than one HTTP method, but I don't think they can be used to change the name of the verb, as far as I can tell.

Issue #1708, opened 2018-04-04
User menu needs tweaks, and it should close when i click outside of it


ericronne:
### Current
![image](https://user-images.githubusercontent.com/9259185/38326113-8b1697fe-380a-11e8-9fef-679ab3998212.png)

 - [x] Update styles as follows:
#### Default
![image](https://user-images.githubusercontent.com/9259185/38326246-edff17ec-380a-11e8-988d-adc7894c99df.png)

#### Hovers
![image](https://user-images.githubusercontent.com/9259185/38326055-670f4e8c-380a-11e8-9d14-972e314d9aee.png)
**Didn't bother with this ☝️ one, since it causes the line to appear immediately upon open, which is a bit inelegant imo**

- [x] Also, the user has to click their email address or the triangle, in order to close the menu. Ideally it would also close when the user clicked anywhere on the page outside of the menu itself. #1804 fixes

hbillings:
Odd, mine is green by default?
<img width="277" alt="screen shot 2018-04-04 at 12 24 10 pm" src="https://user-images.githubusercontent.com/509309/38326477-20fbaf1e-3803-11e8-97d7-508791f93614.png">

Also, I worry that without any sort of color or underline, the only indication that there's a menu there is the little triangle. It's pretty subtle; is it enough?

(I also have questions about whether we should throw the site admin options in this menu, but that's probably a different convo.)

ericronne:
Green by default is correct, but agreed re subtlety. Broke my own rule: links amid non-link text should be underlined by default. How about these states? …

### Logged in, default
![image](https://user-images.githubusercontent.com/9259185/38332790-68e4ca04-381c-11e8-94cd-7ce3a404f145.png)

### Hover (underscore disappears)
![image](https://user-images.githubusercontent.com/9259185/38332811-7664ef56-381c-11e8-9585-bf6509db6119.png)

### Menu clicked open
![image](https://user-images.githubusercontent.com/9259185/38332824-8417e6da-381c-11e8-9612-960fead6ace4.png)

### Hovers, option A (email and triangle turn green) …
![image](https://user-images.githubusercontent.com/9259185/38332847-92be55a2-381c-11e8-9aef-a3cb8fe7f460.png)

### Or option B (email and triangle stay white)?
![image](https://user-images.githubusercontent.com/9259185/38333121-496a8d48-381d-11e8-8c8c-c843bc8ce321.png)

PR #1804, opened 2018-04-19
Hide the usermenu on outside click

jseppi:
Modifies the behavior of the header user menu to hide not only on click of its trigger link, but also on click of anywhere outside of its descendants.

Part of the tweaks requested in https://github.com/18F/calc/issues/1708 cc @ericronne

### DIFF
--- a/frontend/source/js/common/usermenu.js
+++ b/frontend/source/js/common/usermenu.js
@@ -1,6 +1,7 @@
 // @ts-check
 /* global document */
 
+
 /**
  * Determines if `targetEl` is a descendant of the given `parentEl`.
  *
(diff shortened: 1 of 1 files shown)


ericronne:
Yes, awesome!

frontend/source/js/common/usermenu.js
```diff
@@ -1,26 +1,71 @@
 // @ts-check
 /* global document */
 
-function enableUsermenu() {
+
+/**
+ * Determines if `targetEl` is a descendant of the given `parentEl`.
+ *
+ * @param { HTMLElement|Element } parentEl the parent element
+ * @param { HTMLElement|EventTarget|Element|null } targetEl the element in question
```
hbillings: I've never seen this sort of syntax before -- what's it called so I can google it?

jseppi: They’re jsdoc style comments that are interpreted by typescript for type annotation/checking

PR #1809, opened 2018-04-19
Set memory constraints of docker containers to match deployment

toolness:
To improve dev/prod parity and better anticipate potential problems like #1601 during development, we can use Docker's [user memory constraints](https://docs.docker.com/engine/reference/run/#user-memory-constraints) to simulate the kind of resource limits that our production environment has.

I'm not actually sure if our production environment disables swap like this PR does, but [Brendan Gregg's Linux Performance Tools talk](https://www.youtube.com/watch?v=FJW8nGV4jxY) claims that most production environments do.  It also makes it a lot easier to tell when we're using too much memory, since our app just crashes, instead of slowly thrashing about as the kernel swaps memory to disk.

To test this out, you can create a new Python file with the following content and run it in a container:

```python
from ctypes import c_int

Chunk = c_int * 1_000_000

chunks = []

while True:
    chunks.append(Chunk())
    print(f"Allocated {len(chunks)}M")
```

You should see it get killed by the kernel after it's allocated some amount of memory that's commensurate to its container's `mem_limit`.

I also tried uploading Kelly's latest R10 bulk data with these constraints in-place and it works, so that's a good sign.

In the future, we can also add a configuration test that ensures the container memory limits are identical to the ones specified in the manifests, but I'm going to leave that for a later PR for now.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
     command: sh -c "python manage.py collectstatic --noinput ; gunicorn -b :${DOCKER_EXPOSED_PORT} hourglass.wsgi:application"
     ports:
       - "${DOCKER_EXPOSED_PORT}:${DOCKER_EXPOSED_PORT}"
+    mem_limit: 1G
   db:
     image: postgres:9.5.4
     volumes:
@@ -27,6 +28,7 @@ services:
       - db
       - redis
     command: python manage.py rqworker
+    mem_limit: 512M
   rq_scheduler:
     extends:
       file: docker-services.yml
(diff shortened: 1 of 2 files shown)


toolness:
Uh... I am confused as to why tests are failing, since the changes in this PR only affect docker settings that are used during development, and not during CI.  Weird.

toolness:
Update: I tried changing the batch size of bulk upload processing from 5,000 to 60,000, in an attempt to reproduce the crashes we were seeing on production, but it worked.  Ah well--it's likely that there's other factors involved in the production deployment that might have caused the worker to have even less memory in production than it does in development... but regardless, I still think this PR has brought us closer to dev/prod parity, at least.

PR #1812, opened 2018-04-19
Style user menu to match new design

### LINKED ISSUE #1708: User menu needs tweaks, and it should close when i click outside of it

ericronne:
Closes #1708 

- Added underscore to email address by default, for better visibility/accessibility
- Recolored menu to light green
- Thickened menu-item hover to 2px
- Shifted small triangle (atop menu) leftward
- Shifted entire menu leftward


frontend/source/sass/components/_usermenu.scss
```diff
@@ -34,7 +34,7 @@ $usermenu-width: 150px;
       color: $color-primary-darker;
 
       &:hover {
-        text-decoration: underline;
+        border-bottom: 2px solid $color-primary-darker;
```
hbillings: Why `border-bottom` instead of `text-decoration`? Also, it strikes me as _slightly_ weird now that the user's email gets an underline removed on hover, and these get one added. Hrm.

frontend/source/sass/components/_usermenu.scss
```diff
@@ -44,10 +44,10 @@ $usermenu-width: 150px;
       position: absolute;
       top: -$usermenu-caret-size;
       // aligned in center of menu
-      right: $usermenu-width/2 - $usermenu-caret-size/2;
+      left: $usermenu-width/5 - $usermenu-caret-size/2;
```
hbillings: Plz to remove the comment on line 46 if this isn't center-aligned any more.

frontend/source/sass/components/_usermenu.scss
ericronne: `border-bottom` lets me make the underline **thicker**, which looks better with the bold font.

I generally underline text links which appear adjacent to non-link text, so they stand out more clearly than when we use only color to indicate links. In this case, the down arrow is enough of an indicator, but the underline doesn't hurt.

PR #1813, opened 2018-04-19
Use {% fullpage_example %} tags for card styleguide layout

### LINKED ISSUE #1792: Make all card examples in the styleguide use {% fullpage_example %}
It looks like there are 3 examples under the **Card Layout** section that could use the `{% fullpage_example %}` template tag introduced in #1784!

@hbillings I'm assigning you to this, but I can do it if you're too busy, or @ericronne can take a shot at it too.

hbillings:
Closes #1792, and looks much, much better!


hbillings:
Ugh, so, it looks like aXe wants each iframe to have its own unique title. I'll take a peek tonight and see if I can figure out how to do that, but I might need help.

styleguide/templatetags/styleguide.py
```diff
@@ -242,8 +242,9 @@ def fullpage_example(context, name):
     t = context.template.engine.get_template(
         'styleguide_fullpage_example_iframe.html')
     url = _fullpage_example.get_url(name)
+    id = slugify(url)
```
hbillings: This was the best thing I could come up with to create unique IDs -- it spits out ugly stuff like `styleguidefullpage-examplecard-skinny` tho.

hbillings: oooh, that's much better, thank you! Update coming.

jseppi: I think spaces in `id`s are funky, personally :P I checked on them because I had thought they were not valid, but that was apparently something from HTML4 that was changed in HTML5 (ref https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)

PR #1815, opened 2018-04-20
Run mypy on all files

### LINKED ISSUE #1803: Consider running mypy on all files with reduced strictness
While I was away from CALC working on the OMB Policy Library, we started with the same whitelisting strategy we use for mypy in CALC.

However, it ended up being a hassle to keep the whitelist updated, so in https://github.com/18F/omb-eregs/pull/963 we opted instead to just run mypy on _everything_ but with a very lax criteria.  This ended up being easier because we could just add type annotations to any code we wrote and trust that mypy would make sure everything was still copacetic, without having to worry about whether the file we were editing was in the whitelist or not.

Aside from that, I think that just applying mypy to everything could potentially make editor integration easier, for editors that support it.  (For instance, [VSCode's Python extension supports mypy out-of-the-box](https://code.visualstudio.com/docs/python/linting#_mypy).)

toolness:
This fixes #1803 by upgrading mypy to its latest version, slightly reducing its strictness, running it on all files, and fixing the errors that came up.

## Notes

* The only loosening of mypy's configuration is that [`disallow_untyped_calls`](http://mypy.readthedocs.io/en/latest/command_line.html#additional-command-line-flags) is now `False` instead of `True`.

* We don't run mypy on tests or migrations right now.

* I ended up just getting rid of all our `django` mypy stubs, because they were severely limiting our ability to run mypy on all our files: as far as I can tell, stubs can't just define a subset (or submodule) of a package, they need to define _everything_, which meant that mypy was complaining that a bunch of the Django APIs our code used didn't exist, since we didn't define everything in our stub.  So it was easier to just remove the stub entirely.

    I think this will make our code more maintainable going forward too: had I actually appeased mypy here, we'd continuously run into problems going forward, as mypy would get upset at every new Django API we decided to use.

* I am not a big fan of how mypy complains about assignments to empty lists or dicts as being untyped.  I wish there were a mypy configuration setting to just treat untyped dicts/lists as `Dict[Any, Any]` or `List[Any]`, respectively, but as far as I can tell, there isn't.  It's particularly annoying because it forces us to do a `from typing import List, Dict, Any` at the top of every file that needs it.


api/serializers.py
```diff
@@ -37,4 +37,4 @@ class Meta:
                   'hourly_rate_year1', 'current_price', 'next_year_price',
                   'second_year_price', 'schedule', 'sin', 'contractor_site',
                   'business_size')
-        list_serializer_class = ContractSerializer
+        list_serializer_class = ContractListSerializer
```
toolness: Oof, we were actually defining `ContractSerializer` _twice_ in the same file!  Good catch, mypy.

contracts/models.py
```diff
@@ -98,11 +98,11 @@ def bulk_update_normalized_labor_categories(self):
                 values = []
                 for i in range(num_updates):
                     values.append(r'(%s, %s)')
-                values = ", ".join(values)
+                values_str = ", ".join(values)
                 sql = (  # nosec
                     "UPDATE contracts_contract "
                     "  SET _normalized_labor_category = v.nlc"
-                    "  FROM (VALUES" + values + ") AS v (id, nlc)"
+                    "  FROM (VALUES" + values_str + ") AS v (id, nlc)"
```
toolness: Mypy doesn't like it when a variable changes its type mid-code, so I changed the variable name here to appease it.

data_capture/admin.py
```diff
@@ -258,6 +258,14 @@ def has_delete_permission(self, request, obj=None):
...
 
 
+def short_description(desc: str):
+    def decorator(func):
+        func.short_description = desc
+        return func
+
+    return decorator
+
+
```
toolness: Aside from working around https://github.com/python/mypy/issues/2087, this also makes assigning the short description  easier to read.

data_capture/schedules/registry.py
```diff
@@ -79,7 +87,8 @@ def smart_load_from_upload(classname, f):
                 pass
 
     if pricelist is None:
-        raise original_error
+        default_error = ValidationError('Unrecognized price list!')
+        raise original_error or default_error
```
toolness: I'm not really sure if `default_error` will _ever_ be raised, but mypy was concerned about `original_error` possibly being `None` when it was raised (which is bad, because raised exceptions should always subclass `BaseException`).  So I made `default_error` partly to appease it and partly to cover our backs.  (That said, it's not a great sign when it's hard for both a human and mypy to infer whether `original_error` might be `None` here, but I wanted to limit the scope of this PR.)

docker_django_management.py
```diff
@@ -95,6 +95,8 @@
     # If the Docker host is running on Windows, we don't need this
     # module, so it's OK to not import it.
     import pwd
+else:
+    pwd = None
```
toolness: Bleh. I wanted to make sure `mypy` ran without errors on both Windows and Linux; running it on Windows complained that `pwd` doesn't exist when it's used, so I added some logic to appease mypy on Windows.  I think it is strictly appeasement though, since the code that actually uses `pwd` will only ever be run in the docker container (which uses Linux).

hourglass/changelog.py
```diff
@@ -1,24 +1,45 @@
...
+
+    def __init__(self, name: str, pattern: str) -> None:
+        self.name = name
+        self.regex = re.compile(pattern, re.MULTILINE)
+
+    def search(self, contents: str, filename=CHANGELOG) -> Match:
+        match = self.regex.search(contents)
+        if match is None:
+            raise ValueError(f"{self.name} not found in {filename}!")
+        return match
```
toolness: Decided to do this whole rigmarole because we were just calling `search` on a regexp without checking to see if it returned `None`.  Mypy rightfully complained about this, so I made this little abstraction to make any unsuccessful searches raise helpful errors.

docker_django_management.py
jseppi: all those `if pwd` checks are kinda yucky. Could mypy checking just be disabled for `pwd` within this file? I'm not totally sure of all the escape hatches mypy has, so maybe that isn't possible.

toolness: Yeah, unfortunately disabling type checking for pwd is what I originally tried, which satisfied the windows side, but then the linux side was like "yo dawg u don't need to disable type checking here" which made mypy fail.  Blugh it is annoying.  So I think I am just gonna leave it in this annoying state for now.

Issue #1253, opened 2017-01-18
Implement hover states and check marks to upload step icons


ericronne:
![step indicators](https://cloud.githubusercontent.com/assets/9259185/22077597/67c88b36-dd7a-11e6-9d53-e473bb098f59.png)

ericronne:
dupes #1464

PR #1819, opened 2018-04-20
Implement step indicator styles

### LINKED ISSUE #1464: Step indicator styling: current site colors
Currently the steps indicator has only two states: 
**Not yet visited:**
image
and **You are here / Completed**, which both get identical treatment:
image

## How that looks in action
image

## Update to match this design
image

**Note: we also wanted to add interactivity to the circles, so users could have random access to any page they've already completed. Will do another issue...

hbillings:
Closes #1464.

To do:
- [x] Implement new colors.
~- [ ] Make each step bubble link to its related step.~
- [x] Change color of bubble on hover.
- [x] Replace current step's title with hovered step's title on, er, hover.
- [x] Replace previous steps' numbers with check marks unless hovered over.
- [x] Removes animation of current bubble, because the keyframes animation settings interfere with hover styles.


hbillings:
So getting the URL of all the steps is tough because of the way the step widget is abstracted. It _feels_ like there ought to be a way to calculate the URLs the way we do in `urls.py` or something, but I can't think of a good one. 

@toolness maybe we should pair on that part if you have ideas?

hbillings:
As per @ericronne's excellent suggestion, let's merge this as-is and I'll open another PR for the Link Madness™️.

frontend/source/js/data-capture/step-bar.js
```diff
@@ -0,0 +1,32 @@
+/* global jQuery, document */
+
+const $ = jQuery;
+
+$(document).ready(() => {
```
toolness: Want to just change this to `window.addEventListener('DOMContentLoaded', () => {`, to decouple this whole thing from jQuery (a dependency we'd ideally like to eventually get rid of)?  I believe [`DOMContentLoaded`](https://developer.mozilla.org/en-US/docs/Web/Events/DOMContentLoaded) is the equivalent of jQuery's `.ready()`.

frontend/source/js/data-capture/step-bar.js
```diff
@@ -0,0 +1,32 @@
...
+  const steps = document.querySelectorAll('.step-bar__bubbles li');
+  const stepLabels = document.querySelectorAll('.step-bar__labels li');
+  const currentStep = document.querySelector('.step-bar__labels .current');
```
toolness: Hmm, so one downside of this logic as it currently stands is that it assumes that we have only one step widget per page.  In fact, if we have more than one step widget per page, some really weird stuff might happen.  Granted, having more than one step widget per page probably doesn't make much sense outside of our style guide, but we might want to document it just in case.

Alternatives that could alleviate this might be:

* Wrap all this code in a `document.querySelectorAll('.step-bar').forEach(stepBar => {`, and then use `stepBar.querySelector` instead of `document.querySelector` inside the function.  This will "scope" everything to an individual step bar component, ensuring that multiple step bar widgets on a page don't cause bizarre things to happen.

* Encapsulate this in a custom element (aka web component).  This is my preferred solution because it's how the rest of our data capture JS is structured, but it might be better suited for a pairing session, if you're unfamiliar with how custom elements work.

Oh, aside from just being hygenic, following one of these paths will also probably make the code easier to write unit tests for, too.

frontend/source/js/data-capture/step-bar.js
```diff
@@ -0,0 +1,32 @@
+/* global jQuery, document */
```
toolness: A nice alternative to this that I discovered recently is to use [`/* eslint-env browser*/`](https://eslint.org/docs/rules/no-undef#browser) instead, as this allows you to use a bunch of Web APIs without having to prefix them with `window.`.

An even better thing to do would be to use typescript to type check this code!  You can do that by adding `// @ts-check` to the top of the file.  The great thing about using it is that it will notice incorrect DOM API usage and funky edge cases, like the fact that the `prevLabel` on line 12 might be `null` when you access its class list. But that's up to you.

frontend/templates/frontend/steps.html
```diff
@@ -1,14 +1,21 @@
...
+      </label>
+    </li>
+  {% endfor %}
+  </ol>
+  <ol class="step-bar__bubbles" aria-hidden="true">
     {% for step in steps %}
-      <li title="{{ step.label }}"
-      {% if step.current %} class="current"{% endif %}></li>
+      <li{% if step.current %} class="current"{% endif %}>
+        <a href="javascript:void(0)" title="{{ step.label }}"></a>
```
toolness: I assume this is an `<a>` to make room for future hyperlinking to the appropriate step URL or something?  Right now it exists in the page's <kbd>Tab</kbd> order and is keyboard-focusable, but without any meaningful interactivity, which is weird.  Could we remove it until we actually add a real hyperlink in a later PR?

(Aside from that, I think it's funky to add interactive content in an element with `aria-hidden="true"`. I'm not sure if it's actually a _violation_ of any sort but it definitely feels funky...

toolness:
Hmm, so the interactivity is nice in general but I have a few concerns about it:

* It's _only_ activated by mouseover, which leaves out keyboard and touchscreen users. It would be nice to be inclusive of those users if we can, though I don't know of any patterns that would satisfy that. Not really a big deal though, just more of an observation.

* There's no indication on mouseover that the title of the steps widget is now referring to the name of the step the user is moused-over, rather than the name of they step they're currently on. This means e.g. if they're on the "Validate data" step and happen to leave their mouse cursor on the "Recover costs" step, then get distracted from their computer and come back to it later, they'll see this:

    > screen shot 2018-05-24 at 7 06 32 am

    The step they're _on_ is actually "Validate data" but they might think that they're on "Recover costs", and they could be confused until they move the mouse and notice it flip back to the correct title.

    Anyways, I'm guessing this kind of thing is unlikely to occur in practice, but it does make me wonder if using one of our existing UX patterns, like the tooltips provided by the jQuery tooltipster plugin, would be both more user-friendly _and_ require less custom JS logic.

frontend/source/js/data-capture/step-bar.js
ericronne: You, sir, are nothing if not thoughtful and thorough. Just sayin.

The widget is designed to indicate progress through a multi-page form, so i think by definition there wouldn't ever be more than one per page, no?

frontend/templates/frontend/steps.html
ericronne: It may be wise to disable the hover for pages the user hasn't yet visited (and therefore can't click to visit).

ericronne:
Reminder: this entire feature is labeled as an enhancement :P so if it gets (has gotten?) too hairy to solve in an elegant, accessible way, no harm in dialing it back to the static progress indicator.

-----

Good point re the user inadvertently leaving their mouse on the widget. One possible solve: if we disabled all interactivity for pages not yet visited, the hover label could instead read "Return to: [step]"

ericronne:
Note: the circle for the current page should be filled with green, like so (for step 1):
image

image


thx!

hbillings:
@ericronne wait, this is how it looks for me -- how does it look for you?
screen shot 2018-05-24 at 9 50 45 am

frontend/source/js/data-capture/step-bar.js
toolness: I'd actually argue that we might eventually want to put multiple step widgets on the styleguide, to e.g. show the step widget in different kinds of states.  And as I mentioned earlier, it also makes it a lot easier to unit test.  And if we _do_ keep it a global singleton, I think it actually makes sense to turn `.step-bar` into `#step-bar`, since that would communicate the intent of the style at the CSS level too.

ericronne:
Ah, all "you are here" indicators are A-👌 … *except for page one*
image
☝️ is how it looks when i am on step one.


### Should be …
image

frontend/source/js/data-capture/step-bar.js
hbillings: This is so cool; I didn't even realize this was a thing!

frontend/source/js/data-capture/step-bar.js
hbillings: All we really need to do to show the possible states for the steps widget is display it with the "current" step at neither the beginning nor end.

The unit testing point I totally buy, or maybe I don't know enough about it to debate it with you, lol (which reminds me that I really need to get a walkthrough from you on how our JS tests are set up).

frontend/source/js/data-capture/step-bar.js
hbillings: okay, um, so I'm actually a little confused here -- are `eslint-env browsers` and `//@ts-check` mutually exclusive? It sounds like it from your comment but I don't know enough about this yet. :-/

frontend/templates/frontend/steps.html
hbillings: ahhh you're right; I thought we'd had it on an issue, but since I can't find it, it was probably a thing Ronne and I discussed.

The interactivity thing was way back in #1253, so I'm not surprised you don't remember it!

Since we're prioritizing getting a release out right now, maybe we can all come back to having a conversation about this after we get that shipped?

frontend/source/js/data-capture/step-bar.js
toolness: that is a very good question! personally i wish they were mutually exclusive to avoid confusion, but they are not. it is weird because when enabling typescript with ts-check, you then sometimes have to wrestle with appeasing typescript and eslint at the same time, which is suuuuuper annoying (and part of why i just started aggressively disabling eslint features recently).

that said, using both of the directives i mentioned should appease both typescript and eslint simultaneously. i will try to write something up for the docs that explains this junk better.

frontend/source/sass/components/_steps.scss
```diff
@@ -3,37 +3,39 @@
...
+  li {
+    display: none;
     font-weight: 100;
     margin-top: 0.2rem;
     font-size: 1.4rem;
     line-height: 1.2;
-    color: $color-primary;
+    color: $color-gray-dark;
     font-weight: $font-weight-bold;
+    &.js-show {
```
tbaxter-18f: Rather than toggling a classname, would it be better to toggle the `hidden` attribute?

frontend/source/sass/components/_steps.scss
```diff
@@ -49,33 +51,54 @@ ol.steps {
 
   li {
     list-style: none;
-    border-top: 2px solid $color-primary;
+    border-top: 3px solid $color-gray-lighter;
     padding-top: 14px;
     position: relative;
     float: left;
-    width: 46px; // width of :before psuedoclass + $space-3x (22px + 24px)
+    width: 38px; // width of `a` tag bubble + margin (26px + 12px)
```
tbaxter-18f: I love, love, *love* seeing explanations for magic numbers. Thank you!

frontend/templates/frontend/steps.html
```diff
@@ -1,14 +1,21 @@
 <div class="step-bar">
-  <label>
-    <span class="sr-only">
-      Step {{ current_step.number }} of {{ steps|length }}:
-    </span>
-    {{ current_step.label }}
+  <label class="usa-sr-only">
```
tbaxter-18f: Is `label` appropriate here? There's no `for`, nor any corresponding input that I'm seeing.

frontend/templates/frontend/steps.html
tbaxter-18f: The label is related to the step, right? So rather than two `ol`, would a `dl` make more sense, where you have 

```
<dl>
  <dt>I'm a friendly label for this step</dt>
  <dd>And I'm the step!</dd>
</dl>
```
I could be way off base here, but that feels like it would give you a more semantic relationship and better control of the sibliings. Thoughts?

frontend/templates/frontend/steps.html
toolness: Hmm, that's a good question, but we might want to address it in a separate issue, since it looks like this particular component has been using such a `<label>` since before this PR was created.

frontend/templates/frontend/steps.html
toolness: I realized in https://github.com/18F/calc/pull/1991#discussion_r194037719 that the `href` here can actually probably just be `current_step.number` because it's a relative URL that works due to the way our URL routes are set up.  However, as I later mention in https://github.com/18F/calc/pull/1991#discussion_r194033416, we should probably _not_ use a link for steps that are in the future, since users shouldn't be able to jump to steps that they haven't completed prerequisites for.  (That said, I guess that our view code auto-boots users back to the latest unfinished step when this occurs, so it might not necessarily be a deal-breaker.)

hbillings:
Blocked by #1991

Issue #1818, opened 2018-04-20
Consider changing flake8's max line length to 100 characters


toolness:
I've noticed that a lot of 18F projects (and other projects in general) seem to override PEP8's default line length of 80 characters to be 100 instead.  I wouldn't mind busting this up if other folks are OK with it.

PR #1826, opened 2018-04-30
Improve data capture schedule documentation

toolness:
From the beginning of CALC II, we knew that we'd want to support multiple schedules, so we built this into the architecture.  However, the architecture could use a bit more documentation/explanation, so I've added some here, via comments, docstrings, and type annotations.  I've also simplified a few things to make them easier to maintain in the future.

Other notes:

* I went ahead and changed the maximum line length from 80 characters to 100, as per #1818.

* When we first wrote some of this code, we were using python 3.5, not 3.6.  In this PR I've updated some of our type annotations to use [python 3.6 variable annotations](https://docs.python.org/3/whatsnew/3.6.html#pep-526-syntax-for-variable-annotations) instead of `# type:` comments.

* It seems flake8 / pycodestyle doesn't quite have full support for [python 3.6 variable annotations](https://docs.python.org/3/whatsnew/3.6.html#pep-526-syntax-for-variable-annotations) (specifically, the latest version of pycodestyle supports them, but the latest version of flake8 doesn't seem to support the latest version of pycodestyle).  So I disabled a false positive that was being triggered by flake8, which bizarrely made flake8 notice a bunch of other, unrelated PEP8 violations that it wasn't noticing before, for some reason.  So then I had to fix _those_, which muddles this PR up a bit, unfortunately.

* We now use the [`setting_changed`](https://docs.djangoproject.com/en/2.0/ref/signals/#setting-changed) signal to automatically re-populate the schedule registry whenever tests change it via `override_settings()`.  This relieves the tests from having to manually re-initialize the registry, which was annoying to read, hard to remember to do, and confusing to debug if one forgot to do it.

### DIFF
--- a/data_capture/schedules/registry.py
+++ b/data_capture/schedules/registry.py
@@ -1,21 +1,47 @@
-from typing import List, Dict, Any, Tuple
+from typing import Any, List, Dict, Iterator, NamedTuple, Type, Optional, Tuple
 
 from django.conf import settings
 from django.forms import ValidationError
+from django.core.files.uploadedfile import UploadedFile
 from django.utils.module_loading import import_string
 
 from data_capture.schedules.base import BasePriceList
 
 
-CHOICES: List[Tuple[str, str]] = []
-CLASSES: Dict[str, BasePriceList] = {}
+class Choice(NamedTuple):
+    '''
+    Represents a schedule choice for a price list upload.
+    '''
+
+    # The fully-qualified BasePriceList subclass name that
+    # encapsulates the schedule's business logic, e.g.
+    # "foo.bar.Schedule2PriceList".
+    classname: str
+
+    # The human-readable name of the schedule, e.g. "Schedule 2".
+    title: str
 
 
-def _classname(cls):
+# Global list of all schedule choices.
+CHOICES: List[Choice] = []
+
+# Global dictionary that maps fully-qualified class names to
+# their class instances.
+CLASSES: Dict[str, Type[BasePriceList]] = {}
+
+
+def _classname(cls: type) -> str:
+    '''Return the fully-qualified class name for a class.'''
+
     return '%s.%s' % (cls.__module__, cls.__name__)
 
 
 def _init():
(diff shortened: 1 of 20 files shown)


toolness:
@hbillings we should probably video chat about some of these changes, since the architecture of all this schedule registry stuff can potentially be a bit confusing.  Also, getting your opinion on the architecture will guide how we further document and/or refactor this part of the codebase for the transition, so that will be helpful!

data_capture/schedules/registry.py
```diff
@@ -1,53 +1,114 @@
...
 
     for classname in settings.DATA_CAPTURE_SCHEDULES:
         cls = import_string(classname)
-        CHOICES.append((classname, cls.title))
+        CHOICES.append(Choice(classname, cls.title))
         CLASSES[classname] = cls
 
 
-def get_choices():
+def get_choices() -> Iterator[Choice]:
```
hbillings: This arrow syntax is new to me! What's it called, so I can read up?

toolness: Oh, it's part of [Python 3's type annotation syntax](http://www.daveoncode.com/2017/03/06/writing-better-software-with-python-3-6-type-hints/).  It basically can be translated as "returns a(n)", so in this case it's documenting that `get_choices()` is a function that returns an iterator of `Choice` objects.

It's useful as documentation, so readers don't have to read the whole function body to figure out what it's returning; but also, when used with a tool like [`mypy`](http://mypy-lang.org/), it can be used as a kind of linting mechanism, to ensure that nothing that calls the function ever treats its return value as something _other_ than what we've documented it to be.

And as a further optional bonus, on the text editor/IDE side, it can be used for some fancy code completion stuff.  It looks like there's a [vim-mypy plugin](https://github.com/Integralist/vim-mypy) that brings support to vim (that's what you use, right?) and its readme mentions a few other plugins that add support for mypy too!

toolness:
Oh cool, thanks @hbillings !

PR #1844, opened 2018-05-01
Add help text to admin UI

### LINKED ISSUE #1828: Add help text to admin UI
The admin ui is super confusing, and the least we can/should do is add some guidance via help text to it.

hbillings:
Closes #1828.

This makes use of [django-modeladmin-reorder](https://pypi.org/project/django-modeladmin-reorder/) to rearrange the apps and relabel their models in a more explicit way.  Also adds help text to the data capture review options. It's also ugly as hell and I'm sure @ericronne is gonna have things to say about it (If *I* think it's ugly...man).

Initial state:
screencapture-localhost-8000-admin-2018-05-02-18_01_24

Expanded state:
screencapture-localhost-8000-admin-2018-05-02-18_02_12

@toolness, I need to include the WDS JS on this page to get the accordion to work, but for some reason the way to do that is currently eluding me. Some help would be much appreciated!


hourglass/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
-        </caption>
-        {% for model in app.models %}
-            <tr class="model-{{ model.object_name|lower }}">
-            {% if model.admin_url %}
-                <th scope="row"><a href="{{ model.admin_url }}">{{ model.name }}</a></th>
-            {% else %}
-                <th scope="row">{{ model.name }}</th>
-            {% endif %}
+      <div class="app-{{ app.app_label }} module">
+        {% if app.app_label == 'auth' %}
```
hbillings: This section is kinda gross, but this was the fastest, clearest way I could find to get this info in here. I'm totally open to suggestions on other ways to populate it!

hourglass/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
+          <h2>Manage users and permissions</h2>
+        {% elif app.app_label == 'sites' %}
+          <h2>Update site settings</h2>
+          <p class="usa-alert usa-alert-warning">This section is governed by the <a href="https://docs.djangoproject.com/en/2.0/ref/contrib/sites/">Django Sites Framework</a>, and enables emails sent from this system to reference the correct absolute base URL. You will probably never need to touch this unless the production database gets wiped or the project's URL changes.</p>
```
hbillings: Would love @toolness to check this to see if it's right or not 😬

hourglass/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
+                  <li><strong>To remove user-submitted data:</strong> Select "Approved price lists" and choose the price list you want to remove from CALC. This is referred to as "retiring" a price list. The price list will remain available in the admin view, but will no longer be displayed to users in the data explorer.</li>
+                  <li><strong>To edit data that has already been published:</strong> First, remove the data from public use by retiring it as explained above. View "Retired price lists" and click on the contract number of the desired list. Make any necessary edits, and then return to the "Retired price lists" view to approve and re-publish the pricing data.</li>
+                  <li><strong>To help troubleshoot data submission attempts:</strong> Select "Attempted price list submissions" for a list of submissions that never made it all the way to completion. From here, you can "replay" submission attempts to try to see what glitches the user submitting data might have hit.</li>
```
hbillings: I tried to capture workflows here, but I'm not sure I did it justice.

hourglass/settings.py
```diff
@@ -412,3 +414,23 @@
...
+ADMIN_REORDER = (
+    # Use django-modeladmin reorder to rearrange/rename the apps and models
+    # https://pypi.org/project/django-modeladmin-reorder/
+    { 'app': 'data_capture', 'label': 'User-submitted pricing data', 'models': (
+        {'model': 'data_capture.SubmittedPriceListRow', 'label': 'Mute or unmute submitted price list rows'},
+        {'model': 'data_capture.UnreviewedPriceList', 'label': 'Approve or reject unreviewed price lists'},
+        {'model':'data_capture.ApprovedPriceList', 'label': 'Retire approved price lists' },
+        {'model': 'data_capture.RetiredPriceList', 'label': 'Edit and re-approve retired price lists'},
+        {'model': 'data_capture.RejectedPriceList', 'label': 'Approve rejected price lists'},
+        {'model': 'data_capture.AttemptedPriceListSubmission', 'label': 'Replay uncompleted price list submission attempts'},
```
hbillings: I thought these would be clearer if they were labelled with possible actions rather than just the content, but I'm not sure it *actually* makes things clearer.

ericronne:
Awesome! Design is fine for this sprint; i'll circle back later for tweaks (#1854). I'm not sure i understand the admin section enough (yet!) to give much feedback re content.

calc/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
-            {% endif %}
+      <div class="app-{{ app.app_label }} module">
+        {% if app.app_label == 'auth' %}
+          <h2>Manage users and permissions</h2>
+        {% elif app.app_label == 'sites' %}
+          <h2>Update site settings</h2>
+          <p class="usa-alert usa-alert-warning">This section is governed by the <a href="https://docs.djangoproject.com/en/2.0/ref/contrib/sites/">Django Sites Framework</a>, and enables emails sent from this system to reference the correct absolute base URL. You will probably never need to touch this unless the production database gets wiped or the project's URL changes.</p>
+        {% elif app.app_label == 'data_capture' %}
+          <h2>Manage pricing data</h2>
+          <ul class="usa-accordion">
```
toolness: Er... so something I just realized is that we already actually have the [Expandable Area](https://calc.gsa.gov/styleguide/#expandable-area) component we made for data capture.  One advantage of it is that it actually tracks clicks to google analytics, though I suppose that's not particularly useful in this case.  But another advantage is that it wouldn't require importing all of uswds' JavaScript code (I'm not sure how much weight uswds adds to our bundle though).

What do you think?

calc/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
+            <li>
+              <button class="usa-accordion-button" aria-expanded="false" aria-controls="admin-options">
+                How do I manage pricing data?
+              </button>
+              <div id="admin-options" class="usa-accordion-content">
+                <ul>
+                  <li><strong>To review and exclude rows of user-submitted data:</strong> Select "Submitted price list rows," which contains all of the data that is not yet available in CALC. Note that this view shows data from multiple submitted price lists. Here you can "mute" rows that seem wrong or unusual to prevent them from showing up in CALC's data explorer.</li>
```
toolness: So one thing I noticed when reading this is that the paragraph mentions "Submitted price list rows", but that's not actually a literal title of an admin view... there's "Mute or unmute submitted price list rows", but not "Submitted price list rows".  Hmmmm.

I also wonder if it'd be useful to actually just hyperlink directly to the view in the paragraph, rather than requiring the user to scan down the list of admin views?  I dunno, just a thought, not needed for this first iteration or anything!

calc/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
-        {% endfor %}
-        </table>
-        </div>
+          {% if model.admin_url %}
+              <td><a href="{{ model.admin_url }}" class="changelink">{% trans 'Change' %}</a></td>
+          {% else %}
+              <td>&nbsp;</td>
+          {% endif %}
+          </tr>
+      {% endfor %}
```
toolness: Is this part unmodified from the original Django template?  I wonder if it'd be useful to mention that it is, just to reduce the cognitive load of anyone who's reading it... but then again, it could rot if someone did decide to change this part but forgot to change the comment!  Gahh.

calc/templates/admin/index.html
toolness: Heh, looking at the code more closely now, I realize that "Mute or unmute submitted price list rows" _used_ to be called "Submitted price list rows", so I'm guessing you wrote these instructions before renaming some of the admin views, but I could be wrong.

calc/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
-                <th scope="row">{{ model.name }}</th>
-            {% endif %}
+      <div class="app-{{ app.app_label }} module">
+        {% if app.app_label == 'auth' %}
+          <h2>Manage users and permissions</h2>
+        {% elif app.app_label == 'sites' %}
+          <h2>Update site settings</h2>
+          <p class="usa-alert usa-alert-warning">This section is governed by the <a href="https://docs.djangoproject.com/en/2.0/ref/contrib/sites/">Django Sites Framework</a>, and enables emails sent from this system to reference the correct absolute base URL. You will probably never need to touch this unless the production database gets wiped or the project's URL changes.</p>
+        {% elif app.app_label == 'data_capture' %}
+          <h2>Manage pricing data</h2>
```
toolness: This is a lot of content, and it's related to Data Capture, whereas this template is in the project folder... what do you think of moving this content out to, say, `data_capture/templates/data_capture/admin_help.html` and then [`include`](https://docs.djangoproject.com/en/2.0/ref/templates/builtins/#include)'ing it from here?

calc/templates/admin/index.html
```diff
@@ -16,34 +16,59 @@
...
-        {% for model in app.models %}
-            <tr class="model-{{ model.object_name|lower }}">
-            {% if model.admin_url %}
-                <th scope="row"><a href="{{ model.admin_url }}">{{ model.name }}</a></th>
-            {% else %}
-                <th scope="row">{{ model.name }}</th>
-            {% endif %}
+      <div class="app-{{ app.app_label }} module">
+        {% if app.app_label == 'auth' %}
+          <h2>Manage users and permissions</h2>
```
toolness: I wonder if it'd be useful to link to the [Permission groups](https://calc-dev.app.cloud.gov/docs/auth.html#permission-groups) section of our developer docs here?  Although that's written for a developer audience, the "Permission groups" section does explain what the built-in groups are for... anyways, up to you.

calc/settings.py
```diff
@@ -412,3 +414,41 @@
...
         'django.contrib.admindocs',
     )
+
+ADMIN_REORDER = (
+    # Use django-modeladmin reorder to rearrange/rename the apps and models
+    # https://pypi.org/project/django-modeladmin-reorder/
+    {'app': 'data_capture', 'label': 'User-submitted pricing data', 'models': (
+        {
+            'model': 'data_capture.SubmittedPriceListRow',
+            'label': 'Mute or unmute submitted price list rows'
```
toolness: I was going to say that we ought to be able to change this label from `data_capture/admin.py`, to put it closer to where someone would likely look if they wanted to change it, but apparently there's no built-in way to do that with Django admin other than by changing the `Meta` class of a model, which is super weird! Ah well.

calc/templates/admin/index.html
toolness: Actually... do you think we'd lose much if we _didn't_ rename the admin views?  My one concern is that, after looking at django-modeladmin-reorder's source, it's kind of tightly coupled to Django admin's internals, and I could foresee a day when it stops being maintained (it currently only has 6 contributors).  If that happens, it'd be nice if we could just turn off its middleware and CALC would still be fine, just with a more inconveniently-ordered admin section. But if migrating away from django-modeladmin-reorder means changing a bunch of documentation to refer to differently-named views, that could be hard...

calc/templates/admin/index.html
toolness: Using our fancy [bundle analyzer](https://calc-dev.app.cloud.gov/docs/frontend.html#analyzing-bundles) it looks like including uswds adds about 22k to our bundle, which isn't super awesome... but if you think we'll be using uswds' JS features a lot in the future then I guess it's worth it!

calc/settings.py
```diff
@@ -146,6 +146,7 @@
     'frontend',
     'slackbot.apps.SlackbotConfig',
     'uswds_forms',
+    'admin_reorder',
```
toolness: Looking at [django-modeladmin-reorder's github repo](https://github.com/mishbahr/django-modeladmin-reorder), it seems this line isn't actually required, because all the reordering logic is defined in the middleware (the repo has other files in it, like templates, that it doesn't seem to ever use, which is super weird).  But the package's docs _do_ say to include this in `INSTALLED_APPS`, so I guess we might as well include it... odd.

toolness:
Oh shoot, there is one other thing I wanted to mention... Is `/admin/` being tested in any way?  One way to find out is to add an error to the `admin/index.html` template that makes it throw an exception in some way--this could just be some malformed template syntax, then visit `/admin/` to make sure the page explodes, and then run the test suite to make sure at least one test now fails.

If all the tests _pass_, though, that means we don't have any tests that visit `/admin/`, and we should probably add one to "smoke test" that page, at the very least.

calc/templates/admin/index.html
hbillings: ☝️ This screenshot is what I was after in the first place, but I couldn't figure out how to get help text to show up on the model, only the model's fields. I'm not a fan of the huge wall-o-text thing (though I think maybe it could be helpful to document what a typical workflow looks like). 

Another alternative is to rework the admittedly slapdash text to not specifically refer to the fields by name: instead of "Select 'Submitted price list rows' to review and hide selected data from CALC" something like, "You can review submitted price list rows to make sure the data you're approving for inclusion into CALC makes sense. If you see anything that looks odd, you can 'mute' it to keep it from showing up." (Actually, I should take another pass at the language here regardless, because it's pretty stilted.)

calc/templates/admin/index.html
hbillings: ummmm...maybe? This is unmodified from the last round;I don't know why it's showing up as new. There _is_ a comment at the top of the page that says this template `adapted from https://github.com/django/django/blob/stable/1.8.x/django/contrib/admin/templates/admin/index.html`, do you think that does the trick?

calc/templates/admin/index.html
hbillings: Isn't most of the admin data capture-related? Though I do like your point about factoring the custom stuff out of the Django-derived stuff as much as possible. 

If I can make it happen, I'd like to get more help text on more admin pages (like the individual change list views). So in that case, maybe it makes sense to make a folder in the `/admin/` folder that is `/data_capture/` and put everything like this in there? That way it's adjacent to the thing it describes but also separate.

calc/settings.py
hbillings: Django Admin: A Pain In Your Butt Since 2005

calc/settings.py
hbillings: 🤷‍♀️

calc/templates/admin/index.html
ericronne: The wall of text isn't bad, since it's hidden in an accordion by default, but i agree that it seems handier to situate the help text next to the links, like that mockup shows. Then we can use your plainer-language approach without worrying about noting the explicit task name.

The active voice strikes me as better for the links, btw: "Reject price lists" eg (as opposed to "Rejected")

hbillings: If @toolness has ideas on how to make that happen, I'm all ears!

So the "Rejected price lists" page shows you price lists you've _already_ rejected. This is part of why I changed the language in my version to "Approve rejected price lists." 

It would be bloody amazing to be able to bundle these into views like, "New price list data" (with options to review submitted rows or edit unreviewed price lists), "Public price lists" (which would really just be approved ones, I suppose), and "Price lists not shown publicly" (unsuccessful attempts at submission, rejected, and retired lists), with dropdowns on each of those pages to filter by the status.

ericronne:
Cool. I'm going to try noodling with some changes locally, mostly to help get my head better wrapped around this gordion knot

hbillings:
@toolness okay, so I deleted an opening `{% if %}` tag (line 20 of `admin/index.html`) and ran the tests and they did indeed blow up (on `AdminLoginTest.test_is_staff_user_can_view`). HTML tests are tough for me to get my brain around, so if you think there's something else that should be tested, let me know and I'll take a crack at it!

toolness:
NICE, I really like the new text! 💯 !

And cool, it sounds like your new template code is at least being evaluated/executed by the test suite already, so don't worry about busting more tests.

I do still think we should leave the names of the admin views as-is (e.g. "Submitted price lists" rather than "Mute or unmute submitted price list rows") but it's up to you.

hbillings:
I'm going to leave this as-is for now; Ronne and I decided this really needs to be addressed with more refinement in future sprints. The language will be part of that!

PR #1845, opened 2018-05-01
Rename 'hourglass' to 'calc'

### LINKED ISSUE #93: Replace "hourglass" references in code with "calc"
Specifically:
- [ ] updated README to remove Hourglass references
- [ ] `hourglass` app dir
- [ ] `hourglass_site` app dir
- [ ] `hourglass_site/static/hourglass_site` (I would love to kill this redundant directory while we're at it—[openFOIA](https://github.com/18F/foia-hub/tree/master/foia_hub/static) knows how to do this)
- [ ] various settings in `hourglass/settings.py`
- [ ] database name in `hourglass/local_settings.example.py`
- [ ] `hourglass_site/static/hourglass_site/js/hourglass.js`
- [ ] references to `hourglass.js` in `hourglass_site/templates/base.html` and `hourglass_site/static/hourglass_site/js/tests.js`

toolness:
Fixes #93.

Notes:

* There are still a handful of references to Hourglass in `.about.yml`, but that is because it's describing history, e.g. mentioning that the project was originally called Hourglass and later renamed to CALC.

* I added a step to `update.sh` that clears a bunch of cache directories Python uses, which causes tests to fail when package directories are renamed.  For more details, see https://github.com/18F/calc/pull/1845#discussion_r185360755.


update.sh
```diff
@@ -2,6 +2,10 @@
 
 set -e
 
+echo "----- Clearing Python cache -----"
+# https://stackoverflow.com/a/30659970
+find . | grep -E "(__pycache__|\.pyc|\.pyo$)" | xargs rm -rf
```
toolness: When I renamed `hourglass/` to `calc/`, Python complained about some cached files thinking they were in an `hourglass` directory when they weren't, which broke tests. This clears Python's cache whenever devs run `bash docker-update.sh` so that we don't have to deal with that.

.about.yml
```diff
@@ -59,7 +59,7 @@ links:
   text: CALC production instance
 - url: https:/calc-dev.apps.cloud.gov/
   text: CALC development instance
-- url: https://trello.com/b/LjXJaVbZ/hourglass
+- url: https://trello.com/b/LjXJaVbZ/calc
```
toolness: Interestingly, this link actually works!

Although, uh, since we are moving from Trello to github, maybe we should remove this entirely. I guess I will leave that for a separate issue, which I have filed as #1846.

hourglass/setup.sql
```diff
@@ -1,2 +0,0 @@
-ALTER USER hourglass CREATEDB;
-GRANT ALL ON DATABASE hourglass TO hourglass;
```
hbillings: Just makin' sure that it's okay to totally remove these instead of rename them!

update.sh
hbillings: THANK YOU

Issue #847, opened 2016-10-03
Add documentation for Google Analytics


toolness:
~~In #842 we added two custom dimensions to GA but they need to be configured via the admin UI. We should make sure that we document this in the README or somewhere similar!~~

At this point we should add some documentation about the following custom GA functionality:

* The distinction between the DAP and TTS analytics accounts, and which ones track what:

  * [DAP](https://github.com/digital-analytics-program/gov-wide-code) tracks outbound links and document downloads, and because it's a Google Analytics 360 account gives us access to premium functionality like funnel reports. However, it also lumps all the tracking data into a property for all GSA traffic, which means we have to configure custom reports to pull out only CALC-related data. Also, DAP shouldn't be used to track anything that happens behind a login (see #903).

  * TTS analytics tracks custom events and dynamic/virtual page views, as well as [custom dimensions](https://support.google.com/analytics/answer/2709828?hl=en#example-session). Because CALC is its own GA property, viewing reports is easier, but less featureful since it's not a Google Analytics 360 account.

* Document all TTS custom events and/or dimensions introduced in #842, #869, #1706.

* Either document the DAP custom events or link out to the [DAP reference PDF](https://github.com/digital-analytics-program/gov-wide-code/blob/master/documentation/GSA%20DAP%204.0%20-%20DAP%20Code%20Capabilities%20Summary%20and%20Reference.pdf).

toolness:
Um, we should also make sure we configure the admin UI in our current GA config. Not really a github issue per se but I figured I'd mention it here, since we haven't yet configured it.

PR #1852, opened 2018-05-03
Track virtual pageviews and custom events on DAP

toolness:
As mentioned in #847 and #940, we currently report custom pageviews and custom events only to 18F Google Analytics, but not DAP.  This fixes things so we report them to both.  This will hopefully help make it less confusing to figure out which GA account to look at.

## A note on eslint

After 2 years of struggling with eslint and just now almost tearing my hair out trying to placate eslint and typescript at the same time, I'm officially declaring that I am tired of eslint and don't think it's really worth our time to appease.

In the spirit of ["A foolish consistency is the hobgoblin of little minds"](https://www.goodreads.com/quotes/353571-a-foolish-consistency-is-the-hobgoblin-of-little-minds-adored), I'm aggressively telling eslint to ignore rules it complains about, but that I think are pointless (or better auto-fixed in the indeterminate future by something like [prettier](https://github.com/18F/calc/pull/1493)).


toolness:
Woot thanks!!

PR #1859, opened 2018-05-07
Add sample users and "switch user" menu

### LINKED ISSUE #1842: Make it easier to see how users with different permissions see admin pages
I was asking about showing a given item in the admin UI, and @toolness told me it was only visible to superusers. In order to see how users that _aren't_ superusers see things, you have to create a new user in the admin UI, log out, and log back in as that user. This is unwieldy, particularly if you are (like I am) trying to check multiple views at once.
 
@toolness had this idea, and we both thought it would be helpful for future devs:                       
> it'd be nice to have a little menu somewhere in dev mode that makes it super easy to "change context" to different user types, e.g. "show me how a data admin would view this page", "show me how a contract officer would see this page" etc

toolness:
This fixes #1842 by adding a "Switch user" menu to the development nav, visible only when `DEBUG` and `SHOW_DEBUG_UI` are true (the default on dev instances):

> switch-user

The third entry is highlighted because the mouse is hovering over it.

## Notes

* The text of "Style guide" and "Docs" in the developer navigation has changed from white to improve the contrast ratio.

* I used a [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element to avoid the need for any JS but I'm now realizing that this might not be semantically accurate.  Blarg whatever it's just for devs.

## To do

- [x] Modify the [initial user creation](https://calc-dev.app.cloud.gov/docs/auth.html#initial-user-creation) section of the docs as needed.

- [x] Add tests.


data_explorer/templates/_banner.html
```diff
@@ -11,10 +11,28 @@
...
-      <p id="site-status">
+
+      <div class="developer-nav">
         <a href="{% url 'styleguide:index' %}">Style guide</a> |
         <a href="/docs/">Docs</a>
-      </p>
+        {% if show_debug_ui %}
+        |
+        <details>
+          <summary>Switch user&hellip;</summary>
```
hbillings: oh, interesting, I haven't seen these really used before!

docs/auth.md
```diff
@@ -22,9 +22,9 @@ Running `docker-compose run app python manage.py initgroups` will initialize or
 
 ### Initial user creation
 
-When CALC is run for the first time, there will be no users in its database.
+During development, a menu is available at the top of every page which allows you to quickly log in as a variety of example users, each of which represents a different kind of CALC role.
```
hbillings: Do we expand upon what these user roles are somewhere?

toolness: Yup, in the section right above this one :)

PR #1868, opened 2018-05-08
Disable dumb eslint rules

toolness:
Again in the spirit of [aggressively rolling back eslint](https://github.com/18F/calc/pull/1852), this disables a bunch of eslint rules that haven't provided much benefit and have cluttered our code with `// eslint-disable-line` directives.

@hbillings let me know if you actually find some of these warnings useful, I'm OK with re-enabling them.  I guess I'd just rather move from "our eslint is complaining about this, let's consider disabling it for that line" to "our eslint is complaining about this, let's think about whether it's even a useful rule and disable it if not".


toolness:
O good idea, I will re-enable some of those.

PR #1869, opened 2018-05-08
Validate 'sort' field in API requests

### LINKED ISSUE #257: Sorting CSVs on unindexed columns times out
The `/rates/csv` endpoint allows sorting on arbitrary combinations of columns in the `Contract` model. Since some of those columns aren't indexed, and since this endpoint attempts to load all matching rows in the table, some queries take a long time or time out, throwing a 504. If enough slow queries hit the database at once, other parts of the site might slow down as well.

screen shot 2015-11-10 at 10 31 51 pm

Also, attempting to sort on a string that isn't a column throws a 500: https://api.data.gov/gsa/calc/rates/csv/?sort=vendor_name. The expected response code would probably be a 400 or 422.

Recommendations:
- [ ] Index all columns that should support sorting
- [ ] Disallow sorting on any non-indexed columns
- [ ] Disallow sorting on strings that aren't columns

toolness:
Fixes #257.

## Notes

* This works by introspecting on the [`db_index`](https://docs.djangoproject.com/en/2.0/ref/models/fields/#db-index) attribute of every field in the `Contract` model, ensuring that whatever we're indexing is a sortable field.

* We're raising a [Django REST Framework `ValidationError`](http://www.django-rest-framework.org/api-guide/exceptions/#validationerror) when an invalid sort field is provided. This is a bit funky: as DRF's documentation states, it's _supposed_ to only be raised by DRF serializers and field validators, but because CALC 1's view code is very unconventional, it doesn't use either. So we're just raising it in the actual view, which nonetheless seems to work out okay (tests verify this, so we'll be notified if there's ever a regression).

* I've added indexes on `idv_piid` and `vendor_name`.  We'll probably need the latter anyways for supporting vendor search (which I _think_ was brought up as a potential feature, though I can't find a GH issue for it at the moment).  ~~However, `idv_piid` was only really added to ensure tests didn't break, and I'm not sure whether it'll be useful in practice or not.  That said, there _is_ a comment next to the field that says "index this field", despite the fact that it wasn't indexed before, so 🤷‍♂️ .~~ Both fields are definitely needed because the front-end allows them both to be sorted!


api/tests/test_rates_csv.py
```diff
@@ -2,15 +2,15 @@
 
 from django.test import TestCase
 
-from .test_rates_api import GetRatesTests
+from . import test_rates_api
```
toolness: Changed this to import the module rather than the class: when we import the class, pytest thinks it needs to be run as a test suite!

api/tests/test_rates_csv.py
```diff
@@ -19,3 +19,17 @@ def test_prefixes_excel_formula_chars_in_query(self):
...
+        self.assertEqual(resp.status_code, 400)
+        self.assertEqual(resp.json(), [
+            'Unable to sort on the field "sin"'
+        ])
+
+    def test_unindexed_sort_field_raises_400(self):
+        resp = self.client.get(f'{self.path}/?sort=blarg')
+        self.assertEqual(resp.status_code, 400)
+        self.assertEqual(resp.json(), [
+            '"blarg" is not a valid field to sort on'
```
hbillings: why not?? poor `blargh` never has any fun.

api/views.py
```diff
@@ -216,11 +217,21 @@ def get_contracts_queryset(request_params, wage_field):
     price = request_params.get('price', None)
     price__gte = request_params.get('price__gte')
     price__lte = request_params.get('price__lte')
-    sort = request_params.get('sort', wage_field)
+    sort = request_params.get('sort', wage_field).split(',')
```
hbillings: I'm a little unsure of what the `split` is doing here. (I looked at a sample request at `api/docs`, and I didn't see a `wage_field`, er, field returned, so I'm not sure what it's supposed to be.)

toolness: Good catch--the docs for `sort` just say "the column to sort on", but in reality this can be _multiple_ fields, separated by commas.  So we should update the docs to reflect this. Filed #1889 to address it later.

Issue #69, opened 2015-01-21
Document histogram data


shawnbot:
In order to close out [this card](https://trello.com/c/y1tCNA3V/47-show-a-histogram-of-prices-rather-than-simply-a-range-and-average) we need to update the README with some docs on the histogram data. We should explain:

1. How the `histogram` query string parameter works,
1. What the histograms represent, and
1. How they are used on the frontend

PR #1870, opened 2018-05-08
Document more of the API

toolness:
* Adds documentation for the histogram results, fixing #69.
* Adds documentation for the general result format (e.g. pagination details).

## Notes

* I didn't document the contract details very well, partly because they're very similar to the query arguments, and I didn't want to violate DRY. I think we can improve upon this in a later PR.

## To do

- [x] Add documentation for Contract result fields.
- [x] Add documentation for autocomplete result fields.


api/views.py
```diff
@@ -304,6 +304,53 @@ def quantize(num, precision=2):
...
+            contains the labor rate.
+        * `vendor_name` is the name of the vendor that the
+            labor rate's contract is under.
+        * `education_level` is the minimum level of education
+            for the labor rate.
+        * `min_years_experience` is the minimum years of experience
+            for the labor rate.
+        * `hourly_rate_year1`, `current_price`, `next_year_price`,
+            and `second_year_price` contain pricing information for
+            the labor rate.
```
toolness: Uhhh, I forget what the differences between all these are, so I just lumped them together and provided a vague description. 🙁

hbillings: I *think* what's going on here is that `hourly_rate_year1` is the rate for the year the contract started;  `current_price` is the price based on the escalation rate, if applicable; and `next_year_price` and `second_year_price` pertain to out-year pricing (the +1/+2 buttons in the filtering system).

api/views.py
```diff
@@ -304,6 +304,53 @@ def quantize(num, precision=2):
...
+    * `results` is an array containing the results for the
+    current page. Each item in the array contains the following keys:
+        * `id` is the internal ID of the rate in the CALC database.
+            This can be passed to the `exclude` query parameter to
+            exclude individual rates from search results.
+        * `idv_piid` is the contract number of the contract that
+            contains the labor rate.
+        * `vendor_name` is the name of the vendor that the
+            labor rate's contract is under.
+        * `education_level` is the minimum level of education
```
hbillings: This is different than the way the sort uses it, right?

toolness: Umm how do you mean?

PR #1871, opened 2018-05-08
Document the CSS

### LINKED ISSUE #1840: Gently refactor & document CSS
As part of the "transition sprint," document/clean up parts of the CSS that are old and unwieldy. (There are a few of these that Former Heather and the CALC I Team have in there.) Any components should be documented in the styleguide.

hbillings:
Closes #1840, #1878 and #1891.

- [x] Add a section to `/docs/frontend.md` describing some development practices for CSS/Sass.
- [x] Document order in `main.scss`.
- [x] Document why and where the scattered bits of CSS not in `frontend/source/scss` are.
- [x] Comb through components and document anything funky (especially anything that Previous Heather noted was funky in a comment).


data_capture/templates/data_capture/price_list/step_3_errors.html
```diff
@@ -8,7 +8,7 @@
 
       {% with total=rows|length %}
 
-      <div class="message message-error no-bg" role="alert">
+      <div class="usa-alert usa-alert-error alert__no-bg" role="alert">
```
toolness: For some reason I read `alert__no-bg` as "alert, no biggie."

docs/frontend.md
```diff
@@ -4,7 +4,7 @@ In general, front end code is in the [frontend](../frontend/) directory.
 
 This guide is about the nuts-and-bolts of developing front end code; for
 details on how to use or style individual components, see the
-[style guide][].
+[style guide][https://calc-dev.app.cloud.gov/styleguide/].
```
toolness: Oh, this is actually defined further down in the file... it's actually a [reference-style link](https://daringfireball.net/projects/markdown/syntax#link), which I often like better when I know that folks will be reading the original Markdown text, because it makes the original easier to read (b/c the reader doesn't have to find the end of the URL to keep reading the original text).

Can you remove the URL you added here? Thanks!

docs/frontend.md
```diff
@@ -29,21 +29,61 @@ docker-compose run app yarn gulp build
...
+[U.S. Web Design System][https://github.com/uswds/uswds/blob/develop/src/stylesheets/core/_variables.scss]
+and we define new values for them in
+[frontend/source/sass/base/_uswds_variables.scss](../frontend/source/sass/base/_uswds_variables.scss).
+Other variables are specific to CALC, and those are defined in
+[frontend/source/sass/base/_variables.scss](../frontend/source/sass/base/_variables.scss).
+
+This project generally follows a modified (BEM)[https://frontend.18f.gov/css/naming/#bem] (Block, Element, Modifier)
+naming scheme. This prevents namespace collisions and alleviates the need for too much nested Sass.
+
+The `sass` files follow a few conventions:
```
toolness: Should this be Sass instead of `sass`, since you were using Sass earlier?  Not a big deal either way, just wanted to point it out.  Or I guess you could say "The files in the `sass` directory" if you're referring to those?  Eh, I'm probably overthinking this.

docs/frontend.md
```diff
@@ -29,21 +29,61 @@ docker-compose run app yarn gulp build
...
+- Core styles are divided into four categories: [admin](../frontend/source/sass/admin/) (controls styling for the Django site admin),
+[base](../frontend/source/sass/base/) (styles like resets, grid, and variables), [components](../frontend/source/sass/components/)
+(individual site components like the header and footer), and [libs](../frontend/source/sass/libs/) (vendored CSS).
+- All of the site's core styles are imported in [frontend/source/sass/main.scss](../frontend/source/sass/main.scss).
+- Partials (also called includes or imports) are files that will get compiled into the main CSS file. These are prefixed
+with an underscore (such as `components/_footer.scss`).
+
+A more detailed explanation of how to use specific components can be found in the CALC [style guide][https://calc-dev.app.cloud.gov/styleguide/].
```
toolness: This is actually another advantage of reference links--you don't have to put the URL in there, as the Markdown parser will find the reference and hyperlink everything that uses it to the URL.

Also note that reference links use square brackets rather than parentheses (the other links are called inline links b/c the URL goes "inline" w/ the rest of the link text).

docs/frontend.md
```diff
@@ -29,21 +29,61 @@ docker-compose run app yarn gulp build
...
+the data capture tool (which encapsulates both the workflow that starts at
+`calc.gsa.gov/data-capture/step/1` and the bulk upload available to site admins
+at `calc.gsa.gov/data-capture/bulk/region-10/step/1`), and the site admin
+(the place site admins can approve submitted price lists, manage users, and more,
+available at `calc.gsa.gov/admin/`).
 
 ### Data explorer
 
-The data explorer is a [React][]-based app that uses [Redux][] for data flow and state management. It's located in[frontend/source/js/data-explorer](../frontend/source/js/data-explorer/).
+The data explorer is a [React][https://reactjs.org/]-based app that uses [Redux][https://redux.js.org/] for data flow and state management. It's located in[frontend/source/js/data-explorer](../frontend/source/js/data-explorer/).
```
toolness: React and Redux are reference links, so no need to fill in the URLs here.

Also there's a missing space between `in` and the link at the last sentence.

docs/frontend.md
```diff
@@ -29,21 +29,61 @@ docker-compose run app yarn gulp build
...
 
 #### Testing
 
-The data explorer's test suite uses [Jest][], and the tests are located in [frontend/source/js/data-explorer/tests](../frontend/source/js/data-explorer/tests/).
+The data explorer's test suite uses [Jest][https://facebook.github.io/jest/], and the tests are located in [frontend/source/js/data-explorer/tests](../frontend/source/js/data-explorer/tests/).
```
toolness: Please remove the URL since this is a reference link too.

docs/frontend.md
```diff
@@ -96,7 +136,7 @@ Other parts of CALC are usually stored in either Django templates
 or somewhere under the [frontend/source/](../frontend/source/)
 hierarchy.
 
-When in doubt, see the [style guide][]!
+When in doubt, see the [style guide][https://calc-dev.app.cloud.gov/styleguide/]!
```
toolness: Parens not square brackets!

docs/frontend.md
```diff
@@ -29,21 +29,61 @@ docker-compose run app yarn gulp build
 
 ## Developing the front end
 
+### JavaScript
+Globally, we use [yarn](https://yarnpkg.com/en/) to manage our node dependencies and run node tasks.
+`yarn`, in addition to being faster than using `npm install`, has the
+benefit of locking dependency versions via a `yarn.lock` file.
+Read the [yarn workflow docs](https://yarnpkg.com/en/docs/yarn-workflow) if you are not familiar with how to use it.
```
toolness: Yarn and yarn workflow docs are already reference-style links with their URLs at the bottom of this file, so no need to make them inline links... Although I guess if you prefer inline, please remove the reference links at the bottom of the file.

frontend/source/sass/admin/changelists.scss
```diff
@@ -11,6 +11,7 @@ th.action-checkbox-column {
   }
 }
 
+// same as .sidebar in overrides.scss
```
toolness: Hmm what does this mean?  Is it the same exact CSS?  If so, could we use a mixin or something to reduce the repetition and the need for the comment?  (My main concern is that `.sidebar` in `overrides.scss` could get moved/changed/etc and this documentation could no longer be accurate.)

frontend/source/sass/admin/overrides.scss
```diff
@@ -4,7 +4,15 @@
  * Django admin CSS *
  * *****************/
 
-/* nav ul li temp styles until we can get the proper class added in the back end */
+// These styles are here to make the Django admin interface feel more like the
+// rest of the site. The class names are those automatically generated from
+// Django's templating system, so they don't follow our BEM methodology and
+// sometimes we hae to target things in clumsy ways.
+// There are a number of !important declarations used when trying to override
+// Django styles that are already at the highest possible level of specificity.
```
toolness: Nice explanation!

frontend/source/sass/base/_grid.scss
```diff
@@ -1,10 +1,25 @@
+// This is largely a reimplementation of the US Web Design System's grid.
+// At the time this was created, the WDS was incompatible with this project,
+// having an old version of the Skeleton CSS framework in place. The naming
+// conventions come from swapping that project out but retaining old code.
```
toolness: Noice explanation!

frontend/source/sass/base/_utils.scss
```diff
@@ -1,6 +1,8 @@
 @import 'variables.scss';
 
 // legacy sr-only class
+// This class is applied to things that should ONLY be rendered to screen readers,
+// such as descriptions of form fields that would be redundant for sighted users.
```
toolness: Oh! USWDS actually has a [`.usa-sr-only`](https://designsystem.digital.gov/components/accessibility/) that we could be using instead of this... should we file an issue to migrate to that eventually? I think we originally defined our own b/c of how we weren't originally using USWDS or whatever.

docs/frontend.md
hbillings: I  think I was trying to call out the file extension, but that's `scss` so that doesn't even make sense. I'll make this more standard one way or another.

frontend/source/sass/admin/changelists.scss
hbillings: A mixin's not a bad idea, though I'm not sure where to define it. These are both Django override files, and we're styling two disparately named things to look the same. I'm inclined to let this be for right now, because I'm really hoping to do some more work on the admin interface (including maybe updating it so that these aren't divergent any more).

frontend/source/sass/base/_utils.scss
hbillings: Oh excellent! For some reason I couldn't find this last time I went poking around in the WDS and thought they'd gotten rid of it.

frontend/source/sass/admin/changelists.scss
hbillings: Filed #1899 to look into this.

Issue #1874, opened 2018-05-08
Consider adding more choices to Contract model fields


toolness:
A number of the `CharField` fields in the `Contract` model appear to actually involve choices, but don't have actual [`choices`](https://docs.djangoproject.com/en/2.0/ref/models/fields/#choices) associated with them:

* `schedule` must be one of a finite list of schedules (albeit one that is growing).
* `site` should be `customer`, `contractor`, or `both`.
* `business_size` should be `s` or `o`.

Assigning `choices` to these would improve our data integrity, but we should make sure that none of CALC's existing data actually violates the constraints first.

toolness:
Um, ok, I'm working on this now and discovering that `site` and `business_size` are not very well-defined. Our API view currently has this code in it:

```py
    if site:
        contracts = contracts.filter(contractor_site__icontains=site)
    if business_size == 's':
        contracts = contracts.filter(business_size__istartswith='s')
    elif business_size == 'o':
        contracts = contracts.filter(business_size__istartswith='o')
```

Browsing through the data, we've got `contractor_site` values like `Contractor Site` and `Contractor`, `Customer Site` and `Customer`, etc.

Gaaaaah.

PR #1875, opened 2018-05-09
Add data_quality_report management command.

### LINKED ISSUE #1829: Document less-than-ideal aspects of existing CALC data
e.g.:

* That some data has no education or experience metadata

* That some labor categories are duplicated multiple times w/ differing fields (education, sin number, etc) - see #1021, #1029

* That some data is anomalous (e.g. the $700 Teleprompter Operator) - see #1432

* ~~The `sin` field is messed up - see #1033~~ Addressed this in #1850.

* Depending on when the data was updated, the "current price" column may not actually be accurate - see #545.

toolness:
This attempts to fix #1829 by adding a `data_quality_report` management command and a `/data-quality-report/` view that shows statistics about the less-than-ideal aspects of CALC's data.

## Example output

This is based on a snapshot of the database I took today (May 10) via `manage.py load_api_data`.

> screen shot 2018-05-10 at 8 54 38 am

## To do

- [x] Add tests.
- [x] Document the command somewhere.
- [x] Make the view look a bit more snazzy.
- [x] Add a [`verbose_name`](https://docs.djangoproject.com/en/2.0/ref/models/fields/#verbose-name) for `idv_piid` so that `"contract number"` shows up in the report instead of `idv piid`.  ~~I was going to do this in this PR, but it requires a db migration (even though the db schema itself doesn't change), but a migration is already set up in the unmerged #1869 and I don't want to deal with conflicting migrations.~~

## To file as separate issues

- [ ] Consider linking to the data quality report from the admin or something. (Right now we link to it from the sidebar in `/docs/` which is easy to miss.)
- [ ] Add a metric that reports the number of labor rates with invalid choices (e.g. an unexpected schedule, working site, etc.), to help with #1874.


contracts/models.py
```diff
@@ -279,7 +279,9 @@ class Contract(models.Model):
     #   IDV - "Indefinite Delivery Vehicle"
     #   PIID - "Procurement Instrument Identification"
 
-    idv_piid = models.CharField(max_length=128, db_index=True)  # index this field
+    idv_piid = models.CharField(
+        max_length=128, db_index=True,
```
hbillings: Should `db_index` be on a separate line?

contracts/reports/base.py
```diff
@@ -0,0 +1,82 @@
+from typing import Union
+from textwrap import dedent, fill
+import abc
```
hbillings: import what now

contracts/reports/dupes.py
```diff
@@ -0,0 +1,63 @@
...
+    CORE_FIELDS = [
+        'vendor_name',
+        'labor_category',
+        'idv_piid',
+        'current_price',
+    ]
+
+    CORE_FIELD_NAMES = [
+        Contract._meta.get_field(field).verbose_name
+        for field in CORE_FIELDS
```
hbillings: I've never seen this syntax before -- I'm assuming this `for` executes the preceding line for each `field` item, resulting in a list of `CORE_FIELD_NAMES`?

contracts/reports/incomplete.py
```diff
@@ -0,0 +1,15 @@
+from contracts.models import Contract
+from .base import BaseMetric
+
+
+class Metric(BaseMetric):
+    '''
+    Find labor rates that are missing key data.
```
hbillings: Is the intention to expand this, or for it to prioritize education? If the latter, maybe we should specify that here.

contracts/reports/outliers.py
```diff
@@ -0,0 +1,39 @@
...
+    detection, see:
+
+        https://github.com/18F/calc-analysis/blob/master/contract-analysis/anomaly-detection.ipynb
+    '''
+
+    EDU_LEVEL = 'High School'
+
+    EXPERIENCE_CAP = 5
+
+    MIN_PRICE = 500
```
hbillings: Are these defaults just for demonstration purposes, or are they going to be expanded upon in the future? What are they based on?

contracts/reports/dupes.py
hbillings: oh derp, I just realized that this is confusing me simply because it's on two lines instead of one. Ignore me. 😬

contracts/reports/base.py
toolness: Heh, it's for [abstract base classes](https://docs.python.org/3/library/abc.html).  It's also what our base "schedule importer" class uses... basically it allows us to provide a superclass that says "yo instances of this class should always implement `foo` and `bar`" and then if we ever make a subclass that _doesn't_ implement one of those methods, a helpful exception will be raised.

I think `abc` might do more but at least that's the main reason I like to use it.  It works nicely with python type annotations (e.g. mypy).

contracts/models.py
toolness: Hmm, I didn't intend for it to be. Sometimes with functions/constructors that take lots of arguments, I like specifying the args normally, but on their own line so that they have lots of runway. It also seemed in keeping w/ other fields on the model like `education_level`:

```python
    education_level = models.CharField(
        db_index=True, choices=EDUCATION_CHOICES, max_length=5, null=True,
        blank=True)
```

contracts/reports/dupes.py
toolness: Ah yeah, I like splitting up longer list comprehensions into separate lines, where each clause is on its own line, sort of like SQL statements.

contracts/reports/incomplete.py
toolness: The intent was that it might expand... basically, there are fields that I _thought_ were universal, but that I was surprised to find out weren't.  Right now, education is the only one.

Well, actually, experience level might be too, sort of: I've noticed that there are a few labor categories that are clearly very "senior" but the minimum experience is set to 0, and I think it's because our experience field doesn't actually allow null values.  So I was originally going to add minimum experience to the query until I discovered that.

contracts/reports/outliers.py
toolness: Well, it's too resource-intensive to do the full anomaly detection algorithm I wrote on the calc-analysis repo here, so I basically took our $700 teleprompter operator and made it a bit more generic--it seems a bit extreme for the government to pay someone who's less than 5 years out of high school over $500 per hour.  (The teleprompter operator fits into that category.)

Obviously if this metric returns 0 that doesn't mean the dataset has no anomalies, but it's a decent indicator of whether there _are_ anomalies, I think.

hbillings: This would also be nice to touch on (super briefly; just by way of showing folks what we're measuring) in a potential explainer sentence.

contracts/reports/incomplete.py
toolness: Can we do this either in the description of the metric or its footnote? Because of the modular way I constructed the footnotes, I think it'd be most maintainable to improve either of these, rather than changing something else that might get out-of-sync with the actual metric. For instance, we could change the outliers metric to say something like:

```
{count} labor rates are exorbitantly expensive to the point that they may represent bad data [1].
```

with a footnote that says:

```
[1] This is measured by searching for labor rates that require high school, less than 5 years of experience, yet pay over $500 per hour.  Note that this is just a heuristic; even if it is zero, it doesn't mean that CALC doesn't contain anomalous data.
```

Issue #1848, opened 2018-05-02
Update style of current styleguide


hbillings:
In our most recent chat about the styleguide, we decided to keep the current form and change up the style a bit to make it easier to peruse.

- [ ] Navigation is sticky on the side of the page.
- [ ] Each section can be opened in a new window.
- [ ] More visual separation between sections.
- [ ] Ability to collapse the code when not necessary.

PR #1882, opened 2018-05-09
[WIP] Styleguide add header

### LINKED ISSUE #1880: Add header component to style guide

ericronne:
image

@hbillings at some point, i need help rigging up the header to load on the page

image

closes #1880


styleguide/templates/styleguide.html
```diff
@@ -63,6 +63,19 @@ <h6>Heading 6</h6>
...
   {% endexample %}
 
+  {% guide_section "Sub-page site header" %}
+
+  <p>
+    The header / navigation component on CALC sub-pages is a simplified, compact variation of the header on the home page. Its sub-components include the CALC logo (which functions as a home button), the user login utility (for those not signed in), the user-tools menu (visible once a user signs in), and global navigation (which changes depending on a registered user's level of authority).
+  </p>
+
+  {% example %}
+    {% block header %}
```
hbillings: hey hey! you don't need the `{% block header %}` tags here, I don't think. Those are Django template tags meant to include the header on the actual page. You can just use the `include` tag!

ericronne: neeto, thx! 🤞

ericronne: nixed em, but only a part of the header seems to load (missing the bg, eg) …
image

hbillings: Right so... the header isn't meant to be a standalone thing and is hardcoded into our base template. :-/

There are, altogether, three-four tags that make up the header. From `data_capture/templates/base.html`:
```
{% include '_banner.html' %}
<header>
    <div class="container">
        <div class="row">
        {% include '_header.html' %}
        {# Extended header should be placed in this block in descendant templates #}
        {% block header_extension %}{% endblock %}
        {% block header_nav %}
          {% include '_nav.html' %}
        {% endblock %}
        </div>
    </div>
</header>
```

You'll need to add the `header` tags to the include to duplicate how they appear in the base.

That said, I'm not sure it's the best idea to show this code here -- we don't want people copying and pasting the rendered HTML; we want them to use the Django template tags. @toolness, is it possible to render just the HTML in an example without showing the code?

toolness: Yup, see e.g. the template source for the steps widget in the styleguide: https://calc-dev.app.cloud.gov/styleguide/#steps-widget

That said, I don't think we currently support the iframe-style example without showing the HTML snippet, but i can add support for it.

hbillings:
Covered in #1848.

PR #1892, opened 2018-05-10
Allow fullpage examples to use template tags, optionally hide their HTML

toolness:
This enhances the full-page examples in the style guide:

* They can now use template tags. However, in order to accomplish this, I had to make the files in `styleguide/templates/styleguide_fullpage_examples` regular templates, rather than purely HTML files (which actually makes more sense, since they are in the `templates` directory). A full-page example that once contained this HTML:

  ```html
  <p>boop</p>
  ```

  now looks like this:

  ```html
  {% extends "styleguide_fullpage_example.html" %}

  {% block body %}
  {# BEGIN SNIPPET #}
  <p>boop</p>
  {# END SNIPPET #}
  {% endblock %}
  ```

  This certainly looks less readable, but it's also more powerful. The `{# BEGIN SNIPPET #}` and `{# END SNIPPET #}` tell the styleguide where the beginning of the HTML source it should show to readers is (if they're not found, the whole file is used).

  I considered having it just find `{% block body %}` and `{% endblock %}` instead, but I thought this (A) made things seem a bit too magical and (B) made the system less flexible. But we can change it to that, or something else, if this looks too gross.

* The HTML source code of a snippet can optionally be hidden from readers entirely. This can be done by adding `show_html=False` to the `{% fullpage_example %}` template tag, e.g.:

  ```
  {% fullpage_example "card" show_html=False %}
  ```

### DIFF
--- a/styleguide/fullpage_example.py
+++ b/styleguide/fullpage_example.py
@@ -1,4 +1,6 @@
+import re
 from pathlib import Path
+from typing import Optional
 from django.utils.safestring import SafeString
 from django.shortcuts import render
 from django.urls import reverse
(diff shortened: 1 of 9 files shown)


styleguide/fullpage_example.py
```diff
@@ -7,31 +9,52 @@
...
-EXAMPLES_DIR = MY_DIR / 'templates' / 'styleguide_fullpage_examples'
+EXAMPLES_DIRNAME = 'styleguide_fullpage_examples'
 
+EXAMPLES_DIR = MY_DIR / 'templates' / EXAMPLES_DIRNAME
+
+# We need to be careful to make this regular expression for
+# example names very restrictive, because we use the name
+# as part of file paths; so we don't e.g. want to allow
+# periods or dashes.
 REGEX = r'([0-9A-Za-z_\-]+)'
```
hbillings: Did the refactoring remove the need for this? Or am I just not seeing where it's used?

toolness: Oh good question--it's actually used over in [`urls.py`](https://github.com/18F/calc/blob/develop/styleguide/urls.py#L11):

```py
    url(r'^fullpage-example/' + fullpage_example.REGEX + '$',
        fullpage_example.view, name='fullpage_example'),
```

But it seemed to "belong" more in `fullpage_example.py` rather than in `urls.py`, so I defined it there.

toolness:
Woot thanks!!

PR #1893, opened 2018-05-11
Document API 'sort' field better

### LINKED ISSUE #1889: API "sort" parameter can be multiple comma-separated fields
We should document it properly.

See https://github.com/18F/calc/pull/1869#discussion_r187399571 for more details.

toolness:
Fixes #1889.

This actually subtly changes behavior, too: previously we were allowing any indexed `Contract` field to be sorted on, which included weird fields like `_normalized_labor_category`, but now we're specifically only allowing fields that the API exposes to be sorted on.

This also adds some markdown-y utilities to `calc.util` (couldn't think of a better place to put them) that I think will be useful in the future.


toolness:
Booo wifi! Ok thanks!

PR #1901, opened 2018-05-14
Improve setup documentation, .env.sample

toolness:
This improves the setup docs a bit:

* I removed `DATABASE_URL` from `.env.sample` because this environment variable never needs to be set. In fact, because it's set through the Docker environment (in `docker-services.yml`), any setting in the `.env` file won't actually take. I'm pretty sure this line was just left over from ancient pre-Docker CALC where the setting had to be specified.

* Add Windows-specific instructions where needed.

* I added a commented-out `#USE_POLLING=yup` with explanatory comment above it in `.env.sample` because it's much more discoverable that way (otherwise it's buried in our environment variable docs).  This will be helpful to folks on Windows who may otherwise get frustrated because their changes aren't triggering webpack rebuilds.

* Instead of waiting for `Finished 'default' after 3.7 μs` we can now wait for a _much_ more visible message to be logged:

    ```sh
    gulp_1          | [19:15:54] -----------------------------------------
    gulp_1          | [19:15:54] Visit your CALC at: http://localhost:8000
    gulp_1          | [19:15:54] -----------------------------------------
    ```

* A few other minor copy edits.


toolness:
Um, I just merged #1911 into this, in an attempt to have to wait for less stuff to pass before merging everything into `develop`.  We'll see if this actually works or fails horribly soon.

toolness:
Lol I missed the "comment" button and hit "close and comment" instead.

PR #1902, opened 2018-05-14
Split up environment.md into 'dev' and 'general' sections.

toolness:
This is just a proposal, I'm not sure if it's the best idea yet.

Basically, I've noticed that there's two different reasons one could visit the environment docs:

1. The user is developing locally and wants to configure their instance so it works slightly differently for the purposes of development.

2. The user is deploying to the cloud and wants to configure things so that visitors can use the instance.

So this just divides the docs into "Development settings" and "General settings" sections for each use case.  I dunno, it might be dumb, in which case we can close it unmerged.

What do you think @hbillings?  Would this be useful to you?


hbillings:
I feel like it is helpful to have local dev settings separated out from deployment settings (not having to read through EVERYTHING to find the things relevant to me is 👌), but the "general" moniker is, uh, a little general. 😂 Development settings vs. production settings, maybe? Though I guess they apply to staging too...

toolness:
I was thinking of "production" but I was worried that there were things that weren't really "production" that it could be useful for.  For example, you might be developing locally but still want to test Slack or E-mail integration, in which case you want to look at that second section...

(BTW we are getting rid of staging in #1795 which is 💯!)

PR #1903, opened 2018-05-14
Minor updates to frontend docs.

toolness:
Just a handful of minor updates.


docs/frontend.md
```diff
@@ -9,9 +9,9 @@ details on how to use or style individual components, see the
 ## The static asset generator
 
 If you haven't already, make sure you've followed the
-[setup guide](setup.md); it explains how to get the gulp-based
-static asset generator up and running. It will watch for changes to
-front end code and re-build bundles as needed.
```
toolness: The old setup docs, before #1743, specified how to start gulp in its own terminal, but now that we've standardized on Docker, there's no need for that.  This changes the copy a bit to make that more evident.

Issue #379, opened 2016-07-13
Add a management command to test and lint everything


toolness:
As mentioned in https://github.com/18F/calc/pull/377#issuecomment-232203359, we're now using CodeClimate and QuantifiedCode to ensure that our code is properly linted and stuff. It would be nice if we had some kind of command, e.g. `manage.py ultratest`, that:

1. Linted all Python via `flake8`
2. Linted all JS via `eslint`
3. Ran all tests

Basically just a pre-check command that someone could run to make sure that they could submit a PR without having to worry about one of the automated checks outing them.

In the future it could also do additional stuff like maybe running our CSS through a [css linter](http://csslint.net/) too.

PR #1911, opened 2018-05-15
Add continuous integration docs, simplify code climate config

### LINKED ISSUE #1839: Make it possible to disable mypy, typescript, selenium
e.g.:

* Make it possible to disable mypy, typescript, selenium, and other fancy tools that improve the quality of our software but also raise the barrier to modifying/maintaining CALC

toolness:
This adds documentation for our continuous integration tools, CircleCI and Code Climate.

(Ok, technically Code Climate might not be a CI tool, I'm not sure, but for our purposes I'm treating any of the third-party services listed under the "checks" section of our GitHub PRs to be CI tools.)

This also fixes #1839 by documenting how to disable some of CALC's funky tests.

It also simplifies our Code Climate configuration by:

* Removing `pep8`. This was originally added in #322, before `ultratest` existed. But once `ultratest` was added in #379, running the linter on CC became redundant.

  Not only that, it actually uses a subtly different command-line tool than we do--and one that is more out-of-date than ours--which makes it hard to anticipate or reproduce its errors locally.

  As if that wasn't enough, CC's pep8 config is also stored in a different file than the one we use, which means we need to manually ensure that both configurations stay in-sync.

* Removing `eslint`. This was added way back in https://github.com/vzvenyach/calc-fork/pull/1 (by Dave Z, who I didn't even know worked on CALC). Once `ultratest` was added this became redundant too.

  It potentially uses a different version of eslint than we use, which bit us in the past with #403, so we might as well just disable it to ensure this kind of thing doesn't happen again in the future.

### DIFF
--- a/docs/ci.md
+++ b/docs/ci.md
@@ -1,12 +1,11 @@
 ## Continuous integration
 
-CALC uses a number of third-party services to ensure that its code quality
-is not so bad.
+CALC uses a number of third-party services to ensure consistent code quality.
 
 ### CircleCI
 
-[CircleCI][] is used to ensure that all tests still pass as CALC's codebase
-changes. Its configuration is stored in [.circleci/config.yml][].
+[CircleCI][] ensures that, when CALC's codebase changes, all tests still pass.
+Its configuration is stored in [.circleci/config.yml][].
 
 Aside from running tests, CircleCI is also used to automatically deploy
 CALC; see the [Release process](release.md) documentation for further
@@ -16,7 +15,7 @@ Finally, CircleCI is also used for some [monitoring](monitoring.md) tasks.
 
 #### Disabling tests
 
-Because test passing on CircleCI blocks deployment, it's important to
+Because a single failing test on CircleCI blocks deployment, it's important to
 know how to disable some of them in emergencies. In particular:
 
 * Selenium has a history of acting up on CI and sometimes needs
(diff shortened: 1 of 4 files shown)


docs/ci.md
```diff
@@ -0,0 +1,46 @@
+## Continuous integration
+
+CALC uses a number of third-party services to ensure that its code quality
+is not so bad.
```
hbillings: lol! Maybe, "to ensure consistent code quality"?

docs/ci.md
```diff
@@ -0,0 +1,46 @@
...
+
+### CircleCI
+
+[CircleCI][] is used to ensure that all tests still pass as CALC's codebase
+changes. Its configuration is stored in [.circleci/config.yml][].
```
hbillings: Maybe, "CircleCI ensures that, when CALC's codebase changes, all tests still pass."?

docs/ci.md
```diff
@@ -0,0 +1,46 @@
...
+
+Aside from running tests, CircleCI is also used to automatically deploy
+CALC; see the [Release process](release.md) documentation for further
+details.
+
+Finally, CircleCI is also used for some [monitoring](monitoring.md) tasks.
+
+#### Disabling tests
+
+Because test passing on CircleCI blocks deployment, it's important to
```
hbillings: tests instead of tests?

toolness: Oh right... hmm... I think this actually sounds better as "Because a single failing test on CircleCI blocks deployment", so I changed it to that.  If that sounds confusing we can fix it up in another PR!

toolness:
Uhh I am gonna try something weird to make this merge faster, sorry if anything explodes!

PR #1913, opened 2018-05-16
Add more docs about new relic

toolness:
New Relic is actually a pretty multi-faceted service that offers both user analytics *and* application performance monitoring services.  It can be hard to wrap one's head around, so I added more docs to our analytics and monitoring sections about it.


docs/analytics.md
```diff
@@ -20,12 +22,18 @@ snippets included in HTML templates are in
 Please see these files for further documentation on how to track custom
 events, virtual pageviews, and more.
 
-### Server
+#### New Relic
+
+If the `NEW_RELIC_LICENSE_KEY` environment variable has been set
+and the associated New Relic account has [New Relic Browser][]
+enabled, the New Relic Python agent will automatically
+inject JavaScript analytics script tags into CALC pages.
```
toolness: This behavior is super yucky IMO, I literally had to visit calc.gsa.gov and inspect the page source to remember that this is how NR Browser does its thing.  I wish it worked more like Google Analytics, which requires us to add an explicit script tag to our templates.

hbillings: Oof, yeah, that's gross. Even better to have it documented now!

docs/monitoring.md
```diff
@@ -38,11 +38,24 @@ returned object has the following keys:
...
 ### Production tests
 
 CALC has a suite of tests in [production_tests](../production_tests/) for
 checking that the production site is properly configured.
 
 These tests are run nightly as a workflow build in CircleCI.
 
+### New Relic
+
+The server-side supports New Relic integration via the
```
hbillings: Copy edit nitpick: no hyphen. 

(When you put a hyphen in a phrase like `server side`, it becomes an adjective that needs to preface a noun; e.g., `server-side process`. Same thing happens when combining things like `log in` into one word!)

toolness: Oo that's a great way of putting it, I had not thought of it that way before. Will change to have no hyphen!

PR #1917, opened 2018-05-17
Apply monospace fonts evenly

hbillings:
We lost the `.currency` class somewhere along the way, and we also lost the monospace on most of the tables.


hbillings:
@ericronne how's this looking?

@toolness Could you make sure the way I added the class to `sortable-data-cell.jsx` is kosher?

hbillings:
Yep, currently that's on #1946, because I was doing too many things at once and committed it to the wrong branch 🤦‍♀️

frontend/source/js/data-explorer/components/results-table/sortable-data-cell.jsx
```diff
@@ -6,6 +6,7 @@ const getDataCellClasses = (key, sort) => ({
   cell: true,
   sorted: sort.key === key,
   [`column-${key}`]: true,
+  number: key === 'current_price',
```
toolness: Yup, that's totes kosher!

PR #1918, opened 2018-05-17
Adjust focus ring to apply different colors on different backgrounds

### LINKED ISSUE #1740: Focus state has insufficient contrast over light colors
Update the focus color to the following specs:
- `$color-gray` for links which run over **white**
- `$color-gray-darker` for links which run over light colors (e.g. green in footer)
- `$color-gray-light` for links which run over dark colors (e.g. header)

**Current footer**
image

**Current site admin page**
image

hbillings:
Closes #1740.

The current WDS approach applies the same focus color over all backgrounds. This tries to be smarter by creating a `focus-color` mixin that can be extended for the three different scenarios mentioned in #1740.


ericronne:
Cooooool!

I wonder if we should narrow it to only two colors? …
`$color-gray-darker` for both white and light backgrounds
`color-gray-light` for dark backgrounds

`$color-gray` looks less bad over white, but i think devs might appreciate the simplicity of a binary choice. Additionally, zebra-striped tables pose a problem if we use the three-color construct. `$color-gray` doesn't contrast sufficiently with the gray-filled table rows, so we would have to alternate the focus colors for each row. Tedius, and more prone to error.

y'think?

ericronne:
Thanks i'll take a look

PR #1936, opened 2018-05-18
Improves color mappings and simplifies naming scheme

ericronne:
@hbillings we may want to pair on this review, lots going on. Note that i moved the blues out of `_uswds_variables.scss` and into our variables file, because those color names don't seem to exist in The System. (There is a "cool blue" set, but i don't see a "blue" set.) I may very well be off kilter here, jlmk!


ericronne:
Eye. Roll.

The **only** files which i want included in this pr are [`_uswds_variables.scss`](https://github.com/18F/calc/pull/1936/files#diff-ae87493e4853b83f555b83af5b704189) and [`_variables.scss`](https://github.com/18F/calc/pull/1936/files#diff-a3c99f132240ef3019696364cc346d25).

Github and i need a weekend break from each other.

frontend/source/sass/base/_uswds_variables.scss
```diff
@@ -8,24 +8,23 @@ $asset-path: '../vendor/uswds/' !default;
 
 // Color overrides
 // Addtional colors specific to CALC are defined in _variables.scss
-$color-primary:              #2D6580;
-$color-primary-darker:       #274E60;
-$color-primary-darkest:      #092D39;
+$color-primary:              #0770b5; // Links
+$color-primary-darker:       #136b94; // TODO: apply this as primary button color
```
hbillings: plz to file an issue and remove the comment thx

frontend/source/sass/base/_uswds_variables.scss
```diff
@@ -8,24 +8,23 @@ $asset-path: '../vendor/uswds/' !default;
...
+$color-green-light:          #cddc86;
 $color-green-lighter:        #dae6a3;
+$color-green-lightest:       #e6edc2;
 
-$color-primary-alt:          #ccde61;
-$color-secondary:            $color-primary-alt;
-// Solve later: disconnect between ^this^ and other "secondary" colors
+$color-secondary:            #ccde61;
+$color-primary-alt:          $color-secondary;
+// TODO: solve disconnect between ^this^ and other "secondary" colors
```
hbillings: Same here -- file an issue explaining what the disconnect is.

frontend/source/sass/base/_uswds_variables.scss
```diff
@@ -40,7 +39,7 @@ $color-gray-lightest:        #f0f6fa;
 $color-gold:                 #fdb81e; // No change from USWDS, including for visual reference
 
 $color-gold-lightest:        #fbeebe;
-$color-visited:              #4c2c92; // No change from USWDS, including for visual reference
+$color-visited:              #10428f; // TODO: apply this as visited color
```
hbillings: Hmm... currently we're avoiding styling visited links differently on purpose. I think we should make sure this doesn't accidentally style links strangely. 

Also, please file this as an issue.

frontend/source/sass/base/_variables.scss
```diff
@@ -9,11 +9,13 @@
...
 $color-green-darkest:        #61701c;
-$color-green-medium:         #a0ad64;
+$color-green-dark:           #a0ad64;
 
 $color-green-bright:         $color-secondary;
 
-$color-blue-dark:            #0b6199;
+$color-blue-darkest:         $color-primary-darkest;
+$color-blue-dark:            $color-primary-darker;
+$color-blue:                 $color-primary; // Link color
```
hbillings: `$color-blue` and `$color-blue-darkest` are both [USWDS overrides](https://github.com/uswds/uswds/blob/develop/src/stylesheets/core/_variables.scss#L38) and should be in the `_uswds_variables.scss` file. We could consider renaming `$color-blue-dark` to `$color-blue-darker` if we wanted to make them all WDS overrides.

frontend/source/sass/base/_uswds_variables.scss
ericronne: Issued (#1940) but i'm not sure of protocol: leave comment until issue is solved, to inform anyone who comes across it that it's a known issue? If we're removing the comment once an issue is filed, there are a lot of todo's that need to be turned into issues.

frontend/source/sass/base/_uswds_variables.scss
ericronne: I think this was your comment; i just added "todo"

frontend/source/sass/base/_uswds_variables.scss
ericronne: Why did we want them to be the same?

frontend/source/sass/base/_variables.scss
ericronne: K will fix; i was mistakenly using [this page](https://designsystem.digital.gov/components/colors/) (which excludes the blues) as my guide

frontend/source/sass/base/_uswds_variables.scss
hbillings: In general, it's better to track todos as issues rather than comments. Sometimes small stuff like this is OK, but sometimes comments don't get removed after a thing is actually fixed, making it more confusing for folks who do development after.

frontend/source/sass/base/_uswds_variables.scss
hbillings: My bad! Let's still make it an issue. :)

frontend/source/sass/base/_uswds_variables.scss
hbillings: I *think* originally there were two reasons:
- (CSS History Leak attack)[https://www.brokenbrowser.com/css-history-leak/]
- having a generally applied `:visited` color meant we had to override it _everywhere_. Buttons had visited colors, nav had visited colors, different colored links had visited colors...

Are visited colors generally helpful? On a site like this, where you're doing the same sequence of things over and over, I'm not totally sure they are. But I also don't have strong opinions, just wanted to point out that this wasn't an accident originally!

frontend/source/sass/base/_variables.scss
hbillings: ugh, yeah, that's bitten me before. :-(

frontend/source/sass/base/_uswds_variables.scss
ericronne: Derp i don't recall what the disconnect is / other secondary colors are, sry

ericronne:
OBE (#1944)

PR #1937, opened 2018-05-18
Indent nested lists

ericronne:


frontend/source/sass/base/_typography.scss
```diff
@@ -123,6 +123,11 @@ ol.steps li {
   font-family: $font-family-main;
 }
 
+// Indents nested lists
+ul ul {
+  margin-left: 4.8rem;
+}
```
toolness: Hmm, so this is good for nested `ul`'s, but it won't be any good for e.g. a nested `<ol>` in a `<ul>`, or vice versa!

ericronne:
Your speak truth, not funk.

Github is what's funky: it will not allow me to enter any characters in the reply box above.

_stay tuned …_

frontend/source/sass/base/_typography.scss
```diff
@@ -123,6 +123,19 @@ ol.steps li {
...
+ul ol,
+ol ul {
+  margin-left: 4.8rem;
+}
+
+ul ul ul,
+ul ol ul,
+ol ul ul {
+  margin-left: 9.6rem;
+}
```
hbillings: This rule is unnecessary: `ul ul` will apply an extra `4.8rem` of margin to any `ul` nested within another `ul`, whether they're 2 deep or 30.

Also, why not use a spacing variable?

ericronne: because i'm a visual designer,  not a frontender.
go for it!

hbillings:
@toolness, how's this look?

Issue #1203, opened 2017-01-05
SIN numbers shouldn't break on hyphens


hbillings:
This looks funky and is hard to read
<img width="247" alt="screen shot 2017-01-04 at 5 45 49 pm" src="https://cloud.githubusercontent.com/assets/509309/21664730/aa139ca4-d2a5-11e6-9bfc-72b325825536.png">

jseppi:
heh, interesting. Looks like two ways to accomplish [not-breaking-on-hyphens](https://css-tricks.com/forums/topic/prevent-word-breaks/):

1. use a "non-breaking-hyphen" character: `&​#8209;`, which isn't doable for us without doing some kind of string replacement (like in JS or in responses from the API)
1. use the css rule `white-space: nowrap`, but if a SIN column has multiple numbers, I think it would make the whole cell be a single line instead of wrapping anywhere :\

toolness:
Oh, this is happening on data capture stuff only, right?  Since we don't currently show SIN numbers in the data explorer, I think?

hbillings:
yep! Not the explorer, but the admin pages and the data capture pages. Tagged accordingly. :)

toolness:
In #1938, I added an `unbroken_hyphenize` function that does the string replacement Seppi suggested above, if that's useful.

PR #1938, opened 2018-05-18
Add detail pages for data quality report metrics.

toolness:
This removes the weird footnotes from the data quality report and instead adds "Learn more..." links:

> data_quality_report

Clicking on one of the links takes the user to a detail page that also potentially shows example rows for the data, e.g.:

> detail

The table in the detail view just shows the raw underlying data, it's not meant to make it look particularly attractive or human-readable. There's also a horizontal scroll bar at the bottom of the table that allows users to see the values of every field.

Additionally, this PR fixes a bug in the way duplicate rows are counted (which I filed separately as #1939).

## To do

- [x] Add tests.
- [x] ~~Possibly add more documentation to the `BaseMetric` class.~~ Meh.
- [x] ~~Consider pulling the whole doggone data quality report thing out into its own separate django app, maybe.~~ Meh.
- [x] Consider implementing `get_queryset()` for the `dupes` metric.


contracts/templates/data_quality_report_detail.html
```diff
@@ -0,0 +1,52 @@
...
+      {% endif %}
+
+      {% if examples %}
+      <h2>Examples</h2>
+
+      <p>
+        Below are some of the {{ metric.verbose_name }} currently in the database.
+      </p>
+
+      <div style="overflow-x: scroll; max-height: 90vh">
```
toolness: I added the `max-height: 90vh` here to ensure that a horizontal scroll-bar is always shown if needed (otherwise, the table may be so tall that users can't scroll it horizontally while looking at the top row).

I guess we could move this miniscule amount of CSS to a separate file... But INLINE CSS FOR THE WIN

contracts/templates/data_quality_report_detail.html
```diff
@@ -0,0 +1,52 @@
...
+              {% for field in EXAMPLE_FIELDS %}
+                <th>{{ field }}</th>
+              {% endfor %}
+            </tr>
+          </thead>
+          <tbody>
+            {% for row in examples %}
+              <tr>
+              {% for value in row %}
+                <td>{% if value == None %}<em>None</em>{% else %}{{ value }}{% endif %}</td>
```
toolness: If the value of the field is `None` we do want to style it slightly differently, to distinguish `None` (i.e. Python's `None` object) from `"None"` (i.e. a string containing the text "None").

contracts/views.py
```diff
@@ -1,9 +1,63 @@
...
+    'sin',
+    'contract_start',
+    'contract_end',
+    'contract_year',
+    'hourly_rate_year1',
+    'hourly_rate_year2',
+    'hourly_rate_year3',
+    'hourly_rate_year4',
+    'hourly_rate_year5',
+]
```
toolness: Originally I grabbed these values dynamically by introspecting the model, but I decided against it because it could lead to leakage of sensitive information, e.g. if some funky top-secret field is added to `Contract`.  Since the model isn't changed very often anyways, I figured it'd be OK to just whitelist them here--and anyways, this way we have more control over the order in which the fields are shown in the example table.

contracts/reports/dupes.py
```diff
@@ -28,29 +38,30 @@ class Metric(BaseMetric):
...
-    FROM {CONTRACTS_TABLE} as c1, {CONTRACTS_TABLE} as c2
-    WHERE c1.id < c2.id   /* Use < to avoid counting duplicates */
-    AND {CORE_FIELDS_ARE_EQUAL}
-    '''
+            total += row['count']
+            if total > self.MAX_EXAMPLES:
+                break
+
+        return Contract.objects.filter(q)\
+            .order_by(*self.CORE_FIELDS)[:self.MAX_EXAMPLES]
```
hbillings: I get a little lost in here about what exactly this is looking at, but probably not enough to hold it back.

contracts/templates/data_quality_report_detail.html
hbillings: Yeah, this is actually one of the rare instances I don't have a problem with inline CSS: these rules are so minimal and one-off that it doesn't make sense to have a reusable rule for them. 👍

contracts/views.py
hbillings: That seems reasonable, and also less confusing!

calc/utils.py
```diff
@@ -14,6 +14,17 @@ def get_permissions_from_ns_codenames(ns_codenames):
     ]
 
 
+def unbroken_hyphenize(text: str) -> str:
+    '''
+    Replace all hyphens with non-breaking hyphens.
```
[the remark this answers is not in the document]
toolness: Yeah! Seppi actually mentioned it in #1203, but I discovered them independently because I figured "hey, there are non-breaking spaces, so maybe there are non-breaking hyphens too" and googled and HOORAY

depending on where we need to use these for #1203 (front-end vs backend), we can potentially use this function elsewhere too.

contracts/reports/dupes.py
toolness: Oh that is because I named the variables horribly and should have documented it better!  I just did that, so hopefully it is easier to understand now... let me know if it isn't though.

Issue #1915, opened 2018-05-16
Nested lists aren't indented


toolness:
This is a screenshot from the new changelog entries added in #1914:
 
> <img width="535" alt="screen shot 2018-05-16 at 11 52 26 am" src="https://user-images.githubusercontent.com/124687/40128466-ae0b200c-58ff-11e8-9a35-ac775e4e6942.png">

The second and third bullets are in a nested `<ul>`, and while their markers are different, their indentation is not, which makes them hard to read separately from the list they're nested in.

PR #1946, opened 2018-05-22
Merge indented list change

hbillings:
I merged #1915 without realizing it was against `update-changelog` and not `develop`. Whoops.


frontend/source/sass/base/_variables.scss
```diff
@@ -28,7 +28,7 @@ $medium-large-screen: 900px;
 
 // Fonts
 $font-family-main: 'Libre Franklin', 'Helvetica', 'Arial', sans-serif;
-$font-family-mono: Consolas, 'Liberation Mono', Menlo, Courier, monospace;
+$font-family-mono: 'Roboto Mono', 'Liberation Mono', 'Source Code Pro', Courier, monospace;
```
toolness: This was an intentional change, right?  Just making sure because the title of this PR is "Merge indented list change" but this doesn't have anything to do with indented lists... hmmm.

toolness: Nah it's cool, if you want to merge it into production we can just leave it in here!

PR #1947, opened 2018-05-22
Swaps in hue-based colors for "primary" and "secondary" variables

ericronne:


ericronne:
Dealing with remaining buttons in #1948

frontend/source/sass/base/_uswds_variables.scss
```diff
@@ -10,18 +10,19 @@ $asset-path: '../vendor/uswds/' !default;
 // Addtional colors specific to CALC are defined in _variables.scss
 $color-blue:                 #0770b5; // Link color
 $color-primary:              $color-blue;
-$color-blue-darker:          #136b94;
+$color-blue-darker:          #136b94; // Primary button color
 $color-primary-darker:       $color-blue-darker;
-$color-blue-darkest:         #0c4c6f;
+$color-blue-darkest:         #0c4c6f; // Primary button hover color
```
hbillings: Might be helpful to have these notes in the styleguide instead of in here?

Also, why intermix `blue` and `primary` colors?

frontend/source/sass/base/_uswds_variables.scss
```diff
@@ -10,18 +10,19 @@ $asset-path: '../vendor/uswds/' !default;
...
 $color-green-light:          #cddc86;
 $color-green-lighter:        #dae6a3;
 $color-green-lightest:       #e6edc2;
+$color-green-bright:         #ccde61;
 
-$color-secondary:            #ccde61;
-$color-primary-alt:          $color-secondary;
+$color-secondary:            $color-green-bright;
+$color-primary-alt:          $color-green-bright;
 // TODO: solve disconnect between ^this^ and other "secondary" colors
```
hbillings: If we can't remember what the disconnect is, let's just nix this comment.

frontend/source/sass/components/_upload.scss
```diff
@@ -19,14 +19,15 @@ upload-widget {
...
+    color: $color-blue;
     font-weight: $font-weight-bold;
     display: inline;
+    text-decoration: underline;
     cursor: pointer;
   }
 
   label:hover {
-    text-decoration: underline;
+    text-decoration: none;
```
hbillings: good call

frontend/source/sass/base/_uswds_variables.scss
ericronne: Sure, will add notes to style guide #1951 

Re intermixing: do we need primary and secondary color variables at all any more?

frontend/source/sass/base/_uswds_variables.scss
ericronne: ☑️

frontend/source/sass/base/_uswds_variables.scss
hbillings: Yup, the WDS still expects them, so if we don't define them, it will use its own default colors for them.

ericronne: Hm if our design doesn't reference the primary/secondary/etc colors, why do we need to have them in the variables? We don't include others, such as `$color-secondary-dark` etc.

hbillings:
Our design doesn’t, but we’re including the WDS and it still
references the primary/secondary options.

If either we aren’t overriding the color variable of every WDS
component (which we aren’t) or someone pulls in a new WDS element and
it references a primary/secondary variable, we’re gonna be stuck with
whatever WDS uses as its primary/secondary colors because we’d no
longer be overriding them anywhere.

If we wanted to be REALLY safe, we’d assign new values for all the
existing WDS values, but right now we’re only overriding the most
commonly used ones.

ericronne:
Makes sense. Regrouped! Ready to go?

hbillings:
yasssss

Issue #1905, opened 2018-05-15
Gray button in styleguide is not gray


toolness:
From the [buttons section of our styleguide](https://calc-dev.app.cloud.gov/styleguide/#buttons):

> ![gray-button](https://user-images.githubusercontent.com/124687/40056689-1c635222-581a-11e8-9938-f5e024138247.png)

It looks like we're using `usa-button-gray` for that but it's not doing anything?  Also the secondary button doesn't seem to have a great contrast ratio but it might be a dupe of #1791.

toolness:
Also, unrelatedly, the first 5 buttons in that example are a bit confusing, because it's not clear that they're _supposed_ to look the same (I thought it was a bug at first).  We might want to add some text above them that says "The following buttons use different HTML but are styled identically" or something.

ericronne:
Why do we have links that are buttons?
![image](https://user-images.githubusercontent.com/9259185/40136765-40efb33e-590e-11e8-9fdc-ac7e46363aed.png)

Is this text-link "button …
![image](https://user-images.githubusercontent.com/9259185/40136778-47c35b0c-590e-11e8-986d-fe4923d4bce4.png)

… for elements such as this accordion?
![image](https://user-images.githubusercontent.com/9259185/40136839-800f21bc-590e-11e8-9001-208f91f44605.png)

PR #1953, opened 2018-05-23
Fix button active states

### LINKED ISSUE #1952: Button active states are wonky
need a hand.

eolfad3r56

hbillings:
Against #1905, not develop

Closes #1952 for link buttons.

I also noticed that active states for secondary buttons is off. @ericronne, should I fix this while I'm at it?
screen shot 2018-05-23 at 4 34 48 pm


ericronne:
Tackle during coworking?

PR #1956, opened 2018-05-24
Step indicator er visual tweaks

ericronne:
Makes a few visual adjustments to the step indicator. Items best handled by the likes of @hbillings include:
- [ ] Current should remain green on hover (line 14 begins my attempt here; only works if i add !important; guessing it's a specificity thing; seems it could be added to the list in line 6, but my try at that dint work :/ )
- Connecting rule is too high
- Checkmark is a little high in the circle

### DIFF
--- a/frontend/source/sass/components/_steps.scss
+++ b/frontend/source/sass/components/_steps.scss
@@ -47,24 +47,24 @@ ol.step-bar__bubbles {
 
   li {
     list-style: none;
-    border-top: 2px solid $color-gray-lighter;
+    border-top: 3px solid $color-gray-lighter;
     padding-top: 14px;
     position: relative;
     float: left;
     width: 46px; // width of :before psuedoclass + $space-3x (22px + 24px)
 
     a {
-      width: 2.2rem;
-      height: 2.2rem;
+      width: 2.6rem;
+      height: 2.6rem;
       display: inline-block;
-      padding: 0;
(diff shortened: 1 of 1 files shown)


frontend/source/sass/components/_steps.scss
```diff
@@ -47,24 +52,24 @@ ol.step-bar__bubbles {
...
     a {
-      width: 2.2rem;
-      height: 2.2rem;
+      width: 2.6rem;
+      height: 2.6rem;
       display: inline-block;
-      padding: 0;
+      padding-top: .1rem;
       position: absolute;
       top: -12px;
```
hbillings: FYI, this is the value that needed to be increased to move the bubbles up. I bumped it to -15px.

frontend/source/sass/components/_steps.scss
```diff
@@ -11,6 +11,11 @@ ol.step-bar__bubbles li.current a {
   &::before {
     content: counter(li);
   }
+  &:hover {
+    background-color: $color-green-light;
+    border-color: $color-green-light;
+    color: $color-base;
```
hbillings: This was exactly the right thing to do. Unfortunately, on line 85, there was another rule that explicitly set `.current a:hover` to gray. The rules had the same specificity, so the CSS implemented the one that was defined later in the file (the cascade at work, biting people since the dot-com bust).

PR #1963, opened 2018-05-25
Give data explorer table more room to breathe

### LINKED ISSUE #1764: The results table is cropped by our new design
The explorer results list doesn't fit into the new content area:

Hopefully will be fixed by #1755

hbillings:
Closes #1764.

After playing around with a few things, I think the best way to do this is to remove the `.card` classes from the data explorer for now and simply write similar styles that behave the way we want them to. There's a little inline CSS on the index page now to mimic the card look, which I think is OK because we don't want to encourage use of this elsewhere and we want to rewrite a bunch of this anyhow.

### DIFF
--- a/data_explorer/templates/index.html
+++ b/data_explorer/templates/index.html
@@ -33,7 +33,7 @@
 {% endblock header_extension %}
 
 {% block body %}
-<div data-embed-jsx-app-here class="card">
+<div data-embed-jsx-app-here style="background-color: #fff; border-bottom-color: #c5d6de; margin-bottom: 8rem;">
   <div class="content">
     <noscript>
       Please enable JavaScript to use this page's functionality.
--- a/frontend/source/js/data-explorer/components/app.jsx
+++ b/frontend/source/js/data-explorer/components/app.jsx
@@ -91,7 +91,7 @@ class App extends React.Component {
       >
         <TitleTagSynchronizer />
         <section className="search">
-          <div className="container">
+          <div className="container clearfix">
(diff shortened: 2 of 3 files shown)


ericronne:
👀

ericronne:
Neat solve! Is it possible to restore the padding to the top, so your work applies only to the table, like so? …

image

hbillings:
Yup! To be clear: you want padding added to the top and right, and more padding added to the left?

ericronne:
Ideally …

artboard

ericronne:
I think sticking the filter to the side would be too drastic of a design move for now.

Really want to get as close as possible to the spacing we've set up for the overall site, eg …

image

Let me do some visual nudgery, hangon........

ericronne:
Ok, if we can approximate this, it should suffice for now. Not too far from your layout! :)
image

It breaks the new card padding style, but the alternative would be too jarring of a change for users, by at least one standard deviation.

Biggest difference here is to reposition the filter box so that it vertically aligns with the search tool. (Rumor has it, this might be a pain to accomplish.)

hbillings:
Okay, I've got this looking more or less like what you have here (the big difference being the difference in width between the search bar and the graphs, which aren't easily resolvable).

@toolness, in order to do this I had to really change up the HTML of the data explorer in `app.jsx`; I'd love your thoughts on whether this is wise!

hbillings:
After edits from @tram, this looks like this now:
screen shot 2018-05-29 at 11 30 38 am

Issue #1759, opened 2018-04-12
Explorer search box improvements


ericronne:
The data explorer search tool has a couple quick-fix issues
![image.png](https://images.zenhubusercontent.com/555b449c4f7a08fbb7f2de17/b0a1c2cf-b5e5-446e-ad4a-34316ae6c8f6)

- [x] Add "Search labor rates" as a label above the text field as an `h2`, to make for a high-profile visual anchor
- [ ] ~Change the prompt text to read: "Separate multiple terms with commas (Engineer, Analyst, etc.)"~
- [x] "Clear search" is misleading, because the button also resets filters, etc. Change to: "Reset"
- [ ] Fix the radio-button formatting so they don't run together

**It should look kinda like this:**
![image.png](https://images.zenhubusercontent.com/555b449c4f7a08fbb7f2de17/7c8d5350-45a9-4729-b7df-10fdfeea8353)

hbillings:
Why did we decide against the prompt language change?

tram:
:shrug: Just added it into this week's design studio. Any rejiggering of elements may (will likely?) change the descriptive text.

ericronne:
Wait, added what? We can make this match the test plan. I preserved partly in the interest of changing as little as possible as we move into a research phase which will likely involve dramatic changes, and partly to err on the side of instruction clarity. (Keeping two instructional lines, rather than killing one.)

Issue #1752, opened 2018-04-12
Heuristic fixes to the data explorer (incl visual updates)


ericronne:

ericronne:
#1756 and #1762 are outstanding, but can stand on their own documentation. **And** future teams may not use ZenHub), so closing this.

Onward!

PR #1966, opened 2018-05-29
Search field & filter tweaks

ericronne:
Widened search a little and vertically aligned the filters with the search box.

Would like to push the proposed price tool over to the left, so it right aligns with the search tool, but i couldn't quickly figure out how to do that...

image


frontend/source/sass/data_explorer.scss
```diff
@@ -37,7 +37,7 @@ section.search .container {
 }
 
 #labor_category {
-  width: 55%;
+  width: 63%;
```
hbillings: The issue with this is that it breaks the search bar when the screen is `980px` wide. Our first breakpoint isn't until `900px`. :-/
screen shot 2018-05-29 at 11 14 54 am

frontend/source/sass/data_explorer.scss
```diff
@@ -13,7 +13,7 @@ body {
 section.search,
 section.search + .graph-block,
 .help-text.columns.nine {
-  margin: auto $space-8x;
+  margin: auto 0 auto $space-8x;
```
hbillings: Incorporated into #1963.

frontend/source/sass/data_explorer.scss
ericronne: This line break already happens at 755 on the current site. With the width at 55%, the search box is about 342px, which is ~118 pixels narrower than we have on the live site (~460px). About a 25% loss. At 63% (~375px), the button bumps to the next line for a few more visitors, but we don't seem to get many visitors at those middle-of-the-road resolutions ([resolution stats](https://analytics.google.com/analytics/web/#/report/visitors-browser/a48605964w96540303p100698609/explorer-segmentExplorer.segmentId=analytics.screenResolution&explorer-table.plotKeys=%5B%5D/). Still seems usable to me.

One workaround which preserves a wider search field is to kill "search" from the clear button, or go ahead and relabel it to "Reset" now (because the button also removes search filters, etc.). Proposed already in #1759.

hbillings: On the current site, the filters aren't next to the search field, which gives us more space to play with. Our biggest breakpoints are from 900px to 1201px, so we really need to keep it on one line between them. Otherwise, we'll have to introduce more breakpoints to massage styling, and that becomes a rat's nest pretty quickly. (Above 1201px, the card doesn't change size at all, so it's sort of irrelevant at this point.)

I'm fine with relabeling or whatever as long as we can make the above hold true!

ericronne: Cool.

@tram what are your thoughts re going straight to prod with "clear search" relabeled as "Reset" or simply, "Clear"? Either will give us room to widen the search field. We already planned to do this as part of our hueristic fix set (#1752), because the button does more than simply clear the search field; it resets everything.

### Current button on this branch
image

### Option A: "Clear"
image

### Option B: "Reset"
image

ericronne: @tram button now reads "Reset", but it's re-change-able :)

ericronne:
@hbillings i think a quick pairing to iron out a few alignment / padding niggles might be helpful, jlmk

ericronne:
These changes have all been covered in other PRs, right @hbillings?

hbillings:
yup yup!

ericronne:
OBE!

PR #1979, opened 2018-05-30
Fix accessibility regressions

### LINKED ISSUE #1975: Keyboard accessibility regressions on data explorer
I was using <kbd>Tab</kbd> to navigate the Data Explorer on the latest `develop` and noticed a few regressions from our current `master` (i.e., what's at calc.gsa.gov):

1. The "education level" drop-down no longer has a focus indicator. On `master` there's a fuzzy blue outline, but on `develop` there's no indication.

2. The "experience" slider's knobs are focusable, but pressing a variety of keys (enter, space, arrow keys, etc) doesn't do anything. On `master` these aren't focusable, as the `<select>` widgets below the knobs control the same values (these widgets are also still focusable on `develop`).  If there's nothing interactive that can be done with these knobs, we should probably remove them from the tab order like they are on `master`.

3. The "contract year" radios no longer have any focus indication. On `master` there's a fuzzy blue outline around whatever the currently-selected radio button is.

hbillings:
Closes #1975.

- Fixes focus ring on tab around Education button.
- Adds back focus on contract year. Note that this is funky because the markup here is _super_ odd.
On click it looks like:
screen shot 2018-05-30 at 1 54 06 pm
But on tab it looks like:
screen shot 2018-05-30 at 1 53 52 pm
This is because it's actually two different elements getting focus at different points, and I don't know how to reconcile the styles. >_<

To do:
- [x] Fix focus ring color around radio buttons
- [x] Fix focus ring color on education level checkboxes

Not doing:
- Fixing the tab focus on the slider handles. Because this is a component we're rendering from a library, we don't have much control over it, and nothing I've tried (mostly around `tabindex`) manages to take it out of the tab flow.

### DIFF
--- a/frontend/source/sass/components/_inputs.scss
+++ b/frontend/source/sass/components/_inputs.scss
@@ -10,6 +10,11 @@ input[type="radio"] + label {
   margin-top: 0;
 }
 
+input[type="radio"]:focus + label::before,
+input[type="checkbox"]:focus + label::before {
+  @include focus-color($bg-color: "light-color");
+}
+
 label {
   font-weight: 700;
   margin-top: 0;
--- a/frontend/source/sass/data_explorer.scss
+++ b/frontend/source/sass/data_explorer.scss
@@ -223,6 +223,9 @@ dl.dropdown {
     line-height: 24px;
     overflow: hidden;
     padding: 6px 10px;
+    &:focus {
+      outline: #2d4d59 dotted 2px;
+    }
   }
   span.value {
     display:none;
@@ -585,7 +588,7 @@ label.radio {
 }
 
 label.checkbox :checked + span {
-  background-color: $color-blue-dark;
+  background-color: $color-blue-darker;
   box-shadow: 0 0 0 0 $color-white;
   position: relative;
   &:after {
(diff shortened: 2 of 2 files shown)


ericronne:
I also see that focus state on the three search-term radio buttons is too light in color, and you can't tab between the three options. It only tabs to the selected one.

image

It looks like the intent was for the term and its associated radio button to get the focus, like so:

image

This makes sense, since clicking on the term itself serves to select that option:

m5nxppwizn

hbillings:
Can you pull again, @ericronne? I fixed the color.

hbillings:
Oh, also, these are weird: The focus state you're seeing that selects everything is really due to us having a generic `:focus` class that applies to _anything_. We really only want it to apply to the input (or, in this case, the thing that appears to be the input).

ericronne:
Color around the dot is now correct, but is it tab-key accessible?

hbillings:
It is for me; is it not for you?

ericronne:
Could be my messed-up local environment, but here's what tabbing does for me …

c1nhdwuj7p

hbillings:
Oh, so this is an odd thing with radio buttons: tabbing selects the _group_ of them. Once you're in the group, you use the arrow keys to select the one you want. I don't know why that's how it works?

ericronne:
Education filter is also buggy (see #1982). Not a new problem. NBD if we can't fix this today.

PR #1981, opened 2018-05-31
[WIP] Add links to step indicator

hbillings:
Attempts to make the step indicator linkable to past steps by defining a `base_url` in the view that builds the step indicator on a given page.


data_capture/views/bulk_upload.py
```diff
@@ -17,6 +17,7 @@
 
 steps = Steps(
     template_format='data_capture/bulk_upload/region_10_step_{}.html',
+    url_base="",
```
hbillings: placeholder

data_capture/views/price_list_replace.py
```diff
@@ -21,7 +21,8 @@
 SESSION_KEY = 'data_capture:replace_price_list'
 
 steps = Steps(
-    template_format='data_capture/replace_price_list/replace_step_{}.html'
+    template_format='data_capture/replace_price_list/replace_step_{}.html',
+    url_base='',
```
hbillings: placeholder

data_capture/views/price_list_upload.py
```diff
@@ -27,8 +27,14 @@
 
 replayer = Replayer(SESSION_KEY)
 
+def get_base_url_from_step1(step1_url):
+    # first_step = build_url(step1_url)
```
hbillings: @toolness I'm befuddled here. When I include `build_url` at this point, Django tells me I have circular imports from my URL conf. But the same file uses `build_url` in its views. I also tried adding this whole function to `common.py` and get the same error. I'm guessing this has something to do with how Django is building things out, but I don't really understand why from looking at `build_url`.

hbillings:
Closed in favor of #1991. Was a hell of a decade.

data_capture/views/price_list_upload.py
toolness: Hmm, it's been a while since I wrote the `Steps` class, but I think this is because `Steps` *defines* what the `data_capture:step_1` route actually is.  So ultimately this is like saying "yo, `data_capture:step_1` has as its base URL `data_capture:step_1`."  The views can use `buld_url` because the `data_capture:step_X` views are already defined at that point, but they haven't yet been defined at the time that the `Steps` constructor is being called at line 37.

I'll see if I can figure out a way of doing this that works, because I think it'd be more robust than the JS-only way you're doing things in #1991.

hbillings: Yeah, I went around in circles and decided it wasn't worth the time. Given that this is an enhancement to the static UI that already gets built out, I really don't think using JS for this is a terrible thing as long as it fails gracefully (and functionally this seems to be similar to, like, enhancing the upload widget or date input, which are both JS).

PR #1984, opened 2018-05-31
Fixes color of selected filter options

ericronne:
Was `$color-blue-darker`, should've been `$color-blue-dark`

image


ericronne:
It was fixed, someone broke it i sweartogod

PR #1987, opened 2018-05-31
Update search indexes in ContractListSerializer

### LINKED ISSUE #1974: load_api_data doesn't update search index
I just noticed that the `load_api_data` management command doesn't seem to update search indexes, as I used it last week to load all API data from CALC but just noticed that the search functionality on my local CALC instance wasn't working. Running `python manage.py update_search_field` fixed it.

It probably makes the most sense to update the `ContractListSerializer` to update the search indexes, since that's what's calling `bulk_create` (which bypasses the creation of search indices).

toolness:
This fixes #1974, and also further simplifies our code, by efficiently updating search indexes during bulk operations.  This relieves us of having to remember to manually update the search indexes whenever we call the bulk operations.


contracts/models.py
```diff
@@ -84,11 +84,13 @@ def bulk_update_normalized_labor_categories(self):
         call Contract.save().
         '''
 
+        pks = []
         updates = []
         num_updates = 0
         for contract in self.all().only('id', 'labor_category',
```
tbaxter-18f: I only mention it because in this particular case it improve readability if you don't have to split lines, but you don't need `.all()`. You can just say `self.only(....)`. I'm not sure if the six characters would make enough difference to allow you to have this on a single line, though. Which is really the only good reason to bother changing it.

toolness: Oh good idea, I will change that, thanks!

PR #1988, opened 2018-05-31
Search box hueristic fixes

### LINKED ISSUE #1759: Explorer search box improvements

ericronne:
Closes #1759


ericronne:
Need help narrowing the gap between radio buttons and their labels (by about 6px)
image

Checkboxes have the same problem, can fix here if convenient
image

ericronne:
### Radio buttons in style guide look good with my changes …
image

### But explorer search radio buttons don't …
image

### Checkboxes in style guide are optically about one pixel farther away from their labels than radio buttons are to theirs, tho …
image

☝️ #helpwanted @hbillings 😭

hbillings:
@ericronne still looking into the radio button spacing issue, but I can tell you that the checkboxes are 1px further away because their borders are 1px thinner. :-/

ericronne:
Au contraire! I'm referring to the unselected states
image

tbaxter-18f:
I've looked in _inputs and _forms... is it possible the width difference is being inherited from USWDS?

hbillings:
So this seems to be some bizarre kind of DOM rendering:
This butts right up to the margin on the radio button:
screen shot 2018-06-01 at 2 50 30 pm

This one sorta spaces itself out:
screen shot 2018-06-01 at 2 50 17 pm

I haven't been able to find any rules that seem to be causing this. I even tried rendering the first one as a `ul` like the second is, and it still inherited the wonky spacing.

hbillings:
Oh, and also, @ericronne, sorry about misunderstanding your question about widths -- looks like radios have a width of `1.4rem` and checkboxes have a width of `2rem`, so that would probably be the cause of the discrepancy.

tbaxter-18f:
Heather, what's the markup difference between those examples? That extra inexplicable space reminds me of the space you see when using `display: inline-block`. I wonder if messing with `display` would do anything?

hbillings:
@tbaxter-18f I'm almost positive you're right, but I can't quite pinpoint where it needs to happen.

You can see the markup for the horizontal radios just under the search box here: http://calc.gsa.gov/
and the vertical radios here: https://calc.gsa.gov/styleguide/#radio-buttons-and-checkboxes

The biggest difference I see is the use of `div`s and `spans` in the first and `ul` and `li`s in the second, which is why I was puzzled that changing the first markup to the second didn't have an effect. Could be that something odd is cascading, I suppose...

tbaxter-18f:
OK, `[type=checkbox]+label::before, [type=radio]+label::before` is definitely setting `inline-block`. Just testing with `display:block; float: left` closes it up, but we may want to look at a flex or grid solution. Or just make the margin less, since we know inline-block has a predictable-ish spacing. 

However they both have inline-block so that's at least set the same.

Also seeing: the horizontals there are picking up a 0.75 right margin from `_inputs.scss` that the horizontals don't seem to be getting. But I don't think that's it. 

My current theory is that it has to do with how oversized elements overflow inline elements vs block or inline-block. It looks like the horizontal ones are slipping a bit left of the parent.  I wonder if the wrapping spans had were block and floated if would matter.

frontend/source/sass/data_explorer.scss
```diff
@@ -12,11 +12,20 @@ body {
...
   margin-right: 0;
   margin-left: $space-8x;
 }
 
+.search-header p {
+  margin-bottom: 0;
+  margin-top: 0;
+}
+
+.search-header h2 {
```
hbillings: These can be combined into a nested rule for clarity:
```
.search-header {
  p {...}
  h2 {...}
}

File: frontend/source/js/data-explorer/components/app.jsx [hunk 2]
```diff
@@ -91,11 +91,14 @@ class App extends React.Component {
...
-            (For example: Engineer, Consultant)
-          </p>
+          <div className="search-header columns nine">
+            <h2>Search labor categories</h2>
+            <p>
+              Enter your search terms below, separated by commas.
+              {' '}
+              (For example: Engineer, Consultant)
+            </p>
+          </div>
```
hbillings: We'll need to update the [Jest snapshot](https://facebook.github.io/jest/docs/en/snapshot-testing.html) of this file to accommodate for the markup change. You can do that by running `docker-compose run app jest --updateSnapshot` and committing the file that is changed as a result.

ericronne: okedoke

ericronne: ran it, but the file didn't seem to change :/

hbillings: Oops, actually, it looks like there is no snapshot for `app.jsx`.

hbillings:
The bizarre thing here is that the problem persists _even when the markup_ is the same and any custom CSS is removed:
screen shot 2018-06-04 at 1 07 05 pm

frontend/source/sass/data_explorer.scss
```diff
@@ -12,11 +12,21 @@ body {
...
 }
 
+.search-header {
+  p {
+  margin-bottom: 0;
+  margin-top: 0;
+  }
+  h2 {
+    margin-top: $space-8x;
+  }
```
hbillings: This is making me wonder why we don't have generic heading + p rules. Those would probably be pretty handy.

ericronne: Great minds …

ericronne:
The lion's share of issues in the [epic](https://github.com/18F/calc/issues/1752) cover hueristics, and this one seems to fit, since radio buttons were all running together

ericronne:
After running `./docker-update.sh`, here's how the radios look, compared to uswds …
image

Oddly, our spacing renders as 5(?) pixels less than uswds, despite the fact that both have a right margin of 14px. Where should i make the adjustment?

hbillings:
@ericronne The inspector says this is happening on line 13 of `components/_inputs.scss`

frontend/source/sass/components/_inputs.scss
```diff
@@ -10,7 +10,7 @@ input[type="radio"] + label,
 input[type="checkbox"] + label {
   margin-top: 0;
   &:before {
-    margin-right: .6rem;
+    margin-right: 1rem;
```
tbaxter-18f: I wonder, now that we've figured out the spacing weirdness, what if we had no rule/override here at all and just went with the USWDS 0.75em default?

PR #1992, opened 2018-06-01
Add docs for linting, type checking, and editor integration

### LINKED ISSUE #1799: Add guidelines for enabling local linting for IDEs/editors on the host system
It'd be nice to document our general strategy for using IDEs and text editors with CALC.  Typically this is done by installing a virtualenv and/or npm on the host machine and installing the python and/or node dependencies on the host system, so that local editors can offer syntax completion and such.

However, this strategy isn't well-known, and it's also error-prone.  For instance, right now I've created a virtualenv called `venv` in the root directory of CALC.  This works fine with VS Code, which is able to detect the virtual environment, but unfortunately `pytest` and other tools go digging into it when looking for files to test/lint/etc, which completely messes up the development workflow.  I guess one solution to this might be to put the virtualenv outside of the CALC root directory, but all of this back-and-forth gets cumbersome and it'd be nice if we could document some editor-agnostic best practices for folks.

### LINKED ISSUE #1959: Document linting and strong typing setups
See eg https://github.com/18F/calc/pull/1819#discussion_r190868048 for more context.

toolness:
This PR:

* Fixes #1799.
* Fixes #1959.
* Removes `docs/changelog.md`, which was erroneously added in a2d090e027a52d642ac54ec6f020e76223d8404d. This file used to be copied from the root directory as part of the documentation's `Makefile` but we got rid of that in #1866, and I suspect it was accidentally committed to the repository because `docs/changelog.md` was no longer in our `.gitignore`.


docs/editors.md
```diff
@@ -0,0 +1,59 @@
+## Editor integration
+
+In practice, it's often convenient to install editor integrations for
```
tbaxter-18f: I'd strike "In practice" since there is no "in theory", and just start with "It's often convenient to install. Also, I don't think you need the comma after "languages"

docs/editors.md
```diff
@@ -0,0 +1,59 @@
...
+yarn install --ignore-engines
+```
+
+At this point, you should be able to install any editor integrations for
+JavaScript-related code, and they should work as expected.
+
+### Python
+
+Setting up CALC's Python environment can be accomplished by installing
+the version of Python 3 specified in [runtime.txt](../runtime.txt). You
```
tbaxter-18f: Is they py3 version changing often enough to justify bouncing them to runtime.txt? Could we just say "Python 3.x" or "a version of python greater than 3.x"?

docs/linting.md
```diff
@@ -0,0 +1,97 @@
+## Linting and strong typing
+
+CALC uses a number of linters to maintain the style and hygeine of its
```
tbaxter-18f: Typo: hygiene

docs/changelog.md
```diff
@@ -1,344 +0,0 @@
-# Change log
```
tbaxter-18f: Why was the changelog killed?

docs/editors.md
toolness: Yeah, I try to avoid doc rot when possible, and honestly I don't think this documentation is going to be visited terribly often, as it's fairly advanced functionality.  Links to in-repo files are checked at doc build time to ensure they don't go stale, and the user only has to click on the link to see what the python version is, so I think the trade-off is worth it...

docs/editors.md
toolness: Good point, will change this!

tbaxter-18f:
I approved, but I may not have write access, so my review may not have been counted. Anyways, +1 from me.

toolness:
Oh thanks @tbaxter-18f !  It looks like you don't actually have the status to be "the approver" or whatever, but that's ok, I'd like @hbillings to weigh in on this one too if that's ok.

docs/changelog.md
hbillings: d'oh, sorry about this!

docs/editors.md
```diff
@@ -0,0 +1,59 @@
...
+Setting up CALC's JavaScript environment can be accomplished by installing
+[NodeJS][] and [yarn][] and running the following on your local machine:
+
+```
+yarn install --ignore-engines
+```
+
+At this point, you should be able to install any editor integrations for
+JavaScript-related code, and they should work as expected.
+
```
hbillings: Would it make sense to link to the linters bit here? Reading through this, my first question was, "Do we have any recommended integrations"?

toolness: Err, that is a huge question because it totally depends on what the user's editor preferences are, and it could easily become out-of-date.  And it's further complicated by the fact that we have lots of different linting/typing tools, so that means there could be several plugins that need to be installed per editor.  For instance, on VSCode, I have to use their python plugin, which supports both mypy and flake8 integration, and then I use the official typescript plugin and the eslint plugin on top of that.  Blorg.

docs/changelog.md
toolness: hehe no worries, i figured it'd come back to haunt us somehow :)

docs/editors.md
toolness: I mean, that said, we could provide advice on what to google for, maybe?  Like if i were using editor "SuperEdit" and i wanted flake8 support, I would google for "SuperEdit flake8", and odds are the first hit would be pretty useful (in fact, I just did this for "vscode flake8" and it took me to the docs for the python extension I use).

hbillings: LOL, I was writing a comment about how I just meant that "editor integration" seems like a potentially confusing term, and maybe it would just be useful to explain that a bit further by saying, "for instance! there are integrations for lots of the linters and whatnot we use, and you can read about those here!"

toolness: Hehe totally!  What do you think of the changes I just made in 43900bc?  Good, or too easy to miss?

toolness: I just realized that without the hyperlink, it looks like I'm asking you about changes I made in 43,900 BC.

hbillings: WHAT DID YOU DO TO THE DINOSAURS

hbillings: I think that's good!

Issue #1836, opened 2018-04-30
Data freshness


KellyBailey:
As a Contracting Officer, I want to be confident that CALC's data is current so that my estimates are accurrate (last updated date?).

PR #1993, opened 2018-06-04
[BLOCKED ON DESIGN STUDIO] Add Contracts.objects.get_schedule_stats().

toolness:
This adds a ~~`ScheduleUpdateInfo`~~ `ScheduleStats` named tuple along with a ~~`get_schedule_update_info()`~~ `get_schedule_stats()` method on our `CurrentContractManager` class that allows us to easily get information about when a schedule was last updated, how many contracts it has, and how many labor rates are in those contracts.

In the future, this can be used to determine how current CALC's data is (see #1836), perhaps via a data quality report or some other mechanism.

## Notes

* I could have just had `get_schedule_stats()` return a `dict`, but then I would have had to document what was in the dict, and it felt easier to simply create a self-documenting `NamedTuple` instead.  This is both more convenient, as we can use standard attributes rather than dictionary lookups (e.g. `info.schedule` instead of `info['schedule']`), and it also provides mypy with more type annotations, which can lead to more type-safe code.

### DIFF
--- a/contracts/tests/test_contract.py
+++ b/contracts/tests/test_contract.py
@@ -91,8 +91,10 @@ class ScheduleUpdateInfoTests(TestCase):
             schedule='MOBIS', idv_piid='b', upload_source=mar_source)
 
         self.assertEqual(Contract.objects.get_schedule_stats(), [
-            ScheduleStats('Consolidated', 1, 1, None, None),
-            ScheduleStats('MOBIS', 3, 2, feb_source.updated_at, mar_source.updated_at),
+            ScheduleStats('Consolidated', labor_rates=1, contracts=1),
+            ScheduleStats('MOBIS', labor_rates=3, contracts=2,
+                          earliest_update=feb_source.updated_at,
+                          latest_update=mar_source.updated_at),
         ])
 
 
--- a/contracts/tests/test_contract.py
+++ b/contracts/tests/test_contract.py
@@ -10,7 +10,7 @@ from contracts.mommy_recipes import get_contract_recipe
 
 from ..models import (
     Contract, convert_to_tsquery, CashField, BulkUploadContractSource,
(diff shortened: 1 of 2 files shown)


toolness:
Hmm, we might actually want to rename the method to `get_schedule_stats()` and have it also return the number of labor rates in the schedule, as I'm almost certain that would be useful to folks.

toolness:
I guess I am kind of doing this backwards--should have started with a priority guide or mock-up of what we want to show users, then worked on this back-end logic.

My current thinking for this API is that some schedules like those under Region 10 are uploaded in bulk, so there will only really be one date associated with them, while  others are  uploaded bits at a time (e.g. schedule 70).  We can get super fancy with visualizations if we wanted, e.g. to show that "30% of schedule 70 data was uploaded over 90 days ago" or something, but i figured that for now just showing the earliest and latest date might be enough information.  That said, we should probably have a really brief design studio or something to solidify this before committing this PR.

hbillings:
Totally agree that we should talk about what we want to show before we merge this in! Definitely seems like good design studio material. :)

hbillings:
Gah, the "close and comment" button is just too dang close to the "comment" button!

contracts/tests/test_contract.py
```diff
@@ -55,6 +59,45 @@ def test_abbr_uppercase(self):
         self.assertEqual(_normalize('JR. PERSON'), 'junior person')
 
 
+class ScheduleUpdateInfoTests(TestCase):
```
hbillings: Should probably change the name of this to `ScheduleStatsTests`?

toolness: Oh good catch, will update thanku.

ericronne:
Ice boxing pending design studio

PR #1995, opened 2018-06-04
Radio button markup standardization

hbillings:
So it turns out that the extra space was a literal extra space and had nothing to do with markup, but this still standardizes the markup if we want to do that. _shrugging man made of symbols_


frontend/source/sass/data_explorer.scss
```diff
@@ -534,7 +534,7 @@ a.focus-input {
 #query-types {
   padding: 0 1rem;
   label {
-    display: inline-block;
+    //display: inline-block;
```
toolness: Do you think we should remove this comment entirely?

toolness:
Oops, it looks like this change breaks accessibility actually, which is why the CircleCI build is failing: [Radio inputs with the same name attribute value must be part of a group](https://dequeuniversity.com/rules/axe/2.1/radiogroup).

frontend/source/sass/data_explorer.scss
hbillings: oh hmm...yeah, that looks like cruft from me trying to figure out what was goin' on.

hbillings:
Ohhh, interesting! I took off the `role="group"` designation when I changed the radios' parent from a `div` to a `ul`. Will fix (and, man, I wish some of these things had more clear names...)

ericronne:
@hbillings this still in need of a merge?
(i just merged `1759-explorer-search-box-improvements` into dev)

hbillings:
Yeah, but I was hoping you'd verify it looked right in your branch first.

ericronne:
sorry, was unaware of this pr until now

hbillings:
@ericronne Could you still check out this branch and review the way it looks?

ericronne:
wait, i *was* aware of this branch! 
at one point in time, many moons ago (ie yesterday).
👴 

looks like the conversion to `<li>` added some vertical space, i'll adjust, hold pls...

PR #1999, opened 2018-06-06
Install django-uswds-forms via PyPI

toolness:
I recently published django-uswds-forms to PyPI, so this modifies our `requirements.txt` to install it from there, rather than via git.  It's the exact same code (only some documentation has changed).


tbaxter-18f:
I'm still not sure why I can't review and approve, but LGTM!

Issue #627, opened 2016-08-30
Web component methods should be named according to new version of spec


toolness:
We should probably go ahead and use the new spec's naming methods, e.g. `connectedCallback` rather than `attachedCallback`.  I think our polyfill will work things out so that browsers (i.e. Chrome) supporting only the old spec will still work with our new-spec components, but we should make sure this is the case.

toolness:
Um, I am really confused.  Using the latest (v1.0.7) version of our polyfill, simply renaming all `attachedCallback` methods to `connectedCallback` does not work at all--the callback is never called. 😞 

Since everything works as-is and this isn't as straightforward as I thought it'd be, I'm going to remove this issue from the current milestone.

toolness:
shawnbot just linked to a [blog post about the polyfill's support for v1](https://www.webreflection.co.uk/blog/2016/08/21/custom-elements-v1) on slack, which may be useful.

PR #2006, opened 2018-06-12
Upgrade document-register-element to 1.9.1.

toolness:
This upgrades [document-register-element](https://github.com/WebReflection/document-register-element) to the latest version, but does not change our custom elements to use V1 of the specification.

After doing some research it looks like browser vendors do not agree on some crucial aspects of the Custom Elements specification V1. From the current [Google documentation](https://developers.google.com/web/fundamentals/web-components/customelements#extendhtml):

> Only Chrome 67 supports customized built-in elements ([status](https://www.chromestatus.com/feature/4670146924773376)) right now. Edge and Firefox will implement it, but Safari has chosen not to implement it. This is unfortunate for accessibility and progressive enhancement. If you think extending native HTML elements is useful, voice your thoughts on [509](https://github.com/w3c/webcomponents/issues/509) and [662](https://github.com/w3c/webcomponents/issues/662) on Github.

By "customized built-in elements" we mean the use of the `is` attribute to extend built-in elements, which is critical for progressive enhancement (e.g. `<button is="my-button">`).  It's particularly annoying that browser vendors disagree on this because it makes polyfilling harder, but document-register-element seems up to the task.

That said, another big advantage of document-register-element is that it is continuing to maintain backwards compatibility with V0 of the spec, which is what we use.  And because I don't see many compelling use cases to upgrade to V1, and on the contrary just see a bunch of confusion because different browser vendors disagree on different things, I'm tempted to just stick to V0 and abandon #627.

Regardless of that decision, though, it seems useful to at the very least upgrade document-register-element to its latest version in the interim, so that's what this PR does.


toolness:
Oh wow, apparently Chrome Canary already has a CustomElements implementation that deviates from the way DRE polyfills things, so I'm forcing the use of its [ponyfill](https://github.com/WebReflection/document-register-element#new-ponyfill-in-13) to make things work consistently across all browsers. Oy.

toolness:
Oh yeah, sorry, I didn't realize you'd already approved this a week ago!  I just merged `develop` back into this and will merge the PR once CircleCI is happy.

hbillings:
:party_parrot:

PR #2007, opened 2018-06-12
Fix ga.trackInterestingLink() on IE11

### LINKED ISSUE #2005: ga.autoTrackInterestingLinks() test fails on IE 11

toolness:
Fixes #2005, explanations are in the comments of the PR.

### DIFF
--- a/frontend/source/js/common/ga.js
+++ b/frontend/source/js/common/ga.js
@@ -1,6 +1,8 @@
 // @ts-check
 /* eslint-env browser */
 
+import { parse as urlParse } from "url";
+
 /**
  * This is DAP's weird exported Google Analytics function, documented
  * at:
@@ -92,7 +94,16 @@ function trackInterestingLink(el) {
     return;
   }
 
-  const isExternal = el.host !== window.location.host;
+  // Anchors on most modern browsers actually have properties on
+  // them that contain parsed values of the URLs, but not IE11,
+  // so we're going to parse the URL manually.
+  const { pathname, host, hostname } = urlParse(el.href);
+
+  if (!(pathname && host && hostname)) {
+    return;
+  }
+
+  const isExternal = host !== window.location.host;
 
   if (isExternal) {
     // Note that the format of this event corresponds to
@@ -100,14 +111,14 @@ function trackInterestingLink(el) {
     // means that anyone who wants to e.g. track statistics about
     // outbound links across all federal properties can do so,
     // and statistics from CALC will be included.
-    trackEvent('Outbound', el.hostname, el.pathname);
+    trackEvent('Outbound', hostname, pathname);
   } else {
     const isLocalDownload = LOCAL_DOWNLOAD_URLS.some(
-      url => el.pathname.indexOf(url) === 0
+      url => pathname.indexOf(url) === 0
     );
 
     if (isLocalDownload) {
-      trackEvent('Local Download', el.pathname, el.href);
+      trackEvent('Local Download', pathname, el.href);
     }
   }
 }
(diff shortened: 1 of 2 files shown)


frontend/source/js/common/ga.js
```diff
@@ -92,22 +94,31 @@ function trackInterestingLink(el) {
     return;
   }
 
-  const isExternal = el.host !== window.location.host;
+  // Anchors on most modern browsers actually have properties on
+  // them that contain parsed values of the URLs, but not IE11,
+  // so we're going to parse the URL manually.
```
tbaxter-18f: Are you saying window.location.hash doesn't work as expected in IE11?

toolness: No, it works fine--it's the `HTMLAnchorElement` or whatever that is weird.  Modern browsers have e.g. a `host` property that represents the host of the resolved `href` attribute but IE11 doesn't.

tbaxter-18f: so el.hash and/or el.host aren't working correctly? That's really weird. Links have been `location` objects since forever -- long before IE11. What about el.hostname?

toolness: I'm not sure, I didn't dig into too many details but just made sure the tests passed on IE11...

tbaxter-18f: I guess if it's passing, it's fine, but my concern is two-fold: 1) I think we're importing a module we maybe don't need to be importing and 2) I'm wary of fixes that aren't fully explainable.

That said, clear out the codeclimate issues and I'd be OK with approving.

frontend/source/js/tests/ga_tests.js
```diff
@@ -88,11 +88,16 @@ QUnit.test('autoTrackInterestingLinks() works', (assert) => {
...
-      bubbles: true,
-      cancelable: true
-    }));
+
+    // There's an easier way to synthesize this event, but
+    // this is the only way that works on IE11.
+    const event = document.createEvent("MouseEvent");
+    event.initMouseEvent(
+      'click', true, true, window,
+      0, 0, 0, 0, 0, false, false, false, false, 1, null
```
hbillings: _and they laughed the laugh of the damned_

frontend/source/js/common/ga.js
toolness: That's fair, I can investigate it further... I had run into issues with the cross-browser interface for anchor links in the past and assumed it was due to that, but I could be wrong.  Will poke at this when I get back home to my Win10 machine to verify.

toolness: Ok, so I made this example page:

```html
<!DOCTYPE html>
<meta charset="utf-8">
<title>Weird test page</title>
<p>Here is a link: <a id="link" href="./blah/a">boop</a></p>
<script>
onload = function() {
    var link = document.getElementById('link');
    console.log(link.href, link.host, link.hostname, link.pathname);

    // Now change the href, and we'll even give the browser extra time to get its
    // act together...
    link.setAttribute('href', './blah/b');
    setTimeout(function() {
        console.log(link.href, link.host, link.hostname, link.pathname);
    }, 1000);
};
</script>
```

I then started up a local server and visited the page.

On Chrome, Edge, and probably every other browser except IE, the following was logged to the console, as expected:

```
http://localhost:8080/blah/a localhost:8080 localhost /blah/a
http://localhost:8080/blah/b localhost:8080 localhost /blah/b
```

However, on IE11 the following was logged:

```
http://localhost:8080/blah/a localhost:8080 localhost /blah/a
http://localhost:8080/blah/b   ./blah/b
```

That is super weird; so IE11 _does_ have `host` and `hostname`, but it mysteriously sets them to the empty string if `href` is changed to a relative URL, which I assume is a bug.  But unless you can think of a better solution, I think that the only thing I can do is update the comment to be more accurate, but keep the solution of manually parsing the URL in-place...

PR #2011, opened 2018-06-14
Standardize alert styles

### LINKED ISSUE #1697: Alert messages should be displayed with a consistent style throughout the user experience
We've given some design attention to error alerts, but haven't addressed styling for alerts overall. Let's.

hbillings:
Closes #1697.

Regular alerts look like:
screen shot 2018-06-14 at 3 21 00 pm
screen shot 2018-06-14 at 3 20 49 pm

Django-generated alert markup looks like:
screen shot 2018-06-14 at 3 10 57 pm

No background alerts look like:
screen shot 2018-06-14 at 3 28 00 pm


hbillings:
Hmm, this error message looks weird
screen shot 2018-06-14 at 3 12 31 pm

hbillings:
@ericronne So you want all alerts, including the ones with no background, to have black text? Looks like this currently:
screen shot 2018-06-15 at 9 17 03 am

ericronne:
### Reduced the font on the basic style
image

### Modified form errors, to reduce the amount of pink / red coverage
image

### Removed "no-background" style, for consistency
image

### Outstanding issues:
- [ ] Reduce the icon sizes, too big
image

- [ ] Reduce this second line down from `h4` to the default `<p>` formatting
image
This code is in [price_list_upload.py](https://github.com/18F/calc/blob/1697-alerts/data_capture/views/price_list_upload.py)

- [ ] Double check that there are no other instances like this ☝️ 

- [ ] Do a QA pass to be sure i haven't missed any error instances

data_capture/templates/data_capture/messages.html
```diff
@@ -2,7 +2,11 @@
 <alerts-widget>
   {% for message in messages %}
     <div class="usa-alert usa-alert-{{ message.tags }}">
-      <h4>{{ message }}</h4>
+      {% if message|first == '<' %}
+        {{ message }}
+      {% else %}
+        <h4>{{ message }}</h4>
+      {% endif %}
```
toolness: @hbillings this is basically an allowance I made for situations where the message might want to provide its own HTML that includes e.g. a heading followed by a paragraph, rather than having all the message content being wrapped in a heading.  This is basically what's needed for the "We found an existing price list for contract number XXX" message in `price_list_upload.py` and I'm not sure if there's an easier way of allowing for it while also ensuring that other `messages.add_message()` calls elsewhere in our codebase (including in third-party apps) still work as expected.

hbillings:
I filed an issue for this; I think we should wait to do thorough Q/A until after it's closed.

data_capture/templates/data_capture/messages.html
hbillings: Yeah, I agree that this is a little weird, but I can't think of a better way to do it. :-/

data_capture/templates/data_capture/price_list/step_3_errors.html
```diff
@@ -8,15 +8,15 @@
 
       {% with total=rows|length %}
 
-      <div class="usa-alert usa-alert-error alert__no-bg" role="alert">
+      <div class="usa-alert usa-alert-error" role="alert">
         {% if not gleaned_data.valid_rows %}
-          <h3>No valid rows found!</h3>
+          <h4>No valid rows found!</h4>
```
hbillings: So I'm wondering about all these `h3` > `h4`s. From a markup perspective, it seems pretty likely that we're jumping levels, and that the structure of the page is now likely to be `h1` > `h2` > `h4`. Instead of changing the markup, maybe we should change the size of `h3`s in alerts?

data_capture/views/price_list_upload.py
```diff
@@ -103,9 +103,9 @@ def step_1(request, step):
             details_url = build_url('data_capture:price_list_details',
                                     reverse_kwargs={'id': latest.pk})
             msg = mark_safe(  # nosec
-                "We found an existing price list for contract number {}.</br>"
-                "If you'd like, you may "
-                "<a href='{}'>see its details</a>.".format(
+                "<h4>We found an existing price list for contract number {}.</h4>"
+                "<p>If you'd like, you may "
```
hbillings: Is this the only place we're currently using a `p` tag in a `message`? If so, could we maybe reword this so we just need the `h3` instead of changing the template to allow arbitrary HTML?

frontend/source/sass/components/_alerts.scss
```diff
@@ -42,40 +40,54 @@
...
-  &.alert__no-bg {
-    background-color: transparent;
-    border: none;
-
-    h3:before {
-      width: $h3-font-size;
-      height: $h3-font-size;
-      content: url('../../images/icon-warning.svg');
+// Used for styling paragraphs of alert text where a background doesn't make visual sense.
+.alert__no-bg {
```
hbillings: @ericronne, you mentioned getting rid of this style, but, uh, here it is.

frontend/source/sass/components/_forms.scss
```diff
@@ -75,8 +75,6 @@ fieldset + div + fieldset {
 // Django's default form rendering uses a <ul class="errorlist">
 // for errors, and <span class="helptext"> for help text.
 fieldset.fieldset-error {
-  border-left: 3px solid $color-red-dark;
-  background-color: #fef5f5;
```
hbillings: If we're getting rid of this here, where are we styling `fieldset-error`?

data_capture/templates/data_capture/price_list/step_3_errors.html
ericronne: Do alerts fit into header structure? USWDS styles alerts as h3.

ericronne: Do what you think is wise

data_capture/views/price_list_upload.py
ericronne: The `p` replaced a `br`, both artibrary. Reword away

frontend/source/sass/components/_alerts.scss
ericronne: kill

frontend/source/sass/components/_forms.scss
ericronne: We retained the input-area red border. That plus the adjacent error text should suffice.

data_capture/templates/data_capture/price_list/step_3_errors.html
hbillings: Hmm, good question. I'd assume they do, but it's hard to know where exactly they fall because they can be used in so many places. Maybe this is fine.

data_capture/views/price_list_upload.py
hbillings: @toolness, what say you?

ericronne:
~24px should do the trick

data_capture/views/price_list_upload.py
tbaxter-18f: I think we should wrap the first part in a strong tag (maybe with a class). If we really want it block level and/or larger, we should be able to target `.messages strong` or something similar, and we don't need the kinda-weird if/else in the template.

data_capture/views/price_list_upload.py
```diff
@@ -103,9 +103,9 @@ def step_1(request, step):
...
+                "<a href='{}' target='_blank'>see its details</a>.</p>".format(
```
tbaxter-18f: What's the motivation behind target=blank here?

frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,193 @@
...
+    singleInputs,
+  };
+}
+
+export function domContentLoaded(win) {
+  // there are several forms on the page; get the one within the .content div
+  // TODO: make this a more reliable ID selector or something
+  const form = win.document.getElementsByTagName('form')[0];
+  const inputs = parseInputs(win.document.querySelectorAll('input, select, textarea'), win.document.querySelectorAll('uswds-date'));
+  const submitButton = win.document.querySelector('.submit-group button[type="submit"]');
```
tbaxter-18f: I think `const submitButton = form.querySelector(whateveryouwant)` would be safer to get the elements in the current form. 

Also, were you going to loop through visible forms on the page, or are you thinking that should remain a future to-do?

frontend/source/sass/components/_alerts.scss
```diff
@@ -98,6 +110,7 @@ alerts-widget:focus {
 // These classes are used in the Django admin. They're automatically created
 // by Django, so styling them is more straightfoward than renaming them.
 .messagelist li {
+  padding-left: 7rem !important; // to create space for the icon
```
tbaxter-18f: super-minor, but if you put @extend(s) first, then you may not need the !important, because you'll be overriding an already-declared style.

frontend/source/sass/components/_forms.scss
```diff
@@ -130,13 +149,13 @@ form > .fieldset-error {
...
-    margin-bottom: 15px;
-    font-weight: bold;
-  }
+  //p {
+    //display: inline-block;
+    //padding: 0px 40px;
+    //vertical-align: bottom;
+    //margin-bottom: 15px;
+    //font-weight: bold;
+  //}
```
tbaxter-18f: Remove?

frontend/source/js/data-capture/form-validation.js
hbillings: Oh that's a good call. Will do!

Also...ugh. All the _forms_ are visible; it's the individual inputs in some of them that are hidden. Besides the main form, the other three are the "cancel" link, the buttons in the "are you sure you want to cancel" interstitial, and the button in the "safe mode" banner. None of the actual forms are hidden, but their inputs are.

But you've reminded me that I was going to add a class to the main form itself for more surety. I'll do that instead of the `form[0]` stuff.

frontend/source/sass/components/_forms.scss
hbillings: IT IS MY ART

frontend/source/js/data-capture/form-validation.js
hbillings: Hrm, so, defining `submitButton` in terms of `form` is causing the tests to error, because it's attempting to define `form` and `submitButton` asynchronously. I think it should (should!) be fine to leave the button as is, because there should really only ever be a single `.submit-group` with a single `submit` button on a page, anyway. I've added a class to the form itself, though, and I've appended that class to the button's `querySelector` also.

hbillings:
Augh, something in the intervening weeks has broken the `upload-widget` tests, and I'm not sure why. :sob:

data_capture/views/price_list_upload.py
```diff
@@ -103,7 +103,8 @@ def step_1(request, step):
             details_url = build_url('data_capture:price_list_details',
                                     reverse_kwargs={'id': latest.pk})
             msg = mark_safe(  # nosec
-                "We found an <a href='{}' target='_blank'>existing price list</a> for contract number {}.".format(
+                "We found an <a href='{}' target='_blank'>existing price list</a> "
+                "for contract number {}.".format(
```
tbaxter-18f: I'm not sure what it is, but something in this change is the reason the tests are failing.  Somehow the test thinks the "We found an existing price list for contract number _whatever_" changed. I'm not seeing it, but the test is.

hbillings: This is the weirdest thing. I can't figure out how to make it _not_ fail. I thought it was the inclusion of the `a` tag, but that doesn't seem to be it. Maybe we could pair Monday?

data_capture/tests/test_price_list_upload_views.py
```diff
@@ -159,8 +159,11 @@ def test_duplicate_contract_post_returns_html(self):
         self.assertEqual(res.status_code, 200)
         self.assertContains(res, 'A price list with this contract number has '
                                  'already been submitted.')
-        self.assertContains(res, 'We found an existing price list for '
-                                 'contract number gs-boop.')
+        self.assertContains(res, 'We found an <a href=')
+        self.assertContains(
+            res,
+            'existing price list</a> for contract number gs-boop.'
```
hbillings: OMG I can't believe this was the fix. THANK YOU.

PR #2013, opened 2018-06-15
Merge pull request #2012 from 18F/v2.9.1-rc

toolness:
Tag and release v2.9.1 to production

Um I am not sure why this merge is necessary, but right now GitHub thinks that `master` is ahead of `develop` and it shouldn't be. Blegh.


toolness:
Oh, i guess when you merged #2012 into `master`, GitHub made a merge commit, which meant that `master` had one merge commit in it that wasn't part of `v2.9.1-rc`...  Step 10 of the developer instructions say to merge `v2.9.1-rc` into `develop`, instead of merging `master` into `develop`, which means that `master` still has that one lingering commit that isn't in `develop`.

So maybe step 10 of the release instructions should merge `master` into `develop` instead of `v2.9.1-rc`?

Gah I am so confused.

PR #2015, opened 2018-06-15
Step indicator colors

hbillings:
OMG.

Okay, so, in the wake of #1819 and #1991, we've realized we need a mess of logic about form resubmission and tracking current steps. This is too much for right now; it's not a priority. So this is literally _just_ the colors, no hovers, no interactivity, no nuthin'.

screen shot 2018-06-15 at 4 05 10 pm


hbillings:
It's all good! I'm hopeful that we'll be able to circle back around to the functionality at some point, and it was _definitely_ a useful overview. :)

PR #2021, opened 2018-06-22
Enhance HTML5 validation with JS

### LINKED ISSUE #2017: Enhance HTML5 form validation with Javascript
In the pursuit of #2011, we realized we're using HTML5 form validation for as many forms as possible, which is great except that it means we lose our default error styling. 

Form fields with errors should look like this:
screen shot 2018-06-18 at 4 35 45 pm

Instead, they look like this:
screen shot 2018-06-18 at 4 33 45 pm

This is particularly a problem with radio buttons and checkboxes. In order to style them, USWDS hides the actual `input` element, which means no HTML5 error message shows up:
screen shot 2018-06-18 at 4 37 24 pm

It looks like [we can use Javascript](https://pageclip.co/blog/2018-02-20-you-should-use-html5-form-validation.html) to hook into [the events that are bubbling up](https://developer.telerik.com/featured/building-html5-form-validation-bubble-replacements/) from the HTML5 form fields and add styles as needed. This would not only allow us to apply our error styles evenly, but would also help us make sure that the errors are accessible properly. (Some folks said they couldn't get VoiceOver to read the HTML5 error, and I haven't been able to find anything definitive about whether they're accessible or not -- @toolness, do you know?)

hbillings:
**Note: Against #2011, not develop. Should be merged before #2011 is.**

Closes #2017.

To do:
- [x] Only show error message once for radios, checkboxes, and date widgets.
- [x] Finish styling the errors appropriately.
- [x] Tests


frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,99 @@
...
+
+        if (!input.validity.valid && input.validationMessage) {
+          errorContainer.className = INVALID_MESSAGE_CLASS;
+          errorContainer.textContent = input.validationMessage;
+
+          // TODO: only if input is not of type radio, checkbox,
+          // or has a parent with the class usa-form-group
+
+          if (options.showErrorMsg) {
+            parent.insertBefore(errorContainer, fieldsetLabel);
```
tbaxter-18f: I'm not sure, but could this throw an undefined error if it didn't find a legend on L48?

frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+
+      input.addEventListener('invalid', function (e) {
+        // prevent showing the HTML5 tooltip
+        e.preventDefault();
+        checkValidity({showErrorMsg: true});
+      });
+
+      // 'input' will fire each time the user types.
+      input.addEventListener('input', function () {
```
tbaxter-18f: What about textareas? And are we comfortable there won't be too much typing going on?

frontend/source/sass/components/_forms.scss
```diff
@@ -111,6 +112,13 @@ form > .fieldset-error {
   }
 }
 
+// Applied by form-validation.js as an enhancement for html5 validation styles
+.form--invalid {
+  border: 3px solid $color-red-dark;
+  border-left: 3px solid $color-red-dark;
```
tbaxter-18f: We don't need both these rules do we?

frontend/source/js/data-capture/form-validation.js
hbillings: yep, sure could! I updated it to grab a label if there's no legend, which _should_ always be truthy -- but I suppose there's still the possibility that an input could exist without either... maybe I should throw in a check to make sure things exist like I think they do in there.

frontend/source/js/data-capture/form-validation.js
hbillings: `input` is just the name of the event in this case, which is weird. I actually removed this, because it causes jumpiness with our styles, so now we just check validity on submit.

frontend/source/sass/components/_forms.scss
hbillings: ALL THE RED :all-the-things:

hbillings:
Okay, one weird thing that's happening (I suspect because I've got `preventDefault` on `invalid` events) is that the "Next" button only validates fields but doesn't submit them after the user clicks the button. The user has to click the button _twice_ to submit the field. I'm sort of out of brainpower right now, but I suspect I need to listen for whatever event is emitted there to allow the submission to go through on the first click.

hbillings:
@tbaxter-18f if you could review this again, 'twould be most appreciated!

frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,183 @@
...
+    }
+  }
+}
+
+// Radio buttons, inputs and USWDS date fields have deep nesting structures.
+// We need to find the `fieldset` that contains the input in question and
+// set the error message on that, otherwise we end up with multiple messages.
+// Because we're using django-uswds-forms, every form should be encapsulated
+// in a fieldset automatically.
+export function findParentNode(node) {
```
tbaxter-18f: Just curious, why'd you decide to go this route rather than just `node.closest('fieldset')`?

frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,183 @@
...
+    // just in case we try to run this on something not in a fieldset, prevent infinite loop
+    return null;
+  } else {
+    return node.parentNode;
+  }
+}
+
+export function documentCreateElement(window, type) {
+  return window.document.createElement(type);
+}
```
tbaxter-18f: I don't think I'm clear on the need for this function?

frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,183 @@
...
+
+export function toggleErrorMsg(window, options) {
+  if (options.parent) {
+    // I'm tired of typing `options`
+    const parent = options.parent;
+    const errorContainer = parent.querySelector(`.${INVALID_MESSAGE_CLASS}`)
+      || documentCreateElement(window, 'p');
+    // grouped inputs like radios have legends; single inputs just have labels
+    const fieldsetLabel = parent.getElementsByTagName('legend')[0] || parent.getElementsByTagName('label')[0];
```
tbaxter-18f: would `parent.queryselector('legend')` offer any advantage? Probably not. Just thinking out loud.

frontend/source/js/data-capture/form-validation.js
```diff
@@ -0,0 +1,183 @@
...
+  }
+  return {
+    combinedInputs: combinedInputs,
+    singleInputs: singleInputs
+  }
+}
+
+export function domContentLoaded(win) {
+  // there are several forms on the page; get the one within the .content div
+  // TODO: make this a more reliable ID selector or something
```
tbaxter-18f: Wouldn't you want to run this on all forms on a given page?

frontend/source/js/data-capture/form-validation.js
hbillings: It needs a polyfill in IE11. :-(

frontend/source/js/data-capture/form-validation.js
hbillings: It's pretty much solely for testing, ugh. I'm doing so much with the DOM that I have to mock up a fake window in order to test all the methods I'm using. 🤦‍♀️

frontend/source/js/data-capture/form-validation.js
hbillings: Oh! Derp, yeah, that would probably be easier.

frontend/source/js/data-capture/form-validation.js
hbillings: The page I was looking at had four(!) `form` elements, one of which was the hidden form that contains the data that was already entered. That data _should_ already be valid since it's run through this, but if it's not, there's no way to fix errors on the user's end.

hbillings: (That said, I'm not crazy about this solution and would be totes open to suggestions.)

hbillings: `selectors in mirror are larger than they appear`

PR #2029, opened 2018-06-29
Add favicon

### LINKED ISSUE #1096: Add Favicon
We don't really have a logo icon to use, but maybe we could just use the GSA or TTS favicon for the time being? Or even just a blue square using one of the blues from the CALC banner?

Or the donkey...

ericronne:
Closes #1096 

_Note: We may not need all of the sizes that i uploaded. Erred on the side of extravagance._


ericronne:
A cast of thousands

image

ericronne:
@tbaxter-18f not sure what this means / what status it's referring to …
"ci/circleci: build Expected — Waiting for status to be reported"

hbillings:
Merged with admin privileges because according to CircleCI, the check had actually passed, but they had an outage that prevented the UI on GitHub from updating.

PR #2033, opened 2018-07-02
Adding OG and other meta tags to head

tbaxter-18f:
I pondered different approaches to this for quite a while. If anyone has concerns about this approach, holler.


tbaxter-18f:
Currently blocked on merging in #2029

calc/context_processors.py
```diff
@@ -1,15 +1,9 @@
 from django.conf import settings
 from django.utils.safestring import mark_safe
 
-from .site_utils import get_canonical_url
 from .sample_users import SAMPLE_USERS
 
 
-def canonical_url(request):
-    '''Include the request's canonical URL in all request contexts'''
-    return {'canonical_url': get_canonical_url(request)}
```
hbillings: I _thought_ this was necessary for creating the right URLs in automatically generated emails and slack posts, but I can't trace it back to those things now, so 👍

data_explorer/templatetags/data_explorer_tags.py
```diff
@@ -0,0 +1,65 @@
...
+    template and on through this tag. In this case, the usage would simply be:
+
+    {% head_meta %}
+
+    Note that the variables should be named the same as the arguments below, 
+    and you must set them in the view. If you don't, it will fall back on the defaults.
+
+    You can also use named arguments:
+    {% head_meta title=title %} or
+    {% head_meta title="My most excellent page" %}
```
hbillings: Copacetic, dude ☮️

calc/context_processors.py
tbaxter-18f: canonical URLs are pretty good to have, especially when your system can do a lot of redirects or have a lot of routes resolving to the same content (both of which Django will happily let you do). I only removed it here because I rolled the functionality into the template tag rather than as a separate context processor. My thinking is that it made sense to centralize any code inserting into `head`

data_explorer/templatetags/data_explorer_tags.py
tbaxter-18f: image

tbaxter-18f:
OK, I think this is ready for f'realsies review now. Tests are passing locally.

hbillings:
WHOOOOOOOOOO

Issue #2032, opened 2018-07-02
Upgrade minimatch


hbillings:
Snyk reports a [high-risk vulnerability in `minimatch`](https://snyk.io/vuln/npm:minimatch:20160620), a library `gulp` uses. The recommended solution is to upgrade `minimatch`.

tbaxter-18f:
@heather I went ahead and knocked this out.

PR #2044, opened 2018-07-11
#2032: Updated gulp to hopefully make snyk happier

tbaxter-18f:
Should resolve the minmatch and lodash issues here: https://snyk.io/org/calc/project/fcdd61ae-d494-42e9-9ffc-0d5feb8c33e1


hbillings:
Two suggestions from @cmc333333, in case it's helpful:

>1. try re-running with ssh so you can poke around the file system
>2. when you build locally, maybe try removing the `built` dir first? could be that a previous run output the files and those are the ones being used now

I'm not sure what I'd be poking around for in SSH, but it does sort of make sense that it's not blowing up locally because the built files already exist.

tbaxter-18f:
Not sure why Github is lagging, but here's the Circle build: https://circleci.com/gh/18F/calc/1368

PR #2056, opened 2018-07-17
Hopefully more helpful, more logical tool tip content

### LINKED ISSUE #2055: Update the "Contract year" tool tip content
### It says this …
image

… which is untrue (downloaded .csv files only include the current contract year and two subsequent years), and also _doesn't answer the "What's this?" call to action_.

### Update the content to something that seems more useful and appropriate

It might be helpful if the content explained what the tool does. Here's an option: "Show rates for the initial, second, or third year of the contract"

ericronne:
Closes #2055


hbillings:
Looks like you need to update the Jest snapshots: `docker-compose run app jest --updateSnapshot`

ericronne:
image

Issue #2052, opened 2018-07-16
"Contains Words" search not working


KellyBailey:
![image](https://user-images.githubusercontent.com/12072003/42775414-23a3b4dc-88e9-11e8-8751-e90400485fdb.png)

tbaxter-18f:
Possibly related to https://github.com/18F/calc/issues/209

tbaxter-18f:
OK, this is interesting. 

We know that there are results for "copy editor" in the database. If you search for "editor" you can see them. 
We know there are also other results that have "copy" in them. "copy writer", for example. Search for "cop" and you'll see those, too. If I search the DB directly, with `qs = Contract.objects.filter(labor_category__icontains="copy")` I see 111 results.
But if as soon as you add "y" to "cop" in the search, results drop to 0.

My working theory, which admittedly is a stretch, is that because `copy` is a SQL command, we're either stripping it out or mightily confusing the DB with it. Investigation continues.

tbaxter-18f:
Should resolve #1058 too

PR #2061, opened 2018-07-20
#2052 search reliability

tbaxter-18f:
Refactors api.views to produce more predictable and safer results.


api/views.py
```diff
@@ -204,94 +200,91 @@ def get_contracts_queryset(request_params, wage_field):
...
         years = experience_range.split(',')
-        min_experience = int(years[0])
+        min_experience = years[0]
         if len(years) > 1:
-            max_experience = int(years[1])
-
-    if min_experience:
+            max_experience = years[1]
+    # Ensure the input matches expected numeric format to avoid injections
+    if min_experience and min_experience.isdigit():
```
hbillings: 👌

api/views.py
```diff
@@ -204,94 +200,91 @@ def get_contracts_queryset(request_params, wage_field):
...
+
+    business_size = request_params.get('business_size', None)
+    if business_size and business_size in ('s', 'o'):
+        if business_size == 's':
+            contracts = contracts.filter(business_size__istartswith='s')
+        else:
+            contracts = contracts.filter(business_size__istartswith='o')
+
+    # WE NEED TO DOUBLE CHECK SIN AND PRICE.
+    # THEY DO NOT APPEAR TO BE ON THE SEARCH PAGE.
```
hbillings: That's... really weird. Looks like you can issue an API query for price and SIN, even though I don't think we're using them anywhere?
https://calc.gsa.gov/api/docs/#rates

contracts/models.py
```diff
@@ -22,56 +23,52 @@
...
-        'foo:* | bar:* & baz:*'
+        >>> clean_search("an <script>EVIL()</script> example")
+        ['an evil() example']
 
-    Also, unrecognizable/garbage phrases will be removed:
-
-        >>> convert_to_tsquery_union(['foo', '$@#%#@!', 'bar'])
-        'foo:* | bar:*'
+        >>> clean_search('sr. frog, sme, disco chicken')
+        ['senior frog', 'subject matter expert', 'disco chicken']
```
hbillings: Or, as some call the disco chicken, :party_parrot:

contracts/models.py
```diff
@@ -117,8 +114,46 @@ def bulk_create(self, contracts, *args, **kwargs):
...
+            # We use an or operator here to build up the total queryset from
+            # each exactly matched item they found.
+            for phrase in phrases:
+                matches = matches | qs.filter(_normalized_labor_category__iexact=phrase)
+        else:
+            # Match any: Break phrases down into individual words
+            # So "business manager" finds results with "business" AND "manager"
+            # anywhere in the labor category.
+            # Note this is relatively hard on the DB, producing one query per word,
+            # but given our number of users, small queries and indexing, we should be OK.
```
hbillings: Not particularly relevant now, but I wonder if this is an argument for "match exact" being checked by default.

api/views.py
```diff
@@ -116,16 +96,32 @@ def parse_csv_style_string(s):
...
+        return results:
+
+        * Environmental
+        * AIMS
+        * Logistics
+        * Language Services
+        * PES
+        * MOBIS
+        * Consolidated
+        * IT Schedule 70
```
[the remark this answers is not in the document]
tbaxter-18f: I think I need to double-check this change. It's not mine, and I'm not 100% sure where it was picked up from. I THINK it came in with a merge from develop, but I need to track it down and be sure.

api/views.py
tbaxter-18f: Yeah, exactly. Is it some cruft, or is there something I don't know about?

contracts/models.py
tbaxter-18f: I don't think so. I think most people are doing small phrases, and we have a small DB with a small userbase, so we should be able to absorb these extra queries with no issue. As a user, I think 'match exact' would be Not What I Want in most cases.

api/views.py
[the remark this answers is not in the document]
hbillings: Ah, okay! It might be better to strike it, since the list has the potential to be more dynamic now.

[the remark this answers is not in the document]
tbaxter-18f: On further inspection, this looks like a reversion of Atul's work. Correcting now.

api/views.py
tbaxter-18f: I'll leave them in. Wondering if we want to log an issue to fully determine if they should be in there or not.

contracts/models.py
```diff
@@ -381,6 +395,12 @@ def normalize_labor_category(val):
 
         return val
 
+    def get_readable_business_size(self):
+        if 's' in self.business_size.lower():
```
tadhg-ohiggins: Should this be `if self.business_size.lower().startswith('s')`? Or do we know the the presence of an `s` anywhere in the string is always sufficient? I ask in part because the model gives this field a length of 128.

tbaxter-18f: I think that's a really good question I don't really have the answer to. This is unchanged from L342 in the old code, but I think you're right, something feels off here.

Looking at the model and the view -- and the front end -- I think we _should_ be able to match more accurately than any sort of `in` check. But at the very least, I think you are correct that `startswith` would be better. I would check the form HTML to look at the possible submitted values and ask @KellyBailey or @tram to confirm, but yeah, I think this check could be tightened up.

tadhg-ohiggins: The HTML appears to submit values of either `s` or `o`, so this isn't going to cause problems currently but should likely either use an equality comparison or `startswith`.

tbaxter-18f: Thanks for checking! Do you want to pull the branch and make that change or would you like me to?

tadhg-ohiggins: @tbaxter-18f  If you already have it locally, would you mind doing it? I'm currently tethering from a place with bad connectivity.

tbaxter-18f: it'll be Monday AM before I get to it, but I'm happy to.

hbillings:
Is this good to merge?

tbaxter-18f:
If @tadhg-ohiggins has no objections, I'd say yes!

tadhg-ohiggins:
Looks good to me!

Issue #2034, opened 2018-07-02
Combine the contract ID and vendor name columns in the results table


ericronne:
One of the table improvements we're tackling in #2024.
_note: eric is replacing the doc icon_

![image](https://user-images.githubusercontent.com/9259185/42186261-9e6977c0-7e11-11e8-9335-d9000076c57c.png)

- [x] Merge the two columns
- [ ] ~Color the contract ID text with `$color-gray-dark`~
- [x] Reduce the ID text to `$small-font-size`

ericronne:
Here's [an icon](https://drive.google.com/open?id=1YE2xwzP_R5x3ij8zsKgtr3ADnyjkcva8) to try. Likely will revise.

![image](https://user-images.githubusercontent.com/9259185/42233475-4ec29d2c-7eb7-11e8-94bf-3cbc2a510a2e.png)

ericronne:
@hbillings if this will be time consuming, we can split it out from the rest of the spike things. Don't want it to hold up testing.

abrouilette:
Needs testing

PR #2067, opened 2018-07-23
Contract vehicle labeling

### LINKED ISSUE #1929: As a data administrator, I want the Schedule title in the data explorer to be changed to Schedule / Contract Vehicle

ericronne:
Closes #1929.

### The new label works fine on the filter (yay) …

image

### But causes a line break on the table (boo) …
image

Merging the vendor and contract columns (#2034) should give us the space needed to fix.


ericronne:
Column needs to be wider tho :/

tbaxter-18f:
The column SHOULD widen automatically to fit the wider content. _should_

ericronne:
You may be onto something, sir! Not sure how to do this here, since it's `.jsx`

frontend/source/js/data-explorer/components/results-table/index.jsx
```diff
@@ -24,7 +24,7 @@ const COLUMNS = [
   }),
   createSortableColumn({
     key: 'schedule',
-    title: 'Schedule',
+    title: 'Contract vehicle',
```
tbaxter-18f: If it's going to work, right here is where you'd have `'Contract&nbsp;vehicle'`

But it may not work.

ericronne: Yeah i'd tried that, sigh. It stretches the column, but renders quite literally …

image

hbillings: Hm, I wonder if a unicode non-breaking space would work here? Although... unicode.

We could just set a min-width on this column, also, even though I hesitate to suggest that.

hbillings:
What are we blocked on here?

ericronne:
#2034 Merging vendor and contract columns (on the theory that this would give us the space and resolve things), but if we can figure out a non-breaking space method … unblocked

hbillings:
`min-width: 145px; /* temp fix until we combine vendor/contract # */` should do it.

hbillings:
Looks like you need to update your Jest snapshot, and then you should be good to go!

ericronne:
g2g @tbaxter-18f?

PR #2048, opened 2018-07-12
Upgrade ESlint

hbillings:
Closes #2031.

PR #2027, opened 2018-06-28
Results table improvements

ericronne:
Closes #2024

PR #2068, opened 2018-07-23
Tag and release v2.10.0 to production

hbillings:
### Added
- Made it possible for data administrators to edit and add schedule names, descriptions, and SIN numbers, by centralizing schedule metadata in an API call (#2018).
- Added OpenGraph metadata tags (#2033).
- Added progressive enhancement to HTML5 form validation, allowing a standard error alert design (#2021).

### Changed
- Updated deployment docs (#2016) and styleguide docs (#2008).
- Updated styling for data capture stepper widget (#2015).
- Updated the visual design of the results table (#2027).
- Updated the visual design of alerts (#2017).
- Upgraded the version of `document-register-element` used by our web components (#2006).
- Updated the version of ESLint (#2048).
- Updated the version of gulp (#2044).


.circleci/config.yml
```diff
@@ -81,11 +81,20 @@ jobs:
             source $BASH_ENV
             nvm install $CI_NODE_VERSION
 
+      - run:
```
tbaxter-18f: Heather, this was only needed during the gulp4 update (if then). it doesn't hurt anything, but can also be safely removed for some circleci speed-up. Just FYI.

hbillings: Excellent! Thanks.

hbillings:
@tbaxter-18f Yeah, I'm really not sure why the instructions specify not to squash merge it, because I agree that that seems WAY more readable. Might as well give it a go!

PR #2096, opened 2018-08-17
Style refinements for the 2084 search ui work

ericronne:
Changes:
- Added space below "Search CALC" header
- Reduced padding in selector in order to make its height match the search box
- Lightened the selector-list-item default text color a notch, for better contrast with the selected option (because sketch does a better job representing colors than chrome does!)
- Other vertical spacing adjustments

### DIFF
--- a/frontend/source/sass/components/_forms.scss
+++ b/frontend/source/sass/components/_forms.scss
@@ -258,7 +258,7 @@ input[type].form__inline,
   background: $color-gray-lightest;
   border-radius: $border-radius;
   button {
-    background-position: right $space-2x top 1.9rem;
+    background-position: right .6rem top 1.6rem;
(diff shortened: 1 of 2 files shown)


frontend/source/sass/components/_forms.scss
```diff
@@ -187,7 +187,7 @@ dl.contract-details {
   display: flex;
   padding: 0;
   margin-top: $space-1x;
-  margin-bottom: $space-3x;
```
hbillings: Can we override this elsewhere? `.search-group` gets used in multiple places and I'd like to not override this everywhere.

hbillings: Oops, never mind, I'm thinking of `.form__inline`!

PR #2106, opened 2018-08-24
Break price list analysis errors into their own page

### LINKED ISSUE #2077: Price list analysis: process errors before serving up the analysis page
On the [price list analysis](https://github.com/18F/calc/pull/997) spike, the error review and analysis report are combined onto a single page. This makes for an odd layout, and also prevents users from getting any reasonableness-related feedback about rows which may simply have an error-causing typo.

If there's no special reason it was built in that way, i think it would be better to split out these two functions, like we do in the data capture process.
 1. Perform the error check and let the user deal with any errors, then
 2. Display the analysis on its own page …

image

Error page from data capture, for reference …
image

hbillings:
Closes #2077.

The vast majority of this is done, but I'm running into an issue getting the step 1 form data to render. I'm not sure we _need_ this, but the code is set up to take it. @tadhg-ohiggins, I'd really appreciate an eyeball!


data_capture/views/price_list_analyze.py
```diff
@@ -99,16 +102,46 @@ def analyze_step_2(request, step):
     )
 
 
-@steps.step(label='Analysis')
 @require_http_methods(["GET"])
-def analyze_step_3(request, step):
+def analyze_step_2_errors(request):
+    step = steps.get_step_renderer(2)
```
hbillings: This is just for the step widget to render the correct step bubble.

data_capture/views/price_list_analyze.py
```diff
@@ -99,16 +102,46 @@ def analyze_step_2(request, step):
...
 
+    step_1_post_data = get_nested_item(request.session, (
+        'data_capture:analyze_price_list',
+        'step_1_POST'
+    ))
+
+    step_1_form = AnalyzeStep1Form(step_1_post_data)
+
+    form_kwargs = dict(
+        schedule=step_1_form.cleaned_data['schedule'],
```
hbillings: This seems to be where this is blowing up. I keep getting `'AnalyzeStep1Form' object has no attribute 'cleaned_data'`, even though in `analyze_step_2` the same pattern is followed. 

(I'll be honest, I don't entirely follow how `AnalyzeStep1Form` works or what it needs.)

hbillings:
The ESLint errors here should be corrected in #2108.

PR #2108, opened 2018-08-29
Add exact match checkbox

### LINKED ISSUE #2058: Replace the three search radio buttons with an "Exact match" checkbox option, test
image

hbillings:
Closes #2058.

Since we already removed the "match phrase" option and "match all" is the default, this just simplifies the options further. Required adding a bit of logic to `setQueryType` to return `match_all` by default and `match_exact` on toggle.

screen shot 2018-08-29 at 8 54 29 am

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -19,5 +19,11 @@
     "react/forbid-prop-types": [0],
     "react/destructuring-assignment": [0],
     "jsx-a11y/label-has-for": [0],
+    "jsx-a11y/label-has-associated-control": [ 2, {
+      "labelComponents": ["label"],
+      "labelAttributes": ["htmlFor"],
+      "controlComponents": ["input", "select"],
+      "depth": 3,
+    }],
   }
 }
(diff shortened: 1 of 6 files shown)


frontend/tests/test_selenium.py
```diff
@@ -347,7 +333,10 @@ def set_form_value(self, form, key, value):
             field_type = field.get_attribute('type')
             if field_type in ('checkbox', 'radio'):
                 for _field in fields:
-                    if _field.get_attribute('value') == value:
+                    # the exact match toggle changes value based on querytype,
+                    # but always possesses the name "match_exact"
+                    if _field.get_attribute('value') == value or \
+                            _field.get_attribute('name').find(value):
```
tadhg-ohiggins: Fitting really long multiple conditions onto one “line” is a pain in Python…

hbillings: when the limit is 100 chars but your line is 105 😭

Issue #2084, opened 2018-08-07
Implement contract and vendor search UI


ericronne:
Acceptance criteria:
- [ ] Users can search by contract number.
- [ ] Users can search by vendor name.

These reference images should help us launch the combination contract vehicle / vendor search / contract search tool with minimal impact on the layout of the home page.

![now](https://user-images.githubusercontent.com/9259185/43802840-a7754986-9a5c-11e8-942a-3a530bfd5819.png)

![2](https://user-images.githubusercontent.com/9259185/43803151-a06316e0-9a5d-11e8-9573-e11eb24561b1.png)

![now copy 2](https://user-images.githubusercontent.com/9259185/43803035-43580fa0-9a5d-11e8-893f-66eedf2d2cf9.png)

hbillings:
Q: Are we doing the "exact match" checkbox? I don't see that on this mock but I thought that was the general plan. Thanks!

ericronne:
Sure that's worth a try

![image](https://user-images.githubusercontent.com/9259185/43809706-674d802c-9a79-11e8-8075-2ebc4c9ba239.png)

hbillings:
Hm, do we also want to autocomplete vendor/contract #, like we do now for labor categories?

hbillings:
Another thought: As the number of schedules in CALC grows, the vendor and contract search options are going to get pushed further and further down. I'm not sure what to propose to fix this, but wanted to note it somewhere.

hbillings:
@ericronne When you get a chance, could you post the version with the dropdown expanded? thx!!

ericronne:
New Revision (provided @tram feels it merits testing), designed to improve flow of actions. In essence:
1. Make selection from pull down, telling us what sort of thing you are searching for  
2. Enter the text of your search
3. GO

## Default
![image](https://user-images.githubusercontent.com/9259185/43977245-17b9f612-9ca9-11e8-90bd-49c126c3a13b.png)

## Expanded
![image](https://user-images.githubusercontent.com/9259185/43977370-98cf4aae-9ca9-11e8-92c7-f59be6bd4726.png)

tbaxter-18f:
Related to @hbillings  question about autocomplete above: I think autocomplete is nifty (and probably expected) but there probably needs to be some visual mechanism to indicate where it's autocompleting from, tied back to where you're searching in. If that makes any sense at all.

ericronne:
Makes total sense, and agreed.

To the "Design Cave"!
[...]

PR #2125, opened 2018-09-06
Add contract number and vendor UI

hbillings:
While we're waiting on the search returns to be available, I'm rejiggering the schedule selector to be able to accommodate that information when it becomes available.

Done looks like:
~- [ ] Vendor and contract each have a Redux store~ We don't need to capture these currently; just keep track of which one we're showing, which we're now doing with a single new `searchType` store.
~- [ ] The `SearchCategory` component does some logic to first determine if vendor and contract are empty, and, if no `selectedSchedule` exists, defaults to `search in all schedules` if so.~ Also obviated by having a single `searchType` store.
~- [ ] The `SearchCategory` component sets `selectedSchedule` to empty when vendor or contract is chosen.~ No longer necessary, as the `selectedSchedule` check now only happens if `searchType` is `schedule`.
- [x] The search UI's markup can accommodate vendor and contract labels as well as schedule.
- [x] The search UI is accessible.
- [x] Vendor and contract have tests written for them.


hbillings:
@ericronne, this is ready for your stylistic massaging.

Question about the display text. It stands to reason that when vendor or contract is chosen, the button (collapsed) displays like this:
screen shot 2018-09-12 at 9 12 16 pm

But this means that when it's open, it displays like this (plz to ignore the errant checkmark; that will get fixed next):
screen shot 2018-09-12 at 9 11 31 pm

How should I address this? I think we probably want both the currently selected text and the `Search labor categories` headings to display, but I'm not sure how that should look.

ericronne:
closes #2084

ericronne:
Ah! I could've more explicitly detailed the styling. The text of the selected option should be `$color-base`, and it should get a checkmark. Eg …

image

ericronne:
Ok i'm grokking your point. Since a user who sees a label other than the default will have already used the selector, i doubt there would be a usability hit if we kept the label as you have it, but here's a solution that might be a bit more clear and (dare i say?) eleganté …

image

ericronne:
The gray prompt text should also update to reflect the option which has been chosen …

image

ericronne:
Imma split a branch and take a crack at some of these changes! 🤞 🙏

PR #2129, opened 2018-09-06
allows for schedule 736 price list upload

tram:


calc/settings.py
```diff
@@ -410,6 +411,10 @@
     'data_capture.panels.ScheduledJobsPanel',
 ]
 
+PRICE_LIST_ANALYSIS_FINDERS = [
```
hbillings: do we want to add this analysis bit to the upload, or is this left over from earlier conflation between the two?

data_capture/schedules/s736.py
```diff
@@ -0,0 +1,306 @@
...
+        ][0]
+
+    def contract_model_base_year_rate(self):
+        return self.cleaned_data['price_including_iff']
+
+
+class Schedule736PriceList(BasePriceList):
+    # TODO: This class should be DRY'd out since it is nearly verbatim
+    # from the Schedule70PriceList class, but since this feature
+    # is somewhat experimental, I'm focusing more on implementation speed
```
hbillings: Are the comments in this file relevant, or are they copied over also?

tadhg-ohiggins: I think they're relevant and should be left in.

calc/settings.py
tadhg-ohiggins: This is a leftover, which I'll remove.

tadhg-ohiggins:
I decided to rebase instead of merge to bring this into line with `develop` and created a new branch and a new PR here: https://github.com/18F/calc/pull/2140

hbillings:
Closed in favor of #2140

PR #2133, opened 2018-09-13
Add tests for Schedule 736, plus code tweaks to make tests pass.

tadhg-ohiggins:
BA/Bachelors?
It's hard to say, in winter,
As the snowflakes fall.

Add some tests (based on the Schedule 70 tests) for Schedule 736, and tweak the Schedule 736 code to make those tests pass.


hbillings:
This also passes locally for me -- wondering if it's possible there's a version mismatch somewhere. Looks like CircleCI is running Python 3.6.6, while my local docker is on 3.6.4. Doesn't seem like a patch version should cause something like this, but I'm running out of ideas. :-/

hbillings:
Okay, so when I used Trammell's original files, the tests ran (though obviously failed because the sheet has the wrong value in it). When I opened the file, edited it, and closed it, I got this warning in iTerm: `warning: CRLF will be replaced by LF in data_capture/static/data_capture/s736_example.xlsx.
The file will have its original line endings in your working directory.` I know @toolness worked primarily on a Windows machine, which would use `CLRF` endings.

I don't know enough about the `zipfile.BadZipFile: Bad magic number for file header` error that's showing up when the modified file tries to run, but I'm wondering if we'd still have this issue if we edited the file on a Windows box.

PR #2143, opened 2018-09-19
Update the histogram to be the "new" site colors

hbillings:
We were still rocking the old histogram under the assumption that we'd merge it when we merged the rest of the data explorer restructure, but we're not going to have time to do the work required for the latter. This updates the visual style of the graph to be consistent with the new styles, since in its current iteration it uses colors no longer used anywhere else on the site.

TODO:
- [x] Add the dark green line atop the light green bars for 508 compliance.


hbillings:
Looks like: 
screen shot 2018-09-19 at 1 55 11 pm

PR #2149, opened 2018-09-20
Fix jest timeouts

### LINKED ISSUE #2145: Figure out why Jest is timing out
On Circle CI, almost every other test run times out with the error message:
```
Summary of all failing tests
 FAIL  frontend/gulp/tests/webpack-util.test.js (182.247s)
  ● webpackify › performs dead code removal in production

    Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.
      
      at node_modules/jest-jasmine2/build/queue_runner.js:68:21

  ● webpackify › sets NODE_ENV to empty string when not in production

    Timeout - Async callback was not invoked within the 5000ms timeout specified by jest.setTimeout.
      
      at node_modules/jest-jasmine2/build/queue_runner.js:68:21
```

I've tried overriding `jest.setTimeout` in the tests themselves, but I think the override must not be in the right place because nothing happens.

hbillings:
Closes #2145


hbillings:
I've run this six times, and it's failed twice with the error "Too long with no output," which seems to be [an error that sometimes occurs with Docker on CircleCI](https://discuss.circleci.com/t/too-long-with-no-output-while-building-docker-image/12275/13)? I haven't triggered the Jest timeout error that we previously encountered.

hbillings:
Looks like we could also up the CircleCI runtime: https://soetani.com/en/blog/9

PR #2182, opened 2018-09-26
Restores a return-to-top link

hbillings:
There was a regression that happened somewhere where we lost work already done on this. Redid it real quick.


hbillings:
Unable to determine why Jest tests are failing, so closing since we're out of time to work on this.

PR #2194, opened 2018-09-28
Add NLTK's Averaged Perceptron Tagger .pickle file

tadhg-ohiggins:
Why reach for the clouds
When what we need for language
Could be right here now.

Add NLTK's Averaged Perceptron Tagger .pickle file to repository instead of downloading it on every build, and make changes to build files and to price analysis reflecting those changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -73,18 +73,6 @@ jobs:
           paths:
             - ~/venv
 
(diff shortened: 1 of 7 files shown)


tadhg-ohiggins:
Merging without review because we need the data from how it works on dev sooner rather than later.

PR #2522, opened 2020-10-19
adding footer email changes for PROD

dsomalaraju:


theodesta (approved):

theodesta (changes requested):
We don't want to push the Fake UAA stuff.  We need to clean it up first.

PR #2631, opened 2021-03-05
BRMP-12162: fixed fetch with an npm lib, basically a polyfill, fetch …

ccslowmo:
…is not compatible with IE 11


chinchalinchin (approved):

ccslowmo:
closing, created a new PR using yarn instead of npm to install

PR #2650, opened 2021-04-14
BRMP-14294: swapped plus logo out for calc logo white svg

theodesta:
Can you please add frontend/images/calc_logo_white.svg if it's not there already?


ccslowmo (commented):
The calc logo white is already available, https://github.com/GSA/calc/blob/feature/BRMP-13944-logo-noplus/frontend/static/frontend/images/calc_logo_white.svg

ccslowmo (approved):

PR #2696, opened 2021-05-13
Making order year changes- pET

dsomalaraju:


data_explorer/templates/estimating_tool.html
```diff
@@ -412,7 +412,7 @@ <h3>Pricing Control Information</h3>
                      </div>
                      <div class="left column valueList ">
                         <select class="add_tooltip" 
-                           tooltip_title="*******************"
+                           tooltip_title="Select a county from the list where you expect work will be performed, or similar to work will be performed. The county list is limited to counties where wage data exists for your selected occupation. County data is based on the MSA"
```
theodesta, line 415: What does "or similar to work will be performed" mean?

theodesta (approved):

data_explorer/templates/estimating_tool.html
dsomalaraju: I am guessing it is because of we are displaying one pricing (MSA value) if they are more counties under a city.

